PR #3, opened 2018-05-24
Better asset handling

briandds:
This PR uses `webassets` to generate unique URLs for assets with checksums attached.  It also improves the pipeline flow for development.


jgarberdds:
Not working for me
```
(.venv) atst webassets % DEBUG=1 ./app.py
Listening on http://localhost:8888
WARNING:tornado.access:404 GET /static/assets/out.css?83b71a4b (::1) 2.99ms
WARNING:tornado.access:404 GET /static/uswds-1.6.3/js/uswds.min.js (::1) 1.04ms
```

.gitignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
jordangov: Should have a slash at the end (although git is pretty smart about it)

briandds: Yeah, I noticed that, too -- fixed it in #4

PR #5, opened 2018-05-24
Fix travis.yml

briandds:
This fixes the yaml in `.travis.yml`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,5 +3,6 @@ python:
   - "3.6"
 before_install:
   - gem install sass
+  - npm install
 script:
   - python -m pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-  python:
-    - "3.6"
-  script:
-    - python -m pytest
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: python
-  python:
-    - "3.6"
-  script:
-    - python -m pytest
+python:
+  - "3.6"
+before_install:
+  - gem install sass
```
jordangov: Where is the `sass` gem invoked? That is, where do we call for the sass to be compiled into css?

briandds: The sass CLI is used by webassets -- when the url is requested, if the css doesn't exist then it is generated.

briandds: i.e. when the `.urls` method is called (not when the HTTP request comes in)

briandds: https://github.com/dod-ccpo/atst/blob/master/templates/base.html.to#L7

jordangov: Interesting... ok, wasn't aware of that!

PR #7, opened 2018-05-25
First pass at Scripts to Rule Them All

ddsdevon:
First pass at Scripts to Rule Them All
- script/setup: first time setup
- script/bootstrap: dependency install and update
- script/update: do the things when the app has been updated
- script/server: launch the app
- script/test: run tests

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# If a command fails, exit the script
+set -e
+
+# Ensure we are in the app root directory (not the /script directory)
+cd "$(dirname "$0")/.."
+
+# Install virtualenv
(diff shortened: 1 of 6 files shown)


script/setup
```diff
@@ -0,0 +1,23 @@
...
+
+# Install virtualenv
+pip install virtualenv
+
+# Create and activate virtual environment
+python3 -m venv .venv
+source .venv/bin/activate
+
+# Install/update pip
+pip install --upgrade pip
```
jakerella: Maybe add `npm install -g npm` to update that as well?

jakerella:
One comment (inline), but otherwise lgtm.

PR #8, opened 2018-05-25
Login screen

luisgov:
This PR updates some routes and adds a login screen that takes you home when selecting 'Sign In with CAC'.

screen shot 2018-05-25 at 14 42 39


templates/blank.html.to
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
```
briandds: I think this file and the one below can be removed.

templates/hello.html.to
```diff
@@ -0,0 +1,23 @@
+{% extends "blank.html.to" %}
```
briandds: I think this one can be removed.

luisgov: @briandds do we even need this file at all? can we remove `hello.html.to`?

PR #12, opened 2018-05-30
Requests

luisgov:
This PR adds a summary table for existing requests under `/requests` while moving the _Create Request_ dialogue to `/requests/new/1`

screen shot 2018-05-30 at 11 30 05

---
screen shot 2018-05-30 at 11 30 18


dandds:
@briandds Curious what pattern you think is best for routers -> handlers. One handler per endpoint? That's what I've done here so Luis could split out those views easily.

briandds:
> @briandds Curious what pattern you think is best for routes -> handlers. One handler per
> endpoint? That's what I've done here so Luis could split out those views easily.

Not necessarily -- I think of them as grouping together related resources (like a 'controller') -- so using the same handler for 'requests' (existing or new) would make sense to me.

dandds:
K, I'll push that functionality into a single handler.

atst/handler.py
```diff
@@ -11,7 +11,8 @@
 css    = Bundle(
                 'atat.scss',
                 filters = 'scss',
-                output  = '../static/assets/out.css')
+                output  = '../static/assets/out.css',
+                depends = ('**/*.scss'))
```
briandds: cool

templates/requests.html.to
```diff
@@ -4,19 +4,48 @@
...
-  <div class="usa-width-three-fourths usa-layout-docs-main_content">
-
-    {% block form %}
-    form goes here
-    {% end %}
-    {% block next %}
-    <a class='usa-button usa-button-primary' href='{{ reverse_url('request_form',next_screen) }}'>Save &amp; Continue</a>
-    {% end %}
-  </div>
+  <h1>Requests <a class="usa-button usa-button-secondary" href="/requests/new/1">New Request</a></h1>
```
briandds: I'd avoid typing the path here, and instead use `reverse_url` to generate it.

atst/handler.py
luiscielak: all set!

dandds:
@luiscielak Brian and I chatted and we're going to leave the handlers as-is for now. I'll defer to @briandds 's final review here. Looks good to me!

PR #14, opened 2018-05-30
Update README and mention virtualenv.

### LINKED ISSUE #13: Setup script may not be capturing full setup
I tried to clone, then set up the Python app.

Steps:
1. Install Python 3 (I only had 2)
1. Run `./script/setup` (no errors, couple warnings, but not relevant here)
1. Run `DEBUG=1 ./app.py`

Here is the output:
```
Traceback (most recent call last):
  File "./app.py", line 3, in <module>
    from atst.app import make_app
  File "/Users/jordan/projects/atat/atst/atst/app.py", line 1, in <module>
    import tornado.web
ImportError: No module named tornado.web
```

briandds:
This mentions the use of virtualenv.

I think documenting this is enough, but let me know if there's a better way.

Resolves #13

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,14 @@
 
 ## Running (development)
 
-To start the app and watch for changes:
+Turn on the virtual environment:
+
+    . .venv/bin/activate
+
+Note: you may want to use something like [direnv](https://direnv.net)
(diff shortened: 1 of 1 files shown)


briandds:
I'm closing this -- a better solution is to use `script/server`

PR #16, opened 2018-05-30
Envrc and gitignore

briandds:
This adds a default `.envrc` and ignores python byte code.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+. .venv/bin/activate
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,1 @@
-. .venv/bin/activate
+source .venv/bin/activate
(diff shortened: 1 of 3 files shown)


.envrc
```diff
@@ -0,0 +1 @@
+. .venv/bin/activate
```
jordangov: Is this the same as `source`? If so can we use `source`? It makes it easier to read and understand by anyone that isn't a shell scripter.

briandds:
```
$ help .
.: . filename [arguments]
    Read and execute commands from FILENAME and return.  The pathnames
    in $PATH are used to find the directory containing FILENAME.  If any
    ARGUMENTS are supplied, they become the positional parameters when
    FILENAME is executed.
$ help source
source: source filename [arguments]
    Read and execute commands from FILENAME and return.  The pathnames
    in $PATH are used to find the directory containing FILENAME.  If any
    ARGUMENTS are supplied, they become the positional parameters when
    FILENAME is executed.
$
```

They are the same, but okay I've changed it.

PR #20, opened 2018-06-04
Removed requirements.txt in favor of Pipenv

richard-dds:
I converted this repo over from pip and requirements.txt to `pipenv`, at Brian's request. This involves a couple new files and a few tweaks to the scripts.

The justification for using pipenv is that it's now the recommended way of handling python dependencies, according to [python.org](https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies). It has dev packages and a lock file, similar to `yarn` and `mix`. It's also a pretty nice developer experience in my opinion.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,1 @@
-source .venv/bin/activate
+layout virtualenv $(pipenv --venv)
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ scss/assets
 .venv/
 __pycache__
 *.pyc
+.cache/
 
 # Ignore static/fonts for now, since it is just symlink
 static/fonts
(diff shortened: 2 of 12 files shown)


ddsdevon:
@rthoward This PR needs a description. Probably worth throwing the justification for the switch in there...

richard-dds:
@ddsdevon Sorry about that! Added a description.

ddsdevon:
Looks fine to me, but you can fit what I know about Python into a shoebox. For baby shoes. So I'll defer to @briandds or @jordangov or @kristiandds or @dandds or... whomever for actual approval :)

jordangov:
Looks good, but can we update the README file to remove the bit about `source .venv/bin/activate` since it's not needed any more?

richard-dds:
@jordangov Good catch. Replaced that part of the README.

PR #30, opened 2018-06-12
Basic authentication

dandds:
This adds a basic integration with authnid and protects internal routes within the app. The happy path workflow is:

- User visits ATST home/login screen
- User clicks "Sign in with CAC"
- User gets certificate prompt, etc.
- If successfully authenticated in authnid, they are redirected to ATST `/login?bearer-token=[token]`
- ATST validates the bearer token with authnid and, if valid, creates a session for the user

Right now the ATST session is just a dummy cookie. Session management will be its own story.

**Note** you don't actually have to authenticate when the server is running in development. You can either visit `/login-dev` or click the Monty Python gif. This route is only available in dev and test environments.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,3 +1,4 @@
+import re
 import pytest
 import tornado.web
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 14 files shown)


dandds:
NB: this isn't deployed yet, so you'll still see the old behavior on the staging sites at the moment.

tests/test_auth.py
```diff
@@ -0,0 +1,53 @@
...
+            follow_redirects=False,
+            raise_error=False
+        )
+        assert response.headers["Set-Cookie"].startswith("atst")
+        assert response.headers['Location'] == '/home'
+        assert response.code == 302
+
+
+@pytest.mark.gen_test
+def test_login_with_via_dev_endpoint(app, monkeypatch, http_client, base_url):
```
richard-dds: I don't think the `monkeypatch` fixture is used here. You should be able to remove it. Same goes for the next test on line 47.

dandds: Yup, good call.

PR #34, opened 2018-06-18
Reap child processes

briandds:
Terminate the child process upon receiving a TERM or INT.

### DIFF
--- a/script/server
+++ b/script/server
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+reap() {
+   kill -TERM $child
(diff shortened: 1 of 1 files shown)


script/server
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+reap() {
+   kill -TERM $child
```
dcorrigan: Do you need to set the `$child` below? This is still leaving the orphan when I try it locally.

briandds: D'oh, yep.

PR #43, opened 2018-06-26
Use black to autoformat python code

richard-dds:


jordangov:
So many double quotes...

PR #44, opened 2018-06-28
Session management

dandds:
This was mostly @richard-dds , I'm just tying off the PR.

This adds session management to ATST using Redis. Session management is simple:

- When a user logs in with a valid bearer-token, ATST validates the token against authnid. If the token is valid, authnid will send back JSON user data.
- ATST creates a random UUID for the session ID. We add the session ID to Redis and set its value to the JSON user data.
- We set the session ID as a secure cookie on the client side.
- TTL for the session is configurable and set to an hour by default right now. Every time the user takes an action on a protected route in ATST, the TTL for their session is bumped to 60 minutes again. If they're inactive for an hour, they'll get the boot.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -9,6 +9,7 @@ webassets = "==0.12.1"
 Unipath = "==1.1"
 wtforms-tornado = "*"
 pendulum = "*"
+redis = "*"
 
 [dev-packages]
 pytest = "==3.6.0"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "5f91978d61968a720e4edc404ea36123ff96d857c1d7ff865f5aaef9b133db37"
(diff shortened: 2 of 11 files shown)


ddsdevon:
@dandds FYI, the TTL will need to be 10 minutes or less for production (its in the list of web app STIGs). Not sure if you want to bother adjusting it now and just making a note for later.

Also, is there a max session length? If not, we should make a ticket to implement one... after X minutes, we want to invalidate the session to force a fresh auth.

dandds:
@ddsdevon Okay, no harm making that the default.

I added a PT story for max session length [here](https://www.pivotaltracker.com/n/projects/2160940/stories/158697362).

richard-dds:
@ddsdevon @dandds Would it make sense to only have a static session length rather than bumping the TTL after each action? Seems like it could be a little simpler, but I'm not sure if it would pass your security requirements.

richard-dds:
@ddsdevon Made a ticket for the session expiration.

PR #46, opened 2018-06-29
Compile assets to versioned file for production

jgarberdds:
Run `pipenv run python ./precompile.py` to compile assets into the `static/assets` folder.

[https://www.pivotaltracker.com/story/show/157561578](https://www.pivotaltracker.com/story/show/157561578)


script/setup
```diff
@@ -25,5 +28,5 @@ fi
 script/bootstrap
 
 # Symlink uswds fonts into the /static directory
-rm -f ./static/fonts
+rm -rf ./static/fonts
```
ddsdevon: If it turns out static/fonts is a directory versus of the expected symlink, this would force nuke the directory and everything in it and put the symlink there instead... is that what we want? Or do we want these commands to complain because the local state isn't what we expect?

jgarberdds: Good question. I assumed it was just "whatever's there, replace it with this," but you're right to question that assumption. I'll put it back since it's not related.

ddsdevon:
This looks fine based on my limited understanding, but you'll probably want to snag an actual Python person.
It also looks like GitHub has flagged a merge conflict, so you might need to close this, rebase, and open a new PR.

PR #47, opened 2018-07-02
Make using a pipenv environment optional

ddsdevon:
Moving virtual environments between systems is problematic... and unnecessary when creating container images. These modifications allow the ability to install app dependencies into the system python, and use that instead of a virtual environment.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -9,13 +9,24 @@ set -e
 # Ensure we are in the app root directory (not the /script directory)
 cd "$(dirname "${0}")/.."
 
+if [ -z "${CIBUILD+xxxx}" ]; then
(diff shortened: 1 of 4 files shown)


briandds:
Looks basically okay -- but two questions:  Should this change be made to all the repos?  And should `script/setup` be changed to not install `pipenv`?

ddsdevon:
1 - Probably. I was going to role it into cibuild changes, if the concept itself seems solid
2 - Unfortunately, as far as I know, you still need pipenv to do the actual dependency install. It uses the --system flag to ensure the packages go into the main system Python. I could not find anything about using regular pip to process a Pipfile... though I might have missed it.

PR #48, opened 2018-07-02
CIbuild and publish

ddsdevon:
First whack at building and testing ATST in a docker image, and then creating a production ready image to push to our private container repository.

- Add docker files for creating tester and production ready images
- Add script/cibuild which the tester container runs by default
- Minor tweak to script/setup and script/bootstrap (move font symlink step to after npm install)
- Update Travis config to build and test in docker, and then push passing master images to our repo

### DIFF
--- a/docker/prod/Dockerfile
+++ b/docker/prod/Dockerfile
@@ -1,0 +1,40 @@
+FROM python:3.6.5-alpine
+
+# Overridable default config
+ARG APP_USER=atst
+ARG APP_GROUP=atat
+ARG APP_DIR=/opt/atat/atst
+ARG APP_PORT=8000
+ARG SITE_PACKAGES_DIR=/usr/local/lib/python3.6/site-packages
+
+ENV APP_DIR "${APP_DIR}"
+ENV SKIP_PIPENV true
+
+# Copy installed python packages from the tester image
+COPY --from=atst-tester:latest "${SITE_PACKAGES_DIR}" "${SITE_PACKAGES_DIR}"
+
+# Copy the app directory contents from the tester image (includes node modules)
+COPY --from=atst-tester:latest "${APP_DIR}" "${APP_DIR}"
+
+# Set working dir
+WORKDIR ${APP_DIR}
+
+# Add required system packages and app user
+RUN set -x ; \
+  script/alpine_setup "${APP_USER}" "${APP_GROUP}"
+
+# Update file ownership
+RUN set -x ; \
+  chown -R atst:atat "${APP_DIR}"
+
+# Set port to open
+EXPOSE "${APP_PORT}"
+
+# Run as the unprivileged APP user
+USER "${APP_USER}"
+
+# Use dumb-init for proper signal handling
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+# Default command is to launch the server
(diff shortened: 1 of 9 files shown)


script/cibuild
```diff
@@ -0,0 +1,16 @@
...
+
+# Ensure we are in the app root directory (not the /script directory)
+cd "$(dirname "${0}")/.."
+
+# Run lint/style checks and unit tests
+script/test
+
+# Run static code analysis security checks
+# (excluding the tests and node_modules subdirs)
+bandit -r . -x node_modules,tests
```
briandds: Does `bandit` need to be added to the dependencies someplace?

docker/prod/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+ENV SKIP_PIPENV true
+
+# Set port to open
+EXPOSE "${APP_PORT}"
+
+# Use dumb-init for proper signal handling
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+# Default command is to launch the server
+CMD ["bash", "-c", "${APP_DIR}/script/server"]
```
jordangov: I don't know Docker... can you tell me why the start up command is up here versus lower? Is this not procedural?

script/cibuild
jordangov: What will make this fail the CI build? Also, there is a nifty security check for node modules we could implement if we want: https://github.com/nodesecurity/nsp

jordangov: He added it to the atat-app-abuilder repo: https://github.com/dod-ccpo/atat-app-builder/blob/master/roles/atat.builder/defaults/main.yml

ddsdevon: It has been added to the CI atat-app-builder base image, and is currently only run by CI....
Alternatively, we could add it as a dev package maybe?

ddsdevon: Finding vulnerabilities will fail the CI build currently. We can adjust it to only be moderate and above, or something like that if we want. Currently there are 0, so I didn't bother to hedge at all 😸 

For the node_modules, we can certainly add something. Obviously the Python checker isn't the way to go.
We will also have to decide what happens if we find something... it is all upstream code since we are not writing any ourselves. Retesting other peoples code can get a bit sticky, so we'll want a defined mitigation strat probably.

docker/prod/Dockerfile
ddsdevon: Yes... sort of? This just sets the command that the docker image will run by default after being booted up in a container. So when it gets defined doesn't really affect the build of the image itself.

It is my current understanding that each step in a Dockerfile can result in a layer. Each layer can be cached and re-used when rebuilding the image, if there have not been any changes to it or its parent. So, I tried to put everything that will hardly ever change before the stuff that will change very frequently. I think it should allow Docker to be able to re-use cached versions of the early layers fairly often. Once you have a cache miss for a layer, that layer and all the ones after it have to be built fresh.

jordangov: Interesting... ok then. This is fine

script/cibuild
briandds: Yeah, I think so -- if it's a prerequisite for deployment then I think the test should happen sooner, like during testing on travis.

PR #55, opened 2018-07-06
Ui/panels and layout

andrewdds:
This PR sets up a global layout system with simple flexbox-powered Row/Columns.

It also adds a basic .panel component which provides the basis for styled blocks, as seen in various UI components.

Also:
- A styleguide page for demonstrating and testing out UI components (`/styleguide`)
- Some simple documentation of the rows/cols in the style readme

screen shot 2018-07-09 at 9 26 06 am


scss/core/_grid.scss
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  &.row--max {
+    max-width: $site-max-width;
+  }
+}
+
+.col {
+  &.col--pad {
+    @inclue grid-pad;
```
briandds: typo?

PR #62, opened 2018-07-10
Styleguide updates

luisgov:
This PR applies the new grid layout to the request form and adds some examples on headings for the styleguide page.

### DIFF
--- a/templates/styleguide.html.to
+++ b/templates/styleguide.html.to
@@ -1,47 +1,82 @@
 {% extends "base.html.to" %}
 
 {% block sidenav %}
-  <nav class='sidenav'>
-    <ul>
-      <li>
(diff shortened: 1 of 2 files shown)


templates/styleguide.html.to
```diff
@@ -1,47 +1,82 @@
...
+  <div class='panel'>
+    <p>This is a panel</p>
   </div>
+</div>
 
-  <div class='col col--grow'>
-    <div class='panel'>
-      <p>This is a panel</p>
+<div class='col col--grow'>
+  <div class='panel panel__content'>
```
andrewdds: `.panel__content` should be a child of `.panel`, no?

luisgov: ya. changed.

PR #64, opened 2018-07-10
Script standardization and shared functions

ddsdevon:
This is phase 1 of reworking and standardizing all the scripts. Phase 2 will be pulling everything in the /script/include directory over to the scriptz repository, and then updating all of the other apps.

My understanding of the relevant items that came out of the tech discussion prompting this change:
- A virtual environment should always be used
- The virtual environment should be installed to APPDIR/.venv
- The base app scripts focus solely on setting up and running the app in the local environment
- Docker and Kubernetes files should live under /deploy
- All of the apps should use the same set of base functions, setting app specific variables or having alternate logic only when necessary


script/include/run_setup
```diff
@@ -0,0 +1,32 @@
...
+if [ -z "${INSTALL_SASS+is_set}" ]; then
+  INSTALL_SASS="false"
+fi
+
+## Main
+# Remove any existing node modules as part of initial app setup or reset
+rm -rf ./node_modules
+
+if [ "${CREATE_VENV}" = "true" ]; then
+  install_pipenv
```
briandds: I think installing pipenv should be out of scope for the scripts and we should leave this out -- we've had a lot of trouble automating a pipenv install in a reliable and portable way, so I think better just to leave pipenv as an external dependency, and put instructions in the README (or code in the Dockerfile) that is specific to the environment in which pipenv will be running.

ddsdevon: Can we just check to see if it is running and then error out if not present?
The scripts assume it is installed and functional, so if we are not installing ourselves we should probably at least check that.

briandds:
Oh -- I just realized -- one change that we need is to add `pipenv run pip install --upgrade pip` to `script/setup`.  We're seeing cases where this fixes issues that occur when the system pip is too old.  (@andrewcroce )

ddsdevon:
@briandds Sometimes upgrading pip can also break things though... didn't updating to pip 10 end up breaking a number of things for people?
Maybe we can do a minimum version check instead?
Or maybe the install could focus on just a specific major revision, requiring a code change to bump that up? That way it is a conscious choice when to go from pip 10.x to 11.x or whatever, and someone can test that all the app packages install properly before making the change.

briandds:
> Or maybe the install could focus on just a specific major revision, requiring a code change to bump that up? That way it is a conscious choice when to go from pip 10.x to 11.x or whatever, and someone can test that all the app packages install properly before making the change.

👍 Sure -- specifying a version sounds good.

ddsdevon:
@briandds Pushed the change. Includes modifications to:
script/include/helper_functions.inc.sh
script/include/run_setup

briandds:
```

**Comment by briandds** [comment]
```

briandds:
```
$ script/setup
ERROR: pipenv is required but is not present
$ which pipenv
/usr/local/bin/pipenv
$
```
Something is not right...

PR #65, opened 2018-07-10
Update request form styles

luisgov:
This PR adds a sidebar to the requests page. It also addresses some issues with the language on the form.


atst/forms/financial.py
```diff
@@ -59,29 +59,29 @@ class FinancialForm(Form):
     )
 
     clin_0001 = StringField(
-        "CLIN 0001 - Unclassified IaaS and PaaS Amount", validators=[Required()]
+        "<b>CLIN 0001</b> - Unclassified IaaS and PaaS Amount", validators=[Required()]
```
andrewdds: @luisgov what is the purpose of the `<b>` element here? If this is a term being defined by the text that follows, this should be a `<dl>`.

luisgov: @andrewcroce this is just stylistic change to bold that part. I don't think that sentences is a definition.

PR #66, opened 2018-07-11
Add some tests for form validators

patricksmithdds:
This PR adds some tests for the custom form validators and also adds `pytest-watch` dev dependency to allow running tests after any file is changed.


richard-dds:
Looks good! Thanks for adding some much-needed tests.

PR #67, opened 2018-07-11
Switch /script/include to a submodule

ddsdevon:
This PR contains phase 2 of the shared script rework effort. Everything that was in /script/include is now pulled in from scriptz via a git submodule. The README has been updated to reflect this change.

The scripts in /script now set ATST specific configuration variables and then call the shared scripts.


briandds:
sigh

```
script/setup
Creating virtual environment using Python version 3.6...
Creating a virtualenv for this project…
Using /usr/local/bin/python3.6m (3.6.5) to create virtualenv…
⠋Warning: it looks like virtualenv is not in your PATH. We cannot continue until this is resolved.
```

yet

```
$  which virtualenv
/usr/local/bin/virtualenv
```

This works fine:

```
 pipenv --three
Creating a virtualenv for this project…
Using /usr/local/bin/python3 (3.6.5) to create virtualenv…
⠋Running virtualenv with interpreter /usr/local/bin/python3
Using base prefix '/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6'
New python executable in /Users/bduggan/.local/share/virtualenvs/new-Q5nM7WSd/bin/python3.6
Also creating executable in /Users/bduggan/.local/share/virtualenvs/new-Q5nM7WSd/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /Users/bduggan/.local/share/virtualenvs/new-Q5nM7WSd
Creating a Pipfile for this project…
```

But when I set `PIPENV_VENV_IN_PROJECT=True` it does not work.

ddsdevon:
@briandds Do you have any legacy .venv related things maybe? I'm not sure why it wouldn't work, but it sounds like a local issue if you cannot get the flag to work manually.... does it work in the current master branch, where the script is not a submodule?

README.md
```diff
@@ -3,19 +3,46 @@
...
+requirements and guidelines for all ATAT applications.
+Additionally, ATST requires a redis instance for session management. Have redis 
+installed and running. By default, ATST will try to connect to a redis instance 
+running on localhost on its default port, 6379.
+
+### Cloning
+This project contains git submodules. Here is an example clone command that will 
+automatically initialize and update those modules:
+
+    git clone --recurse-submodules git@github.com:dod-ccpo/atst.git
```
patricksmithdds: For an existing clone, I had to run:

```
    git submodule update --init
```

in order to fetch the submodule. It wasn't obvious to me how to get it working, so I think we should add this command to README as well.

ddsdevon: @patricksmithdds I added something for now, but I do not think this kind of information would normally go in the README. This seems more like the type of information that would go into changelog notes and upgrade/version migration type of documentation (since it would only affect users going from an older version to this one).

ddsdevon:
Just pushed a rebase off current master

README.md
patricksmithdds: @ddsdevon That makes sense -- thanks for adding it for now!

PR #68, opened 2018-07-12
Fix for broken layout breakpoint

andrewdds:
Fixes an issue where the panels would not wrap correctly between breakpoints.

| Before | After |
| -- | -- |
| breakpoint_broken | breakpoint_fixed |

Also, fixes a small issue with sidenav panel having double bottom-margin.


andrewdds:
To be fair, it was my bug, not USWDS.

PR #69, opened 2018-07-12
rename some handlers with more specificity

dandds:
Some small changes to handler names and one endpoint to better describe what they do:

- `Home` handler becomes `Root`, because it exclusively serves the app root route, not the `/home` route
- `Login` becomes `LoginRedirect`, and the `/login` route `/login-redirect`. This route does not present the user login page (the root route does that), but instead is the place users are redirected from authnid with a bearer token _after_ they've successfully authenticated.
- `MainHandler` to `Main` because it's the only handler with `Handler` in its name.

Let me know if you disagree with these.


dandds:
If this gets approved I'll make a tiny corresponding PR in authnid for the login redirect route.

atst/app.py
```diff
@@ -22,23 +22,23 @@
 def make_app(config, deps, **kwargs):
 
     routes = [
-        url(r"/", Home, {"page": "login"}, name="main"),
+        url(r"/", Root, {"page": "login"}, name="root"),
```
patricksmithdds: Should we rename the `login` template to `root` as well? I'd expect that template to be shown at the (non-existant) `/login` route.

dandds: I wasn't sure about that one. It _is_ the UI login, but it's also confusing that it's named that way. I'm okay with renaming it if you are.

One other possibility I thought of is to have a real `Login` handler and `/login` route. When a user hits the root route, we automatically redirect them to `/login`. I thought that was big lift for the moment though. 🤷‍♂️

patricksmithdds: I say rename to root and maybe do the redirect later, after we hammer out some of the public facing pages some more.

dandds: Sounds good. New commit.

PR #70, opened 2018-07-12
Add missing fields to financial verification step

patricksmithdds:
Adds the treasury code and ba code fields to the financial verification step and makes the PE number a simple string input instead of a multi-line input.

image


patricksmithdds:
Just re-ran the tests on CI and all is well :sparkles:

PR #76, opened 2018-07-17
Icons!

andrewdds:
This adds SVG icons.

Render icons by using `{% module Icon('name') %}` in a template, where `name` is the filename of an SVG file in `static/icons`.

By default, icons are 16x16 square. New variants and sizes will be added as needed.

SVG icons should come from the Noun Project, specifically [this collection](https://thenounproject.com/monstercritic/collection/tinicons-a-set-of-tiny-icons-perfect-for-ui-elemen/). Custom SVG can be created for specific icons, if needed.

Refer to `/styleguide` for a reference of the currently available icons. Please add to the reference as new icons are added.


atst/ui_modules.py
```diff
@@ -0,0 +1,7 @@
+from tornado.web import UIModule
+
+class Icon(UIModule):
+    def render(self, name, classes=''):
```
patricksmithdds: It looks like we're not adding the `classes` to the rendered element. Can we either add those classes or remove that argument here?

andrewdds: ah forgot to do that...

andrewdds: @patricksmithdds fixed

PR #79, opened 2018-07-19
WIP: Curl httpclient

dandds:
I'm putting this up mostly so the attempt is documented. Pycurl and pipenv don't play nice together. This adds pycurl to the `Pipfile` but excludes it on macos installations (nice feature, thanks pipenv). `script/bootstrap` tries a more specific macos install using pip directly in the pipenv shell. I think something like this worked for @ddsdevon , but my tests still don't pass even after I blow away my virtualenv and rebuild it. I'm open to suggestions if anyone has further ideas.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,7 +10,7 @@ Unipath = "==1.1"
 wtforms-tornado = "*"
 pendulum = "*"
 redis = "*"
-pycurl = "*"
+pycurl = {version = "*", sys_platform = "!= 'darwin'"}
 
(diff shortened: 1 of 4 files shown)


ddsdevon:
@dandds Not sure if it is work noting or not, but the install I did also requires pip 10.* be installed into the virtual environment.

dandds:
@ddsdevon Yup, I've got pip 10.0.1 in my virtualenv.

PR #80, opened 2018-07-19
Validate PE number

patricksmithdds:
This PR validates the PE ID entered in the financial verification step. When the PE id could not be found, the user is warned about it:

image

The user can then submit the form again to continue.

The PE number on the form is only checked when the value is changed. So, if a form were saved with an invalid PE number, the error will be shown only if the PE number was changed to a different, invalid value.

This PR also redirects to the `/requests` page after the financial verification step is submitted (previously, there was an error page, since there is no next step).

### DIFF
--- a/atst/forms/forms.py
+++ b/atst/forms/forms.py
@@ -3,5 +3,5 @@ from wtforms_tornado import Form
 
 class ValidatedForm(Form):
 
-    def validate(self, requests_client=None):
-        return super(ValidatedForm, self).validate()
+    def validate_warnings(self, requests_client=None):
+        return True
--- a/atst/forms/forms.py
+++ b/atst/forms/forms.py
@@ -1,7 +1,12 @@
+import tornado
+from tornado.gen import Return
 from wtforms_tornado import Form
 
 
 class ValidatedForm(Form):
 
-    def validate_warnings(self, *args, **kwargs):
-        return True
+    @tornado.gen.coroutine
+    def perform_extra_validation(self, *args, **kwargs):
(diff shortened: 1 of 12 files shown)


patricksmithdds:
@dandds 7f972f8 should fix that error, thanks for catching it!

atst/forms/forms.py
```diff
@@ -0,0 +1,12 @@
...
+from wtforms_tornado import Form
+
+
+class ValidatedForm(Form):
+
+    @tornado.gen.coroutine
+    def perform_extra_validation(self, *args, **kwargs):
+        """A coroutine that performs any applicable extra validation. Must
+        return True if the form is valid or False otherwise."""
+        raise Return(True)
```
richard-dds: Is there a particular reason you did `raise Return(True)` here instead of just `return True`? I think the `raise` method is only necessary for python2, but that seems like something you would have already tried.

patricksmithdds: I had hoped `return True` would work, but [according to the docs](http://www.tornadoweb.org/en/stable/guide/coroutines.html#native-vs-decorated-coroutines)
> Decorated coroutines always return a Future object

When I had `return True`, I was getting an error attempting to `yield` on the result of the call. As far as I can tell, `raise Return(True)` is the necessary way to do it.

richard-dds: Yeah, I thought the point of the `@tornado.gen.coroutine` decorator was that it wrapped your return value in a future automatically. Weird that you were getting an error, though.

PR looks good though. :shipit:

PR #88, opened 2018-07-23
UI Requests

luisgov:
This PR adds a _New Request_ section to the sidebar navigation and updates the _Requests_ screen with the panel actions and a placeholder for an alert message. 

screen shot 2018-07-24 at 13 38 30


templates/navigation/global_navigation.html.to
```diff
@@ -5,6 +5,7 @@
     {% end %}
 
     {% module SidenavItem("Requests", href="/requests", icon="document", active=matchesPath('/requests')) %}
+    {% module SidenavItem("+ New Request", href="/requests/new", icon="", active=matchesPath('/requests/new')) %}
```
andrewcroce: Hmm. I think we need to figure a better way to do this:
1. This is not nested with "Requests" as its parent, as a sub `ul` item.
2. This should only appear when the parent "Requests" is active.

scss/elements/_panel_actions.scss
```diff
@@ -0,0 +1,33 @@
+/*
+ * Panels Actions
+ * Form inputs and other actions for panels
+ */
+
+ .panel__actions {
```
andrewcroce: I think the class names here are not properly descriptive, and the wrong bem syntax. This implies its the "actions" sub-component of a  "panel", which isn't right. Maybe ".list-actions" or ".list-tools" since they are acting upon lists.

scss/elements/_panel_actions.scss
```diff
@@ -0,0 +1,33 @@
...
+  display: flex;
+  padding: $gap;
```
andrewcroce: I would only apply this padding if its a child of a panel.

scss/elements/_panel_actions.scss
```diff
@@ -0,0 +1,33 @@
...
+
+  &--filter {
+    width: 50%;
+  }
+
+  &--search {
```
andrewcroce: These are not really modifiers, but sub-components.  And we probably shouldn't specify exact widths; what if there are 3 tools? or one? Flex layout should allow them to take up the room they require without explicit widths.

scss/elements/_tables.scss
```diff
@@ -16,6 +16,11 @@ table {
 
     th {
       @include block-list-header;
+
+      &:first-child {
+        padding: $gap * 2 !important; // This overrides USWDS's no padding
```
andrewcroce: This will be unnecessary if you remove the `usa-table-borderless` class. Thats not needed anymore.

scss/elements/_panel_actions.scss
andrewcroce: In fact, I'm not even sure this needs to be its own named component. Can we achieve the same thing with ordinary rows and columns?

PR #90, opened 2018-07-24
Suggest a PE number when PE number is not found

patricksmithdds:
Using the defined format of a PE number (https://www.militaryperiscope.com/terms/t0000241.html), we can suggest a corrected format of the PE number if the user omitted the leading zeros for either the DoD program or category.

Example:

image

### DIFF
--- a/tests/forms/test_financial.py
+++ b/tests/forms/test_financial.py
@@ -1,0 +1,14 @@
+import pytest
+
+from atst.forms.financial import suggest_pe_id
+
+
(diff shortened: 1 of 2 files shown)


tests/forms/test_financial.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+from atst.forms.financial import suggest_pe_id
+
+
+@pytest.mark.parametrize("input,expected", [
```
dandds: I didn't know about this feature. 😍

PR #101, opened 2018-07-27
Add filter dropdown to panel actions

luisgov:
This PR adds a filter dorpdown to the requests action bar panel.

screen shot 2018-07-27 at 10 47 42

### DIFF
--- a/scss/core/_grid.scss
+++ b/scss/core/_grid.scss
@@ -15,7 +15,7 @@
 }
 
 @mixin grid-pad {
-  @include padding($site-margins-mobile);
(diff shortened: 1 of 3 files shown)


scss/core/_grid.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 @mixin grid-pad {
-  @include padding(null $site-margins-mobile);
```
luisgov: @andrewcroce I removed this so that the whole row will have padding. You ok with this?

andrewdds: I dont get it.

PR #105, opened 2018-07-27
Workspace members

luisgov:
This PR adds fields to the members table. It also includes placeholders for search, filter by role, and filter by status functionality.

screen shot 2018-07-27 at 14 38 13


dandds:
@luisgov I'm seeing a small difference in the column header border:

screen shot 2018-07-30 at 10 56 41 am

luisgov:
@dandds thanks, good catch. Issue fixed.

PR #111, opened 2018-07-30
Ui/empty states

andrewdds:
Adds an EmptyState module, and some mock conditional template logic for displaying it on the Requests, WS Members, and WS Projects pages.

screen shot 2018-07-30 at 3 59 15 pm

screen shot 2018-07-30 at 4 05 19 pm

screen shot 2018-07-30 at 4 02 39 pm

### DIFF
--- a/templates/requests.html.to
+++ b/templates/requests.html.to
@@ -1,5 +1,9 @@
 {% extends "base.html.to" %}
 
+{% block template_vars %}
+  {% set requestsEmpty = False %}
+{% end %}
+
 {% block modal %}
   {% if modalOpen() %}
     {% apply modal %}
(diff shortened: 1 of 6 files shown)


templates/requests.html.to
```diff
@@ -31,67 +35,79 @@
 
 {% block content %}
 
+{% if requestsEmpty %}
```
dandds: I think a more python-ish(?) thing to do is just check if the array is empty in these cases:
```
{% if not requests %}
```
Empty lists evaluate to false in Python.

andrewdds: Right that makes sense. I was using a dummy var to demo the screen.

andrewdds:
@dandds fixed, please review again at your convenience

PR #116, opened 2018-07-31
Merge fundz

dandds:
This merges the fundz service and its accompanying domain logic into ATST.


script/ingest_pe_numbers.py
```diff
@@ -0,0 +1,40 @@
...
+from atst.models import PENumber
+
+
+def get_pe_numbers(url):
+    response = urlopen(url)
+    t = response.read().decode("utf-8")
+    return list(csv.reader(t.split("\r\n")))
+
+
+def insert_pe_numbers(db, list_of_pe_numbers):
```
richard-dds: Not urgent, but I think we should consider putting some of this logic into the PENumbers domain object.

PR #121, opened 2018-08-03
Cleanup

andrewdds:
Move `js` and `styles` out to the root
Reorganize macros into separate component files
Clean up a few other things


static/assets/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
patricksmithdds: Can we add this back and not have all the assets under source control?

static/assets/angle-arrow-down-primary.1c0886c7.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="512" height="512" viewBox="0 0 284.929 284.929"><path d="M282.082 76.51L267.808 62.24c-1.902-1.906-4.093-2.856-6.57-2.856-2.47 0-4.66.95-6.563 2.856L142.465 174.44 30.263 62.24c-1.903-1.905-4.093-2.855-6.567-2.855-2.475 0-4.665.95-6.567 2.856L2.856 76.516C.95 78.417 0 80.607 0 83.082c0 2.473.953 4.663 2.856 6.565L135.9 222.693c1.9 1.903 4.092 2.854 6.566 2.854s4.66-.95 6.562-2.854L282.082 89.647c1.902-1.903 2.847-4.093 2.847-6.565 0-2.475-.946-4.665-2.848-6.57z" fill="#0071bc"/></svg>
```
patricksmithdds: @andrewdds All these assets need to be removed after the gitignore was added back -- we shouldn't be tracking all these assets.

PR #125, opened 2018-08-03
Add Parcel front end stuff

andrewdds:


script/ingest_pe_numbers.py
```diff
@@ -7,7 +7,7 @@
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
 sys.path.append(parent_dir)
 
-from atst.app import make_deps, make_config
+from atst.app import make_app, make_config
```
dandds: Are these changes to `ingest_pe_numbers.py` in this PR on purpose?

js/thing.js
```diff
@@ -0,0 +1,7 @@
+console.log('hanlo again')
```
dandds: Are we leaving this in on purpose for demo purposes?

script/ingest_pe_numbers.py
patricksmithdds: Yeah, this branch has some of my changes to get CI working and the changes are necessary to get that script working again.

PR #126, opened 2018-08-03
testing out vue stuff

andrewdds:


andrewdds:
@patricksmithdds thats ok I think, we don't have too many modal. We can rethink it if we need to.

dandds:
It seems like Vue is intercepting all button-click events (or something like that). I can't click past the first page of the requests form anymore.

PR #129, opened 2018-08-06
Authnid

dandds:
This merges authnid's functionality into ATST and adds authorization protection to all endpoints (thanks @richard-dds ). The checks necessary for CAC client auth and CRL checking are implemented.

We didn't duplicate the bearer-token generation logic from authnid since that's not part of the monolith authentication workflow anymore.


jordangov:
Can you point out any changes that are _not_ copies of files from the other repo? Basically, what was needed to make this merge or what should we need to look at closely?

dandds:
@jordangov Sure. You could review the files related to login and auth checks:

```
atst/app.py (set up Redis session, CRL validator)
atst/domain/auth.py (new auth decorator)
atst/routes/__init__.py (the login-redirect endpoint)
```

The user model had to be expanded for the authnid user data and we now interact with it through the Users repository (`atst/domain/users.py`). 

Most of the authentication tests were re-written to match the new authentication workflow. Notably, there's a "test_routes_are_protected" test that confirms that every endpoint we'd expect to be under auth is actually under auth.

atst/routes/requests/financial_verification.py
```diff
@@ -4,9 +4,11 @@
 from . import requests_bp
 from atst.domain.requests import Requests
 from atst.forms.financial import FinancialForm
+from atst.domain.auth import login_required
 
 
 @requests_bp.route("/requests/verify/<string:request_id>", methods=["GET"])
+@login_required
```
patricksmithdds: I believe _all_ the requests routes require a login. Could we use `@requests_bp.before_request` ([for example](https://stackoverflow.com/a/13451233)) to enforce the login requirement instead of decorating each route?

dandds: Oh yeah, d'oh. That makes sense. We can probably do that on the workspaces blueprint too.

atst/routes/dev.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@bp.route("/login-dev")
+def login_dev():
+    role = request.args.get("username", "amanda")
+    user_data = _DEV_USERS[role]
+    basic_data = {k:v for k,v in user_data.items() if k not in ["dod_id", "atat_role"]}
+    user = _set_user_permissions(user_data["dod_id"], user_data["atat_role"], basic_data)
+    session["user_id"] = user.id
+    app.logger.warning(session)
```
richard-dds: We should probably remove this log message, or at least change the severity.

templates/unauthorized.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
```
patricksmithdds: The `base.html` template includes all the logged-in navigation. Can we update this so the unauthorized page doesn't render any of that, since the user won't be logged in.

tests/domain/authnid/test_utils.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+import atst.domain.authnid.utils as utils
+
+DOD_SDN_INFO = {
```
patricksmithdds: I think we should re-use the `DOD_SDN_INFO` added in `tests/mocks.py` here.

atst/routes/dev.py
dandds: 👍

atst/routes/requests/financial_verification.py
dandds: After thinking about it a little, I think it makes sense to apply auth to all the routes and check if the request endpoint is one of the four exceptions that don't need auth (the sign in page, the login-redirect, login-dev, and the unauthorized page).

atst/domain/auth.py
```diff
@@ -0,0 +1,32 @@
+from flask import g, redirect, url_for, session, request
+
+from atst.domain.users import Users
+
+
+UNPROTECTED_ROUTES = ["atst.root", "dev.login_dev", "atst.login_redirect", "atst.unauthorized", "static"]
```
jordangov: Hmm... I could see a scenario where some static assets need to be behind auth. What's our plan for that? I'm thinking about diagrams/images that show architecture or CCPO structure or... I dunno... something.

atst/domain/users.py
```diff
@@ -1,54 +1,65 @@
...
+        try:
+            user = db.session.query(User).filter_by(dod_id=dod_id).one()
+        except NoResultFound:
+            raise NotFoundError("user")
+
+        return user
+
+    @classmethod
+    def create(cls, atat_role_name="developer", **kwargs):
+        atat_role = Roles.get(atat_role_name)
```
jordangov: What happens if a Role is not found?

atst/domain/users.py
```diff
@@ -1,54 +1,65 @@
...
 
         try:
-            user = User(id=user_id, atat_role=atat_role)
-            self.db_session.add(user)
-            self.db_session.commit()
+            user = User(atat_role=atat_role, **kwargs)
+            db.session.add(user)
+            db.session.commit()
         except IntegrityError:
             raise AlreadyExistsError("user")
```
jordangov: Will ALL IntegrityErrors be because the user already exists? Is there no other integrity checks?

atst/domain/users.py
```diff
@@ -1,54 +1,65 @@
...
+
+    @classmethod
+    def create(cls, atat_role_name="developer", **kwargs):
+        atat_role = Roles.get(atat_role_name)
 
         try:
-            user = User(id=user_id, atat_role=atat_role)
-            self.db_session.add(user)
-            self.db_session.commit()
+            user = User(atat_role=atat_role, **kwargs)
```
jordangov: Maybe I just don't understand how Python works, but this seems like a security hole. If we pass in all args, then could we override a value we don't really want? I guess my question is... are we just doing this because it's easier,  should we be more explicit with each arg?

atst/domain/users.py
```diff
@@ -1,54 +1,65 @@
...
+    @classmethod
+    def get_or_create_by_dod_id(cls, dod_id, **kwargs):
         try:
-            user = self.get(user_id)
+            user = Users.get_by_dod_id(dod_id)
         except NotFoundError:
-            user = self.create(user_id, *args, **kwargs)
-            self.db_session.add(user)
-            self.db_session.commit()
+            user = Users.create(dod_id=dod_id, **kwargs)
```
jordangov: See my comment about about creating with args passed in...

atst/domain/auth.py
dandds: That's a fair point. I don't think the app will be responsible for serving static assets in production anyway (NGINX will), so off the top of my head, maybe protected assets will need to live in a different directory.

jordangov: ok

atst/domain/users.py
dandds: `get_or_create_by_dod_id` only gets called in the event someone logs in with CAC, in which case we parse the CAC info and assign it to a Python dictionary with hard-coded keys. That dictionary is what gets passed here. (`Users.create` isn't called in production code, it just breaks out the more basic functionality from `get_or_create_by_dod_id`).

jordangov: Ok, so the dictionary can only contain what we've mapped on the CAC? In that case why do we call it kwargs? Is this a convention? Why not call it `cacdict` or something that identifies it? I know... this is small potatoes now that I understand what you're doing, but someone looking at this (open source) code may have the same reaction as me...

atst/domain/users.py
dandds: For now, yes. They'll all be because some unique constraint was broken (DOD ID already exists, email already exists). We could get more fine-grained if need be.

atst/domain/users.py
dandds: Right. `kwargs` is Python convention (keyword arguments). We can be more specific.

atst/domain/users.py
jordangov: ok

atst/domain/users.py
dandds: For now, the only context this gets called in is when a user first signs in and we give them a "developer" role by default. I don't think there will be a case where it can receive an arbitrary string value (`atat_role_name="foo"`) during user creation.

jordangov: ok

PR #132, opened 2018-08-06
UI/requests kpi

luisgov:
This PR adds support for KPI (Key Performance Indicators) for the request form.

screen shot 2018-08-06 at 14 21 37


patricksmithdds:
> I think Trigget Travis was a country singer, right?

I think he did a duet with Sam Seeceepio in the 80s, too

PR #136, opened 2018-08-07
implement database sessions for test factories

dandds:
This sets a database session for every test factory so that each factory's `create` method will automatically commit the entry.


patricksmithdds:
🎉 tests look so much simpler now, thanks!

PR #137, opened 2018-08-07
Ui/field validation setup

andrewdds:


js/components/text_input.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+    value: {
+      type: String,
+      default: () => ''
+    }
+  },
+
+  data: function () {
+    return {
+      valid: false,
```
patricksmithdds: Seems like this isn't actually used anywhere -- can we remove it and just use `showValid`?

andrewdds: good point

templates/styleguide.html
```diff
@@ -130,15 +130,29 @@ <h6>Heading H6</h6>
...
+              v-on:input='onInput'
+              v-on:change='onChange'
+              v-model='renderedValue'
+              id='basic-text-1'
+              type='text'
+              ref='input'
+              placeholder='$'
+              :guide="true"
+              :mask="mask">
+            </masked-input>
```
patricksmithdds: Will this setup work when we have multiple masked inputs on the page?

andrewdds: Yes I tested it out and it should. Some of the attributes there will need to be parameterized to the component, such as the placeholder, etc.

My worry was having multiple `inline-template` declarations, but Vue doesn't seem to care. Something to keep in mind incase we see weirdness.

PR #138, opened 2018-08-07
159414499 request auth

dandds:
This is a blunt implementation of authorization logic for viewing requests ([PT story](https://www.pivotaltracker.com/story/show/159414499)).

Once we have authorization in a few places, I think we'll be able to generalize and refactor things into a policy pattern.

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -15,6 +15,7 @@ from atst.routes import bp
 from atst.routes.workspaces import bp as workspace_routes
 from atst.routes.requests import requests_bp
 from atst.routes.dev import bp as dev_routes
+from atst.routes.errors import make_error_pages
 from atst.domain.authnid.crl.validator import Validator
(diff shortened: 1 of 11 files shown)


dandds:
Going to make some quick changes based on changes to the PT story.

dandds:
Okay, I updated this with the following changes:

- We let exceptions bubble and add Flask `errorhandler` decorators to catch them and return appropriate HTTP codes/pages to the user. For instance, any uncaught `NotFoundError` raised by a domain module anywhere in the codebase will now return a 404 to the user.
- When a user is not authorized to view a request, we return a 404 so that we don't leak valid request IDs to unauthorized users, per the changes to the PT card.

dandds:
@patricksmithdds Thanks, see the new commits.

dandds:
@richard-dds added a commit for the status event factory thing we talked about.

PR #139, opened 2018-08-07
Update request status names to reflect document

richard-dds:
This also includes some refactoring to move the valid request statuses into an enum.

### DIFF
--- a/atst/domain/requests.py
+++ b/atst/domain/requests.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from sqlalchemy import exists, and_
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.attributes import flag_modified
@@ -26,10 +27,11 @@ def deep_merge(source, destination: dict):
     return _deep_merge(source, dict(destination))
 
 
-class RequestStatuses(object):
-    @classmethod
-    def new(cls, status_name):
-        return RequestStatusEvent(new_status=status_name)
+class RequestStatus(Enum):
+    STARTED = "started"
+    PENDING_FINANCIAL_VERIFICATION = "pending_financial_verification"
+    PENDING_CCPO_APPROVAL = "pending_ccpo_approval"
+    APPROVED = "approved"
 
 
 class Requests(object):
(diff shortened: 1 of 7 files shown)


atst/models/request.py
```diff
@@ -22,11 +22,6 @@ class Request(Base):
...
 
-    @property
-    def action_required_by(self):
-        return {
-            "incomplete": "mission_owner",
-            "pending_submission": "mission_owner",
-            "submitted": "ccpo",
-            "approved": "mission_owner",
-        }.get(self.status)
+    def set_status(self, status):
```
patricksmithdds: Should `status` be `status_event` or are we expecting it to be a `RequestStatus` object?

atst/domain/requests.py
```diff
@@ -100,18 +102,27 @@ def update(cls, request_id, request_delta):
...
         # since it doesn't track dictionary mutations by default.
         flag_modified(request, "body")
 
         db.session.add(request)
         db.session.commit()
 
+    @classmethod
+    def set_status(cls, request: Request, status: RequestStatus):
+        status_event = RequestStatusEvent(new_status=status)
+        request.status_events.append(status_event)
```
patricksmithdds: Should we use `request.set_status` here?

atst/models/request.py
richard-dds: I removed this helper method since it was unused (because I didn't like it). The `Requests.set_status` classmethod has a type annotation to remove any confusion.

atst/models/request_status_event.py
```diff
@@ -1,17 +1,25 @@
...
 
 from atst.models import Base
 from atst.models.types import Id
 
 
+class RequestStatus(Enum):
+    STARTED = auto()
+    PENDING_FINANCIAL_VERIFICATION = auto()
+    PENDING_CCPO_APPROVAL = auto()
+    APPROVED = auto()
```
patricksmithdds: Can we add the other states from the doc -- "expired" and "deleted"? I know we're not using them yet, but it'll be nice to have them encoded here.

atst/models/request_status_event.py
```diff
@@ -1,17 +1,25 @@
...
+from sqlalchemy.types import DateTime, BigInteger
 from sqlalchemy.schema import Sequence
 from sqlalchemy.dialects.postgresql import UUID
 
 from atst.models import Base
 from atst.models.types import Id
 
 
+class RequestStatus(Enum):
+    STARTED = auto()
```
patricksmithdds: I've not used this `auto` before -- it seems kinda magical. Would it be safer to manually define the values here to a string representation of the status?

PR #140, opened 2018-08-08
Prod env

dandds:
Small update to do things:

- regularize the use of "prod" for the production environment, since we use "dev" for development
- add a config value for session cookies in production (so if a cookie is set on a subdomain of "atat.codes", it's available to other instances of the host domain)


dandds:
@patricksmithdds I don't think so; I can't find any reference to FLASK_ENV in scriptz. 

I just alphabetized the base config on this PR to make it more searchable, too, fyi.

PR #141, opened 2018-08-08
Kube deploy

ddsdevon:
This PR... does a lot of things. First the non-kubernetes stuff:
- Remove .venv from dockerignore so that the virtual environment gets properly copied from the testing container to the final one
- Switch final container base from python:3.6.5-alpine to alpine:3.8 and use Alpine's python3 package, due to uwsgi installation dependency issues
- Add copying of `/usr/bin` contents from the testing container, so the wrappers for things like pipenv make it over
- Add a `script/uwsgi_server` script and use that as the final container entry point
- Update `script/alpine_setup` so it adds postgres client libraries, python3, and uwsgi to the list of packages to install

New Kubernetes related files:
- `atst-envvars-configmap.yml`: Defines a configmap containing all of the environment variables that will be defined in all containers
- `atst-configmap.yml`: Defines a configmap for app related items (just the uwsgi ini for now)
- `atst-nginx-configmap.yml`: Defines a configmap for nginx related items (naming the virtual host config)
- `atst.yml`: Defines the atat namespace, app deployment, app services, and the app ingress... its "The Big One"
- `set_X_secret.sh`: Handy reference scripts for managing the items stored as Kubernetes secrets, the contents of which are not to be stored in the code base


deploy/kubernetes/atst-nginx-configmap.yml
```diff
@@ -0,0 +1,76 @@
...
+            include uwsgi_params;
+            uwsgi_pass unix:///var/run/uwsgi/uwsgi.socket;
+        }
+        location @appbasicauth {
+            include uwsgi_params;
+            uwsgi_pass unix:///var/run/uwsgi/uwsgi.socket;
+            auth_basic "Developer Access";
+            auth_basic_user_file /etc/nginx/.htpasswd;
+        }
+    }
```
dandds: Do we want the atat.codes server config to redirect any requests to `/login-redirect` on that host? I don't think it's a big deal for now, but the app doesn't have any internal handling for those cases yet.

deploy/kubernetes/atst-nginx-configmap.yml
```diff
@@ -0,0 +1,76 @@
...
+    server {
+        server_name auth.atat.codes;
```
dandds: We probably need to set the `CAC_URL` config value to something like `http://auth.atat.codes/login-redirect` for the relevant environment.

ddsdevon: Yep, that part is done.... You can't see it because it is part of atst-overrides.ini, and that whole file is currently a secret due to the SECRET related entries.

deploy/kubernetes/atst-nginx-configmap.yml
ddsdevon: Yes, I'll add that.
We should probably also add at least a basic catch in the app... maybe if uwsgi is being used, autoredirect if the HTTP_X_SSL_CLIENT_CERT header isn't defined? Or just add the hostname related checks. I suppose that isn't much harder.

deploy/kubernetes/atst-nginx-configmap.yml
dandds: 👍

PR #143, opened 2018-08-08
Add migrate db before server startup

ddsdevon:
This PR adds a migrate step before launching the UWSGI server.
This ensures a fresh deploy properly updates the DB before starting the
server and accepting traffic

### DIFF
--- a/script/uwsgi_server
+++ b/script/uwsgi_server
@@ -4,5 +4,8 @@
 
 source "$(dirname "${0}")"/../script/include/global_header.inc.sh
 
+# Before starting the server, apply any pending migrations to the DB
+migrate_db
+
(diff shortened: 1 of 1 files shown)


patricksmithdds:
Should we be running `script/update` (which I believe runs the migrate command) before the `script/uwsgi_server` script? I think we'd need to do that to install any new dependencies anyway, right?

ddsdevon:
@patricksmithdds A fresh deploy of container with a new code version comes with all of the dependencies already installed, since that is part of the build process during CI. This extra step takes care of updating the database associated with whatever environment the container is deployed to.

Down the line, this might become more of a pre-deploy/sidecar type of task, but this should cover us for now.

PR #144, opened 2018-08-08
Fix fields in requests listing

richard-dds:
- Show actual username in "Requestor" column
- Display human-readable statuses
- Display total apps (from `num_software_systems` in the request)

Also included some refactoring (sorry):
- Rename Request.creator_id to user_id for consistency
- Created `seed.py`, which can be used in dev to populate users and requests
- Fixed some migrations that were preventing a full downgrade
- Now populate Request.body in the RequestsFactory


patricksmithdds:
> Created seed.py, which can be used in dev to populate users and requests

😍

richard-dds:
@dandds Could definitely use your review here since I had a non-trivial merge with some of your changes.

PR #145, opened 2018-08-08
UI/request form labels

luisgov:
This PR removes labels from choice inputs (radio buttons) and style the legend the same way as input text fields.

### DIFF
--- a/styles/elements/_inputs.scss
+++ b/styles/elements/_inputs.scss
@@ -100,6 +100,10 @@
       padding: 0 0 $gap 0;
       @include h4;
 
+      label {
+        font-weight: $font-bold;
+      }
+
       .icon {
         vertical-align: middle;
       }
--- a/styles/elements/_inputs.scss
+++ b/styles/elements/_inputs.scss
@@ -65,19 +65,11 @@
   }
 
   label {
-    padding: 0 0 $gap 0;
+    padding: 0;
     margin: 0;
     @include h4;
     @include line-max;
     position: relative;
-
-    .usa-input__help {
-      display: block;
-      @include h5;
-      font-weight: normal;
-      padding-top: $gap / 2;
-      @include line-max;
-    }
 
     .icon {
       position: absolute;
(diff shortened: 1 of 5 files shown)


templates/components/text_input.html
```diff
@@ -1,6 +1,5 @@
 {% macro TextInput(field, placeholder='') -%}
   <div class='usa-input {% if errors %}usa-input--error{% endif %}'>
-    <label for={{field.name}}>
```
andrewdds: Im not sure CSS is sufficient for fixing this. The `field.description` no longer has any accessible association with the input. By being inside the label, it had this association. We'll need to add some aria attributes to create that association, `labeledby` or some such thing. Same for the error messages, now that I think of it.

Also, did you check what happens to the icon positioning? It was absolutely positioned relative to the parent label, which it no longer has as a parent. I imagine it lost its positioning context and might be positioned somewhere weird now.

andrewdds:
Also @luisgov could your rebase this? I'm having trouble running it since its so far behind master

styles/elements/_inputs.scss
```diff
@@ -100,9 +100,17 @@
...
+      label {
+        font-weight: $font-bold;
+      }
+
       .icon {
         vertical-align: middle;
       }
+
+  .usa-input__help {
+    font-weight: $font-bold;
```
andrewdds: Shouldn't the help text be normal weight?

andrewdds:
I think a lot of these CSS changes can be avoided if we simply strip the HTML from the `{{ field.label }}` tag from WTForms. 

Looks like jinja already has a `striptags` filter that might do the trick: `{{ field.label | striptags }}`

styles/elements/_inputs.scss
luisgov: 👍

luisgov:
@andrewcroce, all set.

screen shot 2018-08-09 at 10 26 46

PR #147, opened 2018-08-08
Ui/input field frontend validation

andrewdds:
This adds the vue validation to text_input jinja macro.
Adds a few more field validation types (dollars, date, integer, gigabytes)
Adds these validations to request form page 1


patricksmithdds:
@andrewdds should we close or merge #137 before this one gets merged?

templates/requests/screen-2.html
```diff
@@ -19,16 +19,16 @@ <h2>Information About You</h2>
...
+{{ TextInput(f.email_request, placeholder='jane@mail.mil', validation='email') }}
+{{ TextInput(f.phone_number, placeholder='e.g. (123) 456-7890', validation='usPhone') }}
 
 <p>We want to collect the following information from you for security auditing and determining priviledged user access</p>
 
 {{ TextInput(f.service_branch,placeholder='e.g. US Air Force, US Army, US Navy, Marine Corps, Defense Media Agency') }}
 {{ OptionsInput(f.citizenship) }}
 {{ OptionsInput(f.designation) }}
-{{ TextInput(f.date_latest_training) }}
+{{ TextInput(f.date_latest_training, validation='date') }}
```
dandds: Does this need the same placeholder as the other date field?

dandds: I also am getting validation errors on the backend for the string this submits:

```
'date_latest_training':  ['Unable to parse string [04/27/1995]'
```

patricksmithdds: I'm also seeing an issue with this field. Since it needs to be within the past year, if you enter a valid date like `12/12/2020` and submit, the page says there's an error, but the error isn't shown.

js/components/text_input.js
```diff
@@ -72,7 +72,11 @@ export default {
     //
     _checkIfValid: function ({ value, invalidate = false}) {
       // Validate the value
-      const valid = this._validate(value)
+      let valid = this._validate(value)
+
+      if (this.initialErrors && this.initialErrors.length) {
+        valid = false
+      }
```
patricksmithdds: This causes the error to stick around even after the user corrects any error. I think it might be better to move the check into the `mounted` method (wrapping the `checkIfValid` call on line 44 with a conditional to only check if there's not already errors) so that the user sees that beautiful green checkmark after fixing their mistake.

andrewdds: Just fixed that with a `modified` var.

I'm not sure if your suggestion would work, since the errors (server side or otherwise) wont show unless `_checkIfValid` is called.

patricksmithdds:
Travis was (correctly) failing because there was a conflict with the auto-populating of user info work -- I merged master into this branch and fixed the test, so this should be all good now.

patricksmithdds:
@dandds Can you take another look at this? My branch is based off of this, so I'd like to get this merged before making my pull request.

PR #153, opened 2018-08-09
Financial verification modal 159014770

dandds:
This should finish [this](https://www.pivotaltracker.com/story/show/159014770) PT story.

We've added a mechanism to trigger the new Vue modal on page load when necessary, and now we do it when a submitted request is below the $1 million threshold. I've also updated the copy for the modal and alert box per the story, but they could probably use a UI pass (@andrewcroce or @luisgov ).

screen shot 2018-08-09 at 2 12 48 pm

### DIFF
--- a/atst/routes/requests/index.py
+++ b/atst/routes/requests/index.py
@@ -3,6 +3,7 @@ from flask import render_template, g, url_for
 
 from . import requests_bp
 from atst.domain.requests import Requests
+from atst.models.request_status_event import RequestStatus
 
 
 def map_request(request):
@@ -33,4 +34,6 @@ def requests_index():
 
     mapped_requests = [map_request(r) for r in requests]
 
-    return render_template("requests.html", requests=mapped_requests)
+    pending_fv = any(r["status"] == RequestStatus.PENDING_FINANCIAL_VERIFICATION.value for r in mapped_requests)
+
+    return render_template("requests.html", requests=mapped_requests, pending_financial_verification=pending_fv)
--- a/atst/routes/requests/index.py
+++ b/atst/routes/requests/index.py
@@ -3,7 +3,6 @@ from flask import render_template, g, url_for
 
 from . import requests_bp
 from atst.domain.requests import Requests
-from atst.models.request_status_event import RequestStatus
 
 
 def map_request(request):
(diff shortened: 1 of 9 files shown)


templates/requests.html
```diff
@@ -46,9 +40,26 @@ <h1>Your request is now approved!</h1>
 
 {% else %}
 
-  {{ Alert('Pending Financial Verification',
-    message="<p>Your next step is to create a Task Order (T.O.) associated with JEDI Cloud. Please consult a Contracting Officer (KO) or Contracting Officer Representative (COR) to help with this step.</p>"
-  ) }}
+  {% if pending_financial_verification %}
```
patricksmithdds: Do we need this alert in addition to the modal? Seems a bit odd having essentially the same content twice.

dandds: I agree, but that was my understanding from discussion on the story: https://www.pivotaltracker.com/n/projects/2160940/stories/159014770

@kristiandds or @andrewcroce do you remember if the intention was to display both simultaneously?

patricksmithdds: I think it makes sense to show the alert when the user goes to the page at any time and show the modal after submission. It's likely easier to always show the alert than hide it when the modal is shown, so it can be kept as-is.

templates/requests.html
```diff
@@ -46,9 +40,26 @@ <h1>Your request is now approved!</h1>
...
+        Please contact a Contracting Officer (KO), Contracting Officer
+        Representative (COR), or a Financial Manager to help with this step.
+      </p>
+      <p>
+        Once the Task Order has been created, you will be asked to provide
+        details about the task order in the Financial Verification step.
+      </p>
+      <p>
+        <i>Learn more</i> about the JEDI Task Order and the Financial Verification process.
+      </p>
```
patricksmithdds: Can we pull out this and the modal text into a variable/component so we don't have the text repeated twice?

atst/routes/requests/index.py
```diff
@@ -33,4 +34,6 @@ def requests_index():
 
     mapped_requests = [map_request(r) for r in requests]
 
-    return render_template("requests.html", requests=mapped_requests)
+    pending_fv = any(r["status"] == RequestStatus.PENDING_FINANCIAL_VERIFICATION.value for r in mapped_requests)
```
patricksmithdds: We have `Requests.is_pending_financial_verification(request)` in `atst/domains/request.py`, so I think this should be `any(Requests.is_pending_financial_verification(r) for r in requests)`. I'd prefer to not have to reach into the `RequestStatus`'s value.

templates/requests.html
dandds: What do you think of [this](https://github.com/dod-ccpo/atst/pull/153/commits/251bef6bb88585ca06b43ff13d20640ece1a34e0)?

patricksmithdds: Perfect!

PR #158, opened 2018-08-13
Over one mill 159007546

dandds:
Addresses [this](https://www.pivotaltracker.com/story/show/159007546) PT story.

When submitted, requests with the "Pending CCPO Approval" status should display a modal with more information and an alert at the top of the requests index page. This is very similar to the behavior for requests with the "Pending Financial Verification" status.
screen shot 2018-08-13 at 10 28 18 am
screen shot 2018-08-13 at 10 28 35 am

### DIFF
--- a/atst/routes/requests/index.py
+++ b/atst/routes/requests/index.py
@@ -9,8 +9,10 @@ def map_request(request):
     time_created = pendulum.instance(request.time_created)
     is_new = time_created.add(days=1) > pendulum.now()
     app_count = request.body.get("details_of_use", {}).get("num_software_systems", 0)
-    update_url = url_for('requests.requests_form_update', screen=1, request_id=request.id)
-    verify_url = url_for('requests.financial_verification', request_id=request.id)
+    update_url = url_for(
+        "requests.requests_form_update", screen=1, request_id=request.id
+    )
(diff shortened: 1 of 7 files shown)


atst/routes/requests/index.py
```diff
@@ -33,6 +37,15 @@ def requests_index():
...
     mapped_requests = [map_request(r) for r in requests]
 
-    pending_fv = any(Requests.is_pending_financial_verification(r) for r in requests)
-
-    return render_template("requests.html", requests=mapped_requests, pending_financial_verification=pending_fv)
+    pending_fv_count = [
+        True for r in requests if Requests.is_pending_financial_verification(r)
+    ]
+    pending_fv = len(pending_fv_count) > 1
+    pending_ccpo = len(pending_fv_count) != len(mapped_requests)
```
patricksmithdds: I don't think this logic is correct. If, for example, a request was only partially filled out, it's status is "started", so that would cause the "pending CCPO" modal to be displayed.

dandds: Oh, yup, you're right. I forgot you could stop and come back to one. Will fix.

dandds: Fixed. Please take another look when you get a chance.

PR #159, opened 2018-08-13
Details of use form

patricksmithdds:
This PR handles toggling some of the JEDI migration-related questions on the Details of Use portion of the request form.

Some general notes:

* I added a new Vue component, `details-of-use`, to encapsulate the logic in the details of use form. If other forms require similar functionality, they can be made into to separate components.

* The entire form data is passed to the Vue form from the Flask form via the `initialData` property:

     <details-of-use inline-template v-bind:initial-data='{{ f.data|tojson }}'>

* The Vue form pulls out the pieces it needs to track in the `data` function -- otherwise, if a property is accessed, Vue complains that it was not defined.

* Text inputs and options inputs (this PR adds an `optioninput` Vue component, similar to the existing `textinput`) will emit _on the root element_ a `field-change` event when their field's value has changed (with the raw value of the field)

* The Vue form registers a handler for the `field-change` event _on the root element_ that updates the field's value on the form.

* The logic for showing/hiding fields is wrapped in a `computed` method on the form to simplify the template.

The functionality included in this pr:

  * if the user does not select an option for the JEDI migration question ("Are you using the JEDI Cloud to migrate existing systems?"), no other questions for the "cloud readiness" section are shown
  * if the user selects "no" to the migration question, they are shown the "cloud native" question.
  * if the user selects "yes" to the migration question, the other questions relating to the migration are shown.
  * if the user selects "yes" to the "are you working with a technical support team", the question regarding the type of support is shown.
  * the annual estimated cost is shown and updated as the user enters the monthly cost.
  * if the annual estimated cost is over one million dollars, the user is required to fill out the traffic/sessions questions.

This PR also updates the server side form to allow the hidden fields to be left blank, rather than requiring input each time.

### DIFF
--- a/atst/forms/request.py
+++ b/atst/forms/request.py
@@ -1,6 +1,6 @@
 from wtforms.fields.html5 import IntegerField
 from wtforms.fields import RadioField, TextAreaField, SelectField
-from wtforms.validators import Optional
+from wtforms.validators import Optional, Required
 from .fields import DateField
 from .forms import ValidatedForm
 
@@ -19,6 +19,15 @@ class RequestForm(ValidatedForm):
             if self.technical_support_team.data == 'no':
                 self.organization_providing_assistance.validators.append(Optional())
             self.cloud_native.validators.append(Optional())
+
+        try:
+            annual_spend = int(self.estimated_monthly_spend.data or 0) * 12
+        except ValueError:
+            annual_spend = 0
+
+        if annual_spend > 1_000_000:
+            self.number_user_sessions.validators.append(Required())
+            self.average_daily_traffic.validators.append(Required())
 
         return super(RequestForm, self).validate(*args, **kwargs)
 
--- a/atst/forms/request.py
+++ b/atst/forms/request.py
@@ -1,8 +1,10 @@
 from wtforms.fields.html5 import IntegerField
 from wtforms.fields import RadioField, TextAreaField, SelectField
 from wtforms.validators import Optional, Required
+
 from .fields import DateField
 from .forms import ValidatedForm
+from atst.domain.requests import Requests
 
 
 class RequestForm(ValidatedForm):
@@ -25,7 +27,7 @@ class RequestForm(ValidatedForm):
         except ValueError:
             annual_spend = 0
 
-        if annual_spend > 1_000_000:
+        if annual_spend > Requests.AUTO_APPROVE_THRESHOLD:
             self.number_user_sessions.validators.append(Required())
             self.average_daily_traffic.validators.append(Required())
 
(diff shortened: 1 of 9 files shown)


atst/forms/request.py
```diff
@@ -1,11 +1,36 @@
...
+            if self.technical_support_team.data == 'no':
+                self.organization_providing_assistance.validators.append(Optional())
+            self.cloud_native.validators.append(Optional())
+
+        try:
+            annual_spend = int(self.estimated_monthly_spend.data or 0) * 12
+        except ValueError:
+            annual_spend = 0
+
+        if annual_spend > 1_000_000:
```
dandds: Maybe use `atst.domain.requests.AUTO_APPROVE_THRESHOLD`?

atst/forms/request.py
```diff
@@ -1,11 +1,36 @@
...
+from wtforms.validators import Optional, Required
 from .fields import DateField
 from .forms import ValidatedForm
 
 
 class RequestForm(ValidatedForm):
 
+    def validate(self, *args, **kwargs):
+        if self.jedi_migration.data == 'no':
+            self.rationalization_software_systems.validators.append(Optional())
```
dandds: This seems okay for now, but if we need a lot of this kind of branching logic in the form we should probably come up with a nice way to abstract or encapsulate it.

atst/forms/request.py
```diff
@@ -1,11 +1,36 @@
...
 from wtforms.fields import RadioField, TextAreaField, SelectField
+from wtforms.validators import Optional, Required
 from .fields import DateField
 from .forms import ValidatedForm
 
 
 class RequestForm(ValidatedForm):
 
+    def validate(self, *args, **kwargs):
+        if self.jedi_migration.data == 'no':
```
richard-dds: Would it make sense to create a new Validator that made the field optional unless the `self.jedi_migration.data == 'no'`?

atst/forms/request.py
patricksmithdds: Did that in 581f908c1a934bac3f620cc360214c3e70396a16

atst/forms/request.py
patricksmithdds: My understanding was that a validator didn't have access to the whole form, so we aren't able to have one field's validator dependent on another field.

atst/forms/request.py
patricksmithdds: I agree and if you have ideas, I'm all 👂s

atst/forms/request.py
richard-dds: That's too bad, but not too surprising I guess.

PR #161, opened 2018-08-13
Create RequestQuery class to house Request db operations

richard-dds:
This is an exploratory refactor that I did on Friday while working on a bugfix.

My feeling is that `domain.requests` exists to house the JEDI request business logic in one place, but that the module was getting crowded with SQLAlchemy-specific code. I thought it would be a good idea to factor that out for clarity.


tests/routes/test_financial_verification.py
```diff
@@ -2,8 +2,11 @@
 import pytest
 import urllib
 from tests.mocks import MOCK_REQUEST, MOCK_USER
-from tests.factories import PENumberFactory
+from tests.factories import PENumberFactory, RequestFactory
```
dandds: These are the unrelated test changes you mentioned, right?

richard-dds: Unfortunately these test changes are required for this refactor to work properly. Monkeypatching `Requests.get` was no longer sufficient for these tests to pass, since the tests were dependent on the implementation of `Requests.update` calling `Requests.get`.

richard-dds:
@patricksmithdds I'd like to get your opinion on this before I merge.

atst/domain/requests/query.py
```diff
@@ -0,0 +1,76 @@
...
+from sqlalchemy.orm.attributes import flag_modified
+
+from atst.database import db
+from atst.models import Request
+from atst.domain.exceptions import NotFoundError
+
+
+class RequestQuery(object):
+
+    @classmethod
```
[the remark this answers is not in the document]
richard-dds: That makes sense to me in some ways, but I think that it would be strange to have `models/request.py:Request` to be a place both for helper properties (like the two we currently have) that act on a single request instance, and static methods that act on requests in aggregate. Would we move those properties someplace else?

richard-dds:
I think I'm gonna close this for now, since I think it may be a little premature. Later implementation should give us a clearer picture.

PR #165, opened 2018-08-13
Update copy on Technical Point of Contact form

patricksmithdds:
Updates the copy on the "Technical Point of Contact" form, henceforth known as the "Workspace Owner":

image

I've left the section of the stored form unchanged for backwards compatibility, so it'll still be `primary_poc` in the form data.

### DIFF
--- a/atst/forms/poc.py
+++ b/atst/forms/poc.py
@@ -6,11 +6,11 @@ from .validators import IsNumber
 
 
 class POCForm(ValidatedForm):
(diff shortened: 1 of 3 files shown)


patricksmithdds:
> after MVP I think we should try to update the app's variable and file names (poc -> workspace owner, etc.)

I agree. In this particular case, I think we can rename it when we move away from using the JSONB column for storing request data (https://www.pivotaltracker.com/story/show/159719829).

PR #166, opened 2018-08-13
Ui/emphatic subnav

andrewdds:
Per discussions with DDS, this makes the subnav items more emphatic, so for instance, its clearer where you can do "+ Add Thing" actions.

Also fixes a weird layout behavior for navigations on smaller screens.

screen shot 2018-08-13 at 3 56 32 pm

screen shot 2018-08-13 at 3 56 36 pm

### DIFF
--- a/styles/components/_workspace_layout.scss
+++ b/styles/components/_workspace_layout.scss
@@ -6,13 +6,18 @@
 
 .workspace-navigation {
   @include panel-margin;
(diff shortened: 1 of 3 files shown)


andrewdds:
@patricksmithdds thats a good bit of feedback, which is more a question of whether we should have the "+ Add" button on the edit screens. Probably not.

PR #168, opened 2018-08-14
CCPO should not see request status alerts

dandds:
Very small PR related to these two PT stories:

- https://www.pivotaltracker.com/story/show/159014770
- https://www.pivotaltracker.com/story/show/159007546

I realized that the CCPO should not see the status alerts on the requests index page. Those are for the PSO/MO who created the request.

CCPO view:

screen shot 2018-08-14 at 10 33 48 am

Other view:

screen shot 2018-08-14 at 10 33 27 am

### DIFF
--- a/atst/routes/requests/index.py
+++ b/atst/routes/requests/index.py
@@ -39,8 +39,8 @@ def requests_index():
 
     mapped_requests = [map_request(r) for r in requests]
 
-    pending_fv = any(Requests.is_pending_financial_verification(r) for r in requests) and not is_ccpo
-    pending_ccpo = any(Requests.is_pending_ccpo_approval(r) for r in requests) and not is_ccpo
+    pending_fv = not is_ccpo and any(Requests.is_pending_financial_verification(r) for r in requests)
+    pending_ccpo = not is_ccpo and any(Requests.is_pending_ccpo_approval(r) for r in requests)
 
     return render_template(
         "requests.html",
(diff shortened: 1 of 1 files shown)


atst/routes/requests/index.py
```diff
@@ -30,15 +31,16 @@ def map_request(request):
...
+    if is_ccpo:
         requests = Requests.get_many()
     else:
         requests = Requests.get_many(creator=g.current_user)
 
     mapped_requests = [map_request(r) for r in requests]
 
-    pending_fv = any(Requests.is_pending_financial_verification(r) for r in requests)
-    pending_ccpo = any(Requests.is_pending_ccpo_approval(r) for r in requests)
+    pending_fv = any(Requests.is_pending_financial_verification(r) for r in requests) and not is_ccpo
```
richard-dds: Probably doesn't matter much yet, but it might be worth it to prevent these two `any()`s from being evaluated if the user is a CCPO, since for CCPOs the `requests` collection could be pretty large.

We could either duplicate the `pending_fv` and `pending_ccpo`s and put them in the if block, or just move `not is_ccpo and` to the beginning of those two boolean expressions.

patricksmithdds:
Should this hide the KPI panel from non-CCPO users, too?

dandds:
@patricksmithdds Yeah, I have that change on an upcoming PR for https://www.pivotaltracker.com/story/show/159274521.

PR #170, opened 2018-08-14
Requests index view 159274521

dandds:
Relates to [this](https://www.pivotaltracker.com/story/show/159274521) PT story. 

This adjusts the requests index view depending on whether the user is a CCPO or PSO/MO. (Screenshots [here](https://www.pivotaltracker.com/story/show/159274521/comments/193023582)).

A PSO/MO should see the following columns:

- JEDI Cloud Request ID
- Date request initiated / created
- Projected Annual Usage ($) 
- Request Status
- Actions

A CCPO should see:

- the KPI counts
- the search and filter bars
- additional "requester" and "reason flagged" request columns (the latter will be iterated on and implemented later, per https://www.pivotaltracker.com/story/show/159274521/comments/193021706)

### DIFF
--- a/templates/requests.html
+++ b/templates/requests.html
@@ -101,7 +101,6 @@
             {% endif %}
             <th scope="col">Projected Annual Usage ($)</th>
             <th scope="col">Request Status</th>
-            <th scope="col" class="table-cell--shrink">Actions</th>
           </tr>
         </thead>
         <tbody>
@@ -119,9 +118,6 @@
             {% endif %}
             <td>${{ r['annual_usage'] }}</td>
             <td>{{ r['status'] }}</td>
-            <td class="table-cell--shrink">
-              <a href="/request/approval" class='icon-link'>Approval</a>
-            </td>
           </tr>
           {% endfor %}
         </tbody>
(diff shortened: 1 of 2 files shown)


templates/requests.html
```diff
@@ -106,11 +113,13 @@ <h1>Request submitted!</h1>
...
-            <td>{{ r['app_count'] }}</td>
+            {% if extended_view %}
+              <td>{{ r['full_name'] }}</td>
+              <td></td>
+            {% endif %}
+            <td>${{ r['annual_usage'] }}</td>
             <td>{{ r['status'] }}</td>
             <td class="table-cell--shrink">
-              <a href="" class='icon-link'>Download</a>
               <a href="/request/approval" class='icon-link'>Approval</a>
```
patricksmithdds: Can we switch out this action based on the CCPO-ness of the user? A PSO/MO shouldn't see "approve"

dandds: Yeah, the story also mentions removing the download link, but I don't know if that means for just PSO/MO view or all users. If we did both, that means that column is empty for PSO/MO until we implement this spreadsheet: https://www.pivotaltracker.com/story/show/159274521/comments/192986599

dandds:
Removing "Actions" column for now after discussion with @patricksmithdds .

PR #171, opened 2018-08-14
Ui/conditional field display

andrewdds:
Adds stylized grouping to conditional field sets.
Adds animated transitions when groups enter/leave.

screen shot 2018-08-14 at 11 33 24 am

screen shot 2018-08-14 at 11 33 31 am

note: GIF is below is doing something kind of weird with background color.
fieldset


styles/components/_forms.scss
```diff
@@ -0,0 +1,88 @@
...
+  @include alert;
+  @include alert-level('default');
+
+  // @include media($medium-screen) {
+  //   padding: $gap * 2;
+  // }
+
+  // @include media($large-screen) {
+  //   padding: $gap * 4;
+  // }
```
patricksmithdds: Can we drop these unused styles?

templates/requests/screen-1.html
```diff
@@ -31,31 +31,50 @@ <h2>Cloud Readiness</h2>
...
+            </template>
+          </transition>
+          {{ OptionsInput(f.engineering_assessment, tooltip="An engineering assessment is an evaluation to convert your application architecture from on-premises to using the commercial cloud") }}
+          {{ OptionsInput(f.data_transfers) }}
+          {{ OptionsInput(f.expected_completion_date) }}
+        </fieldset>
+      </transition>
+
+      <template v-if='!isJediMigration' v-cloak>
         {{ OptionsInput(f.cloud_native, tooltip="Cloud native is architecting and designing your application to use all the benefits of the commercial cloud. Specifically, designing applications so that they are decoupled from a physical resource.") }}
```
patricksmithdds: Should this have a transition as well?

dandds:
Agree with Patrick!

One thing I noticed is that the very first time I clicked "yes" for "Are you using the JEDI Cloud to migrate existing systems?", the additional fields just appeared with no transition. After their first appearance, I could toggle them in and out with the transitions as I'd expect. Is that the expected behavior? (I'm on FF 61.0.1.)

templates/requests/screen-1.html
andrewdds: good catch thanks

andrewdds:
@patricksmithdds no thats not desired behavior, thanks. I'll look

templates/requests/screen-1.html
```diff
@@ -30,32 +30,52 @@ <h2>General</h2>
...
+          <p>So this means you are spending approximately <span class="label label--info">!{ annualSpendStr }</span> annually</p>
+        </div>
+      </div>
+    </div>
+
     <template v-if="annualSpend > 1000000">
-      {{ TextInput(f.number_user_sessions, validation='integer') }}      
-      {{ TextInput(f.average_daily_traffic, tooltip="Requests are the client-to-server network traffic that is being transferred to your systems",validation="integer") }}
-      {{ TextInput(f.average_daily_traffic_gb, tooltip="GB uploaded is the gigabyte amount of data traffic that is being transferred to your systems",validation="gigabytes") }}      
+      <transition name='slide'>
```
patricksmithdds: I believe this needs to move outside the parent `template` as well. Currently, these will pop in instead of sliding in.

PR #172, opened 2018-08-14
fix border and text color in hover/focused

andrewdds:
When a field is focused or hovered, the active blue color will override success or error states, preventing a weird combination of borders.

screen shot 2018-08-14 at 11 43 26 am
screen shot 2018-08-14 at 11 43 43 am


andrewdds:
I hate using important, but in this case, it is

PR #176, opened 2018-08-14
Ui/review and submit

andrewdds:


templates/requests/screen-4.html
```diff
@@ -14,172 +19,201 @@ <h2>Review &amp; Submit</h2>
...
     </div>
 
     <div>
       <dt>Total Spend</dt>
-      <dd>${{data['details_of_use']['dollar_value']}}</dd>
+      <dd>{{ data['details_of_use']['dollar_value'] or RequiredLabel() }}</dd>
     </div>
 
     <div>
       <dt>Number of User Sessions</dt>
```
dandds: I don't think it's a problem with this PR, but I noticed here that these fields need to be wrapped in the same "over a million" conditional on this form. They're showing up with the new "Response Required" label for requests under that threshold.

PR #177, opened 2018-08-14
Designate technical poc 159495040

richard-dds:
Allow a MO to designate themselves as a technical POC. This involves a kind of gross addition to the (already gross) JEDIRequestFlow class that modules the request body data before updating. My opinion is that this is ok for now, but once we finalize the Request body schema we should tackle the big hunk of tech debt known as `atst.routes.requetsts`.

screen shot 2018-08-14 at 4 32 48 pm

screen shot 2018-08-14 at 4 32 55 pm

### DIFF
--- a/js/components/checkbox_input.js
+++ b/js/components/checkbox_input.js
@@ -1,0 +1,17 @@
+export default {
+  name: 'checkboxinput',
+
+  props: {
+    name: String,
+  },
+
+  methods: {
+    onInput: function (e) {
+        console.log(e)
+      this.$root.$emit('field-change', {
+        value: e.target.checked,
+        name: this.name
+      })
+    }
+  }
+}
--- a/js/components/checkbox_input.js
+++ b/js/components/checkbox_input.js
@@ -7,7 +7,6 @@ export default {
 
   methods: {
     onInput: function (e) {
-        console.log(e)
       this.$root.$emit('field-change', {
         value: e.target.checked,
         name: this.name
--- a/tests/routes/test_request_new.py
+++ b/tests/routes/test_request_new.py
@@ -1,7 +1,4 @@
 import re
-import pytest
-import urllib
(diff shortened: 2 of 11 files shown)


luisgov:
@richard-dds maybe this should be a checkbox instead?

richard-dds:
@luisgov Made the input into a checkbox.

js/components/checkbox_input.js
```diff
@@ -0,0 +1,16 @@
+export default {
+  name: 'checkboxinput',
```
dandds: 🎊

tests/routes/test_request_new.py
```diff
@@ -103,6 +101,58 @@ def test_non_creator_info_is_not_autopopulated(monkeypatch, client, user_session
...
+
+
+def test_not_am_poc_allows_user_to_fill_in_poc_info(client, user_session):
+    creator = UserFactory.create()
+    user_session(creator)
+    request = RequestFactory.create(creator=creator, body={})
+    response = client.post(
+        "/requests/new/3/{}".format(request.id),
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+        data="am_poc=no&fname_poc=test&lname_poc=user&email_poc=test.user@mail.com&dodid_poc=1234567890",
```
dandds: I might use `urllib.parse.urlencode` on a dictionary instead to make it clearer to read/update.

tests/routes/test_request_new.py
```diff
@@ -103,6 +101,58 @@ def test_non_creator_info_is_not_autopopulated(monkeypatch, client, user_session
...
+    )
+    assert ERROR_CLASS not in response.data.decode()
+
+
+def test_not_am_poc_allows_user_to_fill_in_poc_info_for_new_request(client, user_session):
```
dandds: Is this test name right?

PR #178, opened 2018-08-14
Add counts for KPI (159705475)

dandds:
Relates to [this](https://www.pivotaltracker.com/story/show/159705475) PT story.

This wires up the KPI that @luisgov designed on a previous PR and changes the category names per discussion with DDS. Thanks to @richard-dds for helping me figure out the gnarly SQL. (If either you or @patricksmithdds want to take a crack at using the ORM, be my guest.)

**notes**

- I've separated rendering the CCPO and non-CCPO requests index view into separate functions.
- I updated one of the RequestStatus names, per @luisgov on the PT story.

screen shot 2018-08-14 at 4 56 25 pm

### DIFF
--- a/atst/domain/requests.py
+++ b/atst/domain/requests.py
@@ -1,4 +1,6 @@
+from enum import Enum
 from sqlalchemy import exists, and_, exc
+from sqlalchemy.sql import text
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.attributes import flag_modified
 
@@ -154,3 +156,21 @@ class Requests(object):
     @classmethod
     def is_pending_ccpo_approval(cls, request):
         return request.status == RequestStatus.PENDING_CCPO_APPROVAL
+
+    @classmethod
+    def count_status(self, status):
+        raw = text("""
+SELECT count(requests_with_status.id)
+FROM (
+    SELECT DISTINCT ON (rse.request_id) r.*, rse.new_status as status
+    FROM request_status_events rse JOIN requests r ON r.id = rse.request_id
+    ORDER BY rse.request_id, rse.sequence DESC
+) as requests_with_status
+WHERE requests_with_status.status = :status;
+        """)
+        if isinstance(status, Enum):
+            status = status.name
+        results = db.session.execute(raw, {"status": status}).fetchone()
+        (count,) = results
+        return count
+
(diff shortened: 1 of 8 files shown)


atst/domain/requests.py
```diff
@@ -154,3 +156,37 @@ def is_pending_financial_verification(cls, request):
...
+    @classmethod
+    def status_count(cls, status):
+        raw = text("""
+SELECT count(requests_with_status.id)
+FROM (
+    SELECT DISTINCT ON (rse.request_id) r.*, rse.new_status as status
+    FROM request_status_events rse JOIN requests r ON r.id = rse.request_id
+    ORDER BY rse.request_id, rse.sequence DESC
+) as requests_with_status
+WHERE requests_with_status.status = :status;
```
patricksmithdds: Does this select just the current status of a request?

atst/models/request_status_event.py
```diff
@@ -12,9 +12,10 @@ class RequestStatus(Enum):
     STARTED = "Started"
     PENDING_FINANCIAL_VERIFICATION = "Pending Financial Verification"
     PENDING_CCPO_APPROVAL = "Pending CCPO Approval"
+    CHANGES_REQUESTED = "Changes Requested"
     APPROVED = "Approved"
     EXPIRED = "Expired"
-    DELETED = "Deleted"
+    CANCELED = "Canceled"
```
patricksmithdds: I was going to say "cancelled" has two "l"s, but apparently its fine both ways. Yay english

atst/domain/requests.py
```diff
@@ -154,3 +156,37 @@ def is_pending_financial_verification(cls, request):
     @classmethod
     def is_pending_ccpo_approval(cls, request):
         return request.status == RequestStatus.PENDING_CCPO_APPROVAL
+
+    @classmethod
+    def status_count(cls, status):
```
patricksmithdds: I'm a bit concerned that this selects all of the requests without scoping it to the current user. It's fine for now since the CCPO user can see all the requests, but that could change in the future.

atst/domain/requests.py
dandds: Yup.

atst/domain/requests.py
dandds: @patricksmithdds I added an optional creator scope. Can you take a look when you have a chance?

PR #179, opened 2018-08-15
Run tests on Travis

patricksmithdds:
This PR resurrects the previous `script` command to run the tests on the docker tester image and moves the other commands to `after_success` so they are only run if the tests pass.

It _appears_ that the last step of `after_success` (`- docker run --add-host "postgreshost:${postgres_ip}" --add-host "redishost:${redis_ip}" "${TESTER_IMAGE2_NAME}"`) should run the tests, but that is not the case -- I added a test that always fails and [the travis build](https://travis-ci.org/dod-ccpo/atst/builds/416324949?utm_source=github_status&utm_medium=notification) succeeded without error. Updating the script step correctly makes the [build fail](https://travis-ci.org/dod-ccpo/atst/builds/416329881?utm_source=github_status&utm_medium=notification).

It seems the `TESTER_IMAGE2_NAME` image is not actually used elsewhere, so I'm not sure if the `after_success` block can be removed entirely.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,11 @@ before_script:
 script:
   - docker run -d --entrypoint='/bin/sh' -t --name current-atst-tester "${TESTER_IMAGE1_NAME}"
   - docker container exec -t current-atst-tester script/sync-crls
-  - docker commit current-atst-tester "${TESTER_IMAGE2_NAME}"
+  - docker commit
+    --change='ENTRYPOINT ["/usr/bin/dumb-init", "--"]'
+    --change='CMD ["bash", "-c", "${APP_DIR}/script/cibuild"]'
+    current-atst-tester
+    "${TESTER_IMAGE2_NAME}"
   - docker cp current-atst-tester:/opt/atat/atst/crl/. ./crl/
   - docker container stop current-atst-tester
   - docker run --add-host "postgreshost:${postgres_ip}" --add-host "redishost:${redis_ip}" "${TESTER_IMAGE2_NAME}"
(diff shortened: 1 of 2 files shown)


ddsdevon:
Hrm,that's odd. `- docker run --add-host "postgreshost:${postgres_ip}" --add-host "redishost:${redis_ip}" "${TESTER_IMAGE2_NAME}"` should be failing if the tests don't pass... I'm not sure why it wouldn't.
Everything before that is just populating the /crl directory of IMAGE2, which if we never need for tests, can happen after the fact.

IMAGE2 is what the prod image is built from, so it is used (it is just that it is only mentioned in the prod Dockerfile)

patricksmithdds:
The changing cmd/entrypoint was exactly the problem. I updated them back to what they are in the tester dockerfile when the second image is created.

I _really_ didn't want to duplicate those lines, but `docker inspect` does not give the values in any way that's usable by `docker commit` or `docker run`. 

I wanted to do:

```
  - cmd="$(docker inspect ${TESTER_IMAGE1_NAME} -f "{{ .Config.Cmd }}")"
  - entrypoint="$(docker inspect ${TESTER_IMAGE1_NAME} -f "{{ .Config.Entrypoint }}")"
  - docker commit
    --change="ENTRYPOINT ${entrypoint}"
    --change="CMD ${cmd}"
    current-atst-tester
    "${TESTER_IMAGE2_NAME}"
```

But this results in something like:

    --change="CMD [bash -c ${APP_DIR}/script/uwsgi_server]"

where we need

    --change='CMD ["bash", "-c", "${APP_DIR}/script/cibuild"]'

There's some Travis failures available for what this does: https://travis-ci.org/dod-ccpo/atst/builds/416995025.

PR #181, opened 2018-08-16
CRL loading

dandds:
Previously, we were keeping the parsed CRLs in-memory while the app is running so that we could have an `x509Store` ready to go for client authentication. Now, we keep a lookup table of every CRL's issuer and file path. When a client certificate needs to be authenticated, we get its issuer, get the appropriate CRL from the lookup table, parse the CRL, and create a one-off `x509Store`.

We might take a slight performance penalty during authentication, but I didn't notice a big one locally. We should keep an eye on it.

Before, the app used about 16% of my machine's memory with all the CRLs; now it's down to 2.3%.

One other note: previously, the crl validation method would return a boolean. Following the pattern we've established, in now raises an exception.


atst/domain/authnid/crl/__init__.py
```diff
@@ -13,116 +17,76 @@ def sha256_checksum(filename, block_size=65536):
...
+        with open(crl_location, "rb") as crl_file:
+            crl = crypto.load_crl(crypto.FILETYPE_ASN1, crl_file.read())
+            store.add_crl(crl)
+            store = self._add_certificate_chain_to_store(store, crl.get_issuer())
+            return store
+
+    # this _should_ happen just twice for the DoD PKI (intermediary, root) but
+    # theoretically it can build a longer certificate chain
+    def _add_certificate_chain_to_store(self, store, issuer):
+        ca = self.certificate_authorities.get(issuer.der())
```
patricksmithdds: Is it possible the issuer is not one of the cached certificate authorities? `store.add_cert(None)` raises a `TypeError`, so we may need to dance around that if it is possible.

atst/domain/authnid/crl/__init__.py
```diff
@@ -5,6 +5,10 @@
 from OpenSSL import crypto, SSL
 
 
+class CRLException(Exception):
+    pass
+
+
 def sha256_checksum(filename, block_size=65536):
```
patricksmithdds: As you mentioned, it looks like this can be removed now.

atst/domain/authnid/crl/__init__.py
```diff
@@ -13,116 +17,76 @@ def sha256_checksum(filename, block_size=65536):
     return sha256.hexdigest()
 
 
-class Validator:
+def crl_check(cache, cert):
```
patricksmithdds: Can this be a method on the cache? Calling it via `cache.crl_check(cert)` seems more intuitive to me than `crl_check(cache, cert)`. We wouldn't have more than one cache to be able to pass in, right?

atst/domain/authnid/crl/__init__.py
```diff
@@ -5,6 +5,10 @@
 from OpenSSL import crypto, SSL
 
 
+class CRLException(Exception):
```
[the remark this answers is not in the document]
dandds: Sure. `CRLRevocationException`? The `X509StoreContextError` exception we catch is actually pretty generic: it could mean that the cert is revoked, that it's expired, or that the `x509Store` configuration doesn't match the certificate issuer. (In practice, the latter two would not make it past the NGINX SSL validation.)

atst/domain/authnid/crl/__init__.py
dandds: Sort of, technically, yes. In practice a cert with an issuer not in our store would not make it past the initial pass/fail SSL validation, but it's probably best practice to catch and raise that exception anyway. (We would have to have those certificate authorities available to NGINX but not available to the app.)

atst/domain/authnid/crl/__init__.py
dandds: Sure, that makes sense. I guess I wanted it to be similar to the other top-level utility functions under authnid (`parse_sdn`, etc.) but we already have the object.

dandds:
@patricksmithdds Pushed some commits to address your feedback. Let me know what you think.

atst/domain/authnid/crl/__init__.py
```diff
@@ -71,22 +71,31 @@ def _load_crl(self, crl_location):
...
         with open(crl_location, "rb") as crl_file:
             crl = crypto.load_crl(crypto.FILETYPE_ASN1, crl_file.read())
             store.add_crl(crl)
             store = self._add_certificate_chain_to_store(store, crl.get_issuer())
             return store
 
+    def _get_crl_location(self, issuer):
+        crl_location = self.crl_cache.get(issuer)
+
+        if not crl_location:
```
patricksmithdds: If there's no CRL, does that necessarily mean that there's no revoked certificates and therefore the certificate is OK? 

Put another way -- can we be sure that we have a CRL from every valid issuer?

dandds: My understanding is that yes, we should have a CRL for every valid issuer. (CRLs can be empty.) 

Maybe @ddsdevon or @jordangov knows differently?

ddsdevon: Yes, we should have a CRL for all validate authorities (few of the current ones are indeed empty).

PR #182, opened 2018-08-16
Set up continuous deployment

patricksmithdds:
Example of the deployment working (from https://travis-ci.org/dod-ccpo/atst/builds/416981634): 

image

amazing

With this PR, Travis will deploy the `master` branch with the `atat-deployer` user and will wait until the deployment is successful before completing the build. This also bumps up the memory requirement to 6gb, but after #181 gets merged, we can draw that down a bit and add a second replica to fully take advantage of the rolling deployment.


.travis.yml
```diff
@@ -18,6 +20,8 @@ before_install:
   - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules
   # Manually initialize submodules
   - git submodule update --init --recursive
+install:
+  - curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.10.4/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
 
 before_script:
   - rm -rf ./crl/*
```
ddsdevon: @patricksmithdds You should be able to remove this line now. It was to clear out "bad" cached entries before the changes to the sync crl script (assuming this has been rebased off master recently or whatever)

patricksmithdds: Excellent, thanks! I was a bit curious why we were caching the files only to remove them before we could use it.

PR #184, opened 2018-08-20
review and submit page bugs (159795524 and 159889979)

dandds:
This fixes two bugs on the review and submit page.

1. https://www.pivotaltracker.com/story/show/159795524

We should only show the data usage fields on the "review and submit" form if they're necessary for the request (i.e., it's total value is greater than one million).

2. https://www.pivotaltracker.com/story/show/159889979

The alert telling you that you have errors or missing data on the "review and submit" is displayed if you successfully submit a request and then come back to that page. If a user submits a valid request and then comes back to that form page (changes requested, bored, etc.), they shouldn't see that alert. 

It's displaying in those circumstances because `Requests.should_allow_submission` has `STARTED` as the only acceptable status value. I've updated it to accept a range of statuses.

### DIFF
--- a/atst/domain/requests.py
+++ b/atst/domain/requests.py
@@ -139,8 +139,6 @@ class Requests(object):
 
     _VALID_SUBMISSION_STATUSES = [
         RequestStatus.STARTED,
-        RequestStatus.PENDING_FINANCIAL_VERIFICATION,
-        RequestStatus.PENDING_CCPO_APPROVAL,
         RequestStatus.CHANGES_REQUESTED,
     ]
 
--- a/atst/domain/requests.py
+++ b/atst/domain/requests.py
@@ -137,6 +137,13 @@ class Requests(object):
 
         return dollar_value < cls.AUTO_APPROVE_THRESHOLD
 
+    _VALID_SUBMISSION_STATUSES = [
+        RequestStatus.STARTED,
+        RequestStatus.PENDING_FINANCIAL_VERIFICATION,
+        RequestStatus.PENDING_CCPO_APPROVAL,
+        RequestStatus.CHANGES_REQUESTED,
+    ]
+
     @classmethod
     def should_allow_submission(cls, request):
         all_request_sections = [
@@ -145,7 +152,7 @@ class Requests(object):
             "primary_poc",
         ]
         existing_request_sections = request.body.keys()
-        return request.status == RequestStatus.STARTED and all(
+        return request.status in Requests._VALID_SUBMISSION_STATUSES and all(
             section in existing_request_sections for section in all_request_sections
         )
 
--- a/templates/requests/screen-4.html
+++ b/templates/requests/screen-4.html
@@ -21,7 +21,7 @@
 
   <p>Before you can submit your request, please take a moment to review the information entered in the form. You may make changes by clicking the edit link on each section. When all information looks right, go ahead and submit.</p>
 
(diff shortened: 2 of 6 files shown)


templates/requests/screen-4.html
```diff
@@ -114,20 +114,23 @@ <h2>
       <dd>{{ data['details_of_use']['dollar_value'] or RequiredLabel() }}</dd>
     </div>
 
-    <div>
-      <dt>Number of User Sessions</dt>
-      <dd>{{ data['details_of_use']['number_user_sessions'] or RequiredLabel() }}</dd>
-    </div>
+    {% set dollar_value = data['details_of_use']['dollar_value'] | int %}
```
patricksmithdds: These values are required if the _annual spend_ is more than 1,000,000 -- the `dollar_value` here is total budget. We should be using the `monthly_spend` to determine if we need to show/hide these fields.

atst/domain/requests.py
```diff
@@ -137,6 +137,13 @@ def should_auto_approve(cls, request):
 
         return dollar_value < cls.AUTO_APPROVE_THRESHOLD
 
+    _VALID_SUBMISSION_STATUSES = [
+        RequestStatus.STARTED,
+        RequestStatus.PENDING_FINANCIAL_VERIFICATION,
+        RequestStatus.PENDING_CCPO_APPROVAL,
```
patricksmithdds: If the request is in a pending approval/verification state, the user shouldn't be able to access the review & submit page (unless they navigate directly to the URL). I also think we could have issues if the user could change a request while it's currently being reviewed by the CCPO. Could we limit this to only started/changes requested for now?

templates/requests/screen-4.html
```diff
@@ -114,20 +114,23 @@ <h2>
...
+    {% if dollar_value > 1000000 %}
```
richard-dds: I think we should find a way to provide this threshold value from the backend so that we don't need to duplicate this constant.

templates/requests/screen-4.html
dandds: Gotcha. 👍

atst/domain/requests.py
dandds: Sure.

dandds:
@patricksmithdds Okay, updated for when you get a minute. Sorry for my misunderstanding. I added a derived `annual_spend` property to the requests model since we need it in a couple places.

@richard-dds suggested putting the annual spend in a constant, but I couldn't decide on a good place to do that. We need it in both Vue and non-Vue template contexts. Normally I would make this a boolean on the model (`request.is_above_annual_spend_threshold`) but that's more Rails-y than we want, I think.

dandds:
Both builds are green but it's not updating for me for some reason... going to merge.

templates/requests/screen-4.html
```diff
@@ -114,20 +114,22 @@ <h2>
       <dd>{{ data['details_of_use']['dollar_value'] or RequiredLabel() }}</dd>
     </div>
 
-    <div>
-      <dt>Number of User Sessions</dt>
-      <dd>{{ data['details_of_use']['number_user_sessions'] or RequiredLabel() }}</dd>
-    </div>
+    {% if jedi_request and jedi_request.annual_spend > 1000000 %}
```
kristiandds: oh, hard coding the 100000 here might not be a good idea because this should be the common variable that's used in one place?

PR #187, opened 2018-08-20
Ui/review and submit

andrewdds:
Changes default value of all fields to `None` rather than an arbitrary string.
Cleans up display of various strings, with filters and label dictionaries.
Adds conditional display of various values, mimicking the conditional display in the form.
Adds a small macro for displaying a "Required" label if no value is present

### DIFF
--- a/templates/requests/screen-4.html
+++ b/templates/requests/screen-4.html
@@ -8,6 +8,23 @@
 {% from "components/text_input.html" import TextInput %}
 {% from "components/icon.html" import Icon %}
 
+
+{% set dod_component_choices = {
+  "null": "Select an option",
+  "us_air_force": "US Air Force",
+  "us_army": "US Army",
+  "us_navy": "US Navy",
+  "us_marine_corps": "US Marine Corps",
+  "joint_chiefs_of_staff": "Joint Chiefs of Staff"
+} %}
+
+{% set organization_providing_assistance_choices = {
+  "in_house_staff": "In-house staff",
+  "contractor": "Contractor",
+  "other_dod_organization": "Other DoD organization",
+  "none": "None"
+} %}
+
 {% block subtitle %}
 <h2>Review &amp; Submit</h2>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


templates/requests/screen-4.html
```diff
@@ -72,62 +101,110 @@ <h2>
...
+          {{ RequiredLabel() }}
+        {% endif %}
+      </dd>
     </div>
 
-    <div>
-      <dt>Total Spend</dt>
-      <dd>{{ data['details_of_use']['dollar_value'] or RequiredLabel() }}</dd>
-    </div>
+    {% if (data['details_of_use']['estimated_monthly_spend'] and ((data['details_of_use']['estimated_monthly_spend'] * 12) > 1000000)) %}
```
dandds: I think `dollar_value` is always shown. You might want to check or rebase against master. I added a condition for this here: https://github.com/dod-ccpo/atst/blob/master/templates/requests/screen-4.html#L117

templates/requests/screen-4.html
```diff
@@ -162,7 +239,13 @@ <h2>
 
     <div>
       <dt>Phone Number</dt>
-      <dd>{{ data['information_about_you']['phone_number'] or RequiredLabel() }}</dd>
+      <dd>
+        {% if data['information_about_you']['phone_number'] %}
```
dandds: Is there an easy way to do a macro for these, where you pass in the filter as an argument? Like `FilteredDataOrRequiredLabel(data['information_about_you']['phone_number'], usPhone)` ?

andrewdds: Possibly. I'm stretching my Python skills as it stands.

Not all of them take a single-word filter, some of them have multiple filters, etc. I don't have a sense of how hard it would be to macro-ize all that logic, or if it would be worth the effort for a single screen.

dandds: Gotcha. Yeah, I don't think it's worth a big overhaul on this PR.

PR #189, opened 2018-08-20
Option labels

andrewdds:
Adds a template filter for getting an option Label from the Value
Import `SERVICE_BRANCHES` into request form templates
Add a few other option-sets as lists, for consistency
Use `getOptionLabel` filter to display various labels


atst/routes/requests/requests_form.py
```diff
@@ -44,6 +48,10 @@ def requests_form_update(screen=1, request_id=None):
         request_id=request_id,
         jedi_request=jedi_flow.request,
         can_submit=jedi_flow.can_submit,
+        service_branches=SERVICE_BRANCHES,
+        assistance_org_types=ASSISTANCE_ORG_TYPES,
+        data_transfer_amounts=DATA_TRANSFER_AMOUNTS,
+        completion_date_ranges=COMPLETION_DATE_RANGES,
```
patricksmithdds: If we need to do this in more places, we could use a context processor to make the data available in every template: http://flask.pocoo.org/docs/1.0/templating/#context-processors

PR #190, opened 2018-08-20
Ui/reports summary layout

andrewdds:
Dummy route and layout for Funding & Reports screen summary panels.
screen shot 2018-08-20 at 4 46 30 pm

screen shot 2018-08-20 at 4 46 40 pm


templates/workspace_reports.html
```diff
@@ -0,0 +1,88 @@
...
+
+            <div>
+              <dt>Remaining</dt>
+              <dd>$60,000,000</dd>
+            </div>
+          </dl>
+        </div>
+
+        <div>
+          <meter value='40000000' min='0' max='100000000' title='$40,000,000 Total spend to date'></meter>
```
patricksmithdds: Looks like `meter` isn't supported at all in IE: https://caniuse.com/#search=meter

PR #191, opened 2018-08-21
Task Order number not found #159087193

dandds:
This implements [this](https://www.pivotaltracker.com/n/projects/2160940/stories/159087193) PT story.

The only valid TO number at the moment is "DCA10096D0052". This is defined in the `MockEDAClient`.

**notes**
- I decided to implement this as a standard validation. This means that the form has to be initialized with an instance of the EDA client so that `TaskOrders` can query the client.
- There are now two financial form classes. `ExtendedFinancialForm` expects the manual task order data but does not need a valid TO number, while `FinancialForm` expects the more limited field set but _does_ expect a valid TO number. The app decides which form object to use based on the presence of an `extended` URL parameter. This seems simpler to me than a bunch of branch logic within a single form.
- If the EDA client successfully returns TO data, we store the TO number in our database. 
- I standardized our internal names to refer to the TO number as "number" rather than "id". (That's what it is on the model.)
- I'm kicking the can a little bit on deciding how to maintain the distinction between manually-entered TO data and TO data that comes from EDA. (It seems like a distinction we need, but even that I'm not completely sure about.) There's a follow-on [story](https://www.pivotaltracker.com/n/projects/2160940/stories/158436096) to this one. I think this PR actually implements most of that follow-on, but we should use it as an opportunity to decide how we're storing the two different data sources.

screen shot 2018-08-21 at 10 27 21 am

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -8,12 +8,32 @@ from .exceptions import NotFoundError
 class TaskOrders(object):
 
     @classmethod
-    def get(self, order_number):
+    def get(cls, order_number, client=None):
         try:
             task_order = (
                 db.session.query(TaskOrder).filter_by(number=order_number).one()
             )
         except NoResultFound:
+            if client:
+                task_order = TaskOrders._get_from_eda(order_number, client)
+            else:
+                raise NotFoundError("task_order")
+
+        return task_order
+
+    @classmethod
+    def _get_from_eda(cls, order_number, client):
+        to_data = client.get_contract(order_number, status="y")
+        if to_data:
+            return TaskOrders.create(to_data["contract_no"])
+        else:
             raise NotFoundError("task_order")
+
+    @classmethod
+    def create(cls, order_number):
+        task_order = TaskOrder(number=order_number)
+
+        db.session.add(task_order)
+        db.session.commit()
 
         return task_order
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -1,4 +1,5 @@
 from sqlalchemy.orm.exc import NoResultFound
+from flask import current_app as app
 
 from atst.database import db
 from atst.models.task_order import TaskOrder
@@ -8,22 +9,22 @@ from .exceptions import NotFoundError
 class TaskOrders(object):
 
     @classmethod
-    def get(cls, order_number, client=None):
+    def get(cls, order_number):
         try:
             task_order = (
                 db.session.query(TaskOrder).filter_by(number=order_number).one()
             )
         except NoResultFound:
-            if client:
-                task_order = TaskOrders._get_from_eda(order_number, client)
+            if TaskOrders._client():
+                task_order = TaskOrders._get_from_eda(order_number)
             else:
                 raise NotFoundError("task_order")
 
         return task_order
 
     @classmethod
-    def _get_from_eda(cls, order_number, client):
-        to_data = client.get_contract(order_number, status="y")
+    def _get_from_eda(cls, order_number):
+        to_data = TaskOrders._client().get_contract(order_number, status="y")
         if to_data:
             return TaskOrders.create(to_data["contract_no"])
         else:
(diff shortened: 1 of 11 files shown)


templates/requests/financial_verification.html
```diff
@@ -9,6 +9,24 @@
...
+      level='warning'
+    ) }}
+  {% endif %}
+
+  {% if f.is_missing_task_order_number %}
+    {% set extended_url = url_for('requests.financial_verification', request_id=request_id, extended=True) %}
+    {{ Alert('Task Order not found in EDA',
+        message="We could not find your Task Order in our system of record, EDA.
+                 Please confirm that you have entered it correctly.
+                 <a href=\"%s\">Otherwise enter TO information manually.</a>
```
patricksmithdds: When the user clicks this to go to the extended form, all of their inputs are lost. Is there a way we can maintain what they entered?

dandds: I think we'd have to make it a POST. I was going to tackle that in the [follow-on story](https://www.pivotaltracker.com/story/show/159087193), but I can do it here if you think it's better to make this PR feature-complete.

patricksmithdds: Ok, it makes sense to do that in the story for manually entering the information instead (I assume that's the "follow-on story" -- the currently linked one is the one about adding the notice)

dandds: My bad. Yes, you're right. This one: https://www.pivotaltracker.com/n/projects/2160940/stories/158436096

atst/domain/task_orders.py
```diff
@@ -8,12 +8,32 @@
 class TaskOrders(object):
 
     @classmethod
-    def get(self, order_number):
+    def get(cls, order_number, client=None):
```
patricksmithdds: Rather than pass around the EDA API client, could we just access the `current_app` : `from flask import current_app`/ `current_app.eda_client`?

dandds: We could. I was talking about this with @richard-dds , and I think we favored dependency injection here, rather than have the domain objects know too much about Flask. I think I'll do `current_app` for now and we can revise later if we feel it's becoming too tightly coupled.

dandds: @patricksmithdds https://github.com/dod-ccpo/atst/pull/191/commits/7daa19e28348d7fa82c93e898d72faa4a8eb8a7e

atst/routes/requests/financial_verification.py
```diff
@@ -3,25 +3,35 @@
 
 from . import requests_bp
 from atst.domain.requests import Requests
-from atst.forms.financial import FinancialForm
+from atst.forms.financial import FinancialForm, ExtendedFinancialForm
+
+
+def financial_form(data):
+    if http_request.args.get("extended"):
```
patricksmithdds: I think this is fine, but just to clarify: if the URL is something like `?extended=False`, `http_request.args.get("extended")` evaluates to the string `"False"`, which is truthy. The only thing that could be falsy here is an empty string: `?extended=`.

atst/domain/task_orders.py
```diff
@@ -8,12 +9,36 @@
...
+    def create(cls, order_number):
+        task_order = TaskOrder(number=order_number)
+
+        db.session.add(task_order)
+        db.session.commit()
+
         return task_order
+
+    @classmethod
+    def _client(cls):
```
patricksmithdds: Is having this as a classmethod mainly for testing purposes?

atst/routes/requests/financial_verification.py
dandds: Yeah, I realized that in testing. I think it's fine if we're careful never to set "extended" to a value when we don't intend for it to be `True`. I can make it more robust if you want. (Might be useful to have a `getboolean` like `argparse` has.)

atst/domain/task_orders.py
dandds: Yup. Might also make it easier to change the implementation later if we don't want to rely on `current_app` anymore.

PR #194, opened 2018-08-21
Show currently active workspace name in label

patricksmithdds:
This branch updates the nav header to include the name of the workspace, if one is currently active:

image

This PR accomplishes this by adding a `context_processor` to the workspaces blueprint that checks for a `workspace_id` in the URL and uses that to look up the workspace. 

If there is no current workspace, then `workspace` will be undefined in the template. We can use this knowledge instead of manually defining a `navigationContext`, so I removed that variable.

Since we're no longer hard-coding `123456`, some of the regex's setting the active flag on links was no longer valid. Instead of a regex, I use something like `request.url_rule.rule.startswith('/workspaces/<workspace_id>/projects')`

### DIFF
--- a/atst/routes/workspaces.py
+++ b/atst/routes/workspaces.py
@@ -1,29 +1,19 @@
-from flask import Blueprint, render_template
+from flask import Blueprint, render_template, request as http_request
 
-from atst.domain.workspaces import Projects, Members
+from atst.domain.workspaces import Members, Projects, Workspaces
 
 
 bp = Blueprint("workspaces", __name__)
 
-mock_workspaces = [
-    {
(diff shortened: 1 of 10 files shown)


atst/routes/workspaces.py
```diff
@@ -1,45 +1,35 @@
...
 
-mock_workspaces = [
-    {
-        "name": "Unclassified IaaS and PaaS for Defense Digital Service (DDS)",
-        "id": "5966187a-eff9-44c3-aa15-4de7a65ac7ff",
-        "task_order": {"number": 123456},
-        "user_count": 23,
-    }
-]
+@bp.context_processor
```
richard-dds: Very cool!

PR #196, opened 2018-08-21
Update README with better setup instructions

patricksmithdds:
This updates the README with more clear requirements based off of feedback from Montana getting set up with the project. 

I think that README can now be used instead of the [now-archived wiki in `atat-stack`](https://github.com/dod-ccpo/atat-stack/wiki/Developer-Getting-Started).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,31 @@ This is the user-facing web application for ATAT.
 
 ## Installation
 
-### Requirements
-See the [scriptz](https://github.com/dod-ccpo/scriptz) repository for the shared
-requirements and guidelines for all ATAT applications.
+### System Requirements
+ATST uses the [Scripts to Rule Them All](https://github.com/github/scripts-to-rule-them-all)
+pattern for setting up and running the project. The scripts are located in the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,15 +8,31 @@ This is the user-facing web application for ATAT.
...
+`scripts` directory and use script fragments in the
+[scriptz](https://github.com/dod-ccpo/scriptz) repository that are shared across
+ATAT repositories.
+
+Before running the setup scripts, a couple of dependencies need to be installed
+locally:
+
+* `pipenv`
+  ATST requires `pipenv` to be installed for python dependency management. `pipenv`
+  will fetch and install the appropriate versions of Python and `pip`. [See
```
dandds: Wait, will it actually install a Python version your system doesn't have? I've been using `pyenv` for that.

patricksmithdds: You are correct -- `pipenv` does not actually install the version of Python necessary. I'll add Python 3.6 as a dependency as well.

patricksmithdds: Updated in 1af8331

PR #197, opened 2018-08-22
Create workspaces and projects

richard-dds:
Now when financial verification is submitted, a request is approved and a new workspace is created.

Users can view workspaces that they're assigned to, and create new projects and environments if they have the authority to do so.


js/components/forms/new_project.js
```diff
@@ -0,0 +1,46 @@
...
+  },
+
+  methods: {
+    addEnvironment: function (event) {
+      this.environments.push('')
+    },
+
+    removeEnvironment: function (index) {
+      if (this.environments.length > 1) {
+        this.environments.splice(index, 1)
```
dandds: This seems to always delete the environment you added most recently in the form, not the one you clicked the "X" for. i.e., if I have two environment and click the X for the first one, the second one disappears.

atst/routes/workspaces.py
```diff
@@ -1,35 +1,63 @@
...
+def update_project(workspace_id):
+    workspace = Workspaces.get_for_update(g.current_user, workspace_id)
+    form = NewProjectForm(http_request.form)
+
+    if form.validate():
+        project_data = form.data
+        project = Projects.create(
+            workspace, project_data["name"], project_data["description"]
+        )
+        Environments.create(project, project_data["environment_name"])
```
dandds: This only creates one environment per project, even if the user enters multiple environment names. I think we need `environment_name` to be a [`FieldList`](https://wtforms.readthedocs.io/en/latest/fields.html#wtforms.fields.FieldList) and then batch create them.

atst/domain/workspaces.py
```diff
@@ -1,71 +1,101 @@
...
-        return cls.MOCK_WORKSPACES
+    def get(cls, user, workspace_id):
+        try:
+            workspace = db.session.query(Workspace).filter_by(id=workspace_id).one()
+        except NoResultFound:
+            raise NotFoundError("workspace")
 
-class Projects(object):
+        if user not in workspace.users:
+            raise UnauthorizedError(user, "get workspace")
```
dandds: So the idea is that domain classes will be responsible for auth, and will rely on the `Authorization` to check specific permissions? I like that, I think. We should do a small follow-up PR to move the CCPO checks currently in the routes functions.

atst/routes/workspaces.py
```diff
@@ -1,35 +1,63 @@
...
 
 @bp.route("/workspaces")
 def workspaces():
-    return render_template("workspaces.html", page=5, workspaces=Workspaces.get_many())
+    workspaces = Workspaces.get_many(g.current_user)
+    return render_template("workspaces.html", page=5, workspaces=workspaces)
 
 
 @bp.route("/workspaces/<workspace_id>/projects")
 def workspace_projects(workspace_id):
```
dandds: Low priority, but maybe we want a `/workspaces/<workspace_id>` route that redirects here, since it's the default view for a workspace?

atst/routes/workspaces.py
richard-dds: Yeah, I should have mentioned that the reason the environments section of the new project form is half-baked is because Patrick reminded me that it's actually out of scope for the story I was working on. My story only included creating one environment, while another upcoming story calls for creating multiple. I got as far as adding the Vue component and some of the logic for making the environments field list dynamic.

dandds: Oh, I gotcha. Disregard, then.

atst/domain/workspaces.py
richard-dds: That's what I prefer as well, but I'm open to discussing it (@patricksmithdds ).

I think that some of the domain classes are covering two big responsibilities now: handling domain logic and dealing with the database. Because of that, I think we had a tendencey to treat those classes more as CRUDy DB repositories. My opinion is that we should put just about all of the non-HTTP logic into domain methods so that we can test those methods as our top-level public interfaces. That way we can refactor everything under that layer without breaking tests. Integration tests can take care of everything HTTP-related.

atst/routes/workspaces.py
richard-dds: Good call! I'll add that.

atst/domain/workspace_users.py
```diff
@@ -21,7 +21,8 @@ def get(cls, workspace_id, user_id):
 
         try:
             workspace_role = (
-                WorkspaceRole.query.join(User)
+                db.session.query(WorkspaceRole)
```
patricksmithdds: Just curious -- is there a difference between `WorkspaceRole.query` and `db.session.query(WorkspaceRole)`?

atst/domain/workspaces.py
```diff
@@ -1,71 +1,101 @@
...
-        {
-            "name": "Unclassified IaaS and PaaS for Defense Digital Service (DDS)",
-            "id": "5966187a-eff9-44c3-aa15-4de7a65ac7ff",
-            "task_order": {"number": 123456},
-            "user_count": 23,
-        }
-    ]
+    # will a request have a TO association?
+    # do we automatically create an entry for the request.creator in the
+    # workspace_roles table?
```
patricksmithdds: Seems like the answer to this question is "yes" -- the requestor is added as the owner of a workspace.

atst/domain/workspaces.py
```diff
@@ -1,71 +1,101 @@
...
     @classmethod
-    def get(cls, project_id):
-        pass
+    def get_by_request(cls, request):
+        try:
+            workspace = db.session.query(Workspace).filter_by(request=request).one()
+        except NoResultFound:
+            raise NotFoundError("workspace")
+
+        return workspace
```
patricksmithdds: I think this needs to do the permissions check as well before returning.

atst/domain/workspaces.py
patricksmithdds: I like the idea of relying on `Authorization` to check permissions.

In that case, is it possible to do that here? Instead of `if user not in workspace.users` do something like `Authorization.has_workspace_permission(user, workspace, Permissions.VIEW_WORKSPACE)`?

atst/domain/workspaces.py
```diff
@@ -1,71 +1,101 @@
...
+        workspaces = (
+            db.session.query(Workspace)
+            .join(WorkspaceRole)
+            .filter(WorkspaceRole.user == user)
+            .all()
+        )
+        return workspaces
 
+
+class Projects(object):
```
patricksmithdds: Should these be moved out into their own `domain` files?

atst/models/workspace.py
```diff
@@ -0,0 +1,33 @@
...
+    id = Id()
+    name = Column(String, unique=True)
+    request_id = Column(ForeignKey("requests.id"), nullable=False)
+    request = relationship("Request")
+    projects = relationship("Project", back_populates="workspace")
+    roles = relationship("WorkspaceRole")
+
+    @property
+    def owner(self):
+        return self.request.creator
```
patricksmithdds: In the future, we'll be able to transfer ownership of a workspace between users. If it's not too much effort now, could we find the user with some "owner" workspace role instead?

atst/routes/workspaces.py
```diff
@@ -1,35 +1,78 @@
...
+def show_workspace(workspace_id):
+    return redirect(url_for("workspaces.workspace_projects", workspace_id=workspace_id))
 
 
 @bp.route("/workspaces/<workspace_id>/members")
 def workspace_members(workspace_id):
-    members = Members.get_many(workspace_id)
-    return render_template("workspace_members.html", members=members)
+    workspace = Workspaces.get(g.current_user, workspace_id)
+    members_repo = Members()
```
patricksmithdds: Does this still need to be instantiated? Can we just do `Members.get_many(workspace_id)`?

atst/routes/workspaces.py
richard-dds: Added

templates/workspace_project_new.html
```diff
@@ -0,0 +1,62 @@
...
+        </header>
+
+        <ul>
+          <li v-for="(_, i) in environments" class="block-list__item project-list-item__environment">
+            <span class="project-list-item__environment">
+              {{ TextInput(form.environment_name) }}
+              <span v-on:click="removeEnvironment(i)">{{ Icon('x') }}</span>
+            </span>
+          </li>
+        </ul>
```
patricksmithdds: It looks like some of the styles around this page could be tweaked. For example:

image

Is there any plan to get some polish/sparkle from @luisgov or @andrewdds ?

js/components/forms/new_project.js
richard-dds: What I'm seeing on my end is that I can remove environments correctly in all cases except for the first one. When I try to remove the first one it looks like the last one is removed for some reason. I'm not sure why that's happening though. Am I using `splice()` incorrectly?

atst/domain/workspaces.py
richard-dds: Yeah, definitely. I'll abstract that and put it into `Authorization`.

atst/domain/workspaces.py
richard-dds: So far this method isn't being called anywhere except some tests. Not sure if it's better to update it or just remove it.

js/components/forms/new_project.js
patricksmithdds: I think `splice` is correct, but it may have to do with how Vue renders lists of items -- you may need to add a key to each item so that the correct value is removed. We could push that to story about adding+deleting multiple environments when creating a project.

templates/workspace_project_new.html
```diff
@@ -0,0 +1,62 @@
...
+            "Each environment created within a project is an enclave of cloud resources that is logically separated from each other for increased security.",
+            title="learn more"
+          )}}
+
+        </header>
+
+        <ul>
+          <li v-for="(_, i) in environments" class="block-list__item project-list-item__environment">
+            <span class="project-list-item__environment">
+              {{ TextInput(form.environment_name) }}
```
andrewdds: Do we need to use the TextInput macro here? Its main responsibility other than uniform markup is the masking/validation. Looks like the validation icons are screwing with the layout.

templates/workspace_project_new.html
andrewdds: Please... don't say sparkle

andrewdds: Also, I had already created more specific markup for this screen. Looks like it was not used. https://github.com/dod-ccpo/atst/pull/108

templates/workspace_project_new.html
richard-dds: Yeah, I think we'll end up dropping that macro when we focus on the "add multiple environments" story. In fact, if we don't mind ignoring some wonky markup and styling, it may make sense to wait until that story (this sprint) to finish polishing this form's UI.

templates/workspace_project_new.html
patricksmithdds: sparkle

andrewdds:
Perhaps this is better for another PR, but I think the "Workspaces" navigation item should not appear unless you have active workspaces. Unlike Projects or Members, you can't create them directly, so it doesn't make sense to have an empty state screen for this.

atst/domain/workspace_users.py
richard-dds: db.session.query works and WorkspaceRole.query doesn't :)

It might be something that worked before our big refactor, but doesn't anymore. There are a bunch of different ways to query in SQLAlchemy, which can make things a little confusing.

templates/workspace_project_new.html
luisgov: @andrewcroce i've made some updates using that markup. running into an issue with the remove button, looking into it.

templates/workspace_project_new.html
luisgov: @richard-dds might be better to remove it then and add it back when we are ready for that story. i think the issues with the styling will just create more confusion.

patricksmithdds:
> Perhaps this is better for another PR, but I think the "Workspaces" navigation item should not appear unless you have active workspaces.

That's a good idea. I created a separate story for that in pivotal: https://www.pivotaltracker.com/story/show/159967260

atst/domain/workspaces.py
patricksmithdds: I'd say we can remove until we need it.

PR #198, opened 2018-08-22
Ui/reports spend table

andrewdds:
Adds dummy markup for spend table.
Adds general-purpose `toggler` vue component which is used to implement the env row hiding/showing.

screen shot 2018-08-22 at 11 18 34 am

screen shot 2018-08-22 at 11 18 41 am


templates/workspace_reports.html
```diff
@@ -85,4 +86,109 @@ <h2 class='to-summary__heading'>Task Order</h2>
...
+          <td class='table-cell--align-right previous-month'>$58,000</td>
+          <td class='table-cell--align-right previous-month'>$60,000</td>
+          <td class='table-cell--align-right current-month'>$62,000</td>
+          <td class='table-cell--expand current-month meter-cell'>
+            <meter value='62000' min='0' max='62000'></meter>
+          </td>
+        </tr>
+      </tbody>
+
+      <tbody is='toggler' class='spend-table__project'>
```
patricksmithdds: Why is it necessary to use `<tbody is='toggler'` instead of `<toggler`? I've not seen `is` before, but it looks like it's not supported in IE: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/is

templates/workspace_reports.html
```diff
@@ -85,4 +86,109 @@ <h2 class='to-summary__heading'>Task Order</h2>
...
+        <template slot-scope='{ isVisible, toggle }'>
```
patricksmithdds: `slot-scope` is pretty nifty!

js/components/toggler.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+  },
+
+  render: function (createElement) {
+    return createElement( this.$vnode.data.tag, [
+      this.$scopedSlots.default({
+        isVisible: this.isVisible,
+        toggle: this.toggle
+      })
+    ])
```
patricksmithdds: Is this magic to support the `is` attribute or is there another reason this is necessary?

js/components/toggler.js
```diff
@@ -0,0 +1,28 @@
...
+
+  methods: {
+    toggle: function (e) {
+      this.isVisible = !this.isVisible
+    }
+  },
+
+  mounted: function () {
+    console.log(this)
+  }
```
patricksmithdds: We can axe this function.

js/components/toggler.js
andrewdds: I used the render method so I could selectively render the same element that was assigned as the component. It makes the `toggler` more general purpose without explicitly needing to repeat the element type as a prop, for instance.

Instead of something like:
```
<tbody is="toggler" element="tbody">
```
This render method is letting me just detect the actual element thats used, and render that. So if you do:
```
<button is="toggler">
```
it'll render a `button` element.

templates/workspace_reports.html
patricksmithdds: Makes sense, thanks. Perhaps in IE, we could default to showing the expanded version that won't be interactive?

js/components/toggler.js
patricksmithdds: ✨ Magic. :sparkles: Thanks for the explanation!

PR #200, opened 2018-08-22
Manually enter task order information #158436096

dandds:
This is the PT story: https://www.pivotaltracker.com/n/projects/2160940/stories/158436096

If a user enters a task order that can't be found in the EDA API, they can navigate to an extended version of the financial verification form and enter the task order information manually.

This PR fleshes out the task order table and stores the Task Order data from the financial verification form there (funding type, CLIN info). 

I talked to @richard-dds about this, but the flow feels clunky to me for a couple reasons:
- The financial verification form now contains data for two models (`Request` and `TaskOrder`) , and so `Requests.update_financial_verification` is responsible for splitting the data appropriately.
- Since we do specialized validation for the PE number and Task Order number on the form objects, the domain layer can't automatically determine if financial verification is complete or not, which feels wrong. I think we should refactor with an eye toward moving those specialized validations into the domain layer somehow.

Right now, if you enter the one Task Order number the app can fetch from the mock EDA client (DCA10096D0052), it only populates the `number` column of the resulting task order model. This is because the data we were getting from the EDA early on and have mocked out in the client does not completely match the data we'd like now. I added a chore for resolving this: https://www.pivotaltracker.com/story/show/159969717

screenshot_2018-08-22 jedi cloud


alembic/versions/0845b2f0f401_add_request_task_order_relationship.py
```diff
@@ -0,0 +1,30 @@
+"""add request -> task order relationship
```
patricksmithdds: I believe these migrations are going to conflict with migrations added in #197, so we have to be careful about updating the migrations after either one gets merged first.

atst/domain/requests.py
```diff
@@ -201,3 +215,47 @@ def pending_ccpo_count(cls):
...
+            request.task_order = task_order
+            db.session.add(task_order)
+
+        Requests._merge_body(request, {"financial_verification": request_data})
+
+        db.session.add(request)
+        db.session.commit()
+
+    @classmethod
+    def _get_or_create_task_order(cls, number, task_order_data={}):
```
patricksmithdds: We [shouldn't use mutable objects](http://effbot.org/zone/default-values.htm) as a default value. `None` should be used instead of `{}` (even though we're not modifying `task_order_data`, it's best to be safe)

alembic/versions/0845b2f0f401_add_request_task_order_relationship.py
dandds: Yeah, I'm probably going to have some conflicts with #197. We should merge that first and I'll resolve them here.

atst/domain/requests.py
dandds: Oh weird, thanks. That's a python quirk I didn't know.

dandds: Updated: https://github.com/dod-ccpo/atst/pull/200/commits/a1c5a0a30fbd86a8b502f684baf65287a2d0145c

dandds:
I rebased this and fixed the migration chain and some other stuff. Should be merge-able now.

atst/forms/financial.py
```diff
@@ -154,35 +170,41 @@ def validate(self, *args, **kwargs):
     clin_0001 = StringField(
         "<dl><dt>CLIN 0001</dt> - <dd>Unclassified IaaS and PaaS Amount</dd></dl>",
         validators=[Required()],
-        description="Review your task order document, the amounts for each CLIN must match exactly here"
+        description="Review your task order document, the amounts for each CLIN must match exactly here",
+        filters=[formatted_number_to_int]
```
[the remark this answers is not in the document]
dandds: You're right, they don't come in with commas from the form itself. Some of our test fixture data has commas, so it needs to be updated. In either case, we still need to coerce the strings to integers.

atst/domain/requests.py
```diff
@@ -201,3 +215,47 @@ def pending_ccpo_count(cls):
...
+    def update_financial_verification(cls, request_id, financial_data):
+        request = Requests._get_with_lock(request_id)
+        if not request:
+            return
+
+        request_data = financial_data.copy()
+        task_order_data = {k: request_data.pop(k) for (k,v) in financial_data.items() if k in Requests._TASK_ORDER_DATA}
+        task_order_number = request_data.pop("task_order_number")
+
+        task_order = Requests._get_or_create_task_order(task_order_number, task_order_data)
```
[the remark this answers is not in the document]
dandds: I suppose we can. It felt like a `Requests` concern to me since we're deciding based on whether or not there's any task order data there.

atst/domain/requests.py
```diff
@@ -201,3 +215,47 @@ def pending_ccpo_count(cls):
...
+
+        if task_order:
+            request.task_order = task_order
+            db.session.add(task_order)
```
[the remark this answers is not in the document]
dandds: I think so? The request is the one that holds the reference to the task order. I'll try it without.

atst/domain/requests.py
```diff
@@ -201,3 +215,47 @@ def pending_ccpo_count(cls):
...
+        db.session.add(request)
+        db.session.commit()
+
+    @classmethod
+    def _get_or_create_task_order(cls, number, task_order_data={}):
+        if task_order_data:
+            return TaskOrders.create(**task_order_data, number=number, source=TaskOrderSource.MANUAL)
+        else:
+            try:
+                return TaskOrders.get(number)
```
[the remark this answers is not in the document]
dandds: I think that's right. I was trying to minimize the number of database queries, since in our current workflow we _know_ that we need to make the task order if this method is given any data in that dictionary. But you're probably right insofar as we should privilege design at this point.

atst/domain/requests.py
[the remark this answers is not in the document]
dandds: Oh, oh, I misunderstood you. No, you're right, the save cascades.

dandds:
@patricksmithdds I made some updates based on your feedback. 

I think it feels better in general, but we still need [this](https://github.com/dod-ccpo/atst/blob/f32a3eb90dbb19fefe60b479d6e7c70fde3b1a0b/atst/domain/task_orders.py#L55) guard so that we don't add an invalid task order number if the user is filling out the initial financial verification page. I don't love that and I'm wondering if we can improve the workflow there.

atst/domain/requests.py
```diff
@@ -201,11 +213,13 @@ def status_count(cls, status, creator=None):
 
     @classmethod
     def in_progress_count(cls):
-        return sum([
-            Requests.status_count(RequestStatus.STARTED),
-            Requests.status_count(RequestStatus.PENDING_FINANCIAL_VERIFICATION),
-            Requests.status_count(RequestStatus.CHANGES_REQUESTED),
-        ])
+        return sum(
```
richard-dds: I don't think it should hold up this PR, but we should consider refactoring `Requests.status_count` to use SQLAlchemy, accept a list of statuses to filter by, and then call that from this method. That way we can make one query instead of three.

atst/domain/requests.py
```diff
@@ -215,3 +229,43 @@ def pending_ccpo_count(cls):
...
+            for (k, v) in financial_data.items()
+            if k in TaskOrders.TASK_ORDER_DATA
+        }
+        task_order_number = request_data.pop("task_order_number")
+
+        task_order = TaskOrders.get_or_create_task_order(
+            task_order_number, task_order_data
+        )
+
+        if task_order:
```
richard-dds: Is this check required here? It seems like `get_or_create_task_order` always returns a TaskOrder.

atst/domain/requests.py
dandds: I agree. That's from a previous PR, though. I think it's only showing up in the diff because I ran the formatter.

atst/domain/requests.py
dandds: No, it won't return one if there task order number can't be found and there's no task order data. This is because a user might submit the initial financial verification form with an invalid task order number, and we don't want to create a task order model in that case.

PR #204, opened 2018-08-22
Internet explorer fixes

andrewdds:
This fixes a lot of the glaring display bugs in Internet Explorer. It's focused on IE10/Windows 7, and IE11/Windows 10.

Also adds babel polyfill which should fix some javascript errors we were seeing.

There are undoubtedly more bugs, but this should be a good start.

### DIFF
--- a/.postcssrc
+++ b/.postcssrc
@@ -1,0 +1,5 @@
+{
+  "plugins": {
+    "autoprefixer": true
+  }
+}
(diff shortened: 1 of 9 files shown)


patricksmithdds:
There's a couple things I noticed when poking around on this branch on browserstack.

Looks like something funky happened to the requests index:

image

The date is in the wrong format for some reason: 

image

The new project page is empty:

image

The edit member page is even more empty:

image

PR #205, opened 2018-08-23
UI/members

luisgov:
This PR adds a static member edit/profile screen. Right not it is currently hardcoded for a dummy user profile. 

screen shot 2018-08-23 at 16 27 41


templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
...
+      <div class='project-list-item__environment__actions'>
+        <span class="label label--success">Billing</span><a href="#" class="icon-link">set role</a>
+      </div>
+    </li>
+  </ul>
+</div>
+
+<div class='block-list project-list-item'>
+  <header class='block-list__header'>
+    <h2 class='block-list__title'>{{ Icon('arrow-right') }} Digital Dojo</h2>
```
patricksmithdds: Is it possible to use the `toggler` jawn that @andrewdds added in #198 here?

templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
+{% extends "base.html" %}
+
+{% from "components/alert.html" import Alert %}
```
patricksmithdds: This can be removed if we're not using an Alert anywhere.

templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
...
+        <dt>Email:</dt>
+        <dd>knight@mil.gov</dd>
+      </div>
+    </dl>
+    <a href='#' class='icon-link'>edit account details</a>
+  </div>
+</div>
+
+
+<form class='search-bar'>
```
andrewdds: I think I agree with @kristiandds comments in Invision that this would work better directly above the list of projects.

templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
...
+    <input type='search' id='project-search' name='project-search' placeholder="Search by project name"/>
+    <button type="submit">
+      <span class="hide">Search</span>
+    </button>
+  </div>
+</form>
+
+<div class="panel">
+  <div class="panel__heading">
+    <h3>Manage Access</h3>
```
andrewdds: The heading levels are off I think. This should be an `h2`, and each project should be an `h3`.

templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
...
+    <div class="subtitle"><h3>Grant access to an environment</h3></div>
+  </div>
+</div>
+
+<div class='block-list project-list-item'>
+  <header class='block-list__header'>
+    <h2 class='block-list__title'>{{ Icon('arrow-down') }}Code.mil</h2>
```
andrewdds: For consistency with the toggleable table rows in Reports page, should we:
1) Use an `.icon-button .icon-button__large` here?
2) Decide which arrow to use? (I'm using the carat in the Report table, not sure I prefer one or the other).

Also, should we incorporate the toggle behavior yet? or save that for another PR?

templates/member_edit.html
luisgov: 👍

templates/member_edit.html
luisgov: 👍

templates/member_edit.html
luisgov: 👍

templates/member_edit.html
luisgov: good call

templates/member_edit.html
luisgov: i think you meant `.icon-link .icon-link__large`. done and added toggle behavior.

templates/member_edit.html
```diff
@@ -0,0 +1,98 @@
...
+
+{% from "components/alert.html" import Alert %}
+{% from "components/icon.html" import Icon %}
+
+{% block content %}
+
+<div class='panel member-card'>
+  <div class='member-card__header'>
+    <h1 class='member-card__heading'>Danny Knight</h1>
+    <dl><dt>Workspace Role</dt> <dd><span class='label label--info'>Billing Auditor</span></dd></dl>
```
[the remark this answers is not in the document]
luisgov: good call. added.

patricksmithdds:
@luisgov What does the "Save" button on the bottom of the page apply to?

luisgov:
@patricksmithdds right now the save button is just a placeholder. it should save a new member, or save updates from a  new member if edits are made.

PR #206, opened 2018-08-23
Allow a user to add multiple environments when creating a project

richard-dds:
screen shot 2018-08-23 at 11 58 06 am

### DIFF
--- a/templates/workspace_project_new.html
+++ b/templates/workspace_project_new.html
@@ -34,13 +34,14 @@
         </header>
 
         <ul>
-          <li v-for="(_, i) in environments" class="block-list__item">
-              {{ TextInput(form.environment_name) }}
+          <li v-for="(environment, i) in environments" class="block-list__item">
+              {{ form.environment_names.label }}
+              <textarea :name="'environment_names-' + i" :id="'environment_names-' + i" v-model="environment.name"></textarea>
               <span class="icon-link icon-link--danger icon-link--vertical" v-on:click="removeEnvironment(i)">{{ Icon('x') }} Remove</span>
-
           </li>
         </ul>
 
+        {{ form.environment_names.errors }}
         <div class="block-list__footer">
           <a v-on:click="addEnvironment" class="icon-link">Add another environment</a>
         </div>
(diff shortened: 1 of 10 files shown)


templates/workspace_project_new.html
```diff
@@ -34,10 +39,15 @@ <h2 class="block-list__title">Project Environments</h2>
...
-          <li v-for="(_, i) in environments" class="block-list__item">
-              {{ TextInput(form.environment_name) }}
-              <span class="icon-link icon-link--danger icon-link--vertical" v-on:click="removeEnvironment(i)">{{ Icon('x') }} Remove</span>
-
+          <li v-for="(environment, i) in environments" class="block-list__item project-edit__env-list-item">
+            <div class="usa-input">
+              <label :for="'environment_names-' + i">Environment Name</span>
+              <input type="text" :name="'environment_names-' + i" :id="'environment_names-' + i" v-model="environment.name"></input>
+            </div>
+              <button v-on:click="removeEnvironment(i)" v-if="i > 0" type="button" class='project-edit__env-list-item__remover'>
```
patricksmithdds: This should allow deleting any environment if there is more than one environment, so I think the `v-if="i > 0"` should be something like `v-if="environments.length > 1"`.

tests/domain/test_projects.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def test_create_project_with_multiple_environments():
+    workspace = WorkspaceFactory.create()
+    project = Projects.create(workspace, "My Test Project", "Test", ["dev", "prod"])
+
+    assert project.workspace == workspace
+    assert project.name == "My Test Project"
+    assert project.description == "Test"
+    assert [e.name for e in project.environments] == ["dev", "prod"]
```
patricksmithdds: Is the order of environments the same every time? Is it possible this could return `["prod", "dev"]` instead?

richard-dds: I did wonder about whether or not the order was deterministic, but I ran the tests 5 times and they passed each time. I'll find a better solution. :)

PR #207, opened 2018-08-23
Pull workspace owner information from Information About You Step

richard-dds:
If the user designates themselves as the workspace owner (on step 3 of the form), and they've entered their personal information on the "Information About You" section, then that information will be used to populate the workspace owner fields.

### DIFF
--- a/tests/assert_util.py
+++ b/tests/assert_util.py
@@ -1,0 +1,5 @@
+def dict_contains(superset, subset):
+    """
+    Returns True if dict a is a superset of dict b.
+    """
+    return all(item in superset.items() for item in subset.items())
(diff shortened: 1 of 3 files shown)


tests/assert_util.py
```diff
@@ -0,0 +1,5 @@
+def dict_contains(superset, subset):
+    """
+    Returns True if dict a is a superset of dict b.
+    """
+    return all(item in superset.items() for item in subset.items())
```
patricksmithdds: We could use sets for this as well: `return set(subset.items()) <= set(superset.items())` or `return set(subset.items()).issubset(superset.items()))`.

PR #213, opened 2018-08-24
Redirect to projects/new after FV-related request approval

richard-dds:
After a user updates their financial verification information and the request is approved, they are now redirected right to the new project form, rather than the project listing page. The form also now shows the "workspace created!" modal.

The copy that was hidden in the tooltips is now inline.

I also added a bonus validation which ensures that environment names are unique. Includes a small tweak to `components/alert` to allow us to pass blocks rather than just a message string.

screen shot 2018-08-24 at 2 27 36 pm

### DIFF
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -64,8 +64,7 @@ def ListItemRequired(message="Please provide at least one.", empty_values=("", N
 
 def ListItemsUnique(message="Items must be unique"):
     def _list_items_unique(form, field):
-        sorted_values = sorted(v for v in field.data)
-        if sorted_values != sorted(set(sorted_values)):
+        if len(field.data) > len(set(field.data)):
             raise ValidationError(message)
 
     return _list_items_unique
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -60,3 +60,12 @@ def ListItemRequired(message="Please provide at least one.", empty_values=("", N
             raise ValidationError(message)
 
     return _list_item_required
+
+
+def ListItemsUnique(message="Items must be unique"):
+    def _list_items_unique(form, field):
+        sorted_values = sorted(v for v in field.data)
+        if sorted_values != sorted(set(sorted_values)):
+            raise ValidationError(message)
+
+    return _list_items_unique
--- a/templates/workspace_project_new.html
+++ b/templates/workspace_project_new.html
@@ -9,6 +9,14 @@
 {% block workspace_content %}
 
 {% set modalName = "newProjectConfirmation" %}
+{% if request.args.get("newWorkspace") %}
+  {{ Alert('Workspace created!',
+    message="\
+      <p>You are now ready to create projects and environments within the JEDI Cloud.</p>
+    ",
+    level='success'
+  ) }}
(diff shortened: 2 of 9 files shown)


atst/forms/validators.py
```diff
@@ -60,3 +60,12 @@ def _list_item_required(form, field):
             raise ValidationError(message)
 
     return _list_item_required
+
+
+def ListItemsUnique(message="Items must be unique"):
+    def _list_items_unique(form, field):
+        sorted_values = sorted(v for v in field.data)
+        if sorted_values != sorted(set(sorted_values)):
```
grubnubble: so mathy! 👍

montana-mil: so mathy! 👍

tests/forms/test_validators.py
```diff
@@ -51,3 +51,22 @@ def test_Alphabet_rejects_non_letters(self, invalid, dummy_form, dummy_field):
...
+    @pytest.mark.parametrize("valid", [["a", "aa", "aaa"], ["one", "two", "three"]])
+    def test_ListItemsUnique_allows_unique_items(self, valid, dummy_form, dummy_field):
+        validator = ListItemsUnique()
+        dummy_field.data = valid
+        validator(dummy_form, dummy_field)
+
+    @pytest.mark.parametrize(
+        "invalid", [["a", "a", "a"], ["one", "two", "two", "three"]]
+    )
+    def test_ListItemsUnique_rejects_non_letters(
```
montana-mil: isn't this being rejected because there are mulitple "a"s and "two"s? why is it called `test_ListItemsUnique_rejects_non_letters`?

montana-mil: oh maybe a copy/paste error?

atst/forms/validators.py
richard-dds: Just realized that I could make this a little simpler by comparing `len()` to`len(set())`.

atst/forms/validators.py
```diff
@@ -64,8 +64,7 @@ def _list_item_required(form, field):
 
 def ListItemsUnique(message="Items must be unique"):
     def _list_items_unique(form, field):
-        sorted_values = sorted(v for v in field.data)
-        if sorted_values != sorted(set(sorted_values)):
+        if len(field.data) > len(set(field.data)):
```
montana-mil: 💯

templates/workspace_project_new.html
```diff
@@ -32,30 +40,31 @@ <h1>Are you sure you want to create <strong>!{ name }</strong> project?</h1>
...
         {{ TextInput(form.name) }}
         {{ TextInput(form.description, paragraph=True) }}
       </div>
     </div>
 
     {% if form.environment_names.errors %}
-      {{ Alert("Missing Environments", message="Provide at least one environment name.", level="error") }}
+      {% call Alert("Missing Environments", level="error") %}
+        {% for error in form.environment_names.errors %}
+          <p>{{ error }}</p>
```
dandds: It's a little weird that you can have an alert like

```
Missing Environments
Environment names must be unique.
```

but I don't think it should hold up this PR.

PR #214, opened 2018-08-24
Add new members

luisgov:
This PR adds the ability to create and add new members to a workspace. 
Currently this is only static form fields and does not create a member record.

screen shot 2018-08-24 at 14 31 33


patricksmithdds:
@luisgov Is it possible to rebase this branch on master so the diff shows only the relevant changes?

PR #215, opened 2018-08-27
record invalid task order number on the request body

dandds:
This addresses a problem noted [here](https://www.pivotaltracker.com/story/show/158436096/comments/193453188) in PT.

If a user submits an invalid TO number on the basic financial verification form, we don't save it anywhere right now. That means that the field is blank when they move to the extended financial verification form. This PR changes the behavior so that we record the invalid task order number on the request body. That way the field is populated when the user moves to the extended form.

This is probably a temporary fix until we add a real schema for `requests`.


dandds:
We have a separate chore for this: https://www.pivotaltracker.com/story/show/160061299

PR #216, opened 2018-08-27
UI/copy edits and new project form accessibility

luisgov:
This addressed the following items:
* A fix for _Change the Request submission to 3-day Response Time_ https://www.pivotaltracker.com/story/show/159808710
* Adds accessibility support for navigating through the _New Project_ from using the keyboard (tabbing)
* A more descriptive confirmation modal message when creating a new project

---

screen shot 2018-08-27 at 13 37 17

### DIFF
--- a/templates/workspace_project_new.html
+++ b/templates/workspace_project_new.html
@@ -21,13 +21,14 @@
 <new-project inline-template v-bind:initial-data='{{ form.data|tojson }}'>
   <form method="POST" action="{{ url_for('workspaces.update_project', workspace_id=workspace.id) }}" >
     {% call Modal(name=modalName, dismissable=False) %}
-      <h1>Are you sure you want to create <strong>!{ name }</strong> project?</h1>
+      <h1>Create project !{ name }</h1>
 
       <p>
-      When you click "Create Project" the following environments will be created as individual cloud resource groups under <strong>!{ name }</strong> project:
+      When you click <em>Create Project</em>, the environments
       <span v-for="(environment, index) in environments">
         <strong>!{environment.name}</strong><template v-if="index < (environments.length - 1)">, </template>
       </span>
+      will be created as individual cloud resource groups under <strong>!{ name }</strong> project.
       </p>
 
       <div class='action-group'>
--- a/templates/workspace_project_new.html
+++ b/templates/workspace_project_new.html
@@ -32,8 +32,8 @@
       </p>
 
       <div class='action-group'>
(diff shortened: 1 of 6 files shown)


templates/workspace_project_new.html
```diff
@@ -21,18 +21,19 @@
...
         <strong>!{environment.name}</strong><template v-if="index < (environments.length - 1)">, </template>
       </span>
+      will be created as individual cloud resource groups under <strong>!{ name }</strong> project.
       </p>
 
       <div class='action-group'>
-        <button type='submit' class='action-group__action usa-button'>Create Project</button>
-        <a v-on:click="closeModal('{{ modalName }}')" class='action-group__action'>Cancel</a>
+        <button type='submit' class='action-group__action usa-button' tabindex='0'>Create Project</button>
+        <button v-on:click="closeModal('{{ modalName }}')" class='icon-link action-group__action' tabindex='0'>Cancel</button>
```
patricksmithdds: Any `button` that's not a `submit` button needs to specify a type, otherwise it'll submit the form. Right now, clicking "cancel" is the same as clicking "create project"

templates/workspace_project_new.html
```diff
@@ -21,18 +21,19 @@
...
-      <h1>Are you sure you want to create <strong>!{ name }</strong> project?</h1>
+      <h1>Create project !{ name }</h1>
 
       <p>
-      When you click "Create Project" the following environments will be created as individual cloud resource groups under <strong>!{ name }</strong> project:
+      When you click <em>Create Project</em>, the environments
       <span v-for="(environment, index) in environments">
         <strong>!{environment.name}</strong><template v-if="index < (environments.length - 1)">, </template>
       </span>
+      will be created as individual cloud resource groups under <strong>!{ name }</strong> project.
```
patricksmithdds: I'm just curious -- where did this copy change come from? The previous copy is the one from the pivotal story, so I'm just wondering if that was previously out of date or if there's another process I'm unaware of.

luisgov: The original text was confusing and not user friendly. For example we should not prompt users with messages with _Are you sure..._ it shifts blame to the user and adds cognitive load. The revised copy was approved by @anil-dds and @jordangov through Slack. However, I do think we need better process to get this stuff revised...

**NEW:**
screen shot 2018-08-27 at 13 37 17


**OLD:**
screen shot 2018-08-27 at 10 25 43

templates/workspace_project_new.html
luisgov: 👍

PR #218, opened 2018-08-27
Empty states and some permissions

andrewdds:
This started as just adding the correct `EmptyState` component to the Projects list. It morphed into checking permissions in the template; implied in that functionality is the necessity of checking whether a user has permission to do the thing in the button, such as "Add a New Project".

This adds some context to the templates:
- List of workspace permission constants, i.e. `permissions.DO_A_THING`
- `user_can` method that accepts one of the above permission constants, and checks whether the current user, in the current workspace, can do said thing.

The parameters of the `EmptyState` components on the Members and Projects lists are modified depending on whether the user has permission to invite members or create projects.

The Workspace Navigation sidebar subnavs are also modified, so the "+ Add Project" or "+ Add Member" buttons only appear if the user has permission.

Also, change params to camel_case in EmptyState component

screen shot 2018-08-27 at 2 00 44 pm

screen shot 2018-08-27 at 2 40 32 pm

### DIFF
--- a/atst/routes/workspaces.py
+++ b/atst/routes/workspaces.py
@@ -17,6 +17,7 @@ from atst.models.permissions import Permissions
 
 bp = Blueprint("workspaces", __name__)
 
+
 @bp.context_processor
 def workspace():
     workspace = None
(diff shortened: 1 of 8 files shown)


atst/routes/workspaces.py
```diff
@@ -21,7 +23,15 @@ def workspace():
         workspace = Workspaces.get(
             g.current_user, http_request.view_args["workspace_id"]
         )
-    return {"workspace": workspace}
+
+    def user_can(permission):
```
patricksmithdds: This is great!

andrewdds:
@patricksmithdds I think writing tests in python is beyond my comfortable skillset. Do you want to (or want someone else to) add tests to this PR?

montana-mil:
@andrewcroce i can do tests!

PR #221, opened 2018-08-27
move pe number ingestion for setup into a migration

dandds:
This should take care of [this](https://www.pivotaltracker.com/story/show/160063695) bug. It moves PE number ingestion into a migration so that it will get run during our CD. I think both schema and state are fair game for migrations, and having the list of PE numbers is required initial state for the app. I got rid of related code in `PENumbers` that's not used anywhere else.

### DIFF
--- a/script/update
+++ b/script/update
@@ -9,6 +9,3 @@ MIGRATE_DB="true"
 
 # Run the shared update script
 source ./script/include/run_update
-
-# Fetch and import/update the PE numbers
-run_command "python script/ingest_pe_numbers.py"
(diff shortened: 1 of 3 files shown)


script/ingest_pe_numbers.py
```diff
@@ -1,33 +0,0 @@
-from urllib.request import urlopen
```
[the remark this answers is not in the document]
dandds: Sure, I put it back.

script/update
```diff
@@ -7,8 +7,5 @@ source "$(dirname "${0}")"/../script/include/global_header.inc.sh
 # Enable DB migration
 MIGRATE_DB="true"
 
-# Run the shared update script
-source ./script/include/run_update
```
patricksmithdds: @dandds Was this removal intentional?

dandds: No... not sure how I managed that, will fix.

dandds:
Need to rebase and resolve the migrations.

PR #222, opened 2018-08-27
Create a workspace automatically if TO is found in EDA

richard-dds:
For this story: https://www.pivotaltracker.com/story/show/158469897

Since workspace creation was already implemented, this PR adds a check to ensure that workspaces aren't created automatically unless the task order was found in EDA. It also includes a UI change that turns `Approved` request labels into links to the request's workspace.

I'd love to hear anyone's thoughts on the gruesome `financial_verification` route. Is it worth our time to refactor it?


atst/routes/requests/financial_verification.py
```diff
@@ -41,15 +41,16 @@ def update_financial_verification(request_id):
...
-            Requests.submit_financial_verification(request_id)
-            new_workspace = Requests.approve_and_create_workspace(updated_request)
-            return redirect(
-                url_for(
-                    "workspaces.new_project",
-                    workspace_id=new_workspace.id,
-                    newWorkspace=True,
+            submitted_request = Requests.submit_financial_verification(updated_request)
+            if updated_request.is_financially_verified:
+                new_workspace = Requests.approve_and_create_workspace(submitted_request)
```
patricksmithdds: Does this need to check if the source of the task order is EDA vs manual before creating the workspace?

atst/models/request.py
```diff
@@ -17,6 +17,8 @@ class Request(Base):
         "RequestStatusEvent", backref="request", order_by="RequestStatusEvent.sequence"
     )
 
+    workspace = relationship("Workspace", uselist=False, back_populates="request")
```
dandds: It seems like the other side of this relationship should also specify `back_populates`, according to the docs: http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html#sqlalchemy.orm.relationship.params.back_populates
Is there an advantage over `backref` here?

richard-dds: Good call! I'll try using backref instead.

PR #226, opened 2018-08-28
IE fixes, et al

andrewdds:
This was not actually in IE bug, per se. Just that IE doesn't "fix" misaligned tags as well as other browsers.

Just a missing closing tag due to a misplaced `endif`.

Note: has commits from #223 so that needs to be merged first.


patricksmithdds:
Good catch!

PR #229, opened 2018-08-29
Adding the encrypted token to the yml file

kristiandds:
Environment variables don't work so... putting the encrypted token specific to the build channel which is generated using 

https://docs.travis-ci.com/user/notifications/#configuring-slack-notifications


patricksmithdds:
@kristiandds Should we merge this PR, or are we going to hold off for the switch to Circle?

kristiandds:
I think we can merge! apologies for the delay :)

PR #232, opened 2018-08-30
Allow user to view submitted request

patricksmithdds:
Previously, a user could click on a request that was pending approval and edit it without submitting. This behavior is not intended and confusing. Now, the user will be taken to a page that summarizes the request and a notice that the request is being reviewed:

image

This PR also includes a small refactoring of the authorization check when attempting to view a request. The refactoring adds a method `Autorization.can_view_request` to be more consistent with the authorization checks around viewing workspaces.

### DIFF
--- a/atst/domain/requests.py
+++ b/atst/domain/requests.py
@@ -63,7 +63,7 @@ class Requests(object):
     def get(cls, user, request_id):
         try:
             request = db.session.query(Request).filter_by(id=request_id).one()
-        except NoResultFound:
+        except (NoResultFound, exc.DataError):
             raise NotFoundError("request")
 
         if not Authorization.can_view_request(user, request):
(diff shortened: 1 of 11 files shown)


atst/domain/requests.py
```diff
@@ -59,12 +60,15 @@ def exists(cls, request_id, creator):
             return False
 
     @classmethod
-    def get(cls, request_id):
+    def get(cls, user, request_id):
         try:
             request = db.session.query(Request).filter_by(id=request_id).one()
-        except NoResultFound:
+        except (NoResultFound, exc.DataError):
```
richard-dds: What's the reason for catching a `DataError` here?

patricksmithdds: A `DataError` will get raised if we try to filter by an ID that is not a valid UUID.

PR #233, opened 2018-08-30
UI/Selector Vue component

andrewdds:
This creates a jinja macro/vue component which is designed to replace the wtforms Select input.
It expects as a parameter a wtforms select field "object". Array of choices tuples can take a third value to supply a description.

role-picker

### DIFF
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -50,8 +50,6 @@ def mixedContentToJson(value):
         value["task_order"] = value["task_order"].filename
     return app.jinja_env.filters["tojson"](value)
 
-def toJson(data):
-    return json.dumps(data)
 
 def register_filters(app):
     app.jinja_env.filters["iconSvg"] = iconSvg
@@ -60,4 +58,3 @@ def register_filters(app):
     app.jinja_env.filters["readableInteger"] = readableInteger
     app.jinja_env.filters["getOptionLabel"] = getOptionLabel
     app.jinja_env.filters["mixedContentToJson"] = mixedContentToJson
-    app.jinja_env.filters["toJson"] = toJson
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -1,4 +1,5 @@
 import re
+import json
 from flask import current_app as app
 from werkzeug.datastructures import FileStorage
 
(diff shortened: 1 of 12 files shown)


atst/filters.py
```diff
@@ -46,6 +47,8 @@ def mixedContentToJson(value):
         value["task_order"] = value["task_order"].filename
     return app.jinja_env.filters["tojson"](value)
 
+def toJson(data):
+    return json.dumps(data)
```
patricksmithdds: There's a [`tojson` filter](http://jinja.pocoo.org/docs/2.10/templates/#tojson) built into Jinja. Does this do anything different?

atst/forms/data.py
```diff
@@ -102,3 +102,11 @@
     ("3-6 months", "3-6 months"),
     ("Above 12 months", "Above 12 months"),
 ]
+
+WORKSPACE_ROLES = [
+    ("admin", "Admin", "Lorem ipsum dolor sit amet, consectetur adipiscing elit."),
+    ("billing_auditor", "Billing Auditor", "Nullam sit amet sem volutpat, congue est nec, euismod leo."),
+    ("developer", "Developer", "Aenean suscipit facilisis eros, vitae malesuada dui egestas in."),
+    ("owner", "Owner", "Nullam luctus a arcu non placerat."),
+    ("security_auditor", "Security Auditor", "In ut leo vitae libero convallis finibus nec ac massa. Aliquam bibendum nulla at eros dapibus tristique."),
```
patricksmithdds: We have copy for these descriptions ([see the description of this issue](https://www.pivotaltracker.com/story/show/158072111)) -- can we add them now? It seems those descriptions have newlines in them, so I'd be interested to see that they place nice with the new display component.

atst/filters.py
andrewdds: nope TIL

atst/forms/data.py
andrewdds: I'll give it a try

andrewdds: It does not appear to respect `\n` but I don't think its a big deal.

styles/elements/_tooltip.scss
```diff
@@ -68,7 +68,7 @@
       background: $color;
       color: black;
       padding: 24px;
-      border-radius: 5px;
+      //border-radius: 5px;
```
montana-mil: did you want to keep this here?

PR #237, opened 2018-08-31
Add a member to a workspace

richard-dds:
https://www.pivotaltracker.com/story/show/158072111

With @montana-mil.

Hooks up the new member form to a new route that creates the member in the workspace.

Includes a change to the Selector component to make it play nicely with WTForms. Since the WTForms SelectField always expects a tuple with a 2-arity, I collapsed the second and third elements into an object with keys `name` and `description`.

### DIFF
--- a/atst/forms/data.py
+++ b/atst/forms/data.py
@@ -106,27 +106,37 @@ COMPLETION_DATE_RANGES = [
 WORKSPACE_ROLES = [
     (
         "owner",
-        "Workspace Owner",
-        "Can add, edit, deactivate access to all projects, environments, and members. Can view budget reports. Can start and edit JEDI Cloud requests.",
+        {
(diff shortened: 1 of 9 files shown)


atst/forms/data.py
```diff
@@ -106,27 +106,37 @@
 WORKSPACE_ROLES = [
     (
         "owner",
-        "Workspace Owner",
-        "Can add, edit, deactivate access to all projects, environments, and members. Can view budget reports. Can start and edit JEDI Cloud requests.",
+        {
+            "name": "Workspace Owner",
+            "description": "Can add, edit, deactivate access to all projects, environments, and members. Can view budget reports. Can start and edit JEDI Cloud requests.",
+        },
```
patricksmithdds: I was surprised when this initially worked -- good catch!

PR #238, opened 2018-09-04
Request schema #159719829

dandds:
This introduces a schema for request data, which are now stored in `request_revisions`. Every time a request is updated, we create a new revision. The request model can still serialize its body as the nested JSON/dictionary object the rest of the app is written to expect, but now it does that by serializing the latest revision.

I know we'd like to refine the schema more, particularly around using booleans for yes/no questions. I think it's worth making sure the PR as it stands doesn't introduce regressions and then handle that refinement separately.

There's also a migration for moving existing request bodies to revisions. A fast way to test if this is working (if you aren't very attached to the data in your local databse) is:

```
dropdb atat
createdb atat
kubectl -n atat exec -i --tty postgres-master-7694bfbb9c-dxjhk -- pg_dump --user postgres atat | psql atat
pipenv alembic upgrade head
```


atst/models/request_revision.py
```diff
@@ -0,0 +1,77 @@
...
+    # primary_poc
+    am_poc = Column(Boolean, default=False)
+    dodid_poc = Column(String)
+    email_poc = Column(String)
+    fname_poc = Column(String)
+    lname_poc = Column(String)
+
+    # details_of_use
+    jedi_usage = Column(String)
+    start_date = Column(Date())
```
patricksmithdds: For some reason, when I visit the details of use form after filling it out, the date I'm being shown is parsed incorrectly. In this example, I successfully saved January 2, 2023 as the start date and can verify it in the ipython console:

```
In [22]: req.body['details_of_use']['start_date']
Out[22]: datetime.date(2023, 1, 2)

In [23]: req.latest_revision.start_date
Out[23]: datetime.date(2023, 1, 2)
```

But visiting the page gives me an incorrect date:

image

I'm curious if it has something to do with the `parse_date` helper we're using and if we still need to do that now that we can store dates instead of dates-as-strings-in-JSONB.

atst/domain/requests.py
```diff
@@ -103,7 +100,10 @@ def submit(cls, request):
     @classmethod
     def update(cls, request_id, request_delta):
         request = Requests._get_with_lock(request_id)
-        request = Requests._merge_body(request, request_delta)
+
+        new_body = deep_merge(request_delta, request.body)
+        revision = create_revision_from_request_body(new_body)
+        request.revisions.append(revision)
```
patricksmithdds: I was initially imaging this working a bit differently. I was thinking that a new revision would be created only when the request was created, submitted, or rejected, mainly to make it easier to view what was changed between request submissions. If we make a new revision for every update, how can view changes since the last submission? Also, do we need to create a revision if nothing changed?

richard-dds: Not sure if it would be too complicated, but we could also tie revisions to a status change since these two concepts are related.

alembic/versions/090e1bd0d7ce_remove_revisions_body_column.py
```diff
@@ -0,0 +1,65 @@
...
+        return create_revision_from_request_body(body)
+
+
+def upgrade():
+    Session = sessionmaker(bind=op.get_bind())
+    session = Session()
+    for request in session.query(Request).all():
+        (body,) = session.execute("SELECT body from requests WHERE id='{}'".format(request.id)).fetchone()
+
+        # this data should already exist as a task_order
```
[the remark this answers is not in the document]
dandds: I thought that data should already exist on a related `task_orders` entry, but in fact only some of it does. I'll fix.

atst/domain/requests.py
dandds: Yeah, I was wondering if status should be held on the revision or if they should have some kind of relationship.

atst/models/request_revision.py
dandds: This seems to be a problem with how WTForms is serializing the date for the HTML. Even if I revert to using their built-in `DateField` and specify the format, it does that. I'll see if I can find a simple solution.

dandds: Update: It's not actually WTForms fault. We serialize the date field data to a string when we build the template in `TextInput`. I added a `DateInput` component to make sure we serialize the `datetime.date` object correctly.

dandds:
@patricksmithdds I think I addressed most of your comments. Let me know if not. Thanks!

dandds:
@patricksmithdds making that column `ARRAY(String)` seems to solve the problem.

atst/models/request_revision.py
```diff
@@ -9,6 +9,7 @@
     Sequence,
 )
 from sqlalchemy.orm import relationship
+from sqlalchemy.types import ARRAY
```
patricksmithdds: The note from the docs (https://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.ARRAY) suggests:

> However, currently only the PostgreSQL backend has support for SQL arrays in SQLAlchemy. It is recommended to use the postgresql.ARRAY type directly when using ARRAY types with PostgreSQL, as it provides additional operators specific to that backend.

Is there any reason not to use the `postgresql.ARRAY` type?

dandds: Nope, i'll change.

dandds: Amended that commit.

PR #239, opened 2018-09-04
Edit user role

montana-mil:
## Description
Allows users with `ccpo`, `owner`, or `admin` permissions to update the role of members in their workspaces. 

Worked with @richard-dds on this.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/159010528

### DIFF
--- a/atst/domain/workspaces.py
+++ b/atst/domain/workspaces.py
@@ -82,6 +82,15 @@ class Workspaces(object):
         return workspace_user
 
     @classmethod
+    def update_member(cls, user, workspace, member, role_name):
+        if not Authorization.has_workspace_permission(
+            user, workspace, Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE
+        ):
+            raise UnauthorizedError(user, "update workspace member")
+
+        return WorkspaceUsers.update_role(member, workspace.id, role_name)
+
+    @classmethod
     def _create_workspace_role(cls, user, workspace, role_name):
         role = Roles.get(role_name)
         workspace_role = WorkspaceRole(user=user, role=role, workspace=workspace)
--- a/atst/domain/workspaces.py
+++ b/atst/domain/workspaces.py
@@ -39,10 +39,7 @@ class Workspaces(object):
     def get_for_update(cls, user, workspace_id):
         workspace = Workspaces.get(user, workspace_id)
         Authorization.check_workspace_permission(
-            user,
-            workspace,
-            Permissions.ADD_APPLICATION_IN_WORKSPACE,
-            "add project",
+            user, workspace, Permissions.ADD_APPLICATION_IN_WORKSPACE, "add project"
         )
 
         return workspace
@@ -92,7 +89,7 @@ class Workspaces(object):
             user,
             workspace,
             Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
-            "update workspace member",
+            "edit workspace member",
         )
 
         return WorkspaceUsers.update_role(member, workspace.id, role_name)
--- a/atst/domain/workspaces.py
+++ b/atst/domain/workspaces.py
@@ -38,10 +38,13 @@ class Workspaces(object):
     @classmethod
     def get_for_update(cls, user, workspace_id):
         workspace = Workspaces.get(user, workspace_id)
-        if not Authorization.has_workspace_permission(
-            user, workspace, Permissions.ADD_APPLICATION_IN_WORKSPACE
-        ):
-            raise UnauthorizedError(user, "add project")
+        Authorization.check_workspace_permission(
+            user,
+            workspace,
+            Permissions.ADD_APPLICATION_IN_WORKSPACE,
+            "add project",
+        )
+
         return workspace
 
     @classmethod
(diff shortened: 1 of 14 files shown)


atst/models/workspace.py
```diff
@@ -71,4 +72,4 @@ def task_order(self):
 
     @property
     def members(self):
-        return MOCK_MEMBERS
+        return [WorkspaceUser(role.user, role) for role in self.roles]
```
patricksmithdds: Can we also get rid of `MOCK_MEMBERS` now that it's no longer used anywhere?

atst/routes/workspaces.py
```diff
@@ -119,3 +122,44 @@ def create_member(workspace_id):
...
+def view_member(workspace_id, member_id):
+    workspace = Workspaces.get(g.current_user, workspace_id)
+    Authorization.check_workspace_permission(
+        g.current_user,
+        workspace,
+        Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
+        "edit this workspace user",
+    )
+    member = WorkspaceUsers.get(workspace_id, member_id)
+    form = NewMemberForm(http_request.form)
```
patricksmithdds: Should this be `form = EditMemberForm()`?

atst/routes/workspaces.py
```diff
@@ -119,3 +122,44 @@ def create_member(workspace_id):
...
+    return render_template(
+        "member_edit.html", form=form, workspace=workspace, member=member
+    )
+
+
+@bp.route(
+    "/workspaces/<workspace_id>/members/<member_id>/member_edit", methods=["POST"]
+)
+def update_member(workspace_id, member_id):
```
patricksmithdds: We should enforce the Authorization check here as well.

templates/member_edit.html
```diff
@@ -2,223 +2,141 @@
...
-        <option value="Billing Auditor" selected="selected">Billing Auditor</option>
-        <option value="CCPO">CCPO</option>
-        <option value="Developer">Developer</option>
-        <option value="Owner">Owner</option>
-        <option value="Security Auditor">Security Auditor</option>
-      </select>
-    </div>
+  <div class='panel member-card'>
+    <div class='member-card__header'>
+      <h1 class='member-card__heading'>Danny Knight</h1>
```
patricksmithdds: Can we put in the real member data now that we're actually passing the data in?

templates/member_edit.html
```diff
@@ -2,223 +2,141 @@
...
 
+        <div class="usa-input member-card__input">
+          <label for="filter-status">Workspace Role</label>
+          {{ Selector(form.workspace_role) }}
```
patricksmithdds: I think some of the styles might need to be tweaked here. Previously this looked like:

image

But now I'm seeing this:

image

Also, when the selector is expanded, the role names should be bolded (see the selector on the new member form).

tests/domain/test_workspaces.py
```diff
@@ -119,3 +119,38 @@ def test_need_permission_to_create_workspace_user():
...
+        "last_name": "User",
+        "email": "new.user@mail.com",
+        "workspace_role": "developer",
+        "dod_id": "1234567890",
+    }
+    member = Workspaces.create_member(owner, workspace, user_data)
+    role_name = "developer"
+
+    updated_member = Workspaces.update_member(owner, workspace, member, role_name)
+    assert updated_member.workspace == workspace
```
patricksmithdds: The `role_name` isn't different from the original role, so I think we should select a different role to update it to and also add an assertion to verify that the member's role was updated.

templates/member_edit.html
patricksmithdds: This should also pre-select the user's current role. Currently, it shows "select an option" regardless of what the user's role is.

atst/routes/workspaces.py
montana-mil: yes

atst/domain/workspaces.py
```diff
@@ -81,6 +81,15 @@ def create_member(cls, user, workspace, data):
         )
         return workspace_user
 
+    @classmethod
+    def update_member(cls, user, workspace, member, role_name):
+        if not Authorization.has_workspace_permission(
+            user, workspace, Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE
+        ):
+            raise UnauthorizedError(user, "update workspace member")
```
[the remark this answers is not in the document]
montana-mil: oh good catch. @richard-dds we can didn't have any reason against using `check_workspace_permission`, right? I'm going to switch it to use that.

atst/models/workspace.py
montana-mil: yep. done!

atst/domain/workspaces.py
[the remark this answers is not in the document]
montana-mil: actually there are a number of `has_workspace_permissions` calls in that file. Should I switch them all out? @dandds @richard-dds

templates/member_edit.html
```diff
@@ -2,223 +2,140 @@
...
-  <div class="panel__heading panel__heading--tight">
-    <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environment</div></h2>
+  <div class="panel">
+    <div class="panel__heading panel__heading--tight">
+      <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environment</div></h2>
+    </div>
   </div>
-</div>
 
-<form class='search-bar'>
```
patricksmithdds: Looks like removing this form changed the way the search part renders:

image

Was there an issue with having multiple forms on the page?

montana-mil: that part isn't wired up at all. is it important to make it look nice?

PR #240, opened 2018-09-04
Update empty request screen

luisgov:
This PR updates the copy for the empty request screen.

Story: 
https://www.pivotaltracker.com/story/show/159499449

screen shot 2018-09-04 at 09 38 48

### DIFF
--- a/templates/requests.html
+++ b/templates/requests.html
@@ -28,7 +28,8 @@
 {% if not requests %}
 
   {{ EmptyState(
-    'There are currently no active requests for you to see.',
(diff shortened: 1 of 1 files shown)


patricksmithdds:
@luisgov I think this looks fine, but the pivotal issue says the sub-message should be in a tooltip. Are we intentionally not using tooltips here?

luisgov:
@patricksmithdds correct. no need to put it on the tooltip for extra hovering. this info should be exposed. i'll update the pivotal story.

PR #243, opened 2018-09-04
Action required flag

richard-dds:
Display an "action required" flag on requests that require action from the current user. When there are requests that require action, also display an alert on the top of the page.

screen shot 2018-09-04 at 3 01 08 pm

screen shot 2018-09-04 at 3 01 01 pm

### DIFF
--- a/atst/routes/requests/index.py
+++ b/atst/routes/requests/index.py
@@ -6,84 +6,96 @@ from atst.domain.requests import Requests
 from atst.models.permissions import Permissions
 
 
-def map_request(request, viewing_role):
+class RequestsIndex(object):
+    def __init__(self, user):
+        self.user = user
 
-    time_created = pendulum.instance(request.time_created)
-    is_new = time_created.add(days=1) > pendulum.now()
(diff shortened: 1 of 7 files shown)


atst/routes/requests/index.py
```diff
@@ -6,78 +6,96 @@
...
+            "order_id": request.id,
+            "is_new": is_new,
+            "status": request.status_displayname,
+            "app_count": app_count,
+            "last_submission_timestamp": last_submission_timestamp,
+            "full_name": request.creator.full_name,
+            "annual_usage": annual_usage,
+            "edit_link": edit_link,
+            "action_required": request.action_required_by == viewing_role,
+        }
```
patricksmithdds: I was curious about the difference between using `{}` vs `dict()` in this file and [according to stack overflow](https://stackoverflow.com/a/6612024) `{}` is twice as fast as `dict()`. Totally not a concern here, just a TIL.

rthoward: Interesting! I had no idea. I was just being lazy here, but I hate to waste  cycles. I'll convert this to a dict literal

PR #245, opened 2018-09-04
Redirect CCPO to requests index after login

richard-dds:
Now a CCPO is redirect to the requests index after logging in. 

I don't think that the /home route is being used anymore. Not sure if we should delete it or not.

### DIFF
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -17,7 +17,7 @@ def root():
 
 @bp.route("/home")
 def home():
-    return render_template("home.html")
+    return redirect(url_for("requests.requests_index"))
 
 
 @bp.route("/styleguide")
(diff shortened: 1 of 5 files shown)


atst/routes/__init__.py
```diff
@@ -47,7 +47,7 @@ def login_redirect():
     session["user_id"] = user.id
 
     if user.atat_role.name == "ccpo":
```
dandds: Can we just get rid of the conditional here and in `routes/dev`?

patricksmithdds:
> I don't think that the /home route is being used anymore. Not sure if we should delete it or not.

Not sure if this just makes things more complicated, but could we always redirect to `/home` on login and have that route perform the redirect? In the future, we'll have to redirect not only based on the type of user but also which resources they have access to (how many requests/workspaces and their roles on those workspaces) and I think putting that logic in the home route is the best place for it.

rthoward:
That makes perfect sense. 👍

PR #249, opened 2018-09-05
Fix seed script

richard-dds:


andrewdds:
Maybe some documentation on this?

PR #252, opened 2018-09-06
View member list as workspace owner

richard-dds:
Allow users with the correct permissions to view a workspace's members. Most of the complexity of this PR came from adding two new permissions:

`VIEW_WORKSPACE`: we were just calling `user in workspace.users` to determine if a user could view a workspace, but that check wasn't providing CCPOs with superuser access. Using this new permission gives us more fine-grained control.

`VIEW_WORKSPACE_MEMBERS`: the permission required to view /workspaces/<id>/members. Only bestowed up on the owner and ccpo roles.

I also added workspace members to the seed script.

### DIFF
--- a/alembic/versions/ad30159ef19b_add_view_workspace_members_permission.py
+++ b/alembic/versions/ad30159ef19b_add_view_workspace_members_permission.py
@@ -7,6 +7,8 @@ Create Date: 2018-09-05 11:17:17.204089
 """
 from alembic import op
 from sqlalchemy.orm.session import Session
+from sqlalchemy.orm.attributes import flag_modified
+
 from atst.models.role import Role
 from atst.models.permissions import Permissions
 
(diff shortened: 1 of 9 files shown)


atst/domain/workspaces.py
```diff
@@ -53,6 +54,17 @@ def get_by_request(cls, request):
 
         return workspace
 
+    @classmethod
+    def get_with_members(cls, user, workspace_id):
```
dandds: Two things:
- The name of this makes me think it will return a dictionary or something that includes both the workspace and its members, in a way that `Workspace.get` wouldn't.
- The only real difference between this and `Workspace.get` is the additional auth check. Is that a deliberate pattern? I think the implementation might be a little confusing. Maybe a private `_get` that accepts these args plus a list of permissions to check, and then the public methods rely on that? Or maybe accumulating perms on top of the basic `get` is good, I don't know.

tests/domain/test_workspaces.py
```diff
@@ -155,3 +155,27 @@ def test_need_permission_to_update_workspace_user_role():
...
+    owner = UserFactory.create()
+    workspace = Workspaces.create(RequestFactory.create(creator=owner))
+    workspace = Workspaces.get_with_members(owner, workspace.id)
+
+    assert workspace
+
+
+def test_ccpo_can_view_workspace_members():
+    workspace = Workspaces.create(RequestFactory.create(creator=UserFactory.create()))
+    ccpo = UserFactory.from_atat_role("ccpo")
```
dandds: I missed that you added this factory method, nice.

atst/domain/workspaces.py
richard-dds: Yeah, it is a deliberate pattern, but I've been struggling with what to call this. I know that we discussed keeping auth checks in our domain methods, which is why I have this separate method. I'll think about how to architect this, but in the meantime, do you have a good name in mind?

alembic/versions/ad30159ef19b_add_view_workspace_members_permission.py
```diff
@@ -0,0 +1,53 @@
...
+def upgrade():
+
+    session = Session(bind=op.get_bind())
+
+    all_roles_but_default = session.query(Role).filter(Role.name != "default").all()
+    for role in all_roles_but_default:
+        role.add_permission(Permissions.VIEW_WORKSPACE)
+        session.add(role)
+
+    owner_and_ccpo = session.query(Role).filter(Role.name.in_(["owner", "ccpo"])).all()
```
patricksmithdds: Should this also be added to the "admin" user?

richard-dds: Good catch! Added

PR #255, opened 2018-09-06
Review request page #159038263

dandds:
The PT story is [here](https://www.pivotaltracker.com/story/show/159038263).

This plugs in real data for the CCPO review screen.

**notes**

- I refactored the `_review` partial to use a macro for each piece of request info and added some related Jinja filters.
- This also adds the download link for PDFs. There was a nice clean way of doing this with libcloud, but I hit [LIBCLOUD-931](https://issues.apache.org/jira/projects/LIBCLOUD/issues/LIBCLOUD-931?filter=allopenissues) when I tried the Rackspace provider, and so we write the download to a tempfile before streaming it to the user.

screenshot_2018-09-06 jedi cloud

### DIFF
--- a/atst/routes/requests/approval.py
+++ b/atst/routes/requests/approval.py
@@ -36,8 +36,15 @@ def approval(request_id):
 def task_order_pdf_download(request_id):
     request = Requests.get(g.current_user, request_id)
     if request.task_order and request.task_order.pdf:
-        object_name = request.task_order.pdf.object_name
-        generator = app.uploader.download_stream(object_name)
-        return Response(generator, mimetype="application/pdf")
+        pdf = request.task_order.pdf
+        generator = app.uploader.download_stream(pdf.object_name)
+        return Response(
+            generator,
+            headers={
+                "Content-Disposition": "attachment; filename={}".format(pdf.filename)
+            },
+            mimetype="application/pdf",
+        )
+
     else:
         raise NotFoundError("task_order pdf")
--- a/atst/routes/requests/approval.py
+++ b/atst/routes/requests/approval.py
@@ -1,7 +1,9 @@
-from flask import render_template, g
+from flask import render_template, g, Response
+from flask import current_app as app
 
 from . import requests_bp
 from atst.domain.requests import Requests
+from atst.domain.exceptions import NotFoundError
 
 
 def task_order_dictionary(task_order):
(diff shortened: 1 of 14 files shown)


templates/requests/_review.html
```diff
@@ -254,25 +116,72 @@ <h2>
...
 
+{% if financial_review %}
+  <hr>
+  <h2>
+    Financial Verification
+  </h2>
+
+  {% if pdf_available %}
+    <div>
+      <a href="{{ url_for("requests.task_order_pdf_download", request_id=request_id)}}" download>
```
patricksmithdds: Looks like the `download` attribute isn't supported in IE: https://caniuse.com/#feat=download. How does IE handle this link?

atst/routes/requests/approval.py
```diff
@@ -1,9 +1,52 @@
...
+def task_order_dictionary(task_order):
+    return {
+        c.name: getattr(task_order, c.name)
+        for c in task_order.__table__.columns
+        if c.name not in ["id", "attachment_id"]
+    }
+
+
+@requests_bp.route("/requests/approval/<string:request_id>", methods=["GET"])
+def approval(request_id):
```
patricksmithdds: Not sure if this outside the scope of this PR, but can we hide the form on requests that have not been submitted? I believe a CCPO user should not be able to approve/reject a request that hasn't been submitted yet, but it would still be nice for the CCPO user to see a read-only view of the request details.

atst/routes/requests/approval.py
```diff
@@ -1,9 +1,52 @@
...
+    data = request.body
+    if request.task_order:
+        data["task_order"] = task_order_dictionary(request.task_order)
+
+    return render_template(
+        "requests/approval.html",
+        data=data,
+        request_id=request_id,
+        status=request.status.value,
+        financial_review=True,
```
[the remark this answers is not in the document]
dandds: I thought that eventually we'd be automatically populating those fields from the EDA data if the task order could be found there (it would happen [here](https://github.com/dod-ccpo/atst/blob/master/atst/domain/task_orders.py#L32)), so it wouldn't matter if it was manually entered.

I was planning on having `financial_review` set based on the request status in the follow-on [story](https://www.pivotaltracker.com/n/projects/2160940/stories/159038361), but I can do it here.

templates/requests/_review.html
dandds: I'll test it there and revise.

atst/routes/requests/approval.py
dandds: Yeah, I agree. It might be better handled in the follow-on story. I think we need a little more clarity about what the CCPO sees based on different request statuses. For instance, it sounds like they should be able to submit internal notes regardless of the status (I think?).

atst/routes/requests/approval.py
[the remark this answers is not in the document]
dandds: Also, we should show the source for the task order info (EDA or manual), so I'll add that.

templates/requests/_review.html
dandds: The download still works in IE 11 on our window laptop. I think it probably respects the Content-Disposition header even if it's ignoring that attribute.

atst/routes/requests/approval.py
patricksmithdds: > For instance, it sounds like they should be able to submit internal notes regardless of the status (I think?).

I believe the internal notes would only be left when the request is submitted for review. I don't think the CCPO will be viewing a request that is not yet submitted. But, we can leave this as-is for now and can address the intricacies later on.

PR #260, opened 2018-09-07
Investigate requests against EDA

patricksmithdds:
This PR adds the ability to make some requests against the real EDA servers. This PR is the result of investigating EDA and _does not_ change the app to use EDA or include credentials for accessing the API.

This PR includes a script, `example_fetch_from_eda.py`, that makes a couple example requests to EDA (given that you update your config with the appropriate credentials -- see the below google doc) which will allow us to further poke at & investigate the data EDA returns.

For more information, see [this doc on google drive](https://docs.google.com/document/d/1LvpnWHFTdKgwrsENOjgiATFzWjhlmYxZmZ2Ts1_bTck/edit)


dandds:
@richard-dds I don't think I was copied on the email referenced in the Google Doc that has the account password. Were you? If so, can you plug in the creds and test that this works? I updated it with the `defusedxml` library the security audit asked for.

dandds:
@patricksmithdds Richard and I couldn't find the password referenced in the google doc and couldn't test this properly. I made a couple superficial updates. I don't think it will hit conflicts anytime soon, so we'll let it hang until you're back.

PR #262, opened 2018-09-10
Ui/cumulative budget chart

andrewdds:
This adds an SVG budget chart to the reports screen.

Fixture data for a number of months is supplied to the template through the route. Presumably little will need to change when "real" data is supplied.

Notably missing is the end-of-TO projection for how the current and cumulative spend will fall relative to the total budget. I.e. will it be under/over budget.

Also missing are future projections of monthly spend. Currently, the projected monthly spends are duplicates of the most recent actual monthly spend.

We can prioritize if and how we want to stub these out in a later PR, if needed.

screen shot 2018-09-11 at 9 57 33 am
 
Note: An IE/Javascript bug (not related to this work) will prevent this screen from being visible. Bug is tracked here: https://www.pivotaltracker.com/story/show/160417558

### DIFF
--- a/atst/domain/reports.py
+++ b/atst/domain/reports.py
@@ -200,5 +200,7 @@ class Reports:
     @classmethod
     def cumulative_budget(cls, alternate):
         return {
-            "months": CUMULATIVE_BUDGET_BELUGA if alternate else CUMULATIVE_BUDGET_AARDVARK
+            "months": CUMULATIVE_BUDGET_BELUGA
+            if alternate
+            else CUMULATIVE_BUDGET_AARDVARK
         }
--- a/atst/domain/reports.py
+++ b/atst/domain/reports.py
@@ -196,3 +196,9 @@ class Reports:
             "projects": project_totals,
             "workspace": workspace_totals,
         }
+
+    @classmethod
+    def cumulative_budget(cls, alternate):
+        return {
+            "months": CUMULATIVE_BUDGET_BELUGA if alternate else CUMULATIVE_BUDGET_AARDVARK
+        }
--- a/templates/workspaces/reports/index.html
+++ b/templates/workspaces/reports/index.html
@@ -149,7 +149,7 @@
               v-bind:class='{ "budget-chart__block--highlighted": month.isHighlighted }'
               v-bind:width='month.metrics.blockWidth'
               v-bind:x='month.metrics.blockX'
-              v-bind:height='height'/>
+              v-bind:height='height'></rect>
 
             {# budget bar #}
             <rect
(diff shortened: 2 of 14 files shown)


templates/workspaces/reports/index.html
```diff
@@ -93,6 +93,133 @@ <h2 class='to-summary__heading'>Task Order</h2>
...
+                --><template v-if='month.budget.spend'>Spend:</template><!--
+                --><template v-if='month.budget.projected'>Projected Spend:</template><!--
+                --><span v-html='month.spendAmount'></span><!--
+                -->&nbsp;|&nbsp;<!--
+                --><template v-if='month.budget.cumulative'>Total:</template><!--
+                --><template v-if='month.budget.projected'>Projected Total:</template><!--
+                --><span v-html='month.cumulativeAmount'></span><!--
+              --></template><!--
+
+              --><template v-else>No spend for this month</template>
```
dandds: Just curious: what are these HTML comments doing for us?

atst/domain/reports.py
```diff
@@ -196,3 +196,11 @@ def monthly_totals(cls, alternate):
...
             "workspace": workspace_totals,
         }
+
+    @classmethod
+    def cumulative_budget(cls, alternate):
+        return {
+            "months": CUMULATIVE_BUDGET_BELUGA
+            if alternate
+            else CUMULATIVE_BUDGET_AARDVARK
+        }
```
dandds: Is there a reason to nest the data under a "months" key in this dictionary if it's the only value? Could this just return BELUGA or AARDVARK?

andrewdds: Perhaps. This is me attempting to do Python

templates/workspaces/reports/index.html
andrewdds: Oh this fun bit...
This is a little trick I have in my back pocket to remove whitespace characters between inline elements. If you have a line break between the elements, the browser will render a whitespace. Inside a browser title tooltip, this makes it do all kinds of weird indenting.

Alternatively, this could all be one long line, but this makes it more readable... if a little perplexing.

andrewdds:
@dandds thats a great idea about the "reset to today" feature. I'll stick it in the backlog for future consideration.

templates/workspaces/reports/index.html
dandds: Ooh, interesting. That makes sense.

atst/domain/reports.py
dandds: :+1: We can revisit when we wire this with real data. I don't think it should hold up merging the PR.

PR #266, opened 2018-09-11
Review history #159038479

dandds:
Here's the PT story: https://www.pivotaltracker.com/n/projects/2160940/stories/159038479

This wires up the CCPO approval log.

Few notes:
- I implemented this so that the actions are "approve", "accept", or "deny" because of the new request status.
- The timezone in the timestamp is expressed as a UTC offset for right now; that's what we get at the moment without importing more Python dependencies. It won't stay like this. I added a new story, #160419017, to address how we want to format timestamps. I think we should make a deliberate decision about how we want to display timestamps to end-users here.
- The story doesn't specify it, but I have the logs displayed in reverse order so that the most recent action is first.

### DIFF
--- a/atst/routes/requests/approval.py
+++ b/atst/routes/requests/approval.py
@@ -35,7 +35,7 @@ def render_approval(request, form=None):
     return render_template(
         "requests/approval.html",
         data=data,
-        statuses=request.status_events,
+        statuses=reversed(request.status_events),
         request_id=request.id,
         status=request.status.value,
         pending_review=pending_review,
--- a/atst/routes/requests/approval.py
+++ b/atst/routes/requests/approval.py
@@ -35,6 +35,7 @@ def render_approval(request, form=None):
     return render_template(
         "requests/approval.html",
         data=data,
+        statuses=request.status_events,
         request_id=request.id,
         status=request.status.value,
         pending_review=pending_review,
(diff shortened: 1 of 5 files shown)


templates/requests/approval.html
```diff
@@ -142,60 +142,42 @@ <h2 class='h3 request-approval__columns__heading'>Approval Log</h2>
       <div>
         <div class='approval-log'>
           <ol>
-            <li>
-              <article class='approval-log__log-item'>
-                <div>
-                  <h3 class='approval-log__log-item__header'>Denied by Darth Vader</h3>
-                  <p>"You have failed me for the last time, Admiral. Captain Piett. Yes, my lord. Make ready to land out troops beyond the energy shield and deploy the fleet so that nothing gets off that system."</p>
```
montana-mil: goodbye darth :(

dandds: :(

atst/routes/requests/approval.py
```diff
@@ -35,6 +35,7 @@ def render_approval(request, form=None):
     return render_template(
         "requests/approval.html",
         data=data,
+        statuses=reversed(request.status_events),
         request_id=request.id,
         status=request.status.value,
```
montana-mil: is this `status` outside of the scope of the for loop in `templates/requests/approval.html` staring on line 146? Is the inner `status` overwriting this one?

dandds: Good call. The `for` loop should scope the inner `status` correctly, but it's confusing to have it shadow the outer status. I'll change the names to make it clearer.

PR #268, opened 2018-09-11
Fix edit link for new request

richard-dds:
Fixes https://www.pivotaltracker.com/story/show/160355202

### DIFF
--- a/templates/requests/_review.html
+++ b/templates/requests/_review.html
@@ -15,12 +15,20 @@
   </div>
 {% endmacro %}
 
+{% macro EditLink(screen) %}
+  {% if request_id %}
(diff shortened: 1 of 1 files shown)


templates/requests/_review.html
```diff
@@ -15,12 +15,20 @@
   </div>
 {% endmacro %}
 
+{% macro EditLink(screen) %}
+  {% if request_id %}
+    {{ url_for('requests.requests_form_update', screen=screen, request_id=request_id)}}
+  {% else %}
+    {{ url_for('requests.requests_form_update', screen=screen, request_id=None) }}
```
montana-mil: does this mean no permissions? so then it would go to 403, right?

montana-mil: oh i see

PR #270, opened 2018-09-11
Request Internal comments

richard-dds:
screen shot 2018-09-11 at 3 41 00 pm

https://www.pivotaltracker.com/story/show/160141410

Allows CCPO users to leave internal comments. Definitely needs some UI polish.

I implemented Request.internal_comments as a one-to-one relationship rather than a plain field for two reasons:

1. The relationship will probably become a one-to-many in the future, in which case (I think) we can just remove the `uselist=False` flag from the field declaration.

2. Our migration path breaks if we alter the fields on the request model. This is really bad, and we need to fix it soon. One possible fix is to squash all of our migrations before the deployment, which may be a good idea anyway.


dandds:
i feel worse about writing that migration every day :(

PR #276, opened 2018-09-12
CircleCI CD

ddsdevon:
This PR adds deployment capabilities to the CircleCI configuration, and disables image pushing and deployment from Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,13 +45,6 @@ script:
   - docker container stop current-atst-tester
   - docker run --add-host "postgreshost:${postgres_ip}" --add-host "redishost:${redis_ip}" "${TESTER_IMAGE2_NAME}"
 
-before_deploy:
-  - docker build --tag "${PROD_IMAGE_NAME}" . -f deploy/docker/prod/Dockerfile
-  - remote_image_name="${ATAT_DOCKER_REGISTRY_URL}/${PROD_IMAGE_NAME}:${GIT_SHA}"
-  - docker tag "${PROD_IMAGE_NAME}" "${remote_image_name}"
-  - docker images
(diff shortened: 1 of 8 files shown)


patricksmithdds:
@ddsdevon What still needs to be done on this branch? Is this in a state that someone else could pick up?

ddsdevon:
It is very close. There were some issues with the build image, and I _think_ I fixed them all in a mad scramble at the end... but I'm not 100% sure.

Someone should be able to pick this up and finish it out?
I recommend:
- Rebase
- Update the bottom of the `.circleci/config.yml` file so images are pushed for this branch instead of master
- Manually update a Kubernetes deployment with the `-circleci` version of the image to see if it works
- If all is good, we revert the config change and go from there

patricksmithdds:
I've rebased this and tested on CircleCI and everything looked good: https://circleci.com/gh/dod-ccpo/atst/897. I also added a step to generate the build info to be sure it was CircleCI making the magic happen.

I think this is good to go for now -- one change we might want to make in the future is remove the `-circleci` from the image name, but it doesn't really hurt to have it there at the moment.

.travis.yml
```diff
@@ -50,11 +50,11 @@ before_deploy:
   - remote_image_name="${ATAT_DOCKER_REGISTRY_URL}/${PROD_IMAGE_NAME}:${GIT_SHA}"
   - docker tag "${PROD_IMAGE_NAME}" "${remote_image_name}"
   - docker images
-  - docker push "${remote_image_name}"
+ #- docker push "${remote_image_name}"
```
montana-mil: do you still want this guy in here?

patricksmithdds: Good catch -- we can git rid of that.

PR #279, opened 2018-09-12
Fix bug in new request menu

richard-dds:
https://www.pivotaltracker.com/story/show/160430122


richard-dds:
@montana-mil Sorry, I should have written a description. The issue is that the request was being created with some data (am_poc) in the workspace owner section of the body. To check if a section of the report was completed, we checked for the presence of that section. Since the workspace owner section was already populated am_poc, the section was wrongly displayed as completed.

PR #281, opened 2018-09-13
Query refactors

richard-dds:
I thought it might be nice to start moving some of the query logic out of the domain classes. I created two new classes, RequestsQuery and WorkspacesQuery, to house sql-alchemy specific code. I also created a base Query class with a basic CRUD implementation.

I also threw in a refactor to the workspaces domain tests, since we have so many of them and they were getting out of hand.

Let me know if you like the pattern. If not, we can just keep the test cleanup.

### DIFF
--- a/atst/domain/common/__init__.py
+++ b/atst/domain/common/__init__.py
@@ -1,0 +1,1 @@
+from .query import Query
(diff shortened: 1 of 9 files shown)


richard-dds:
@dandds That's true. I kind of forgot about that. I can think of a few instances where we've put queries into the model class, but not many. I can definitely see Patrick's logic. I think the Query class is just a concept I picked up from Elixir.

Honestly I'd be fine putting the queries anywhere. I just want to move them out of the domain. Which way are you leaning toward?

PR #282, opened 2018-09-13
Request status views #160109433

dandds:
For this PT story: https://www.pivotaltracker.com/story/show/160109433

This refines what a user sees for a request depending on the request's status. Details are in the story.

**notes**
- I moved the pre-existing `requests/pending/<request_id>` route to a `details` route and made it generic. That way we can reuse it for approved requests.
- I noted in the PT card that users will not be able to save partial data on the financial verification form if they click over to view the request details (https://www.pivotaltracker.com/story/show/160109433/comments/194161413). One other edge case is if a user gets to the extended financial form, clicks over to see the request details, and clicks back to the financial verification form. In that case, we should show them the extended version, but I don't see a clean way to keep track of whether to do that.

screen shot 2018-09-13 at 11 26 29 am


atst/domain/requests.py
```diff
@@ -203,6 +203,16 @@ def is_pending_ccpo_acceptance(cls, request):
     def is_pending_ccpo_approval(cls, request):
         return request.status == RequestStatus.PENDING_CCPO_APPROVAL
 
+    @classmethod
+    def is_pending_ccpo_action(cls, request):
```
richard-dds: I definitely like having these little properties for checking statuses, as opposed to callers having to do something like `request.status == SomeStatus`. 

I do wonder if they belong on the domain class rather than on the model instance.

dandds: I was wondering about that too. I guess that counts as a derived property, right? I'll move them.

dandds:
@richard-dds I had to rebase this. Can you re-approve when you get a minute?

PR #284, opened 2018-09-13
Ui/request approval screen

andrewdds:
This makes some changes and fixes to the CCPO approval layout:

- An additional toggle is added, asking whether the request is ready for approval, or needs revisions.
- The selection of either of these triggers the Message to Requester to appear, and the label/description will change slightly depending on which.
- Selecting "Ready for approval" will also trigger display of "Authorizing Officials" fields.
- Both field sections are styled in a blue alert-style block to emphasize them when they appear.

Also, adds a `local-timestamp` vue component, which formats a supplied datetime, and should automatically shift the date to the client browser's timezone.

Also fixes a bug in the relationship between request reviews and their parent statuses, so a reuqest's `review.status` will return a single object, rather than a list. Thanks @dandds

### DIFF
--- a/atst/forms/ccpo_review.py
+++ b/atst/forms/ccpo_review.py
@@ -7,7 +7,7 @@ from .validators import Alphabet, PhoneNumber
 
 
 class CCPOReviewForm(ValidatedForm):
-    comment = TextAreaField("Comments (optional)")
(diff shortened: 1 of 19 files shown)


dandds:
@luisgov I rebased. Can you re-approve when you get a sec?

PR #287, opened 2018-09-14
Review comment #160185269

dandds:
For this PT story: https://www.pivotaltracker.com/story/show/160185269

If a request has had changes requested (prior to either acceptance or approval) and the CCPO left a comment, then we display an alert with the CCPO's comment.

screen shot 2018-09-14 at 11 48 50 am

### DIFF
--- a/templates/requests/comment.html
+++ b/templates/requests/comment.html
@@ -1,0 +1,5 @@
+{% from "components/alert.html" import Alert %}
+
+{{ Alert('Changes Requested',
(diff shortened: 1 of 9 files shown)


templates/requests/comment.html
```diff
@@ -0,0 +1,5 @@
+{% from "components/alert.html" import Alert %}
+
+{{ Alert('Changes Requested',
```
richard-dds: The Alert macro has a callable version now, which makes it easier to put markup inside of the alert message:

```
{% call Alert('Changes Requested', level='warning') %}
  <p>Markup here</p>
{% endcall %}
```

PR #288, opened 2018-09-14
display zeroes in the request review template

dandds:
Hardik pointed out that the review page can't display zero for dollar fields: https://www.pivotaltracker.com/story/show/160450606/comments/194225623

This fixes that by checking ensuring that a value is not `None`, instead of just checking if it's falsey.

### DIFF
--- a/templates/requests/_review.html
+++ b/templates/requests/_review.html
@@ -6,8 +6,9 @@
   <div>
     <dt>{{ title | safe }}</dt>
     <dd>
-      {% if data[section] and data[section][item_name] %}
-        {{ data[section][item_name] | findFilter(filter, filter_args) }}
+      {% set value = data.get(section, {}).get(item_name) %}
+      {% if value is not none %}
+        {{ value | findFilter(filter, filter_args) }}
       {% else %}
         {{ RequiredLabel() }}
       {% endif %}


templates/requests/_review.html
```diff
@@ -6,8 +6,9 @@
   <div>
     <dt>{{ title | safe }}</dt>
     <dd>
-      {% if data[section] and data[section][item_name] %}
-        {{ data[section][item_name] | findFilter(filter, filter_args) }}
+      {% set value = data.get(section, {}).get(item_name) %}
+      {% if not value == None %}
```
richard-dds: Would `if value is not None` work here? `is` is the idiomatic (and more performant) operator for null checks.

dcorrigan: I didn't know that! Jinja doesn't like it, though:

```
jinja2.exceptions.TemplateAssertionError: no test named 'None'
```

This is the thing I really don't like about Jinja: it's only sort-of Python.

dandds: I didn't know that! Jinja doesn't like it, though:

```
jinja2.exceptions.TemplateAssertionError: no test named 'None'
```

This is the thing I really don't like about Jinja: it's only sort-of Python.

dandds: @richard-dds Updated with the Jinja version we talked about on Friday.

PR #289, opened 2018-09-14
Move request notifications

richard-dds:
https://www.pivotaltracker.com/story/show/160186174

Move alerts for requests from the index page into the request details pages.

screen shot 2018-09-14 at 2 30 51 pm

screen shot 2018-09-14 at 2 30 39 pm


templates/requests/details.html
```diff
@@ -5,21 +5,28 @@
...
+    {% endcall %}
+
+  {% elif requires_fv_action %}
+
+    {{ Alert('Pending Financial Verification', fragment="fragments/pending_financial_verification.html") }}
     {% include 'requests/review_menu.html' %}
-  {% endif %}
 
-  {% if pending_review %}
-    {{ Alert('Your request is being reviewed',
```
dandds: Are we definitely using the other approval notifications in lieu of this one? That makes sense, I'm just checking.

templates/requests/financial_verification.html
```diff
@@ -6,6 +6,10 @@
 
 {% block content %}
 
+{% if request.is_pending_financial_verification %}
+  {{ Alert('Pending Financial Verification', fragment="fragments/pending_financial_verification.html") }}
+{% endif %}
```
dandds: Very small thing, but I put the alerts below the review menu/tabs in https://github.com/dod-ccpo/atst/pull/287. They should probably be grouped together. @andrewcroce or @luiscielak do you have an opinion about whether alerts should fall above or below the horizontal tab menu we use for financial verification and the request form?

richard-dds: @dandds Good call bringing that up. Hardik mentioned putting them above when he and I synced up on Friday, but obviously it should be up to Andrew and Luis.

templates/requests/details.html
richard-dds: I'm not totally sure what you mean, but we are displaying this info in an alert (which is being included from a fragment).

PR #292, opened 2018-09-17
Audit log

richard-dds:
https://www.pivotaltracker.com/story/show/160103536

Creates a new `AuditLog` table and domain object. All CUD operations on users, requests, workspaces, projects and environments are logged. Includes a new route, `/activity-history`, viewable only by the CCPO and security auditor ATAT roles, which lists all audit events.


atst/domain/requests/requests.py
```diff
@@ -66,7 +67,7 @@ def get_many(cls, creator=None):
         return RequestsQuery.get_many(creator)
 
     @classmethod
-    def submit(cls, request):
+    def submit(cls, user, request):
```
dandds: Should we be logging within `update` too? I don't know if those count as distinct update events, because they happen in quick succession most of the time, but a user can leave a request and come back days later to complete it.

atst/domain/requests/requests.py
```diff
@@ -210,17 +216,23 @@ def _add_review(cls, user, request, review_data):
     def advance(cls, user, request, review_data):
         if request.status == RequestStatus.PENDING_CCPO_ACCEPTANCE:
             Requests.set_status(request, RequestStatus.PENDING_FINANCIAL_VERIFICATION)
+            AuditLog.log_event(user, request, "accept request")
```
dandds: Should `Requests.set_status` just infer the log action from the status and call `log_event` itself?

atst/domain/requests/requests.py
dandds: Ditto `update_internal_comments`, even though that isn't wired to anything.

atst/domain/workspaces/workspaces.py
```diff
@@ -89,8 +95,11 @@ def update_member(cls, user, workspace, member, role_name):
             Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
             "edit workspace member",
         )
-
-        return WorkspaceUsers.update_role(member, workspace.id, role_name)
+        member = WorkspaceUsers.update_role(member, workspace.id, role_name)
+        AuditLog.log_workspace_event(
+            user, workspace, member.workspace_role, "update workspace_role"
+        )
+        return member
```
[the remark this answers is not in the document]
richard-dds: Since we're not catching any exceptions from AuditLog, I think things would just blow up. Unless you mean that not only should the event only be logged if the database update succeeds, but also vice versa (as part of the same transaction)? That would be interesting, and probably good, but I think it would require some significant, but very doable refactoring.

atst/domain/requests/requests.py
richard-dds: It would be nice to have those updates in one place, but I don't think we can do this without putting a `db.session.commit` in this method, since otherwise we can't be sure if the status update actually gets persisted. I'm already a little wary of how often we're calling commit, but maybe it's not a big deal?

patricksmithdds: 👍 I like the idea of pushing the logging into `set_status`.

atst/models/audit_event.py
```diff
@@ -0,0 +1,35 @@
+from sqlalchemy import String, Column, ForeignKey
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import relationship
+
+from atst.models import Base, types, mixins
+
+
+class AuditEvent(Base, mixins.TimestampsMixin):
```
patricksmithdds: There may be some context/conversations I missed, but have we considered creating some sort of mixin to hook into create/update events for the models that we care about? I'm concerned about requiring us to remember to log an audit event every time we perform one of those actions.

atst/domain/audit_log.py
```diff
@@ -0,0 +1,55 @@
...
+from atst.models.audit_event import AuditEvent
+
+
+class AuditEventQuery(Query):
+    model = AuditEvent
+
+    @classmethod
+    def get_all(cls):
+        return (
+            db.session.query(AuditEvent).order_by(AuditEvent.time_created.desc()).all()
```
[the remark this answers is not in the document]
richard-dds: It is necessary for the base class, but `self.model` will always be there, so that should work well. I'll change it.

atst/domain/environments.py
```diff
@@ -2,31 +2,30 @@
...
         )
-        db.session.add(environment_user)
+        db.session.add(environment_role)
         db.session.commit()
 
+        AuditLog.log_workspace_event(
+            user,
+            environment.project.workspace,
+            environment_role,
+            "add environment role",
```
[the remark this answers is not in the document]
richard-dds: Not a good reason, but I think we've named domain methods `add_x` when we're adding a relation to a parent object (environments / projects to a workspace). If we think it's worth standardizing all of them I'd be happy to change it.

atst/models/audit_event.py
richard-dds: @dandds  and I paired on this originally, and we worked through a few different patterns for doing the audit logging more "automatically". We talked a bit about SQLALchemy event handlers, and also a `@log_event` decorator for model methods.

In the end I think we decided that the event handlers were too generic. As for the decorators, we realized that it would force us into certain conventions that we already mostly follow, but might be perplexing. Things like always having a user as the first param, or always returning the updated object (hopefully only one) or throwing an exception. We just settled on making those logs explicit.

My opinion about it all is that logging explicitly is better in the short term, because it's simple and requires less changes to the codebase. In the long term, I think implicitly logging could be made more feasible by other changes we make to the codebase. My hunch is that we're missing an abstraction that would make that approach natural.

I think this is a great discussion, though, and I think we should keep at it until we're all on the same page. I'm in no hurry to merge this.

PR #293, opened 2018-09-17
Redirect from /home depending on requests and workspaces

richard-dds:
https://www.pivotaltracker.com/story/show/160186430

### DIFF
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -23,7 +23,10 @@ def home():
     if num_workspaces == 0:
         return redirect(url_for("requests.requests_index"))
     elif num_workspaces == 1:
-        workspace_id = g.current_user.workspace_roles[0].workspace_id
+        workspace_role = g.current_user.workspace_roles[0]
+        workspace_id = workspace_role.workspace.id
+        is_request_owner = workspace_role.role.name == "owner"
+
         if is_request_owner:
             return redirect(
                 url_for("workspaces.workspace_reports", workspace_id=workspace_id)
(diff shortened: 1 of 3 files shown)


atst/routes/__init__.py
```diff
@@ -17,7 +17,23 @@ def root():
...
 def home():
-    return redirect(url_for("requests.requests_index"))
+    is_request_owner = bool(g.current_user.owned_requests)
+    num_workspaces = len(g.current_user.workspace_roles)
+
+    if num_workspaces == 0:
+        return redirect(url_for("requests.requests_index"))
+    elif num_workspaces == 1:
+        workspace_id = g.current_user.workspace_roles[0].workspace_id
+        if is_request_owner:
```
patricksmithdds: This should check for whether the user has the owner role on the workspace, instead of if they own any requests.

tests/routes/test_home.py
```diff
@@ -17,3 +17,72 @@ def test_user_without_workspaces_has_no_workspaces_nav(client, user_session):
...
+    request = RequestFactory.create()
+    workspace = Workspaces.create(request)
+
+    user_session(request.creator)
+    response = client.get("/home", follow_redirects=False)
+
+    assert "/workspaces/{}/reports".format(workspace.id) in response.location
+
+
+def test_request_owner_with_more_than_one_workspace_redirected_to_report(
```
dandds: Should this be `redirected_to_workspaces`?

atst/routes/__init__.py
```diff
@@ -17,7 +17,26 @@ def root():
 
 @bp.route("/home")
 def home():
-    return redirect(url_for("requests.requests_index"))
+    is_request_owner = bool(g.current_user.owned_requests)
```
patricksmithdds: This declaration can be removed here, since its defined on line 28 instead.

PR #296, opened 2018-09-17
Ie/fix toggler

andrewdds:
Redo the Spend Table as a Vue component. This sidesteps the bug noted here: https://www.pivotaltracker.com/story/show/160417558

Confirmed it works in IE 10 and 11, and Edge


templates/workspaces/reports/index.html
```diff
@@ -251,9 +261,63 @@ <h2 class='spend-table__title'>Total spend per month</h2>
...
+
+          <td class='table-cell--align-right previous-month'>
+            <span v-html='formatDollars(project[prevMonthIndex] || 0)'></span>
+          </td>
+
+          <td class='table-cell--align-right current-month'>
+            <span v-html='formatDollars(project[currentMonthIndex] || 0)'></span>
+          </td>
+
+          <td class='table-cell--expand current-month meter-cell'>
```
patricksmithdds: I'm not sure exactly what's going on here, but after toggling the row, the totals get reset to 0. For example, before toggling anything, the "SF18" has some spend shown:

image

And then after expanding, the total spend for the project is shown as 0:

image

Toggling again leaves the totals as 0:

image

andrewdds: Yep I see that... fixing

andrewdds: @patricksmithdds should be fixed now

patricksmithdds:
One minor bug in IE10: the dashed lines don't extend to underneath bar chart (but they do extend in the expanded sub-rows):

image

andrewdds:
@patricksmithdds fixed
screen shot 2018-09-18 at 11 04 48 am

PR #297, opened 2018-09-17
add expiration_date to manual task order form

dandds:
For this PT story: https://www.pivotaltracker.com/story/show/160326598

Adds a task order expiration field to the manual task order form.

screen shot 2018-09-17 at 4 28 24 pm


alembic/versions/4f4defb7b440_add_expiration_date_to_task_order.py
```diff
@@ -0,0 +1,28 @@
+"""add expiration date to task order
+
+Revision ID: 4f4defb7b440
+Revises: 2572be7fb7fc
```
patricksmithdds: There was a new revision added, `4c425f17bfe8`, so this should be rebased on master and adjusted accordingly.

atst/models/task_order.py
```diff
@@ -32,6 +32,7 @@ class TaskOrder(Base):
     clin_1003 = Column(Integer)
     clin_2001 = Column(Integer)
     clin_2003 = Column(Integer)
+    expiration_date = Column(Date())
```
patricksmithdds: Random q: why do we need to instantiate `Date`, but not `Integer`?

alembic/versions/4f4defb7b440_add_expiration_date_to_task_order.py
dandds: Yup, working on that now.

atst/models/task_order.py
dandds: Hmm, we don't, looks like. I don't know why I started doing that.

PR #302, opened 2018-09-19
Name a request #160396084

dandds:
PT: https://www.pivotaltracker.com/story/show/160396084

This adds request names. The name a user enters is also used for the initial workspace name. The request name is used as the display name for the request everywhere we were previously using the ID.

**note**
- We never reference the request's `name` property directly. We always use the `displayname` property, which has the `id` as a fallback.
- @luiscielak or @andrewcroce , I made very small changes to the copy in some places to accommodate this change. Could you review and let me know if we need more changes?
- I found a bug in an earlier migration I wrote and fixed it here.

screen shot 2018-09-19 at 11 32 41 am


atst/forms/new_request.py
```diff
@@ -147,6 +147,19 @@ def validate(self, *args, **kwargs):
         format="%m/%d/%Y",
     )
 
+    name = StringField(
+        "Name your Request",
```
patricksmithdds: I think this should be `Name Your Request` (capital Y) to be consistent with the other fields.

js/lib/input_validations.js
```diff
@@ -81,8 +81,8 @@ export default {
   },
   workspaceName: {
     mask: false,
-    match: /^.{4,50}$/,
+    match: /^.{4,100}$/,
     unmask: [],
-    validationError: 'Workspace names must be at least 4 and not more than 50 characters'
+    validationError: 'Workspace and request names must be at least 4 and not more than 50 characters'
```
patricksmithdds: The message here should be updated to reflect the 50 -> 100 change.

dandds: good catch, thanks.

PR #304, opened 2018-09-19
Assign and Update Environment Roles for Workspace Users

montana-mil:
## Description
 - Users with permissions to edit workspace users can now assign and update their workspace users' environment roles on each project (and each environment on the project).
 - Each environment has its own modal in order to capture environment role changes to multiple environments. This means that there will be a field on the posted form for every environment in each project that the user being editing is a member of.
 - When a new environment role is chosen, the displayed environment role is updated. Note that the environment role is not actually updated until the user presses `Save`. If a selection was made, pressing `Cancel` will change the display back to the original role. 
 - This PR also wires up the edit member template to use the user's real data instead of hardcoded example data.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/158072276
https://www.pivotaltracker.com/story/show/160298069

## Screenshots
screen shot 2018-09-20 at 11 00 52 am

screen shot 2018-09-19 at 2 49 48 pm

### DIFF
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -27,7 +27,7 @@ class Environments(object):
         db.session.commit()
 
     @classmethod
-    def add_member(cls, user, environment, member, role=None):
+    def add_member(cls, environment, member, role):
         environment_user = EnvironmentRole(
             user=member, environment=environment, role=role
         )
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -2,7 +2,7 @@ from sqlalchemy.orm.exc import NoResultFound
 
 from atst.database import db
 from atst.models.environment import Environment
-from atst.models.environment_role import EnvironmentRole, CSPRole
+from atst.models.environment_role import EnvironmentRole
 from atst.models.project import Project
 from atst.models.permissions import Permissions
 from atst.domain.authz import Authorization
@@ -27,9 +27,9 @@ class Environments(object):
         db.session.commit()
 
     @classmethod
-    def add_member(cls, user, environment, member, role=CSPRole.NONSENSE_ROLE):
+    def add_member(cls, user, environment, member, role=None):
         environment_user = EnvironmentRole(
-            user=member, environment=environment, role=role.value
+            user=member, environment=environment, role=role
         )
         db.session.add(environment_user)
         db.session.commit()
@@ -57,9 +57,9 @@ class Environments(object):
         return env
 
     @classmethod
-    def update_environment_role(cls, ids_and_roles, workspace_user):
+    def update_environment_role(cls, user, ids_and_roles, workspace_user):
         Authorization.check_workspace_permission(
-            workspace_user.user,
+            user,
             workspace_user.workspace,
             Permissions.ADD_AND_ASSIGN_CSP_ROLES,
             "assign environment roles",
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -54,13 +54,13 @@ class Environments(object):
         return env
 
     @classmethod
-    def update_environment_role(cls, environment_data, workspace_user):
+    def update_environment_role(cls, ids_and_roles, workspace_user):
         # TODO need to check permissions?
-        for i in range(len(environment_data)):
-            new_role = environment_data[i]["role"]
-            environment = Environments.get(environment_data[i]["id"])
+        for i in range(len(ids_and_roles)):
+            new_role = ids_and_roles[i]["role"]
+            environment = Environments.get(ids_and_roles[i]["id"])
             env_role = EnvironmentRole.get(
-                workspace_user.user_id, environment_data[i]["id"]
+                workspace_user.user_id, ids_and_roles[i]["id"]
             )
             if env_role:
                 env_role.role = new_role
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -4,7 +4,6 @@ from atst.database import db
 from atst.models.environment import Environment
 from atst.models.environment_role import EnvironmentRole, CSPRole
 from atst.models.project import Project
-from atst.domain.users import Users
 
 from .exceptions import NotFoundError
 
@@ -59,13 +58,13 @@ class Environments(object):
         new_role = environment_data["user_role_name"]
         environment = Environments.get(cls=cls, environment_id=environment_data["id"])
         if workspace_user.has_environment_roles:
-            env_role = EnvironmentRole.get(workspace_user.user_id, environment.id)
+            env_role = EnvironmentRole.get(
+                workspace_user.user_id, environment_data["id"]
+            )
             env_role.role = new_role
         else:
             env_role = EnvironmentRole(
-                user=workspace_user.user,
-                environment=environment,
-                role=new_role
+                user=workspace_user.user, environment=environment, role=new_role
             )
         db.session.add(env_role)
         db.session.commit()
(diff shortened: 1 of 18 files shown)


atst/forms/data.py
```diff
@@ -141,6 +141,45 @@
     ),
 ]
 
+ENVIRONMENT_ROLES = [
+    ("no_access", {"name": "no access", "description": "No environment access."}),
```
patricksmithdds: Does this need to be an explicit role? Could we infer no access from the fact that the user does not have an environment role?

js/components/forms/edit_workspace_member.js
```diff
@@ -0,0 +1,40 @@
...
+    change: function (e) {
+      this.value = e.target.value
+    },
+    readableName: function (role) {
+      return role.replace(/[_]/g, " ")
+    },
+  },
+
+  mounted: function () {
+    console.log(this.initialData, this.choices)
```
patricksmithdds: We can yank this out.

templates/components/modal.html
```diff
@@ -1,7 +1,7 @@
 {% from "components/icon.html" import Icon %}
 
 {% macro Modal(name, dismissable=False) -%}
-  <template v-if='modals.{{name}} === true' v-cloak>
+  <div v-show="activeModal === '{{name}}'" v-cloak>
```
patricksmithdds: Just curious -- why change from `template` to `div`? I believe the benefit of `template` was that nothing gets rendered if v-show is false, but otherwise an empty element would be rendered.

atst/forms/data.py
```diff
@@ -141,6 +141,45 @@
...
+    (
```
patricksmithdds: Looks like we're missing the "Developer" role here.

templates/workspaces/members/edit.html
```diff
@@ -49,167 +50,88 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
...
+                            {% endif %}
+                          </label>
+                          {% endif %}
+                      </li>
+                      {% endfor %}
+                    </ul>
+                    <div class='block-list__footer'>
+                      <div class='action-group'>
+                        <a v-on:click="closeModal('{{ env.name }}RolesModal')" class='action-group__action usa-button'>Select Access Role</a>
+                        <a class='action-group__action icon-link icon-link--danger' v-on:click="closeModal('{{ env.name }}RolesModal')">Cancel</a>
```
patricksmithdds: Is it possible to revert any selected changes when cancelling the modal?

templates/components/modal.html
andrewdds: I think this could probably be switched back to a template. The key here is we needed `v-show` instead of `v-if` so that the form elements inside the modal are not removed from the DOM on submit.

It also might make sense for this behavior (using `v-if` or `v-show`) to be parameterized in the macro, like `{% macro Modal(name, dismissable=False, persistent=True) -%}`. I can imagine this could be a problem for other uses of the Modal.

patricksmithdds: > I can imagine this could be a problem for other uses of the Modal.

What sorts of problems has your imagination brought?

montana-mil: @andrewdds I'm going to leave out the `persistent=True` for now. If it ends up being a problem, we know where to change it

montana-mil: We were having some trouble with combinations besides `div` and `v-show`, so I'm keeping those for now.

montana-mil:
I've taken care of these comments, but I found a few more bugs. They both happen if the modal is opened, a selection is made, and the modal is closed via the `Cancel` button.

- If the `Save` button at the bottom of the page is pressed, the cancelled selection will be saved. This happens even though the display name shows correctly.
- If the modal is opened again, the default selected item is the selection made the last time the modal was opened. It should default to the selection of the user's actual environment role if they have one.

atst/domain/projects.py
```diff
@@ -49,3 +49,21 @@ def for_user(self, user, workspace):
             .filter(EnvironmentRole.user_id == user.id)
             .all()
         )
+
+    @classmethod
+    def get_all(cls, workspace_user, workspace):
+        Authorization.check_workspace_permission(
+            workspace_user.user,
+            workspace,
+            Permissions.VIEW_APPLICATION_IN_WORKSPACE,
```
richard-dds: I'm not sure if this is the right permission here, because I'm logged in as the workspace owner but I'm failing a permission check when I try to edit a member. Maybe `VIEW_WORKSPACE` would be better?

atst/models/environment_role.py
```diff
@@ -23,6 +24,18 @@ class EnvironmentRole(Base):
     user_id = Column(UUID(as_uuid=True), ForeignKey("users.id"))
     user = relationship("User", backref="environment_roles")
 
+    @classmethod
+    def get(cls, user_id, environment_id):
```
richard-dds: Any particular reason that this is on a model, rather than on a domain object? Both work of course, but our convention so far has been to put these kinds of things on the domain objects.

montana-mil: I think i had it in a domain at first, but wasn't sure if it was worth it to create a new domain file for this one method. I can add it in.

montana-mil:
> * If the modal is opened again, the default selected item is the selection made the last time the modal was opened. It should default to the selection of the user's actual environment role if they have one.

This is still happening, but @andrewdds and I agreed we could allow this for now.

atst/domain/projects.py
montana-mil: @richard-dds is the user Danny Knight? I have a feeling he is corrupt. I used these permissions because we used them above in line 26 for the `get` method.

atst/routes/workspaces.py
```diff
@@ -235,6 +253,11 @@ def update_member(workspace_id, member_id):
             role = form.data["workspace_role"]
             Workspaces.update_member(g.current_user, workspace, member, role)
 
+        try:
+            Environments.update_environment_role(g.current_user, ids_and_roles, member)
+        except UnauthorizedError:
+            pass
```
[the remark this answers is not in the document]
montana-mil: i guess that's already taken care of in `update_environment_role`?

atst/models/environment_role.py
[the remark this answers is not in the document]
montana-mil: ha oops

atst/domain/environments.py
```diff
@@ -20,9 +27,9 @@ def create_many(cls, project, names):
         db.session.commit()
 
     @classmethod
-    def add_member(cls, user, environment, member, role=CSPRole.NONSENSE_ROLE):
+    def add_member(cls, user, environment, member, role=None):
         environment_user = EnvironmentRole(
-            user=member, environment=environment, role=role.value
+            user=member, environment=environment, role=None
```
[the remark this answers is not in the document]
montana-mil: isn't it just setting the default?

templates/workspaces/members/edit.html
```diff
@@ -49,173 +50,82 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
...
-    </form>
-
-  </div>
-
-  {% endcall %}
-
+  {% for project in projects %}
   <div is='toggler' default-visible class='block-list project-list-item'>
-    <template slot-scope='props'>
+    <template slot-scope='{ isVisible, toggle }'>
```
patricksmithdds: We can't use destructuring in IE10, so this should be `props` and we need to do `props.isVisible` and `props.toggle` in this `template` tag.

andrewcroce: nice catch

PR #305, opened 2018-09-19
do not divide by zero

dandds:
Don't do it.

This fixes the template bug for requests that have a budget of zero dollars (i.e., they don't have any CLIN data).


dandds:
@andrewcroce The spend amount is always zero for real workspaces at the moment, because we don't have any CSP usage data yet. 

For the demo, I think we're going to hard-code things to return appropriate AARDVARK or BELUGA data for two specific workspaces: https://www.pivotaltracker.com/story/show/160300247

I agree that this isn't great for regular workspaces, though. Maybe we should talk further about what to do for those.

PR #309, opened 2018-09-20
Fix/modal

andrewdds:
Fixes a couple problems with the modals.

Repositions the close button, which had shifted into the wrong position when the modal structure was refactored:
screen shot 2018-09-20 at 1 37 07 pm

Adds an event emitter/handler for modal opening and closing, which adds a `modal-open` class to the document body. Vertical scrolling on the body is then disabled when a modal is open.

Adds some IE specific styles the fix the interior scrolling of the modal body.

screen shot 2018-09-20 at 11 44 03 am


templates/styleguide.html
```diff
@@ -7,6 +7,7 @@
 
 
 {% block content %}
+  <div id='modalOpen' data-modal='styleguideModal'></div>
```
patricksmithdds: Is it intentional to have the styleguide modal open by default?

andrewdds: deeeerp that was for testing

PR #311, opened 2018-09-21
Always use the front-end validation message

andrewdds:
This modifies the TextInput macro/component to only display the front-end defined error message. Even when a server-side error is present, the more detailed front-end message will be displayed.

Addresses https://www.pivotaltracker.com/story/show/160627446


andrewdds:
@dandds that seems like a good idea to me

PR #315, opened 2018-09-21
Audit log v2

richard-dds:
Log audit events automatically when any `Auditable` model is changed, where `Auditable` is a new mixin.


tests/domain/test_audit_log.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.fixture(scope="function")
+def developer():
+    return UserFactory.from_atat_role("default")
+
+
+def test_non_admin_cannot_view_audit_log(developer):
+    with pytest.raises(UnauthorizedError):
+        AuditLog.get_all_events(developer)
```
patricksmithdds: Could we add a simple test here that a CCPO can view audit logs? It would be nice to see a request or workspace created or modified and see that the CCPO can see the automatically created audit events.

atst/models/mixins/auditable.py
```diff
@@ -0,0 +1,65 @@
...
+        _obj = getattr(_obj, item, default)
+    return _obj
+
+
+def camel_to_snake(camel_cased):
+    s1 = re.sub("(.)([A-Z][a-z]+)", r"\1_\2", camel_cased)
+    return re.sub("([a-z0-9])([A-Z])", r"\1_\2", s1).lower()
+
+
+class AuditableMixin(object):
```
patricksmithdds: This mixin is great! 😍

atst/models/audit_event.py
```diff
@@ -0,0 +1,41 @@
...
+class AuditEvent(Base, TimestampsMixin):
+    __tablename__ = "audit_events"
+
+    id = types.Id()
+
+    user_id = Column(UUID(as_uuid=True), ForeignKey("users.id"), index=True)
+    user = relationship("User", backref="audit_events")
+
+    workspace_id = Column(UUID(as_uuid=True), ForeignKey("workspaces.id"), index=True)
+    workspace = relationship("Workspace", backref="audit_events")
```
patricksmithdds: Similar to workspace, I think we should tracking the associated request, if it exists. Being able to see all the requestor's and approver's actions on a request would be good for an auditor, I think.

atst/models/audit_event.py
```diff
@@ -0,0 +1,41 @@
...
+
+    resource_name = Column(String(), nullable=False)
```
patricksmithdds: Is `resource_type` a more accurate name for this? For example, a request has a name & id, so it may be unclear if `resource_name` and `resource_id` should be the name and id of the request instead of "request" and its id.

atst/models/audit_event.py
```diff
@@ -0,0 +1,41 @@
...
+    resource_id = Column(UUID(as_uuid=True), index=True, nullable=False)
+    action = Column(String(), nullable=False)
```
patricksmithdds: I think some of the logged events are hard to read with just ids in there. Does it make sense to log `displayname`, if the property exists?

I'm particularly thinking of the request status event logged event. If we track both the request id and the displayname, we would go from this:

> Amanda Adamson performed create on request_status_event 62b76e4c-7cd7-4a92-88ca-f6981e9c0035

to

> Amanda Adamson performed create on request_status_event 62b76e4c-7cd7-4a92-88ca-f6981e9c0035 ("Submitted") on request fb01955f-2a9d-43f0-b0db-82205b32469c

which is more usable information.

atst/models/audit_event.py
richard-dds: Yep, you're right. Changed

tests/domain/test_audit_log.py
richard-dds: 👍

atst/models/audit_event.py
richard-dds: 👍

PR #316, opened 2018-09-21
Buildinfo

ddsdevon:
This PR adds the creation of static files contain build related information to the containers being built.

- Add `script/generate_build_info.sh`, which creates the two files `static/buildinfo.html` and `static/buildinfo.json`; each file contains information about the commit the code was built from, as well as what is doing the building
- Adds the buildinfo.* files to gitignore
- Updates the Travis config to call `script/generate_build_info.sh`


ddsdevon:
@patricksmithdds Fixed that issue where I left the global includes commented out

PR #318, opened 2018-09-24
Reports task order info #159975038

dandds:
PT story: https://www.pivotaltracker.com/story/show/159975038

This wires real data for the "Task Order" box on the reports page. Re-run the `seed_sample.py` script to get real data for the mock workspace.

**notes**
- I updated the `seed_sample.py` script to add real data for the financial section of a request if we're using to to make a workspace.
- After discussion, Hardik and I decided that "Manage Task Order" should link to the workspace settings form, since that's where users will eventually go to update the TO and other things related to the workspace.

### DIFF
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -191,7 +191,9 @@ class TaskOrderFactory(Base):
     source = Source.MANUAL
     funding_type = FundingType.PROC
     funding_type_other = None
-    number = factory.Faker("md5")
+    number = factory.LazyFunction(
+        lambda: "".join(random.choices(string.ascii_uppercase + string.digits, k=13))
+    )
     expiration_date = factory.LazyFunction(
         lambda: datetime.date(
             datetime.date.today().year + random.randrange(1, 15), 1, 1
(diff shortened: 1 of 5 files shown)


tests/factories.py
```diff
@@ -172,7 +191,9 @@ class Meta:
     source = Source.MANUAL
     funding_type = FundingType.PROC
     funding_type_other = None
-    number = factory.Faker("md5")
+    number = factory.LazyFunction(
+        lambda: "".join(random.choices(string.ascii_uppercase + string.digits, k=13))
+    )
     expiration_date = factory.LazyFunction(
         lambda: datetime.date(
             datetime.date.today().year + random.randrange(1, 15), 1, 1
```
andrewdds: Not sure how much it matters, but do we want to tighten up this range a bit? 15 years in the future is a bit long perhaps.

Also, I think this will always be January 1.

andrewdds: screen shot 2018-09-24 at 11 10 55 am

dandds: What seems reasonable? 5?

andrewdds: perhaps

patricksmithdds: > Also, I think this will always be January 1.
I'm not sure this is a valid assumption. I believe it may be aligned with the fiscal year instead of the calendar year.

PR #324, opened 2018-09-24
Use proper display name for workspace roles

richard-dds:
The members list screen was showing roles using `Role.name`, which presented weird-looking strings like `billing_auditor` and `ccpo`. This PR adds a `Role.display_name` column, seeds it from `seed_roles.py`, and displays it in `members/index.html`.

screen shot 2018-09-24 at 2 22 18 pm


alembic/versions/c19ae79d2521_add_role_display_name.py
```diff
@@ -0,0 +1,45 @@
...
+# revision identifiers, used by Alembic.
+revision = "c19ae79d2521"
+down_revision = "359caaf8c5f1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
```
patricksmithdds: Is this change intended?

script/seed_roles.py
```diff
@@ -13,6 +13,7 @@
 roles = [
     Role(
         name="ccpo",
+        display_name="CCPO",
```
patricksmithdds: There's a couple descriptions available in `forms/data.py` for some of these roles as well. Could we add the descriptions here, and then instead of hard-coding stuff in `forms/data.py`, just look up the subset of available roles where we need them?

alembic/versions/c19ae79d2521_add_role_display_name.py
richard-dds: It wasn't, but alembic added it with `--autogenerate`, and it does seem necessary. I thought about opening up another PR to add it, but it seemed like more trouble than it was worth.

script/seed_roles.py
richard-dds: We could, but I'm not exactly sure what the right way would be to do that with WTForms. Could we punt on it? Currently we're just not using `Role.description`.

alembic/versions/c19ae79d2521_add_role_display_name.py
patricksmithdds: Makes sense to me.

script/seed_roles.py
patricksmithdds: That makes sense. I've put a `tech debt` issue in pivotal for reconciling the possible duplication.

PR #326, opened 2018-09-24
New project validation #160064518

dandds:
PT story: https://www.pivotaltracker.com/story/show/160064518

This moves new project and environment form validation into the Vue component. This way we can trigger validation errors before the user is shown the confirmation modal.

### DIFF
--- a/js/components/forms/new_project.js
+++ b/js/components/forms/new_project.js
@@ -74,7 +74,7 @@ export default {
     },
 
     validateAndOpenModal: function (modalName) {
-      const childrenValid = this.$children.reduce((previous, newVal) => {
+      let isValid = this.$children.reduce((previous, newVal) => {
         // display textInput error if it is not valid
         if (!newVal.showValid) {
           newVal.showError = true
@@ -83,18 +83,17 @@ export default {
         return newVal.showValid && previous
       }, true)
 
-      const hasNames = this.environmentsHaveNames()
-      const uniqNames = this.envNamesAreUnique()
-
-      if (!hasNames) {
+      if (!this.environmentsHaveNames()) {
+        isValid = false
         this.showMissingError = true
       }
 
-      if (!uniqNames) {
+      if (!this.envNamesAreUnique()) {
+        isValid = false
         this.showUniqueError = true
       }
 
-      if (childrenValid && hasNames && uniqNames) {
+      if (isValid) {
         this.openModal(modalName)
       }
     }
--- a/js/components/forms/new_project.js
+++ b/js/components/forms/new_project.js
@@ -3,6 +3,12 @@ import textinput from '../text_input'
 
 const createEnvironment = (name) => ({ name })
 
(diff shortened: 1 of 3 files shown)


js/components/forms/new_project.js
```diff
@@ -50,6 +62,40 @@ export default {
...
+      let isValid = this.$children.reduce((previous, newVal) => {
+        // display textInput error if it is not valid
+        if (!newVal.showValid) {
+          newVal.showError = true
+        }
+
+        return newVal.showValid && previous
+      }, true)
+
+      if (!this.environmentsHaveNames()) {
```
patricksmithdds: We may need a third error -- currently you if you try to submit a project with one (or more) valid environment names and an empty name, the error shown says that you must provide at least one environment name:

image

It should instead say something to the effect that names cannot be empty.

templates/fragments/edit_project_form.html
```diff
@@ -42,11 +42,12 @@ <h1>{{ title_text }}</h1>
...
-      {% for error in form.environment_names.errors %}
-        {{ Alert(error, level="error") }}
-      {% endfor %}
-    {% endif %}
+    <div v-if="showMissingError">
+      {{ Alert("Provide at least one environment name.", level="error") }}
+    </div>
+    <div v-if="showUniqueError">
+      {{ Alert("Environment names must be unique.", level="error") }}
+    </div>
```
patricksmithdds: Will we be showing more than one error at a time? If not, it might be more efficient for the component to have an `error` attribute with a string error message rather than a bool attribute for every type of error.

dandds: I don't really like the multiple bools either. You can have more than one error at a time (in this case, two redundant env names and an empty env), though. I was thinking it might eventually be useful to take a more React-ish approach and keep a list of errors and render child `Alert`s from the parent form, but that seems like a big lift for the moment.

patricksmithdds: I think keeping a list of errors is fine -- I'm not sure what you mean by a "big lift" for that, though?

dandds:
@patricksmithdds I pushed the changes we talked about with the temp fix for #160768940. Give it a look when you have a minute.

js/components/forms/new_project.js
```diff
@@ -50,6 +56,51 @@ export default {
...
+        // more do not have names
+        return this.environments.every((e) => e.name !== "")
+      } else {
+        return true
+      }
+    },
+
+    envNamesAreUnique: function () {
+      const names = this.environments.map((e) => e.name)
+      return [...new Set(names)].length == this.environments.length
```
patricksmithdds: I don't believe we can use the spread syntax in IE: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Browser_compatibility

dandds: bummer

PR #327, opened 2018-09-24
Dummy help page

andrewdds:
This adds a single public '/help' route with some dummy navigation and content.
A handful of "learn more" links throughout are now directing to this page.

Also fixes a few layout issues with the topbar and footer.

Also makes a separate `base_public` template, which is inherited by public pages, such as login and help. The login route now has a separate template inheriting from this.

help

### DIFF
--- a/templates/base_public.html
+++ b/templates/base_public.html
@@ -24,7 +24,7 @@
       </a>
 
       {% if g.current_user %}
-        <a href="/" class="topbar__link">
+        <a href="{{ url_for('atst.home') }}" class="topbar__link">
           <span class="topbar__link-label">{{ g.current_user.first_name + " " + g.current_user.last_name }}</span>
           {{ Icon('avatar', classes='topbar__link-icon') }}
         </a>
@@ -48,7 +48,7 @@
     <div class='app-footer__info'>
       <h5 class='app-footer__info__title'>Joint Enterprise Defense Infrastructure</h5>
 
-      <a href='/help' class='icon-link app-footer__info__link' target='_blank' rel='noopener noreferrer'>
+      <a href='{{ url_for('atst.helpdocs') }}' class='icon-link app-footer__info__link' target='_blank' rel='noopener noreferrer'>
         {{ Icon('help') }}
         <span>JEDI Help</span>
       </a>
--- a/templates/base_public.html
+++ b/templates/base_public.html
@@ -1,0 +1,67 @@
+{% from "components/icon.html" import Icon %}
+
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>{% block title %}JEDI Cloud{% endblock %}</title>
+  {% assets "css" %}
+    <link rel="stylesheet" href="{{ ASSET_URL }}" type="text/css">
+  {% endassets %}
+  <link rel="icon" type="image/x-icon" href="/static/img/favicon.ico">
+</head>
+<body>
+
+<div id='app-root'>
+
+  <header class="topbar topbar--public">
+    <nav class="topbar__navigation">
+      <a href="{{ url_for('atst.home') }}" class="topbar__link topbar__link--home">
+        {{ Icon('shield', classes='topbar__link-icon') }}
+        <span class="topbar__link-label">JEDI Cloud</span>
+      </a>
+
+      {% if g.current_user %}
+        <a href="/" class="topbar__link">
+          <span class="topbar__link-label">{{ g.current_user.first_name + " " + g.current_user.last_name }}</span>
+          {{ Icon('avatar', classes='topbar__link-icon') }}
+        </a>
+
+        <a href="{{ url_for('atst.logout') }}" class="topbar__link" title='Log out of JEDI Cloud'>
+          {{ Icon('logout', classes='topbar__link-icon') }}
+        </a>
+      {% else %}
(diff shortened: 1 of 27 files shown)


templates/base_public.html
```diff
@@ -0,0 +1,67 @@
...
+<div id='app-root'>
+
+  <header class="topbar topbar--public">
+    <nav class="topbar__navigation">
+      <a href="{{ url_for('atst.home') }}" class="topbar__link topbar__link--home">
+        {{ Icon('shield', classes='topbar__link-icon') }}
+        <span class="topbar__link-label">JEDI Cloud</span>
+      </a>
+
+      {% if g.current_user %}
```
patricksmithdds: `current_user` is set on the global context only for authenticated routes: https://github.com/dod-ccpo/atst/blob/04d03f3c93deedc4048ba53e959b9190e8ff32e7/atst/domain/auth.py#L21-L24 

Currently, this is showing the Amanda user because of some leftover test data (which is actually [removed in #315](https://github.com/dod-ccpo/atst/pull/315/files#diff-fc9c4eadc6e0dda96dce62a0c7cc6f9d)). After that's merged this will always just show the "Log in" button for public routes. If that's ok, perhaps we should remove the conditional here? Or, we could set the `current_user` for _all_ routes instead.

templates/footer.html
```diff
@@ -1,3 +1,12 @@
+{% from "components/icon.html" import Icon %}
+
 <footer class='app-footer'>
-  <h5>Joint Enterprise Defense Infrastructure</h5>
+  <div class='app-footer__info'>
+    <h5 class='app-footer__info__title'>Joint Enterprise Defense Infrastructure</h5>
+
+    <a href='/help' class='icon-link app-footer__info__link' target='_blank' rel='noopener noreferrer'>
```
patricksmithdds: Can we use the `url_for` helper here instead of hardcoding `/help`? If we rely on `url_for` for all URLs, we can more easily change URLs in the future.

templates/base_public.html
```diff
@@ -0,0 +1,67 @@
...
+    </nav>
+  </header>
+
+  {% block content %}{% endblock %}
+
+  <footer class='app-footer'>
+    <div class='app-footer__info'>
+      <h5 class='app-footer__info__title'>Joint Enterprise Defense Infrastructure</h5>
+
+      <a href='/help' class='icon-link app-footer__info__link' target='_blank' rel='noopener noreferrer'>
```
patricksmithdds: We should use `url_for` here.

templates/fragments/pending_financial_verification.html
```diff
@@ -7,6 +9,11 @@
   Once the Task Order has been created, you will be asked to provide
   details about the task order in the Financial Verification step.
 </p>
-<p>
-  <i>Learn more</i> about the JEDI Cloud Task Order and the Financial Verification process.
-</p>
+
+<div class='alert__actions'>
+  <a href='/help' class='icon-link'>
+    {{ Icon('help') }}
```
patricksmithdds: Looks like this pattern is repeated a couple times -- is it worth adding a macro for it?

templates/base_public.html
```diff
@@ -0,0 +1,67 @@
...
+<body>
+
+<div id='app-root'>
+
+  <header class="topbar topbar--public">
+    <nav class="topbar__navigation">
+      <a href="{{ url_for('atst.home') }}" class="topbar__link topbar__link--home">
+        {{ Icon('shield', classes='topbar__link-icon') }}
+        <span class="topbar__link-label">JEDI Cloud</span>
+      </a>
```
patricksmithdds: For some reason, this link isn't at the top left on IE10 on Win7:

image

Looks correct in chrome, though.

templates/base_public.html
andrewdds: Its possible to be logged in AND on a public route, so I don't think that will work.

templates/fragments/pending_financial_verification.html
andrewdds: theres an 'actions' param in the Alert macro that takes a list of dicts to build the alert links. But because the fragments were used here for the Alert content, I couldn't see a way to use that. I'm open to suggestions.

templates/base_public.html
andrewdds: Fixed...
screen shot 2018-09-25 at 8 11 53 am

andrewdds:
@dandds thanks for fixing the tests

templates/fragments/pending_financial_verification.html
patricksmithdds: I was thinking we could have a `HelpLink` macro: 

```

{% macro HelpLink() -%}
  <div class='alert__actions'>
    <a href='/help' class='icon-link'>
      {{ Icon('help') }}
      {% if caller %}
        <div class='alert__message'>{{ caller() }}</div>
      {% endif %}
    </a>
  </div>
{%- endmacro %}
```

And then call that macro here, like:

```
{% call HelpLink() %}
  Learn more about the JEDI Cloud Task Order and the Financial Verification process.
{% endcall %}
```

I don't believe there'd be any issue calling a macro inside this fragment, but I may be misunderstanding what you're saying.

andrewdds: ohh I thought you meant the helplinks specifically inside the Alert. I think there might be too much variation in the help links to make this a valuable macro. There are modifier classes, the markup order might be different, the `a` attributes might need to vary, and sometimes theres an onclick handler, rather than an href, etc.

Seems like a bit too much work to parameterize all that.

patricksmithdds: Sounds good -- they all looked the same at first glance, but if there's lots of variability, it makes sense to not have a macro.

PR #328, opened 2018-09-24
Empty state for workspace reports should show existing projects

dandds:
PT story: https://www.pivotaltracker.com/story/show/160732667

We only show the empty state for the reports page if the workspace has no projects. I also updated the reporting methods so that the projects and environments for a workspace are displayed even if there's no spend.


dandds:
@patricksmithdds Good catch. Added it.

andrewdds:
@dandds the month displayed in the selector should correspond to the month selected in the URL, not just be the current month. This behavior is already functioning with the dummy data.

PR #329, opened 2018-09-25
Fix link to github repo in buildinfo.json

patricksmithdds:
The current buildinfo (https://www.atat.codes/static/buildinfo.html) links to the commit with an extra `https://github.com/`, for example: `https://github.com///github.com/dod-ccpo/atst/commit/2bbb859ad8c92ece65807457c56107a58e5f500c`. This change should make a valid link.


ddsdevon:
Hrm... I'm guessing Travis is using an HTTPS based remote URL, versus the SSH one we usually use locally (thus changing format of remote.origin.url). We can probably drop the `cut` related stuff entirely, in that case.
I'll run some local tests...

ddsdevon:
@patricksmithdds Okay, I think this covers us... I checked out a second copy of the repo via HTTPS, and it seems to work for that too. Yay multitasking! \o/

PR #334, opened 2018-09-25
Always set initial text input value when field data is not None

patricksmithdds:
Fixes an issue where $0 entries for CLIN values were not being rendered. Now, they will be shown appropriately:

image


montana-mil:
Were they not being rendered after submitting? The zeros are coming up for me in the form regardless.

patricksmithdds:
@montana-mil Yeah, the error was happening if the form was submitted and re-rendered with errors, then the $0 fields would render with no data.

PR #338, opened 2018-09-25
Make BA code optional

andrewcroce:
When all other fields fail... this field stands alone

screen shot 2018-09-25 at 2 34 22 pm

### DIFF
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -2,7 +2,7 @@ import re
 import pendulum
 from wtforms.fields.html5 import DateField, EmailField
 from wtforms.fields import StringField, FileField
-from wtforms.validators import InputRequired, Required, Optional, Email, Regexp
+from wtforms.validators import InputRequired, Required, Email, Regexp
 from flask_wtf.file import FileAllowed
 
 from atst.domain.exceptions import NotFoundError
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -122,7 +122,7 @@ class BaseFinancialForm(ValidatedForm):
 
     ba_code = StringField(
         "Program Budget Activity (BA) Code (Optional)",
-        description="BA Code is used to identify the purposes, projects, or types of activities financed by the appropriation fund. <br/><em>It should be two digits, followed by a letter.</em>",
+        description="BA Code is used to identify the purposes, projects, or types of activities financed by the appropriation fund. <br/><em>It should be two digits, followed by an optional letter.</em>",
         validators=[Regexp(BA_CODE_REGEX)],
     )
 
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -2,7 +2,7 @@ import re
 import pendulum
 from wtforms.fields.html5 import DateField, EmailField
 from wtforms.fields import StringField, FileField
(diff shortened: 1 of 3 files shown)


atst/forms/financial.py
```diff
@@ -121,9 +121,9 @@ def is_missing_task_order_number(self):
     )
 
     ba_code = StringField(
-        "Program Budget Activity (BA) Code",
+        "Program Budget Activity (BA) Code (Optional)",
         description="BA Code is used to identify the purposes, projects, or types of activities financed by the appropriation fund. <br/><em>It should be two digits, followed by a letter.</em>",
```
patricksmithdds: While we're here, can we update the text here to indicate the letter is optional as well.

montana-mil:
I ended up just taking the Optional() validator out altogether. The form works as expected and the tests pass. When `Optional()` was included (in either argument position), there were about 7 tests not passing inexplicably. @dandds and I went over this and agreed this is the best way to handle it, though it feels uncomfortable since we don't know why it works. (Or rather, why including `Optional()` *didn't* work)

patricksmithdds:
@montana-mil @dandds I poked a little bit more into why `Optional()` wasn't working and I think it's just a problem with the tests. The `Optional` validator [checks whether `raw_data` is set](https://github.com/wtforms/wtforms/blob/283b2803206825158834f1828bbf749c129b7c47/src/wtforms/validators.py#L213-L240), and the field's `raw_data` is only set [when the field receives the `form_data` kwarg](https://github.com/wtforms/wtforms/blob/283b2803206825158834f1828bbf749c129b7c47/src/wtforms/fields/core.py#L335) (which is [passed in from the form](https://github.com/wtforms/wtforms/blob/283b2803206825158834f1828bbf749c129b7c47/src/wtforms/form.py#L110-L116)).

In the routes, `flask_wtf` handles setting the `form_data` for us, but in tests where we operating outside of a request context, we have to do it manually. 

All of that is to say, we can use the `Optional()` validator if we update the tests like:

```diff
diff --git a/tests/forms/test_financial.py b/tests/forms/test_financial.py
index 970e4b6b..dc89ab30 100644
--- a/tests/forms/test_financial.py
+++ b/tests/forms/test_financial.py
@@ -74,8 +74,8 @@ def test_treasury_code_validation(input_, expected):
     ],
 )
 def test_ba_code_validation(input_, expected):
-    form_data = {"ba_code": input_}
-    form = FinancialForm(data=form_data)
+    form_data = ImmutableMultiDict([("ba_code", input_)])
+    form = FinancialForm(form_data)
     form.validate()
     is_valid = "ba_code" not in form.errors
```

andrewdds:
I misunderstood the request. The field is NOT optional, just the letter in the code is optional.
`Required()` validator added back, just changed the description.

atst/forms/financial.py
```diff
@@ -2,7 +2,7 @@
 import pendulum
 from wtforms.fields.html5 import DateField, EmailField
 from wtforms.fields import StringField, FileField
-from wtforms.validators import InputRequired, Required, Email, Regexp
+from wtforms.validators import InputRequired, Required, Optional, Email, Regexp
```
patricksmithdds: This `Optional` import isn't used anywhere anymore.

PR #339, opened 2018-09-25
Hook up the "Add Environment Access" link when adding a member

patricksmithdds:
After adding a member to a workspace, an alert is shown with a link to add environment access that previously did nothing. Now, it will link to the edit workspace member page:

image


patricksmithdds:
@dandds I had to fix some code formatting issues, so could you please re-review at your leisure?

PR #341, opened 2018-09-26
Initial audit log UI cleanup

andrewdds:
Adds timestamp, highlights activity user, adds some layout.

Note: the activity description still needs some work to make it more human readable, but this is a first step mainly focused on layout.

activity-log

### DIFF
--- a/atst/models/audit_event.py
+++ b/atst/models/audit_event.py
@@ -42,7 +42,6 @@ class AuditEvent(Base, TimestampsMixin):
 
         return " ".join([user_str, action_str, display_name_str, scope_str])
 
-
     @property
     def activity_description(self):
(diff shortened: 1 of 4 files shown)


atst/models/audit_event.py
```diff
@@ -42,6 +42,20 @@ def __str__(self):
 
         return " ".join([user_str, action_str, display_name_str, scope_str])
 
+    @property
+    def activity_description(self):
```
dandds: If this is the string representation of the event we're going to use in the log, I think we should just replace `__str__` above with this version.

andrewdds: Its just a description of the activity, not a full representation of the event. I don't think we need the __str__ thing really, but I left it alone for now.

PR #347, opened 2018-09-27
put workspace role definitions in one place

dandds:
PT story: https://www.pivotaltracker.com/story/show/160745296

This puts all of the workspace and atat `Role` definitions in a constant in `atst.domain.roles`.

**notes**
- I thought about putting the definitions in the role model file but decided it was a domain concern.
- I also thought about having the form data pull from the database instead of the constant, in case we update the constant but forget to reseed the database. Maybe that would be better?
- It might be useful to have an attribute on the `roles` table to define what kind of role it is. "ccpo" and "default" are not available as workspace roles. That exclusion is hard-coded at the moment.


atst/forms/data.py
```diff
@@ -104,41 +106,9 @@
...
-    (
-        "security_auditor",
-        {
-            "name": "Security Auditor",
-            "description": "Views only the projects and environments they are granted access to. Can also view activity logs.",
-        },
-    ),
+    (role["name"], {"name": role["display_name"], "description": role["description"]})
+    for role in ROLE_DEFINITIONS
+    if role["name"] not in ["ccpo", "default"]
```
patricksmithdds: Would it be preferable to have the role definitions separate out the ATAT roles from the Workspace roles? Rather than having one list of all the roles, we could have one list of ATAT roles with just default, ccpo and then another list of Workspace roles with everything else. I think that's one step closer to the role "type" without affecting the schema.

dandds:
@patricksmithdds I made the split you suggested. I think a `type` column for roles is a good idea, but maybe premature (maybe?). I wonder if we should have a longer discussion about role management after the demo.

PR #350, opened 2018-09-27
Coverage #160694116

dandds:
This adds `pytest-cov` per PT story https://www.pivotaltracker.com/n/projects/2160940/stories/160694116. Coverage runs by default in pytest and a total coverage score below 90 fails.

We're currently at 89%; I'm waiting on https://github.com/dod-ccpo/atst/pull/343 to be merged, since that should put us over the bar (thanks @montana-mil ).

**note**
This locks our Python version at 3.6.6, instead of 3.6.*. `pytest-cov` breaks on lesser versions. This means if you're running some other minor version you will have to switch. We should prepare everyone and merge at a scheduled time we all know about.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = atst/routes/dev.py
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ config/dev.ini
 
 # uploads
 /uploads
+
+# coverage output
+.coverage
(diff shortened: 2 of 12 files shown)


dandds:
Also noting here that if you're running a different version, `pipenv install --python 3.6.6 --dev` should reinstall your dependencies correctly after you install 3.6.6.

We should probably also change `script/update` and `script/setup` to specify the version number.

dandds:
Aaand we still need more tests to pass. :( I'll get to that this week.

PR #351, opened 2018-09-27
Miscellaneous fixes

patricksmithdds:
This PR fixes a couple things that we noticed while Hardik was running through the demo:

* adding a user was failing because the email entered conflicted an existing email entered for another user. The DOD id is the field that defines the uniqueness of the user and, in practice, the email would not be re-used for multiple DOD ids, so the unique constraint can be dropped.
* the input validator for email addresses was updated to be case-insensitive.
* the modal content for CCPO approval & acceptance was updated. This was also showing the incorrect modal after submitting the financial verification (telling the user to prepare their TO after they submitted a TO)
* if a TO number was not found, the error shown was "Input required". Instead, we should be showing the "TO number not found" error:
image
* The required fields if the JEDI request was over the threshold were not being shown on the review page (number of sessions & data transfer amounts)
* If an error occurs when submitting a CCPO approval (if the phone number is entered incorrectly, for example), the review form should be shown by default.


dandds:
@patricksmithdds 

> the email would not be re-used for multiple DOD ids

What about when users can update their own profiles and email addresses? We know that the address on their email CAC cert won't necessarily be the one they use day-to-day and we've talked about allowing them to edit their email for that reason. Is it a security concern if a user can enter someone else's email and potentially begin receiving email notifications for that account?

atst/routes/requests/financial_verification.py
```diff
@@ -71,9 +71,7 @@ def update_financial_verification(request_id):
                 )
             else:
                 return redirect(
-                    url_for(
-                        "requests.requests_index", modal="pendingFinancialVerification"
```
dandds: my bad :(

patricksmithdds:
> Is it a security concern if two accounts have the same email and receive notifications for both?

That's a good point -- I think I was too hasty in removing the constraint and we should consider this more deeply first. I've removed the commit that removed the constraint, so emails will remain unique.

PR #355, opened 2018-10-01
IE fixes

andrewdds:
Fixes a few things:

https://www.pivotaltracker.com/story/show/160794811
Fixes box layout at the top of the budget page, so they split 50/50 correctly

https://www.pivotaltracker.com/story/show/160794046
Fixes a problem with all layouts that prevented the global layout container from shrinking to fit the window. This was just most apparent on this screen, but was problem everywhere.

Also fixes another problem related to this, where the screen metrics were being measured too early, and rendering the chart wrong. This adds an event handler that runs the screen metrics measurement once again when page loads.

https://www.pivotaltracker.com/story/show/160795011
Adds a missing fallback meter bar

https://www.pivotaltracker.com/story/show/160794726
Restricts text width in EmptyState block, which was flowing off screen in IE

Fixes the header again
https://www.pivotaltracker.com/story/show/160876580

### DIFF
--- a/js/components/charts/budget_chart.js
+++ b/js/components/charts/budget_chart.js
@@ -34,6 +34,7 @@ export default {
   mounted: function () {
     this._setDisplayedMonths()
     this._setMetrics()
+    addEventListener('load', this._setMetrics)
     addEventListener('resize', this._setMetrics)
   },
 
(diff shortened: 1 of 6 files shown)


montana-mil:
looks good! 👍

PR #357, opened 2018-10-01
Display user name and workspace name in audit log

richard-dds:
screen shot 2018-10-01 at 11 40 45 am

### DIFF
--- a/templates/audit_log.html
+++ b/templates/audit_log.html
@@ -20,7 +20,16 @@
               <h2 class='h4 audit-log__item__name'>
                 {{ event.user.full_name if event.user else "ATAT System" }}
               </h2>
-              {{ event.activity_description }}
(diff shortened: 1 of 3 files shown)


templates/audit_log.html
```diff
@@ -20,7 +20,16 @@ <h1 class="block-list__title">Activity History</h1>
...
+              {{ event.action }} {{ event.resource_type }} <code>{{ event.resource_id }}</code>
+              {% if event.display_name %}
+                ({{ event.display_name }})
+              {% endif %}
+
+              <br>
+
+              {% if event.workspace %}
+                in Workspace <code>{{ event.workspace_id }}</code> ({{ event.workspace.name }})
+              {% endif %}
```
patricksmithdds: I think we should also show the request id/name, if there is a request associated with the event.

richard-dds: Whoops! Thanks for the heads up.

PR #358, opened 2018-10-01
End to end test #160690740

dandds:
PT story: https://www.pivotaltracker.com/story/show/160690740

This adds a harness for end-to-end testing with Selenium and BrowserStack. This PR only adds two very basic acceptance tests as examples. I'd like to get eyes on the setup before we dive into adding a lot of new end-to-end tests.

**note**
I'm having the top-level script use a different database, `atat_selenium`. We could write a separate seed script to pre-populate this with data useful for the end-to-end tests without affecting the other databases. It's also possible the test factories will just work in this context, but I haven't gotten that far with it.


dandds:
I take it back, since we're waiting on the other one.

dandds:
I'm just going to keep dropping notes here, I guess.

I did a little research and it seems like it's difficult to get Selenium to supply client certs for SSL transactions. Most people talk about using a Firefox profile with the certs included, like this [SO thread](https://stackoverflow.com/questions/38337976/python-selenium-how-to-specify-a-client-certificate-to-use-in-client-ssl-authe/38498525#38498525). That would really constrain us, obviously. The other option is to _not_ do client auth in the environment we use for Selenium testing. We'd have to be careful to not leave security holes and find a way to stub the CAC info we'd normally get from SSL. (It's possible we could monkeypatch the running application under test and force a reload, but that seems like a big lift.)

config/selenium.ini
```diff
@@ -0,0 +1,4 @@
+[default]
+PGDATABASE = atat_selenium
+CRL_DIRECTORY = tests/fixtures/crl
+WTF_CSRF_ENABLED = false
```
patricksmithdds: Why do we need to disable CSRF protection?

pytest.ini
```diff
@@ -2,3 +2,4 @@
 norecursedirs = .venv .git node_modules
 env =
   D:FLASK_ENV=test
+addopts = --ignore=tests/acceptance/
```
patricksmithdds: Apologies for letting the coverage branch hang out and causing future merge conflicts ☹️

tests/acceptance/test_basic.py
```diff
@@ -0,0 +1,50 @@
...
+
+import cryptography.x509 as x509
+from cryptography.hazmat.backends import default_backend
+
+
+USER_CERT = "ssl/client-certs/atat.mil.crt"
+
+
+@pytest.mark.parametrize("browser_type", BROWSERSTACK_CONFIG.keys())
+@pytest.mark.usefixtures("live_server")
```
patricksmithdds: What is this "live_server" fixture? Is that just a requirement from browerstack?

tests/acceptance/test_basic.py
```diff
@@ -0,0 +1,50 @@
...
+        "first_name": cn_parts[1],
+        "dod_id": cn_parts[-1],
+        "atat_role_name": "developer",
+    }
+    return Users.get_or_create_by_dod_id(**user_info)
+
+
+@pytest.mark.usefixtures("live_server")
+def test_login(drivers, client, app, valid_user_from_cert):
+    driver = drivers["win7_ie10"]
```
patricksmithdds: Any reason not to parameterize the browser type here, too?

tests/acceptance/test_basic.py
```diff
@@ -0,0 +1,50 @@
...
+        "last_name": cn_parts[0],
+        "first_name": cn_parts[1],
+        "dod_id": cn_parts[-1],
+        "atat_role_name": "developer",
+    }
+    return Users.get_or_create_by_dod_id(**user_info)
+
+
+@pytest.mark.usefixtures("live_server")
+def test_login(drivers, client, app, valid_user_from_cert):
```
patricksmithdds: Do we need to get a user for this test? Is this more of an example for the future?

README.md
```diff
@@ -131,6 +131,27 @@ To re-run tests each time a file is changed:
...
     pipenv run ptw
 
+### Selenium Tests
+
+Selenium tests rely on BrowserStack. In order to run the Selenium tests
+locally, you need BrowserStack credentials. The user email and key can
+be found on the account settings page. To run the selenium tests:
+
+```
+BROWSERSTACK_TOKEN=<token> BROWSERSTACK_EMAIL=<email> ./script/selenium_test
```
patricksmithdds: Do we have a shared token/email (I didn't find it in 1Password) that devs can use to test?

script/selenium_test
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+# run BrowserStackLocal in the background
+echo "starting BrowserStack local client..."
+./$BSL_FILE --key $BROWSERSTACK_TOKEN &
+BSL_ID=$!
+
+# run example selenium script that fetches the home page
+echo "running selenium tests"
+pipenv run pytest tests/acceptance -s
```
patricksmithdds: Not sure if this needs to be addressed, but I got an error when running `pytest` here (I had forgotten to install packages) which caused `BrowserStackLocal` to not be killed afterwards. I only noticed because BrowserStack spits out a bunch of information if you try to start a second instance and then I was getting a message that `kill` couldn't find the process (though it continues and runs the tests successfully).

pytest.ini
dandds: Not a problem. Very easy to resolve.

config/selenium.ini
dandds: You're right, we don't. I think I more or less copied the test config.

tests/acceptance/test_basic.py
dandds: No, this is a fixture we get from pytest-flask. It starts a live server in the background. We need this so that Selenium has an instance of the app it can hit with real requests:

https://pytest-flask.readthedocs.io/en/latest/features.html#live-server-application-live-server

We set the `_external=True` kwarg in `url_for` inside the acceptance tests so that it constructs a full URI we can give to selenium (like http://localhost:4567/requests).

README.md
dandds: Yep, it's in the BrowserStack account itself. Go to the dashboard, then "Account" > "Settings", then the token is under "Local Testing". The email is just the email address for the BrowserStack account itself. I'll put more of this info in the readme.

dandds:
@patricksmithdds re:

> Do we want to run these tests as part of CI?

I vote no for now. I think that they'll take too long. Once we start adding substantial selenium tests, we'll be looking at tens of minutes for each test suite run. In my experience, acceptance tests like these are not robust and can fail for a bunch of unexpected reasons. I think it will be very frustrating to have to wait 10 or 20 minutes for CI on a small PR and then have the acceptance tests potentially fail because it couldn't connect to BrowserStack or something like that.

I think that we should find a way to run them as a cron job against master a few times a day to catch regressions, and maybe manually run them against major PRs.

tests/acceptance/test_basic.py
dandds: I put this in more as an example of how to get a valid session than as a real test (ditto for whether we should parameterize it with multliple browsers).

patricksmithdds:
@dandds 

> I think that we should find a way to run them as a cron job against master a few times a day to catch regressions, and maybe manually run them against major PRs.

That makes sense. Both Travis and CircleCI support scheduled builds, so I'd say we can add one after we complete the switch over to CirleCI.

PR #359, opened 2018-10-01
Fix report display for task order with no expiration date

richard-dds:
Update the budget report so that it displays a `-` instead of `Invalid Date` when the workspace's task order has no `expiration_date`.

screen shot 2018-10-01 at 2 26 26 pm

### DIFF
--- a/templates/workspaces/reports/index.html
+++ b/templates/workspaces/reports/index.html
@@ -65,16 +65,26 @@
               <div>
                 <dt>Expires</dt>
                 <dd>
-                  <local-datetime
-                    timestamp='{{ expiration_date }}'
-                    format='MMMM D, YYYY'>
-                    </local-datetime>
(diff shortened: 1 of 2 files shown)


templates/workspaces/reports/index.html
```diff
@@ -65,16 +65,26 @@ <h2 class='to-summary__heading'>Task Order</h2>
...
+                      -
+                  {% endif %}
                 </dd>
               </div>
 
               <div>
                 <dt>Remaining</dt>
-                <dd>{{ remaining_days }} days</dd>
+                <dd>
+                  {% if remaining_days %}
```
patricksmithdds: Not sure it's worth worrying about, but if `remaining_days` is 0, we'd show `-` instead of 0 days, so I think this should be `if remaining_days is not none`.

PR #360, opened 2018-10-01
Redirect after login #160851186

dandds:
PT story: https://www.pivotaltracker.com/story/show/160851186

If a user tries to hit a specific route but they are not logged in, we should redirect them to that route after they authenticate.

With help from Leigh!

### DIFF
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -1,3 +1,4 @@
+import urllib.parse as url
 from flask import Blueprint, render_template, g, redirect, session, url_for, request
 
 from flask import current_app as app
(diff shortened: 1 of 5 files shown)


atst/routes/dev.py
```diff
@@ -63,4 +64,4 @@ def login_dev():
     )
     session["user_id"] = user.id
 
-    return redirect(url_for("atst.home"))
+    return redirect(redirect_url())
```
patricksmithdds: This gives me an error because it's attempting to redirect to use `url_for("dev.home")`. I believe the `url_for` in the `redirect_url` helper needs to be the fully qualified version -- `url_for("atst.home")`.

atst/routes/__init__.py
```diff
@@ -70,20 +80,27 @@ def _make_authentication_context():
     )
 
 
+def redirect_url():
```
patricksmithdds: Can we give this a more descriptive name? It's a little too generic for me to know what it's intent is.

atst/routes/__init__.py
```diff
@@ -15,7 +17,15 @@
 
 @bp.route("/")
 def root():
-    return render_template("login.html")
+    redirect_url = app.config.get("CAC_URL")
+    if request.args.get("next"):
+        redirect_url = url.urljoin(
+            redirect_url, "?{}".format(url.urlencode(request.args))
```
patricksmithdds: This is possibly not an issue, but I think we may want to only encode the `next` arg, instead of putting _all_ the request args in `redirect_url`.

dandds:
Fixed, thanks @patricksmithdds

PR #365, opened 2018-10-02
Reduce alerts on financial form

andrewdds:
This cleans up the alerts display on the financial verification form.

- The blue info alert only shows when there are no errors, or the extended form is not showing.
- The red error alert only shows if the missing TO isn't the only error
- The extended TO alert content is modified to reflect the instructions more clearly, rather than just duplicating the missing TO message. Also, helpdocs link is added to this alert.

screen shot 2018-10-02 at 8 40 10 am

screen shot 2018-10-02 at 8 40 17 am

### DIFF
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -156,7 +156,6 @@ class FinancialForm(BaseFinancialForm):
         return "task_order_number" in self.errors and len(self.errors) == 1
 
 
-
 class ExtendedFinancialForm(BaseFinancialForm):
     def validate(self, *args, **kwargs):
         if self.funding_type.data == "OTHER":
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -151,6 +151,11 @@ class FinancialForm(BaseFinancialForm):
     def is_missing_task_order_number(self):
         return "task_order_number" in self.errors
 
+    @property
+    def is_only_missing_task_order_number(self):
+        return "task_order_number" in self.errors and len(self.errors) == 1
+
+
 
 class ExtendedFinancialForm(BaseFinancialForm):
     def validate(self, *args, **kwargs):
--- a/templates/requests/financial_verification.html
+++ b/templates/requests/financial_verification.html
@@ -9,7 +9,7 @@
 
 {% include 'requests/review_menu.html' %}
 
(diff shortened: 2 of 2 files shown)


templates/requests/financial_verification.html
```diff
@@ -21,8 +21,8 @@
 <div class="col">
 
   {% if extended %}
-    {{ Alert('Task Order not found in EDA',
-      message="We could not find your Task Order in the EDA, our system of record. Please confirm that you have entered the information correctly and resubmit or click on the button below to enter the TO manually.",
+    {{ Alert('Manually enter Task Order information',
+      message="Additional field are displayed below, where you can manually enter financial information as documented in your Task Order.",
       level='warning'
```
dandds: This text is much better. I wonder if this should still be a warning, though. What do you think?

andrewdds: Mainly want to keep they yellow to visually tie it to the extra financial fields

atst/forms/financial.py
```diff
@@ -151,6 +151,10 @@ def perform_extra_validation(self, existing_request):
     def is_missing_task_order_number(self):
         return "task_order_number" in self.errors
 
+    @property
+    def is_only_missing_task_order_number(self):
+        return "task_order_number" in self.errors and len(self.errors) == 1
```
andrewdds: Is there a better way to do this?

patricksmithdds: I, too, think its fine. Only other way I can think of is `self.errors.keys() == ['task_order_number']`, but I think your way is easier to read.

templates/requests/financial_verification.html
```diff
@@ -21,8 +21,8 @@
 <div class="col">
 
   {% if extended %}
-    {{ Alert('Task Order not found in EDA',
-      message="We could not find your Task Order in the EDA, our system of record. Please confirm that you have entered the information correctly and resubmit or click on the button below to enter the TO manually.",
+    {{ Alert('Manually enter Task Order information',
+      message="Additional field are displayed below, where you can manually enter financial information as documented in your Task Order.",
```
patricksmithdds: "field" should be pluralized in the start of the message: "Additional fields are displayed below"

andrewdds: 👍

PR #371, opened 2018-10-03
Handle new-project submission with Enter key

andrewdds:
This fixes the way form submission is handled in the new-project form. Rather than handling on-click of the submit button, submission is handled on-submit of the form itself. This allows the "Enter" button (or any other method of submitting the form) to trigger the modal properly.

Also sets a `readyToSubmit` flag to decide whether to prevent the default submission and show the modal, or allow the submission.

Fixes https://www.pivotaltracker.com/story/show/160940008

### DIFF
--- a/js/components/forms/new_project.js
+++ b/js/components/forms/new_project.js
@@ -16,7 +16,8 @@ export default {
     initialData: {
       type: Object,
       default: () => ({})
-    }
+    },
+    modalName: String
   },
 
   data: function () {
@@ -86,10 +87,10 @@ export default {
       return names.every((n, index) => names.indexOf(n) === index)
     },
 
-    handleSubmit: function (modalName, event) {
+    handleSubmit: function (event) {
       if (!this.readyToSubmit) {
         event.preventDefault()
-        this.validateAndOpenModal(modalName)
+        this.validateAndOpenModal(this.modalName)
       }
     },
 
--- a/js/components/forms/new_project.js
+++ b/js/components/forms/new_project.js
@@ -85,6 +85,13 @@ export default {
       return names.every((n, index) => names.indexOf(n) === index)
     },
 
+    handleSubmit: function (modalName, event) {
+      if (!this.readyToSubmit) {
+        event.preventDefault()
+        this.validateAndOpenModal(modalName)
+      }
+    },
+
     validateAndOpenModal: function (modalName) {
       let isValid = this.$children.reduce((previous, newVal) => {
         // display textInput error if it is not valid
--- a/js/components/forms/new_project.js
+++ b/js/components/forms/new_project.js
@@ -40,6 +40,7 @@ export default {
       errors: [],
       environments,
       name,
+      readyToSubmit: false
     }
   },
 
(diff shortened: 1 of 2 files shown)


templates/fragments/edit_project_form.html
```diff
@@ -9,7 +9,7 @@
   {% set action_text = 'Create' if new_project else 'Update' %}
   {% set title_text = 'Add a new project' if new_project else 'Edit {} project'.format(project.name) %}
 
-  <form method="POST" action="{{ form_action }}" >
+  <form method="POST" action="{{ form_action }}" v-on:submit="handleSubmit('{{ modalName }}', $event)">
```
montana-mil: is `modalName` supposed to be here?

montana-mil:
I'm not able to get this to submit at all now. The modal isn't opening. I looked through the code, but I couldn't find anything super obvious :/

andrewdds:
Hmmmm

andrewdds:
@montana-mil I forgot to change the button type back to submit... seems to work now

js/components/forms/new_project.js
```diff
@@ -99,6 +107,7 @@ export default {
       isValid = this.errors.length == 0 && isValid
 
       if (isValid) {
+        this.readyToSubmit = true
```
patricksmithdds: I think we need to set this to false when the form is invalid. Currently you could fill out the form correctly, press enter to open the modal, cancel out of the modal, change the data to an invalid state (add an environment with an empty name), and then pressing enter will submit the invalid form without showing any modal.

Actually, the form submitting without showing a modal at all may indicate there's something else going on here that needs to be looked into.

montana-mil:
I wasn't able to tab into the modal. If I hit Enter, it creates a new project, but since I can't tab in the modal, I couldn't hit Cancel.

js/components/forms/new_project.js
andrewdds: @patricksmithdds I think you're right about reverting `readyToSubmit` to false.

Re: the second part, thats what this PR is meant to address. The problem was the submit handler was only tied to clicking the button. This ties it to the submit event of the form, which should handle all methods of submitting.

andrewdds:
@montana-mil if thats only about tabbing, I'd say thats an unrelated issue regarding the tabbing context of all modal windows, and I think we already have some bugs filed about it.

andrewdds:
@patricksmithdds I added a cancel handler, which makes `readyToSubmit` false every time.
Also, made the `modalName` a prop of the component, theres no reason to pass it into each function.

js/components/forms/new_project.js
```diff
@@ -85,6 +87,18 @@ export default {
...
+        this.validateAndOpenModal(this.modalName)
+      }
+    },
+
+    handleCancelSubmit: function () {
+      this.readyToSubmit = false
+      this.closeModal(this.modalName)
+    },
+
     validateAndOpenModal: function (modalName) {
```
patricksmithdds: `modalName` is still passed along here. Can we use `this.modalName` in this method instead?

andrewdds: ah missed that one, thanks

PR #372, opened 2018-10-03
Workspace users project permissions

montana-mil:
## Description
Users with no environment access on a project should not be able to see that project at all. The user is not allowed to see any environment that they do not have access to either. If a user is a CCPO, they can see all projects and all environments, regardless of membership in the environments.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/160939536

## Screenshots
screen shot 2018-10-03 at 1 59 16 pm
screen shot 2018-10-03 at 1 58 58 pm


montana-mil:
I think we may need to look at permissions again. This one seems more appropriate, but all users have it: `VIEW_APPLICATION_IN_WORKSPACE`

PR #374, opened 2018-10-03
financial verification routes refactor

dandds:
This refactors the logic of the financial verification routes into a service object, like we do for the request form and the requests index page. I think this will make incremental saving and more nuanced PDF handling, two upcoming PT stories, easier. It's basically done, it just needs some more unit tests for the service object class.

If @patricksmithdds @richard-dds or @montana-mil want it while I'm gone, be my guest.

### DIFF
--- a/atst/routes/requests/financial_verification.py
+++ b/atst/routes/requests/financial_verification.py
@@ -7,8 +7,8 @@ from atst.forms.financial import FinancialForm, ExtendedFinancialForm
 
 
 class FinancialVerification:
-    def __init__(self, user, request_id, extended=False, post_data=None):
-        self.request = Requests.get(user, request_id)
+    def __init__(self, request, extended=False, post_data=None):
+        self.request = request
         self._extended = extended
         self._post_data = post_data
         self._form = None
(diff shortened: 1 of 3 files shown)


dandds:
Thanks for the tests @montana-mil . Putting this up for review now. 

Did this with some help from @leigh-mil .

atst/routes/requests/financial_verification.py
```diff
@@ -6,80 +6,132 @@
 from atst.forms.financial import FinancialForm, ExtendedFinancialForm
 
 
-def task_order_data(task_order):
-    data = task_order.to_dictionary()
-    data["task_order_number"] = task_order.number
-    data["funding_type"] = task_order.funding_type.value
-    return data
+class FinancialVerification:
+    def __init__(self, user, request_id, extended=False, post_data=None):
```
richard-dds: This may be a nitpick, but would it make sense to just pass in a request and do the `Requests.get` outside of this class? That way we don't need to monkeypatch `Requests.get` to test this.

dandds: Yeah, that's better. It shouldn't need the user, either. Fixed on da6c971.

PR #376, opened 2018-10-05
Standardize auth patterns

richard-dds:
Removed domain-specific auth queries from `domain.authz`. More complex auth checks should live inside the relevant domain.


tests/domain/test_authz.py
```diff
@@ -1,20 +0,0 @@
...
-
-    other_user = UserFactory.create()
-    assert not Authorization.can_view_request(other_user, request)
-
-
-def test_ccpo_user_can_view_request():
-    role = Roles.get("ccpo")
-    ccpo_user = UserFactory.create(atat_role=role)
-    request = RequestFactory.create()
-    assert Authorization.can_view_request(ccpo_user, request)
```
patricksmithdds: Can we keep these tests around with the new `RequestsAuthorization` class?

tests/domain/test_requests.py
```diff
@@ -233,3 +234,31 @@ def test_add_internal_comment():
...
+    request = RequestFactory.create(creator=creator)
+
+    assert RequestsAuthorization(creator, request).can_view
+
+
+def test_creator_can_view_own_request():
+    creator = UserFactory.create()
+    request = RequestFactory.create(creator=creator)
+
+    assert RequestsAuthorization(creator, request).can_view
```
patricksmithdds: Looks like this is a duplicate of the test above it.

richard-dds: Whoops! Thanks

PR #377, opened 2018-10-08
Handle AlreadyExistsError on user creation with non-unique email

montana-mil:
## Description
If a user attempted to create a new member with an email that already exists in the system, a `500` server error would appear. This PR will now show the user a friendlier error message.

It also refactors how we show errors and adds a test.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/160836012

## Screenshot
screen shot 2018-10-08 at 10 37 13 am


templates/error.html
```diff
@@ -4,9 +4,8 @@
 
 <main class="usa-section usa-content">
 
-<h1>Not Found</h1>
+<h1>{{ message }}</h1>
```
dandds: Should we supply a default here, just in case? I don't know what it would be. Maybe just "An error occurred"?

montana-mil: good idea

PR #379, opened 2018-10-08
Root url should redirect if user is already logged in

richard-dds:
https://www.pivotaltracker.com/story/show/161024051


atst/routes/__init__.py
```diff
@@ -25,6 +25,9 @@ def root():
             "?{}".format(url.urlencode({"next": request.args.get("next")})),
         )
 
+    if g.current_user:
```
dandds: Could this be the very first thing in the route function?

PR #383, opened 2018-10-11
Add UAT env and move to *.atat.code.mil (from *.atat.codes)

patricksmithdds:
This is all kubernetes config changes and has already been applied to the cluster. This PR accomplishes two things:

* adds a second namespace for a UAT environment (uat.atat.code.mil)
* moves www.atat.codes and auth.atat.codes to www.atat.code.mil and auth.atat.code.mil

Some notes:
* the UAT configuration contains a lot of duplication from the staging environment. In the future, it would be great to use some tool that allows tempated kubernetes configs so we can reuse the same template for both
* the UAT env re-uses the same postgres/redis instances that staging uses to reduce the amount of resources necessary
* these changes have already been applied to the cluster, so these are mostly artifacts of what exists
* one pain point was getting the correct certs for the domains using `kube-lego`. The load balancers were redirecting http:// requests to https:// which would raise an error due to invalid certs (which we were in the process of getting). I temporarily switched the LBs to http instead of https, let `kube-lego` get the certs, and then switched them back to https. Additionally, the auth services needed ingresses to be added temporarily in order for `kube-lego` to know it needed to fetch certs from them as well.

### DIFF
--- a/deploy/kubernetes/uat/uat.yml
+++ b/deploy/kubernetes/uat/uat.yml
@@ -168,11 +168,11 @@ metadata:
     nginx.ingress.kubernetes.io/proxy-body-size: 10m
 spec:
   tls:
-  - hosts:
-    - uat.atat.codes
-    secretName: atst-uat-ingress-tls
+  - secretName: atst-uat-ingress-tls
+    hosts:
+    - uat.atat.code.mil
   rules:
-  - host: uat.atat.codes
+  - host: uat.atat.code.mil
     http:
       paths:
       - path: /
--- a/deploy/kubernetes/uat/uat.yml
+++ b/deploy/kubernetes/uat/uat.yml
@@ -24,7 +24,7 @@ spec:
         fsGroup: 101
       containers:
         - name: atst
-          image: registry.atat.codes:443/atst-prod:d8d05da
+          image: registry.atat.codes:443/atst-prod:76854ac
           resources:
             requests:
                memory: "2500Mi"
--- a/deploy/kubernetes/uat/uat.yml
+++ b/deploy/kubernetes/uat/uat.yml
@@ -79,7 +79,7 @@ spec:
               mode: 0644
         - name: nginx-auth-tls
           secret:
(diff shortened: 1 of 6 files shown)


deploy/kubernetes/uat/uat.yml
```diff
@@ -0,0 +1,181 @@
...
+              mountPath: "/etc/nginx/.htpasswd"
+              subPath: .htpasswd
+            - name: uwsgi-socket-dir
+              mountPath: "/var/run/uwsgi"
+      imagePullSecrets:
+        - name: regcred
+      volumes:
+        - name: atst-config
+          secret:
+            secretName: atst-config-ini
```
ddsdevon: This config contains things like the cookie secret, session secrets, etc., which are items that should be unique per environment.
It does also contain some shared items, but it is probably better to duplicate those in a separate INI entity anyway, so they can easily diverge later if they need to (also, now that the app checks ENV vars for overrides, a lot of those items can be moved into the atst-envars object instead, which is easier to parse and update; there should probably be one of those per environment too).

ddsdevon:
Yeah, the SSL/TLS pain points are annoying and will hopefully go away with a more complete and integrated cloud provider. That whole part feels a bit hacked together because... it is, due to Kubernetes not being able to manage the cloud provider load balancer, which is the more "normal" way this all works.

deploy/kubernetes/uat/uat.yml
patricksmithdds: I took the config file from the staging environment and changed all the secrets to something unique before creating the secret for the UAT env (also needed to update the database name so that we can use the same postgres instance safely). 

I agree that splitting out the things that are likely to change to a separate INI/ENV vars would be good to do in the future.

ddsdevon: Oh right! You made a new namespace. That was the part I forgot... thought they were sharing the same `atst-config-ini` secret which is where my concern came from.

PR #384, opened 2018-10-12
Mailer

dandds:
Covers this PT story: https://www.pivotaltracker.com/story/show/160803625

This adds the ability for the app to send notification emails as background jobs. I made a bunch of opinionated changes on this PR, so discussion is welcome.

**new dependencies**
- [flask-rq2](https://flask-rq2.readthedocs.io/en/latest/): I chose this over the regular `flask-rq` package because that one doesn't seem to be maintained anymore and the repo for flask-rq2 is in the official RQ Github group.
- [honcho](https://github.com/nickstenning/honcho): Honcho is a Python port of the ruby tool Foreman, which I'm used to using to manage multiple development processes. In this case, I thought is would be useful to start `yarn watch`, the queue worker process, and the main app process under a single manager for development. You can see these listed in the `Procfile`. `script/server` invokes Honcho, or you can also just run `pipenv run honcho start`.

**Dev endpoints for email**
In test and development mode, the app stores sent emails in Redis instead of really sending them.

- `/test-email`: In development mode, you can hit this endpoint with URL query params for `to`, `subject`, and `body`. It redirects to the development inbox, `/messages`. Since the email is "sent" in a background job, you'll probably have to refresh the page to see it. 
- `/messages`: This is the fake inbox for dev mode. It will list every email sent while the app is running. It gets cleared every time you restart the app.

**try out real emails**
I can demo this for you if you want. If not, the easiest thing to do is to make a throwaway gmail account with _only_ regular password auth enabled. Once you do that, remove the DEBUG config setting from `base.ini` and add these params:

```
MAIL_SERVER = smtp.gmail.com
MAIL_PORT = 587
MAIL_SENDER = <your email account>
MAIL_PASSWORD = <your email password>
MAIL_TLS = true
```

**notes**
- I don't feel great about the initialization pattern for RQ, but it needs to be globally available like our `db` object.
- @patricksmithdds The deployment doesn't rely on `script/server` at all, does it? If not, I think we should delete `dev_server` and just use `server` (or the plain honcho command) for local dev.
- I'm not catching or re-raising any mail-specific errors at this point. I think that the email errors from the Python standard library are pretty clear for debugging purposes, and we wouldn't have to show any kind of email error to a user synchronously since they happen in a background job. 
- I haven't put much thought into retries, timeouts, etc., for jobs. Open to adding that if you feel it should be on this PR, otherwise we can make a chore for it.


script/server
```diff
@@ -4,8 +4,5 @@
 
 source "$(dirname "${0}")"/../script/include/global_header.inc.sh
 
-# Compile js/css assets
-yarn build
-
 # Launch the app
-run_command "./app.py ${LAUNCH_ARGS}"
+run_command "honcho start"
```
patricksmithdds: I like using honcho here 💯

atst/queue.py
```diff
@@ -0,0 +1,9 @@
+from flask_rq2 import RQ
+from flask import current_app as app
+
+queue = RQ()
+
+
+@queue.job
+def send_mail(to, subject, body):
+    app.mailer.send(to, subject, body)
```
patricksmithdds: I think it is still important to be able to specify a queue for the app/worker to use. In our current setup, for example, both the staging env and UAT env use the same Redis instance, which I believe will cause issues if they both use the same queue name.

After a bit of digging, it looks like an issue is that the `queue.job` decorator requires you to explicitly set the queue name (`@queue.job('atat_dev')`), otherwise it'll just use the "default" queue. We can work around this in a number of ways and this was my attempt at it:

```diff
diff --git a/atst/queue.py b/atst/queue.py
index 8c94bf8f..0b92bf91 100644
--- a/atst/queue.py
+++ b/atst/queue.py
@@ -1,9 +1,43 @@
 from flask_rq2 import RQ
 from flask import current_app as app
 
-queue = RQ()
 
+class ATSTQueue(RQ):
 
-@queue.job
-def send_mail(to, subject, body):
-    app.mailer.send(to, subject, body)
+    """Internal helpers to get the queue that actually does the work.
+
+    The RQ object always uses the "default" queue, unless we explicitly request
+    otherwise. These helpers allow us to use `.queue_name` to get the name of
+    the configured queue and `_queue_job` will use the appropriate queue.
+
+    """
+
+    @property
+    def queue_name(self):
+        return self.queues[0]
+
+    def get_queue(self, name=None):
+        if not name:
+            name = self.queue_name
+        return super().get_queue(name)
+
+    def _queue_job(self, function, *args, **kwargs):
+        self.get_queue().enqueue(function, *args, **kwargs)
+
+    """Instance methods to queue up application-specific jobs."""
+
+    def send_mail(self, to, subject, body):
+        self._queue_job(ATSTQueue._send_mail, to, subject, body)
+
+    """Class methods to actually perform the work.
+
+    Must be a class method (or a module-level function) because we being able
+    to pickle the class is more effort than its worth.
+    """
+
+    @classmethod
+    def _send_mail(self, to, subject, body):
+        app.mailer.send(to, subject, body)
+
+
+queue = ATSTQueue()
diff --git a/atst/routes/dev.py b/atst/routes/dev.py
index 862d602e..0d90397f 100644
--- a/atst/routes/dev.py
+++ b/atst/routes/dev.py
@@ -10,7 +10,7 @@ from flask import (
 
 from . import redirect_after_login_url
 from atst.domain.users import Users
-from atst.queue import send_mail
+from atst.queue import queue
 
 bp = Blueprint("dev", __name__)
 
@@ -78,7 +78,7 @@ def login_dev():
 
 @bp.route("/test-email")
 def test_email():
-    send_mail.queue(
+    queue.send_mail(
         [request.args.get("to")], request.args.get("subject"), request.args.get("body")
     )
 ```

I'm curious what you think of that and if its a step in the right direction or not?

File: Procfile [hunk 3]
```diff
@@ -0,0 +1,3 @@
+assets: yarn watch
+web: PORT=8000 python app.py
+queue: flask rq worker
```
patricksmithdds: Any thoughts on how we would run this on the deployed cluster? The app is currently run with the `uwsgi_server` script and I'm not sure that would be able to run the `flask rq worker` in addition to the app. 

I think we'd need to run it as a separate deployment or another container in the same deployment. In that case, I'm curious if we need to worry about migrations running from the `uwsgi_server` finishing before we can start the worker? Maybe we can rely on retrying the jobs until the migrations are done?

dandds: I think ideally we'd have another container in the same deployment that is exactly the same as the one that serves HTTP requests, but has a different final CMD in the Dockerfile. That way we can tell it to run just the `flask rq worker` command, either from a top-level script or directly. Does that make sense/seem possible?

atst/queue.py
dandds: Yeah, that makes sense. I like encapsulating all the queue behavior in a top-level object so that we can manage queue name directly. I'll apply that change and add the environment-specific naming back.

dandds: Okay, I borrowed your commit and restored the environment-based queue names.

atst/utils/mailer.py
```diff
@@ -0,0 +1,82 @@
...
+    def send(self, recipients, subject, body):
+        pass
+
+    # do not collect messages by default
+    @property
+    def messages(self):
+        return []
+
+
+class Mailer(BaseMailer):
```
richard-dds: I feel like this might be a good place to favor composition over inheritance, by having two implementations of a Connection interface that a single Mailer class accepts as a dependency.

atst/queue.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_queue(self, name=None):
+        if not name:
+            name = self.queue_name
+        return super().get_queue(name)
+
+    def _queue_job(self, function, *args, **kwargs):
+        self.get_queue().enqueue(function, *args, **kwargs)
+
+    # pylint: disable=pointless-string-statement
```
patricksmithdds: 😆 "pointless statement" -- such judgement from a pylint comment!

atst/utils/mailer.py
```diff
@@ -0,0 +1,84 @@
...
+    @contextmanager
+    def _host(self):
+        host = None
+
+        if self.use_tls:
+            host = smtplib.SMTP(self.server, self.port)
+            host.starttls()
+        else:
+            host = smtplib.SMTP_SSL(self.server, self.port)
+        host.login(self.username, self.password)
```
patricksmithdds: Should this be part of the `else` branch? If not, I think a blank line above it will make this code more readable to me.

atst/utils/mailer.py
```diff
@@ -0,0 +1,84 @@
...
+class SMTPConnection(MailConnection):
+    def __init__(self, server, port, username, password, use_tls=False):
+        self.server = server
+        self.port = port
+        self.username = username
+        self.password = password
+        self.use_tls = use_tls
+
+    @contextmanager
+    def _host(self):
```
patricksmithdds: Tiny request -- can we name these (`_host` and `_message` below) with some verb to clarify what they're doing? Otherwise `message = self._message(...)` is a bit confusing to me.

dandds:
@patricksmithdds updates in c81eab0. thanks!

PR #386, opened 2018-10-16
Edit user #160268937

dandds:
PT story: https://www.pivotaltracker.com/story/show/160268937

I have some open questions about the feature, but I think they could be handled on a subsequent PR. (https://www.pivotaltracker.com/story/show/160268937/comments/195330940)

Users can click their profile name, which takes them to a form where they can update their profile information. Any attribute can be updated except the DOD ID. Validations are the same as the equivalent fields within the request.

screenshot_2018-10-16 jedi cloud


dandds:
Made some adjustment based on discussion: https://www.pivotaltracker.com/story/show/160268937/comments/195380845

atst/forms/edit_user.py
```diff
@@ -52,19 +51,17 @@ class EditUserForm(ValidatedForm):
...
         "Latest Information Assurance (IA) Training Completion Date",
         description='To complete the training, you can find it in <a class="icon-link" href="https://iatraining.disa.mil/eta/disa_cac2018/launchPage.htm" target="_blank">Information Assurance Cyber Awareness Challange</a> website.',
         validators=[
-            Required(),
             DateRange(
                 lower_bound=pendulum.duration(years=1),
                 upper_bound=pendulum.duration(days=0),
                 message="Must be a date within the last year.",
-            ),
+            )
```
patricksmithdds: If this field isn't provided, a type error is raised when attempting to check if None is within the date range.

templates/user/edit.html
```diff
@@ -3,21 +3,22 @@
...
 {% block content %}
 <div class='col'>
-  {{ Alert('This form does not yet function',
-    message="<p>Functionality of this form is pending more engineering work. Engineers, please remove this alert when done.</p>",
-    level='warning'
-  ) }}
+
+  {% if updated %}
+    {{ Alert('User information updated.', level='success') }}
+  {% endif %}
```
patricksmithdds: I think we should also show an error alert if there were errors saving the form. Otherwise, the user may not notice that they need to scroll down to see an error (if, for example, only the IA training date was wrong).

atst/forms/edit_user.py
```diff
@@ -25,7 +25,7 @@ class EditUserForm(ValidatedForm):
     phone_number = TelField(
```
patricksmithdds: Is it possible to reduce the duplication between this form and the "information about you" form? Or, maybe that'll cause issues in the future if the form contents diverge?

templates/user/edit.html
```diff
@@ -3,21 +3,22 @@
...
       <h1>
         <div class='h2'>{{ user.first_name }} {{ user.last_name }}</div>
+        <div class='h3'>DOD ID: {{ user.dod_id }}</div>
         <div class='subtitle h3'>Edit user details</div>
       </h1>
     </div>
   </div>
 
-  {% set form_action = url_for('atst.save_user') %}
+  {% set form_action = url_for('users.user') %}
```
patricksmithdds: Probably doesn't make any difference, but it may be clearer to use `users.update_user` here.
```suggestion
  {% set form_action = url_for('users.update_user') %}
```

atst/models/user.py
```diff
@@ -52,3 +57,10 @@ def __repr__(self):
             self.has_workspaces,
             self.id,
         )
+
+    def to_dictionary(self):
```
richard-dds: Would this make sense as a mixin? I see that `TaskOrder` has a `to_dictionary` method, but it also drops the `attachment_id`, so I'm not sure if it's worth it or not.

dandds: I thought about it. I'd say if we find one more use-case, we should make a mixin and find a way to declare the "drop" list per model.

atst/forms/edit_user.py
dandds: I tried this out on  dab5fe8. Let me know how you feel about it.

atst/models/user.py
dandds: Incidentally, I suspect not having an intuitive way to declare what shouldn't be serialized by default is why SQLAlchemy doesn't provide this out-of-the-box.

atst/forms/edit_user.py
dandds: Hmm, that's not actually working how I thought. I'm going to keep tinkering with it.

atst/forms/new_request.py
```diff
@@ -162,58 +163,21 @@ def validate(self, *args, **kwargs):
 
 
 class InformationAboutYouForm(ValidatedForm):
-    fname_request = StringField("First Name", validators=[InputRequired(), Alphabet()])
+    fname_request = inherit_field(USER_FIELDS["first_name"])
 
-    lname_request = StringField("Last Name", validators=[InputRequired(), Alphabet()])
+    lname_request = inherit_field(USER_FIELDS["last_name"])
 
     email_request = EmailField("E-mail Address", validators=[InputRequired(), Email()])
```
patricksmithdds: Any reason not to use `inherit_field(USER_FIELDS["email"])` here?

atst/forms/edit_user.py
```diff
@@ -1,70 +1,86 @@
...
+
+    return unbound_field.field_class(*unbound_field.args, **kwargs)
+
+
+class EditUserForm(ValidatedForm):
+    first_name = inherit_field(USER_FIELDS["first_name"])
+    last_name = inherit_field(USER_FIELDS["last_name"])
+    email = inherit_field(USER_FIELDS["email"])
+    phone_number = inherit_field(USER_FIELDS["phone_number"], required=False)
+    service_branch = inherit_field(USER_FIELDS["service_branch"], required=False)
```
patricksmithdds: I believe this field is supposed to be optional, but not selecting a value gives a form error and prevents updating just the name/email/phone number of the user:

image

tests/routes/test_users.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_user_can_update_profile(user_session, client):
+    user = UserFactory.create()
+    user_session(user)
+    new_data = {**user.to_dictionary(), "first_name": "chad", "last_name": "vader"}
+    new_data["date_latest_training"] = new_data["date_latest_training"].strftime(
+        "%m/%d/%Y"
+    )
+    client.post(url_for("users.user"), data=new_data)
```
patricksmithdds: Minor comment: this is still using `url_for("users.user")`, which confused me for a second. It'll still work because the URL for `users.user` and `users.update_user` are the same, but maybe we can change it for readability.

atst/forms/new_request.py
dandds: Yes, that one does not have a description field. Maybe it should, but I didn't want to change page content.

tests/routes/test_users.py
dandds: whoops, will fix

atst/forms/edit_user.py
dandds: I thought this was working yesterday. :/ That field definitely has the optional validator attached. Frustrating, wtforms.

dandds: It looks like this is because the data had `None` as the value for the "select an option" option. Other, similar fields use an empty string instead of `None`.

PR #387, opened 2018-10-16
Save Financial Verification Draft

richard-dds:
https://www.pivotaltracker.com/story/show/159279565

Allow users to save a draft of the financial verification form to come back to later.

Involved a pretty heavy refactor of the finver routes and form validations. My strategy was to unroll most of the code back into the routes, then factoring out a class per route with a single method that served as a good testing seam. I rebuilt the old http tests using this seam, added some for the draft feature, implemented the draft feature, and then factored out some of the common elements.

In the end, I think we have more duplication, but I think the temporary duplication is easier to deal with while we look for a better way to express these complex forms.

I'm hoping there are no regressions, but since I had to move the tests around a bit there are no guarantees. I've been testing this manually but I could use a little help if someone has time to spare.

screen shot 2018-10-16 at 11 18 38 am


tests/forms/test_financial.py
```diff
@@ -1,22 +1,22 @@
...
+#     [
+#         ("0603502N", None),
+#         ("0603502NZ", None),
+#         ("603502N", "0603502N"),
+#         ("063502N", "0603502N"),
+#         ("63502N", "0603502N"),
+#     ],
+# )
+# def test_suggest_pe_id(input_, expected):
+#     assert suggest_pe_id(input_) == expected
```
patricksmithdds: Can we bring back these tests?

atst/routes/requests/financial_verification.py
```diff
@@ -1,134 +1,246 @@
...
+                submitted = True
+        else:
+            self._raise(form)
+
+        if submitted:
+            return {
+                "state": "submitted",
+                "workspace": workspace,
+                "request": updated_request,
+            }
```
patricksmithdds: I like this refactor, but I think this "context" feels a bit awkward to me -- that each class returns an arbitrary object possibly links too much of how the route uses it to its implementation. I'm curious if it makes sense to just always return the updated_request and keep the workspace creation in the route?

richard-dds: Yeah, I see what you mean. I'll try to figure out something more general.

richard-dds:
@patricksmithdds Just noticed that trying to save a draft with a PDF results in a 500. Fixing.

patricksmithdds:
@richard-dds I'm still getting an error when I try to save a draft of the extended financial verification form.

atst/routes/requests/financial_verification.py
```diff
@@ -1,134 +1,278 @@
...
+        return render_template(
+            "requests/financial_verification.html",
+            jedi_request=request,
+            f=e.form,
+            extended=is_extended,
+        )
 
-    if finver.workspace:
+    if updated_request.is_pending_ccpo_approval:
+        workspace = Requests.approve_and_create_workspace(updated_request)
```
patricksmithdds: I don't think this conditional is correct -- if the updated request is pending approval we _do not_ want to automatically approve and create the workspace. I just tested this and after manually entering the task order information, the workspace was created instead of moving to the approval step. This is probably something we should add a test for as well.

atst/routes/requests/financial_verification.py
```diff
@@ -1,134 +1,278 @@
...
+        return render_template(
+            "requests/financial_verification.html",
+            jedi_request=request,
+            f=e.form,
+            extended=is_extended,
+        )
 
-    if finver.workspace:
+    if updated_request.is_approved:
+        workspace = Requests.approve_and_create_workspace(updated_request)
```
patricksmithdds: Sorry, but I think this went too far in the other direction -- now, submitting financial verification with the found TO (which should automatically create a workspace) requires CCPO approval.

richard-dds: Yeah, that makes sense. I don't know what I was thinking with that conditional. `If the request is approved, then approve it.`

PR #389, opened 2018-10-16
Workspaces dropdown

andrewdds:
This adds a dropdown panel menu for workspace switching.

screen shot 2018-10-16 at 11 43 21 am
screen shot 2018-10-16 at 11 43 36 am

### DIFF
--- a/atst/routes/workspaces.py
+++ b/atst/routes/workspaces.py
@@ -48,7 +48,12 @@ def workspace():
             )
         return False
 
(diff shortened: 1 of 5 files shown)


andrewdds:
@patricksmithdds I'm not sure how that would even work. Click handler on... everything?

patricksmithdds:
> I'm not sure how that would even work. Click handler on... everything?

I'm not entirely sure, but I've seen it done with modals before, I think using some sort of backdrop element, which may not be applicable here.

PR #394, opened 2018-10-17
Revoke members' environment access

richard-dds:
https://www.pivotaltracker.com/story/show/158072194

Clicking "revoke all access" will now remove a user from all environments in that project once that user presses "save". As a side-effect, there's now a "No Access" option in the environment role choices.

Could probably use a styling / copy sweep.

screen shot 2018-10-17 at 4 07 47 pm


js/components/forms/edit_environment_role.js
```diff
@@ -30,25 +32,30 @@ export default {
   methods: {
     change: function (e) {
       e.preventDefault()
-      this.new_role = e.target.value
+      this.new_role = this.revoke ? "" : e.target.value
     },
     cancel: function () {
       this.new_role = this.initialData
     },
+    actualRole: function () {
```
patricksmithdds: Can this be removed?

js/components/forms/edit_environment_role.js
```diff
@@ -30,25 +32,30 @@ export default {
   methods: {
     change: function (e) {
       e.preventDefault()
-      this.new_role = e.target.value
+      this.new_role = this.revoke ? "" : e.target.value
```
patricksmithdds: I think we should use the `newRole` helper method here.

js/components/forms/edit_environment_role.js
```diff
@@ -30,25 +32,30 @@ export default {
...
-        if (this.choices[arr][0] == this.new_role) {
+        if (this.choices[arr][0] == newRole) {
           return this.choices[arr][1].name
         }
       }
-      return this.new_role ? this.new_role : "no access"
     },
     label_class: function () {
-      return this.displayName === "no access" ?
+      return this.displayName === "No Access" ?
```
patricksmithdds: Can this use the value instead -- if the value is an empty string, then there's no access. That will decouple this conditional from the display string.

templates/workspaces/members/edit.html
```diff
@@ -59,15 +61,27 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
...
+        {% endcall %}
       </header>
       <ul v-show='props.isVisible'>
         {% for env in project.environments %}
 
           {% set role = EnvironmentRoles.get(member.user_id, env.id).role %}
 
         <li class='block-list__item'>
-          <edit-workspace-member inline-template initial-data='{{ role }}' v-bind:choices='{{ choices | tojson }}'>
+          <edit-environment-role inline-template initial-data='{{ role or "" }}' v-bind:choices='{{ choices | tojson }}' v-bind:revoke='revoke'>
```
patricksmithdds: I believe passing in `revoke` like this makes it impossible to add a role after revoking. I could imagine a scenario where the user clicks "revoke" to revoke all access and then adds back one or two environment roles, rather than setting "no access" for the environments to remove.

templates/workspaces/members/edit.html
```diff
@@ -51,6 +51,8 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
   </div>
 
   {% for project in projects %}
+  {% set revoke_modal_name = project.name + 'RevokeModal' %}
+  <edit-project-roles inline-template v-bind:name="'{{ project.name }}'">
   <div is='toggler' default-visible class='block-list project-list-item'>
```
patricksmithdds: We don't enforce the linting on these files, but can you indent the children of `edit-project-roles`?

templates/workspaces/members/edit.html
```diff
@@ -59,15 +61,27 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
           <template v-else>{{ Icon('caret_right') }}</template>
           <h3 class="block-list__title">{{ project.name }}</h3>
         </button>
-        <span><a href="#" class="icon-link icon-link--danger">revoke all access</a></span>
+        <span><a v-on:click="openModal('{{ revoke_modal_name }}')" class="icon-link icon-link--danger">revoke all access</a></span>
+        {% call Modal(name=revoke_modal_name, dismissable=False) %}
```
patricksmithdds: Not sure if we can pull this modal into a different place in the file, but it throws off the alignment of the "revoke all access" button when the modal is open:

image

dandds:
Based on the way the story is written, it seems like clicking "confirm" should actually perform the update and return the workspace admin to the members list view with a confirmation modal. I think that what happens now (I click confirm, and then I also have to click "save") is a little confusing. Maybe I missed some discussion around this, though.

js/components/forms/edit_project_roles.js
```diff
@@ -0,0 +1,32 @@
...
+  data: function() {
+    return { revoke: false }
+  },
+
+  methods: {
+    doRevoke: function () {
+      // This is being used to send an event to the edit-environment-role child component.
+      // We'll toggle this back on the next line so that it can be used again.
+      this.revoke = true
+      setTimeout(() => { this.revoke = false }, 25)
```
patricksmithdds: I think we can accomplish this with events and without the need to use `setTimeout` here. We could have a separate event bus for each project, but for now we can use the global event bus. I took a pass at this here: https://github.com/dod-ccpo/atst/commit/d350cf98ce6798f6a9a9562af1e3c0835641578b

richard-dds:
@dandds You're right, it does seem to say that pretty clearly in the PT story. @patricksmithdds How do you think we should proceed?

js/components/forms/edit_project_roles.js
richard-dds: @patricksmithdds Awesome, thanks! I'll give that a shot and probably merge it in.

PR #395, opened 2018-10-17
Dropdown selector keyboard accessibility

andrewdds:
This fixes the custom dropdown Selector component to allow keyboard accessibility. It mimics (as closely as possible) the native keyboard behavior of a normal HTML select.

See: https://www.pivotaltracker.com/story/show/160353685

When focused, the down-arrow key expands the popup. The up/down-arrow keys allow you to select through the different options. The enter key then closes the popup. The behavior is subtly different when using a mouse: clicking the radio option will automatically close the popup (after a small delay).

Some learnings as a result of trying to achieve this behavior:

As it turns out, when the keyboard is used to select a radio button, the browser considers this a click, and a Mouse Event is triggered. Therefore its very difficult to know the difference between a mouse selecting a radio button, and a keyboard doing so.

I attempted various things to achieve this, but what ended up working best was a simple `usingKeyboard` flag that is set when the down-arrow is initially used to open the selector popup. This prevents the popup from closing until you manually do so with the enter key. There is a very narrow use case where someone opens the selector with a mouse, but chooses the radio with the keyboard, in which case it will appear to close as soon as you change it once. I couldn't figure any way to avoid this.

Also, one attempt included refactoring the options as a separate Vue component. I thought this might make it possible to listen to events differently. It didn't, and this refactoring was probably unnecessary. I didn't bother un-refactoring because: why waste time, it doesn't really hurt anything, and it ended up being a good lesson in Vue architecture.

### DIFF
--- a/js/components/selector.js
+++ b/js/components/selector.js
@@ -1,10 +1,41 @@
 import { VPopover } from 'v-tooltip'
 
+const SelectorInput = {
+  name: 'SelectorInput',
+  props: {
+    name: String,
+    value: String,
+    label: String,
+    description: String,
+    selected: Boolean,
+    handleChange: Function,
+    handleEnter: Function
+  },
+
+  computed: {
+    id: function () {
+      return `${this.name}_${this.value}`
+    }
+  },
+
+  methods: {
+    onChange: function (e) {
+      this.handleChange(this.value)
+    },
+
+    onEnter: function (e) {
+      this.handleEnter()
+    }
+  }
+}
+
+
 export default {
   name: 'selector',
 
   components: {
-    VPopover
+    VPopover,
+    SelectorInput
   },
 
   props: {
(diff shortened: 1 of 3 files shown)


patricksmithdds:
@andrewdds Should also handle the up arrow to open the selection dropdown? The icon on the selection box seems to indicate either would work:

image

Also, after opening the dropdown, you must press the down arrow before changing any selection. For example, if this is the opened dropdown:

image

Pressing the up arrow has no effect. Once I hit the down arrow, though, I can properly use the up arrow to make a different selection.

Curious if it's also a good idea to allow using the escape key to close the dropdown (and discard any change)? Might be out of scope for this branch, though.

andrewdds:
@patricksmithdds good calls on the up arrow and esc keys.
I've added these behaviors.

Also fixed the focusing issue where you needed to press down arrow again to focus on the radio buttons. It will now focus on the selected one automatically.

After talking with @luisgov we opted to remove the auto-closing behavior altogether. So in all cases, you have to manually close the popup, either by clicking out of it anywhere, or hitting enter, or esc. Its not ideal, but at least its a uniform behavior that doesn't break when you try to combine keyboard and mouse controls.

PR #399, opened 2018-10-18
Update Alphabet validator to accept hyphens

leigh-mil:
## Description
This PR makes it possible for users to enter first and last names that include special characters!

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/161325890](https://www.pivotaltracker.com/story/show/161325890)

## Screenshots
screen shot 2018-10-18 at 3 50 36 pm

### DIFF
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -50,13 +50,13 @@ def PhoneNumber(message="Please enter a valid 5 or 10 digit phone number."):
     return _is_phone_number
 
 
-def Alphabet(message="Please enter only letters."):
-    def _alphabet(form, field):
+def Name(message="Please enter only letters."):
+    def _name(form, field):
         match = re.match(r"[\w \,\.\'\-]+", field.data)
         if not match or match.group() != field.data:
             raise ValidationError(message)
 
-    return _alphabet
+    return _name
 
 
 def ListItemRequired(message="Please provide at least one.", empty_values=("", None)):
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -52,7 +52,7 @@ def PhoneNumber(message="Please enter a valid 5 or 10 digit phone number."):
 
 def Alphabet(message="Please enter only letters."):
     def _alphabet(form, field):
-        match = re.match(r"[A-Za-z\-\']+", field.data)
+        match = re.match(r"[\w \,\.\'\-]+", field.data)
         if not match or match.group() != field.data:
             raise ValidationError(message)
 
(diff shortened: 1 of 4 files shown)


leigh-mil:
Yes! I will update to include them too.

tests/forms/test_validators.py
```diff
@@ -39,13 +39,13 @@ def test_PhoneNumber_rejects_invalid_numbers(
...
-    @pytest.mark.parametrize("valid", ["a", "abcde"])
+    @pytest.mark.parametrize("valid", ["a", "abcde", "hi mark", "cloud9", "niña"])
     def test_Alphabet_accepts_letters(self, valid, dummy_form, dummy_field):
         validator = Alphabet()
         dummy_field.data = valid
         validator(dummy_form, dummy_field)
 
-    @pytest.mark.parametrize("invalid", ["", "hi mark", "cloud9"])
+    @pytest.mark.parametrize("invalid", [""])
     def test_Alphabet_rejects_non_letters(self, invalid, dummy_form, dummy_field):
```
montana-mil: it might be worth it to rename this test to something like `test_Alphabet_rejects_empty_string`

atst/forms/validators.py
```diff
@@ -52,7 +52,7 @@ def _is_phone_number(form, field):
 
 def Alphabet(message="Please enter only letters."):
```
montana-mil: the message should probably be updated, too. or taken away?

patricksmithdds: We may want to rename the validator altogether, too -- it's not validating that the contents are alphabetical anymore.

leigh-mil: Right now this validator is only used for names, do you think this validator will be used for any additional fields in the future? 
If not we can rename the validator to `Name` or something similar.

patricksmithdds: I'm not sure where else we would use this validator. I think `Name` is a good name for the validator; it more accurately describes what's being validated.

tests/forms/test_validators.py
```diff
@@ -38,16 +38,16 @@ def test_PhoneNumber_rejects_invalid_numbers(
...
+    @pytest.mark.parametrize("valid", ["a", "abcde", "hi mark", "cloud9", "niña"])
+    def test_Name_accepts_letters(self, valid, dummy_form, dummy_field):
+        validator = Name()
         dummy_field.data = valid
         validator(dummy_form, dummy_field)
 
-    @pytest.mark.parametrize("invalid", ["", "hi mark", "cloud9"])
-    def test_Alphabet_rejects_non_letters(self, invalid, dummy_form, dummy_field):
-        validator = Alphabet()
+    @pytest.mark.parametrize("invalid", [""])
```
patricksmithdds: Can we add a couple more invalid test cases -- just some strings containing invalid characters.

leigh-mil:
@patricksmithdds Finished up Montana's and your requested changes!

tests/forms/test_validators.py
leigh-mil: Ignore my other message, I didn't see this before I send it!

PR #402, opened 2018-10-23
cronjob config for syncing CRLs

dandds:
PT story: https://www.pivotaltracker.com/story/show/159074108

This adds a cronjob to our Docker containers for syncing CRLs every 6 hours. One hiccough is that we have to start `crond` in the background, so I added that to the CMD.

I can't find a way to test this thoroughly; getting the docker image completely built and running it locally is difficult. I spent some time trying to get a `docker-compose.yml` together that we could use for testing docker changes, but that felt like a rabbit hole. It might be worth pursuing in the future, though.

### DIFF
--- a/deploy/docker/prod/Dockerfile
+++ b/deploy/docker/prod/Dockerfile
@@ -21,7 +21,7 @@ EXPOSE "${APP_PORT}"
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 
 # Default command is to launch the server
-CMD ["bash", "-c", "crond -b && ${APP_DIR}/script/uwsgi_server"]
+CMD ["bash", "-c", "${APP_DIR}/script/uwsgi_server"]
 
 ### Items that will change almost every build
 #############################################
@@ -40,9 +40,6 @@ WORKDIR ${APP_DIR}
 # Add required system packages and app user
 RUN set -x ; \
   script/alpine_setup "${APP_USER}" "${APP_GROUP}"
-
-# add sync-crl cronjob for atst user
-RUN echo "1 */6 * * * /opt/atat/atst/script/sync-crls tests/crl-tmp" >> /etc/crontabs/atst
 
 # Update file ownership
 RUN set -x ; \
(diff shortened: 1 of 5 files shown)


deploy/docker/prod/Dockerfile
```diff
@@ -21,7 +21,7 @@ EXPOSE "${APP_PORT}"
 ENTRYPOINT ["/usr/bin/dumb-init", "--"]
 
 # Default command is to launch the server
-CMD ["bash", "-c", "${APP_DIR}/script/uwsgi_server"]
+CMD ["bash", "-c", "crond -b && ${APP_DIR}/script/uwsgi_server"]
```
patricksmithdds: Mentioned this IRL, but: I believe this Dockerfile isn't being used and the container is being built by CircleCI in `.circleci/config.yml` instead.

dandds:
@patricksmithdds We talked about this IRL, but getting one of our images to run locally is tricky. I can get it to stay up for a couple minutes though, and can get the running processes:

```
❯ docker exec -i -t atst ps aux
PID   USER     TIME  COMMAND
    1 atst      0:00 /usr/bin/dumb-init -- bash -c crond -b && /opt/atat/atst/s
    7 atst      0:00 {uwsgi_server} /bin/bash /opt/atat/atst/script/uwsgi_serve
    9 atst      0:00 crond -b
   18 atst      0:02 /opt/atat/atst/.venv/bin/python ./script/seed_roles.py
   23 atst      0:00 ps aux
```

Seems like the `crond` background process is up and running.

deploy/docker/tester/Dockerfile
```diff
@@ -27,6 +27,9 @@ WORKDIR ${APP_DIR}
 # Copy over setup scripts
 COPY script/ ./script/
 
+# add sync-crl cronjob for atst user
+RUN echo "1 */6 * * * /opt/atat/atst/script/sync-crls tests/crl-tmp" >> /etc/crontabs/atst
```
patricksmithdds: Since this is in `alpine_setup` as well, the Dockerfiles might cause the `sync-crls` script to be run twice on the same schedule -- I imagine there could be conflicts since the scripts would attempt to write the same files. So, can we just rely on this editing of `crontabs` to be in `alpine_setup`?

dandds: Yeah, that's fine. I thought CD wasn't using these Dockerfiles anymore, though?

patricksmithdds: Yeah, they're not being used, but in case we do want to go back to using them, I could see it causing an issue, so just wanted to mention it.

PR #403, opened 2018-10-24
pin flask_session sub-dependency

dandds:
A `flask_session` sub-dependency, `itsdangerous`, updated its API within the last couple days and now I'm getting errors for `script/update`. This pins that dep at its previous version. We can upgrade when `flask_session` does.


dandds:
Nevermind. I think this only happened after I tried to install `rpdb` and `pipenv` re-calculated the dependency tree. It's not necessary on current master.

PR #406, opened 2018-10-25
Invite email #159201690

dandds:
PT story: https://www.pivotaltracker.com/story/show/159201690

with @leigh-mil 

When a workspace owner adds a new user, a notification email with an invitation link is sent to that user.

*implementation notes*
The logic behind invitations is pretty straightforward, but I think the user invitation flow has some consequences for our databases schema as it relates to how we manage new users and workspace authorization.
- When a workspace owner adds a user that did not previously exist, the user created in the database has a `provisional` attribute set to `True`. When a user who has been added in this way logs in for the first time, `provisional` changes to `False`. This means that in future stories we can allow workspace owners to edit user details as long as `provisional` is still true.
- A workspace_role has an `accepted` column to indicate whether the user has accepted an invitation to the workspace. This is now a precondition for authorizing a regular user to do anything with a workspace. 
I'm not completely sold on this setup as-implemented, so if you have thoughts, please chime in.


atst/domain/invitations.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, invite_id):
+        self.invite_id = invite_id
+
+    @property
+    def message(self):
+        return "{} has expired".format(self.invite_id)
+
+
+class Invitations(object):
+    EXPIRATION_LIMIT = 360
```
patricksmithdds: Can we add a comment or update the variable name to indicate the unit of time this represents?

atst/models/invitation.py
```diff
@@ -0,0 +1,25 @@
...
+from sqlalchemy.orm import relationship
+
+from atst.models import Base, types
+from atst.models.mixins.timestamps import TimestampsMixin
+
+
+class Invitation(Base, TimestampsMixin):
+    __tablename__ = "invitations"
+
+    id = types.Id()
```
patricksmithdds: I'm curious if we want to add a `token` column as well? The token is then what we'd put in the link in the email instead of the id. The benefit would be that we would be able to decouple the invitation token from the kind of ID we're using.

atst/models/invitation.py
```diff
@@ -0,0 +1,25 @@
...
+
+    user_id = Column(UUID(as_uuid=True), ForeignKey("users.id"), index=True)
+    user = relationship("User", backref="invitations")
```
patricksmithdds: I think we should also track the user that created the invitation as well.

atst/models/invitation.py
```diff
@@ -0,0 +1,25 @@
...
+
+    workspace_id = Column(UUID(as_uuid=True), ForeignKey("workspaces.id"), index=True)
+    workspace = relationship("Workspace", backref="invitations")
+
+    valid = Column(Boolean, default=True)
```
patricksmithdds: The `valid` boolean seems a little misleading to me -- even if an invitation were expired, the `valid` column would still be set to `True` until someone attempts to use the token. I'd propose that we instead keep track of a status enum with a couple values: 

* "accepted" - the invited user accepted the invitation
* "revoked" - the inviter revoked the invitation
* "pending" - the invite was sent but not used yet
* "rejected" - the invite was used but there was an error (eg., mismatched user ID or expired)

We could also add a `expiration_time` column that we could use to see if a token is expired or not.

atst/models/workspace_role.py
```diff
@@ -22,6 +22,8 @@ class WorkspaceRole(Base, mixins.TimestampsMixin, mixins.AuditableMixin):
         UUID(as_uuid=True), ForeignKey("users.id"), index=True, nullable=False
     )
 
+    accepted = Column(Boolean)
```
patricksmithdds: Do we need to track this in the database? Could this be a computed property that basically sees if a workspace member has an unaccepted invitation? A user could not accept an invitation but not accept their role in a workspace, so I don't think I see the value of having this as a separate column.

atst/routes/workspaces.py
```diff
@@ -299,3 +317,14 @@ def update_member(workspace_id, member_id):
             workspace=workspace,
             member=member,
         )
+
+
+@bp.route("/workspaces/invitation/<invite_id>", methods=["GET"])
+def accept_invitation(invite_id):
+    invite = Invitations.accept(invite_id)
```
patricksmithdds: This also needs to check the user id of the current user. If we're going to handle that in the followup card, can we just put a `TODO` somewhere to remind us?

atst/models/invitation.py
dandds: Sure, that makes sense. It can still just be a random string, right? We're not encrypting any data in it.

atst/models/invitation.py
dandds: Sure, we can do that.

atst/models/workspace_role.py
dandds: A computed property on the `WorkspaceRole` model? We could do that, it just feels a little like mixing domain logic into the models in a way we've been avoiding. Maybe a class method on the `Workspaces` domain class?

atst/routes/workspaces.py
dandds: Yup, will add a TODO. I was going to leave the specifics of identity-matching for the next story.

atst/models/workspace_role.py
dandds: The problem I'm running into--which maybe isn't a problem--is that we have to join the `invitations` table now to check if there is an accepted invite when we build lists of workspaces for a given user. I don't think this is necessarily bad, but I will have to take some time to cleanup the tests. We need to ensure that there is an "accepted" invite every time we test a user's valid workspaces, which results in some test boilerplate.

dandds: Curious what @richard-dds thinks of the (unfinished) changes on 0cf7800

dandds:
@patricksmithdds This should be ready for another pass. I think I handled all your comments; let me know if not. If we remove the `accepted` column from the workspace role, we have to join and query with the invitations table to get user workspaces. I tried to consolidate the workspace permission logic into the `WorkspaceUsers` domain, which should know how to limit a user's perms if they haven't accepted an invitation. The query we use to get a user's workspaces is also scoped in this way.

alembic/versions/25bcba9b99a9_add_invitiations.py
```diff
@@ -0,0 +1,48 @@
...
+    op.create_index(op.f('ix_invitations_user_id'), 'invitations', ['user_id'], unique=False)
+    op.create_index(op.f('ix_invitations_inviter_id'), 'invitations', ['inviter_id'], unique=False)
+    op.create_index(op.f('ix_invitations_workspace_id'), 'invitations', ['workspace_id'], unique=False)
+    op.create_index(op.f('ix_invitations_token'), 'invitations', ['token'], unique=True)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index(op.f('ix_invitations_workspace_id'), table_name='invitations')
```
richard-dds: I see four calls to `create_index` but only two to `drop index`. Is that an issue, or do the other two get dropped automatically with the table?

atst/domain/workspace_users.py
```diff
@@ -30,6 +31,32 @@ def get(cls, workspace_id, user_id):
...
+                .join(Invitation, WorkspaceRole.workspace_id == Invitation.workspace_id)
+                .filter(Invitation.user_id == WorkspaceRole.user_id)
+                .filter(Invitation.status == InvitationStatus.ACCEPTED)
+                .one()
+            )
+        except NoResultFound:
+            return None
+
+    @classmethod
+    def workspace_user_permissions(cls, workspace, user):
```
patricksmithdds: This change will break all existing users -- since no existing user has an invitation, therefore no user has access to any existing workspace. I know they're not real users yet, but I'm curious if we could approach this a bit differently:

* check if there are any outstanding invitations for this user/workspace. 
    - If it's not accepted, then the user has no permissions
    - If it is accepted, the user has the appropriate workspace/atat permissions
* otherwise, continue as normal -- the user has the appropriate workspace/atat permissions

This would allow us to circumvent creating an auto-accepted invitation for the workspace owner, as well. Curious what you think of that, @dandds ?

atst/domain/workspace_users.py
```diff
@@ -30,6 +31,32 @@ def get(cls, workspace_id, user_id):
...
+    @classmethod
+    def _get_active_workspace_role(cls, workspace_id, user_id):
+        try:
+            return (
+                db.session.query(WorkspaceRole)
+                .join(User)
+                .filter(User.id == user_id, WorkspaceRole.workspace_id == workspace_id)
+                .join(Invitation, WorkspaceRole.workspace_id == Invitation.workspace_id)
+                .filter(Invitation.user_id == WorkspaceRole.user_id)
+                .filter(Invitation.status == InvitationStatus.ACCEPTED)
```
richard-dds: I think we discussed this while we were pairing, so forgive me for being forgetful. Why did you decide to join the Invitation to determine if a workspace role is "active", rather than just keeping an is_active flag on the WorkspaceRole that gets flipped when the invite is accepted?

alembic/versions/25bcba9b99a9_add_invitiations.py
dandds: Good catch, thanks. I squashed a bunch of incremental migrations I made while applying changes and forgot those.

atst/domain/workspace_users.py
dandds: @patricksmithdds had qualms about the flag on WorkspaceRole, since a user can't accept an invitation and also not accept their workspace role.

atst/domain/workspace_users.py
dandds: This would work. The only other option would be a migration to add invitations for existing users, which is complicated in its own way.

I am a little worried about having multiple ways to handle users' relationships to workspaces. That seems like a place security holes might appear. This is, granted, a very vague misgiving.

atst/domain/workspace_users.py
patricksmithdds: Yeah, it seemed to me that the `active` flag on the workspace role was really just duplicating the information in the invitation (since one could not decline a workspace role while accepting an invitation). Unless we're planning on being able to deactivate a workspace role separately from an invitation, it seems like the flag could get out of sync at some point.

PR #407, opened 2018-10-25
Deploy RQ worker

patricksmithdds:
This PR updates the Kubernetes config to include a second deployment to run the RQ worker. Running it as a separate deployment will allow us to scale it up/down in the future, independently of the web service.

I had initially planned on building two different docker images, each with a different `CMD` -- one to run the `wsgi_server` and one to run the new `rq_worker` script. Then, I found that we can specify `args` in the container configuration in Kubernetes to override the `CMD` from the container, which is a bit simpler.

The `atst-worker` deployment is configured similar to the `atst` deployment, but does not add any of the NGINX configuration. I also added an environment variable, `REQUIRE_CRLS`, that can be set to `False` to skip loading all the CRLs, since they require a lot of memory and are not needed for the RQ worker.

These changes are currently deployed to the k8s cluster and I was able to test with the `/test-email` dev route. The test message was then displayed on the `/messages` page and the logs from the worker indicated success:

```
20:23:29 atat_dev: _send_mail(['test@example.com'], 'test', 'test') (be42560a-3cc2-42ba-bc9a-2879fb80530a)
20:23:29 atat_dev: Job OK (be42560a-3cc2-42ba-bc9a-2879fb80530a)
20:23:29 Result is kept for 500 seconds
```

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -28,7 +28,6 @@ from atst.queue import queue
 
 
 ENV = os.getenv("FLASK_ENV", "dev")
-REQUIRE_CRLS = os.getenv("REQUIRE_CRLS", "True")
 
 
 def make_app(config):
@@ -47,7 +46,7 @@ def make_app(config):
     app.config.update({"SESSION_REDIS": app.redis})
 
     make_flask_callbacks(app)
-    if REQUIRE_CRLS == "True":
+    if app.config.get("REQUIRE_CRLS"):
         make_crl_validator(app)
     register_filters(app)
     make_eda_client(app)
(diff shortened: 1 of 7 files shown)


atst/app.py
```diff
@@ -28,6 +28,7 @@
 
 
 ENV = os.getenv("FLASK_ENV", "dev")
+REQUIRE_CRLS = os.getenv("REQUIRE_CRLS", "True")
```
dandds: Are you keeping this out of the app config just for simplicity? 

Doing string comparison for what's basically a bool seems weird, but I think handling bools through environment variables is weird, period. I did discover that if you set something in one of the INI files to `0` (like `DEBUG = 0`) and then use `config.getboolean` to get that value, it will evaluate to `False`.

patricksmithdds: I initially did it this way since it's a lot easier to add an environment variable to the kubernetes config vs copy & tweak the config.ini and keep it in a secret. But, I forgot that env variables will override the values from the config, so I did add it to the base config and take advantage of the `config.getboolean`: e177fe0

PR #410, opened 2018-10-29
Fix project name

andrewdds:
The project name was being used to generate an ID for the access-revocation modal. This was causing an error when it had an apostrophe in the name.

This changes it to use the project ID instead of the name, since it's a safer string. Also cleans up the props attributes, removing unnecessary `v-bind:`s and quotes.

Adds a global template function for casting to string.

### DIFF
--- a/templates/workspaces/members/edit.html
+++ b/templates/workspaces/members/edit.html
@@ -53,7 +53,7 @@
   </div>
 
   {% for project in projects %}
-  {% set revoke_modal_name = string(project.id) + 'RevokeModal' %}
+  {% set revoke_modal_name = (project.id|string) + 'RevokeModal' %}
   <edit-project-roles inline-template name="{{ project.name }}" id="{{ project.id }}">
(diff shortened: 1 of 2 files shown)


templates/workspaces/members/edit.html
```diff
@@ -53,8 +53,8 @@ <h2 class="h3">Manage Access <div class="subtitle">Grant access to an environmen
   </div>
 
   {% for project in projects %}
-  {% set revoke_modal_name = project.name + 'RevokeModal' %}
-  <edit-project-roles inline-template v-bind:name="'{{ project.name }}'" v-bind:id="'{{ project.id }}'">
+  {% set revoke_modal_name = string(project.id) + 'RevokeModal' %}
```
patricksmithdds: I believe you can use the filter directly and then don't need any of the changes in `filters.py`: 
```suggestion
  {% set revoke_modal_name = (project.id|string) + 'RevokeModal' %}
```

PR #411, opened 2018-10-29
Add phone ext

leigh-mil:
## Description
This PR allows users to specify an extension for their phone number. The phone number and extension fields were moved on to their own lines as suggested by @andrewdds. This not only made styling easier but also (hopefully) make these fields more noticeable to users, which was an issue during UAT!

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/161268015](https://www.pivotaltracker.com/story/show/161268015)

## Screenshots
screen shot 2018-10-31 at 4 54 41 pm


patricksmithdds:
@leigh-mil This has some conflicts with master -- can you rebase or resolve those conflicts? One tricky thing may be updating the migration chain, since there were some added in master since this branch was started.

leigh-mil:
@patricksmithdds the branch has been rebased!

js/lib/input_validations.js
```diff
@@ -47,6 +47,12 @@ export default {
     unmask: ['(',')','-',' '],
     validationError: 'Please enter a 10-digit phone number'
   },
+  phoneExt: {
+    mask: [/\w/, /\w/, /\w/, /\w/, /\w/, /\w/, /\w/, /\w/, /\w/, /\w/],
+    match: /^\w*$/,
+    unmask: [],
```
patricksmithdds: I believe `unmask` here should specify the placeholder: `['_']`. Currently, if you enter two numbers, it's saved as `42________` instead of just `42`.

lpassamano:
@patricksmithdds the changes you requested have been made. I also had to rebase to include PR #431 so my tests would pass!

alembic/versions/ea8b5681b37c_make_user_id_nullable_for_request_review.py
```diff
@@ -1,7 +1,7 @@
 """make user_id nullable for request_review
 
 Revision ID: ea8b5681b37c
-Revises: e1081cf01780
+Revises: 4c0b8263d800
```
patricksmithdds: This change actually makes it so that your new migration doesn't get run. As the branch is now, your migration is basically added right behind this migration.

`ea8b5681b37c` (make user_id nullable) is the latest revision and it revises your new migration, `4c0b8263d800`. When we run alembic to upgrade the database, it checks to see what revision the database schema is currently using. Since `ea8b5681b37c` is already the latest revision in master, alembic will see that everything is up to date and not apply any migrations. 

Instead, this revision should be unchanged and the revision should now specify that it revises this revision (`ea8b5681b37c`) instead of what it currently revises (`e1081cf01780`).

leigh-mil:
@patricksmithdds migrations are fixed and correct (I hope)!

PR #412, opened 2018-10-29
Notify PSO or MO of request status change

richard-dds:
Send PSO or MO an email when certain status changes occur.

Example email:

```
From: mailtester1013@gmail.com
To: brandon@test.com
Subject: Your JEDI request status has changed 

<h1>Your JEDI request status has changed</h1>

The status of your JEDI Cloud request - ellis - was recently updated. Log in to see whether this change requires an action or response from you. 

http://localhost:8000/requests/details/673d576d-ac08-448f-8d5c-25807e6a7ebb 
```

### DIFF
--- a/atst/domain/requests/status_event_handler.py
+++ b/atst/domain/requests/status_event_handler.py
@@ -4,24 +4,20 @@ from atst.models.request_status_event import RequestStatus
 
 
 class RequestStatusEventHandler(object):
-    STATUS_TRANSITIONS = set([
-        (
-            RequestStatus.PENDING_CCPO_ACCEPTANCE,
-            RequestStatus.PENDING_FINANCIAL_VERIFICATION,
-        ),
-        (
-            RequestStatus.PENDING_CCPO_ACCEPTANCE,
-            RequestStatus.CHANGES_REQUESTED,
-        ),
-        (
-            RequestStatus.PENDING_CCPO_APPROVAL,
-            RequestStatus.CHANGES_REQUESTED_TO_FINVER,
-        ),
-        (
-            RequestStatus.PENDING_CCPO_APPROVAL,
-            RequestStatus.APPROVED,
-        ),
-    ])
+    STATUS_TRANSITIONS = set(
+        [
+            (
+                RequestStatus.PENDING_CCPO_ACCEPTANCE,
+                RequestStatus.PENDING_FINANCIAL_VERIFICATION,
+            ),
+            (RequestStatus.PENDING_CCPO_ACCEPTANCE, RequestStatus.CHANGES_REQUESTED),
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.CHANGES_REQUESTED_TO_FINVER,
+            ),
+            (RequestStatus.PENDING_CCPO_APPROVAL, RequestStatus.APPROVED),
+        ]
+    )
 
     def __init__(self, queue):
         self.queue = queue
--- a/atst/domain/requests/status_event_handler.py
+++ b/atst/domain/requests/status_event_handler.py
@@ -1,0 +1,41 @@
+from flask import render_template
+
+from atst.models.request_status_event import RequestStatus
+
+
+class RequestStatusEventHandler(object):
+    def __init__(self, queue):
+        self.queue = queue
+
+    def handle_status_change(self, request, old_status, new_status):
+        handler = self._get_handler(old_status, new_status)
+        if handler:
+            handler(request)
+
+    def _get_handler(self, old_status, new_status):
+        return {
+            (
+                RequestStatus.PENDING_CCPO_ACCEPTANCE,
+                RequestStatus.PENDING_FINANCIAL_VERIFICATION,
+            ): self._send_email,
+            (
+                RequestStatus.PENDING_CCPO_ACCEPTANCE,
+                RequestStatus.CHANGES_REQUESTED,
+            ): self._send_email,
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.CHANGES_REQUESTED_TO_FINVER,
+            ): self._send_email,
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.APPROVED,
+            ): self._send_email,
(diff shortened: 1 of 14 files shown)


templates/emails/request_status_change.txt
```diff
@@ -0,0 +1,5 @@
+<h1>Your JEDI request status has changed</h1>
+
+The status of your JEDI Cloud request - {{ request.name }} - was recently updated. Log in to see whether this change requires an action or response from you.
```
dandds: I think this needs to be `request.displayname`. I'm getting blanks here every time.

richard-dds: Whoops! I made that change and forgot to commit it.

templates/emails/request_status_change.txt
```diff
@@ -0,0 +1,5 @@
+<h1>Your JEDI request status has changed</h1>
```
patricksmithdds: If this is a plaintext email, should we skip the `<h1>` tags here?

atst/domain/requests/status_event_handler.py
```diff
@@ -0,0 +1,41 @@
...
+            ): self._send_email,
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.CHANGES_REQUESTED_TO_FINVER,
+            ): self._send_email,
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.APPROVED,
+            ): self._send_email,
+        }.get((old_status, new_status))
```
dandds: Is the idea here that the handlers might change? Is there an advantage over just sending the email if a tuple of the old and new status exist in a known list of tuples?

richard-dds: Yeah, I was wondering about that. Now that you mention it, it's probably best not to try to predict the future. I'll just check for membership rather than doing that unnecessary dispatch logic.

atst/domain/requests/status_event_handler.py
```diff
@@ -0,0 +1,35 @@
...
+            (RequestStatus.PENDING_CCPO_ACCEPTANCE, RequestStatus.CHANGES_REQUESTED),
+            (
+                RequestStatus.PENDING_CCPO_APPROVAL,
+                RequestStatus.CHANGES_REQUESTED_TO_FINVER,
+            ),
+            (RequestStatus.PENDING_CCPO_APPROVAL, RequestStatus.APPROVED),
+        ]
+    )
+
+    def __init__(self, queue):
```
patricksmithdds: Is this passed in for dependency injection reasons (I'm assuming tests)? Or, can we just import the queue instead?

templates/emails/request_status_change.txt
```diff
@@ -0,0 +1,5 @@
+Your JEDI request status has changed
+
+The status of your JEDI Cloud request - {{ request.displayname }} - was recently updated. Log in to see whether this change requires an action or response from you.
+
+{{ url_for('requests.view_request_details', request_id=request.id, _external=True) }}
```
patricksmithdds: I think this follows the ACs of the issue, but I'm not sure this is the best URL to send the user to -- it'd be nice if we could send a link to the next step of the process. Additionally, the request details page does not show the comments from the CCPO user, which would be confusing for the user in half of the scenarios in which we'll be sending the emails. I think if we tailor the URL a bit, we could get around some potential user frustration.

In my mind, the ideal link would be based on the new status of the request, basically what happens when we display the request in the index table: https://github.com/dod-ccpo/atst/blob/d696acbaa4718cad6522be8f611dc68c0d13388a/atst/routes/requests/index.py#L58-L73.

tests/domain/test_requests.py
```diff
@@ -222,3 +220,37 @@ def test_random_user_cannot_view_request():
     request = RequestFactory.create()
 
     assert not RequestsAuthorization(user, request).can_view
+
+
+def test_pending_finver_triggers_notification(queue):
+    request = RequestFactory.create()
+    request = Requests.set_status(request, RequestStatus.PENDING_CCPO_ACCEPTANCE)
+    request = Requests.set_status(request, RequestStatus.PENDING_FINANCIAL_VERIFICATION)
+    assert len(queue.get_queue()) == 1
```
patricksmithdds: Not sure if this is overkill, but could we check that the job is correct? Something like:

```python
job = queue.get_queue().jobs[0]
assert job.func == queue._send_mail
assert job.args[0] == [request.owner]
```

tests/test_queue.py
```diff
@@ -1,17 +1,5 @@
...
-
-
-def test_send_mail():
-    initial = len(queue.get_queue())
+def test_send_mail(queue):
     queue.send_mail(
         ["lordvader@geocities.net"], "death start", "how is it coming along?"
     )
-    assert len(queue.get_queue()) == initial + 1
+    assert len(queue.get_queue()) == 1
```
patricksmithdds: 👍

atst/domain/requests/status_event_handler.py
richard-dds: It is for DI, but we're currently not making use of it for tests. Should I remove it?

templates/emails/request_status_change.txt
richard-dds: Very good call. I can think of two approaches to this: extract this method into a function somewhere where it can be shared, or make this method a route that redirects based on the request status. Do you have a preference?

patricksmithdds: Hmm, that's a good question. I was going to say extract it to a function, but since the link is in an email, the status could change (for example, if the user clicks a link from a previous email). In that case, it might be better to have a route that redirects so that the links always make sense.

atst/routes/requests/requests_form.py
```diff
@@ -144,3 +145,34 @@ def view_request_details(request_id=None):
...
+                "requests.view_request_details", request_id=request.id
+            )
+        else:
+            redirect_url = url_for(
+                "requests.requests_form_update", screen=1, request_id=request.id
+            )
+    elif is_ccpo:
+        redirect_url = url_for("requests.approval", request_id=request.id)
+
+    return redirect(redirect_url)
```
patricksmithdds: Could we add some tests to document this route as well?

PR #413, opened 2018-10-30
update requests

dandds:
This updates `requests` to 2.20.0, which addresses [CVE-2018-18074](https://nvd.nist.gov/vuln/detail/CVE-2018-18074). 

EDIT
<strike>Apparently `pipenv` won't even let you try to update one dependency in isolation. [It's a whole thing.](https://github.com/pypa/pipenv/issues/966) Because of this, I've updated all our Pipfile dependencies. Everything still passes.</strike>

Okay, just manually updated the `Pipfile.lock` so we only update requests. Ignore my comments below.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -322,11 +322,11 @@
         },
         "requests": {
             "hashes": [
-                "sha256:63b52e3c866428a224f97cab011de738c36aec0185aa91cfacd418b5d58911d1",
-                "sha256:ec22d826a36ed72a7358ff3fe56cbd4ba69dd7a6718ffd450ff0e9df7a47ce6a"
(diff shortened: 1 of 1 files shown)


dandds:
screen shot 2018-10-30 at 9 17 51 am

😭

dandds:
I think one of the dub-dependencies for `pendulum` is breaking our `sync-crls` script. This is why it's nice to be able to increment dependencies one at a time.

dandds:
We're seeing this specific issue from `pendulum`: https://github.com/sdispater/pendulum/issues/300

We should wait to merge this until we can bump to the forthcoming pendulum release. I don't want to lock sub-dependencies like I did on the most recent commit.

PR #416, opened 2018-10-30
Searching, filtering and sorting for requests page

richard-dds:
Searching, filtering and sorting are all completed. Could use a styling pass, since I'm not sure how the column headings should look now that they're clickable.

Columns (the first one has been clicked to enable sorting):
screen shot 2018-10-31 at 3 54 27 pm

Possible statuses to filter by:
screen shot 2018-10-31 at 3 54 39 pm

### DIFF
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -77,6 +77,10 @@ def dateFromString(value, formatter="%m/%Y"):
     return datetime.datetime.strptime(value, formatter)
 
 
+def boolean(value):
+    return "true" if value else "false"
+
+
 def register_filters(app):
     app.jinja_env.filters["iconSvg"] = iconSvg
     app.jinja_env.filters["dollars"] = dollars
@@ -88,3 +92,4 @@ def register_filters(app):
     app.jinja_env.filters["renderList"] = renderList
     app.jinja_env.filters["formattedDate"] = formattedDate
     app.jinja_env.filters["dateFromString"] = dateFromString
+    app.jinja_env.filters["boolean"] = boolean
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -77,10 +77,6 @@ def dateFromString(value, formatter="%m/%Y"):
     return datetime.datetime.strptime(value, formatter)
 
 
-def boolean(value):
-    return "true" if value else "false"
-
-
 def register_filters(app):
     app.jinja_env.filters["iconSvg"] = iconSvg
     app.jinja_env.filters["dollars"] = dollars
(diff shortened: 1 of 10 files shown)


atst/filters.py
```diff
@@ -88,3 +92,4 @@ def register_filters(app):
     app.jinja_env.filters["renderList"] = renderList
     app.jinja_env.filters["formattedDate"] = formattedDate
     app.jinja_env.filters["dateFromString"] = dateFromString
+    app.jinja_env.filters["boolean"] = boolean
```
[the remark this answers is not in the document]
richard-dds: I was thinking that it didn't work, but I was struggling with a few Vue things at the time and might be wrong about that. I'll test it out.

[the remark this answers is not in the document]
richard-dds: You're right!

templates/requests/index.html
```diff
@@ -3,8 +3,11 @@
 {% from "components/alert.html" import Alert %}
 {% from "components/modal.html" import Modal %}
 {% from "components/empty_state.html" import EmptyState %}
+{% from "components/icon.html" import Icon %}
 
 {% block content %}
+  <requests-list inline-template v-bind:requests='{{ requests | tojson }}' v-bind:is-extended='{{ extended_view | tojson }}' v-bind:statuses='{{ possible_statuses | tojson }}'>
```
patricksmithdds: Minor comment: if the modals below are before the opening tag of `<requests-list>`, I believe we won't need to include the `Modal` component in the [`RequestsList` component](https://github.com/dod-ccpo/atst/pull/416/files#diff-f67c675bf84c99a595601f541932ae09R13).

js/components/forms/requests_list.js
```diff
@@ -0,0 +1,146 @@
...
+        sort: {
+          columnName: '',
+          isAscending: true
+        },
+        columns: indexBy(prop('attr'), columnList),
+      }
+    },
+
+    mounted: function () {
+    },
```
patricksmithdds: Looks like we can remove this vestigial function.

templates/requests/index.html
```diff
@@ -101,52 +102,49 @@ <h1>Request submitted!</h1>
...
-              <th scope="col">Requester</th>
-            {% endif %}
-            <th scope="col">Projected Annual Usage ($)</th>
-            <th scope="col">Request Status</th>
-            {% if extended_view %}
-              <th scope="col">DOD Component</th>
-            {% endif %}
+            <th @click.prevent="updateSortValue(column.attr)" v-for="column in getColumns()"scope="col">
+              !{ column.displayName }
+              <span v-if="column.attr === sort.columnName && sort.isAscending" class="icon icon--caret_up" aria-idden="true">
```
patricksmithdds: I believe the `class` and `aria-hidden` attrs on this element aren't necessary, since they're in the `Icon` template as well.

PR #417, opened 2018-10-31
Fix manual fin ver #161560975

leigh-mil:
## Description
This PR fixes a regression that required a valid TO number to submit an extended financial verification form. Everything is now validating as it should, and only checks for the presence of a TO number when submitting an extended fin ver form.

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/161560975](https://www.pivotaltracker.com/story/show/161560975)


richard-dds:
👏 Looks good!

PR #421, opened 2018-10-31
Update UAT environment

patricksmithdds:
This PR updates the UAT env with the latest changes to our deployment including:

* the cron job to sync CRLs will now be running
* the container image has been update to the latest code
* the flask-rq worker deployment has been added

This branch also allows specifying the `RQ_QUEUES` via an environment variable so that the UAT env and the staging env can continue to use the same Redis resource without conflicting use of RQ.

These changes have already been applied to the `atat-uat` namespace on the cluster.

### DIFF
--- a/config/base.ini
+++ b/config/base.ini
@@ -16,6 +16,7 @@ PGUSER = postgres
 PORT=8000
 REDIS_URI = redis://localhost:6379
 REQUIRE_CRLS = true
+RQ_QUEUES = atat_%(ENVIRONMENT)s
 SECRET = change_me_into_something_secret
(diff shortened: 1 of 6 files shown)


config/base.ini
```diff
@@ -16,6 +16,7 @@ PGUSER = postgres
 PORT=8000
 REDIS_URI = redis://localhost:6379
 REQUIRE_CRLS = true
+RQ_QUEUES = atat_%(ENVIRONMENT)s
```
dandds: Interesting. Didn't know you could do this in an INI file.

PR #428, opened 2018-11-05
Auto approval logs #160420024

dandds:
https://www.pivotaltracker.com/story/show/160420024

Add CCPO review entries when the system accepts or approves a request automatically.

screen shot 2018-11-05 at 3 27 54 pm


atst/domain/requests/requests.py
```diff
@@ -95,6 +106,16 @@ def approve_and_create_workspace(cls, request):
...
         return workspace
 
+    @classmethod
+    def auto_approve_and_create_workspace(
+        cls,
+        request,
+        reason="Financial verification information found in Electronic Document Access API",
+    ):
+        workspace = Requests.approve_and_create_workspace(request)
+        Requests._add_review(None, request, {"comment": reason})
```
patricksmithdds: Feel free to disregard this comment, but when I use `None` in the argument list like this, I often think it's hard to know _what_ is `None`. In those cases, I like to use keyword args: 

```python
Requests._add_review(user=None, request=request, review_data={"comment": reason})
```

dandds: That's a good idea. I think that's clearer. I feel like every Python function that sticks around for a while winds up using kwargs exclusively :grin:

PR #429, opened 2018-11-05
Fix json serialization bug in finver form

richard-dds:
Oof

### DIFF
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -1,8 +1,6 @@
 import re
 import datetime
 from flask import current_app as app
-from werkzeug.datastructures import FileStorage
-from atst.models.attachment import Attachment
 
 
 def iconSvg(name):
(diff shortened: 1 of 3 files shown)


atst/filters.py
```diff
@@ -47,9 +48,12 @@ def mixedContentToJson(value):
     if (
         isinstance(value, dict)
         and "task_order" in value
-        and isinstance(value["task_order"], FileStorage)
+        and (
+            isinstance(value["task_order"]["pdf"], FileStorage)
+            or isinstance(value["task_order"]["pdf"], Attachment)
```
patricksmithdds: Are there any other possible types here? Rather than checking the instance type, we could check for the `filename` attr: `hasattr(value["task_order"]["pdf"], "filename")`.

richard-dds: Checking for the attr is a great idea. I'll do that instead.

PR #430, opened 2018-11-06
script for generating user test certs for PIVKey cards

dandds:
I made a little script that can generate the cert, key, and PFX package for user certs for the CAC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,9 @@ config/dev.ini
 /crl-tmp
 *.bk
 
+# test CA config
+ssl/client-certs/*.srl
+
 # uploads
 /uploads
 
--- a/script/make-test-cac
+++ b/script/make-test-cac
@@ -13,20 +13,20 @@ set -e
 
 SAN="subjectAltName=email:$2"
 
+openssl genrsa -out $3.key 2048
+
 CSR=$(openssl req \
   -new \
-  -newkey rsa:4096 \
-  -sha256 \
   -nodes \
-  -days 365 \
   -subj "/CN=$1" \
   -reqexts SAN \
   -config <(cat /etc/ssl/openssl.cnf; echo '[SAN]'; echo $SAN) \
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -39,6 +39,9 @@ config/dev.ini
 /crl-tmp
 *.bk
 
+# test CA config
+ssl/client-certs/*.srl
```
patricksmithdds: Should this be `*.crl`? Or, how does a .srl get generated?

script/make-test-cac
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# script/make-test-cac:  Set up a test CAC card.
+# Usage:
+# ./script/make-test-cac [DOD identifier string] [user email] [certificate name]
```
patricksmithdds: I was a little thrown off about "certificate name" -- is that just the filename of the generated cert?

script/make-test-cac
```diff
@@ -0,0 +1,42 @@
...
+# i.e.:
+# ./script/make-text-cac JONES.ANDY.1234567890 andy@example.com andy
```
patricksmithdds: tiny typo: `make-text-cac` => `make-test-cac`

.gitignore
dandds: `openssl` needs the CA to have some kind of serial number in order to sign client certs, and the `-CAcreateserial` option on the `x509` command autogenerates it for us. I added *.srl to the gitignore, too.

script/make-test-cac
dandds: Yeah, it uses that name for all the generated files: `andy.crt`, `andy.key`, `andy.pfx`. I'll update it to "output name".

.gitignore
dandds: Oh that _is_ the gitignore, duh

PR #432, opened 2018-11-06
control error display for options input component

dandds:
PT story: https://www.pivotaltracker.com/story/show/161717158

This expands our `optioninput` Vue component so that it can dismiss errors when a user marks a choice. I think that the only validation error these can have is that nothing has been marked (because they're required), so any kind of input update should be valid. I also made a small adjustment to our sass so that radio buttons and checkboxes get the correct validation color.

### DIFF
--- a/js/components/options_input.js
+++ b/js/components/options_input.js
@@ -2,8 +2,22 @@ export default {
   name: 'optionsinput',
 
   props: {
-    name: String
+    name: String,
+    initialErrors: {
+      type: Array,
+      default: () => []
+    },
   },
+
+
+  data: function () {
+    return {
+      showError: (this.initialErrors && this.initialErrors.length) || false,
+      showValid: false,
+      validationError: this.initialErrors.join(' ')
+    }
+  },
+
 
   methods: {
     onInput: function (e) {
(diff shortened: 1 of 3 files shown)


js/components/options_input.js
```diff
@@ -2,15 +2,31 @@ export default {
...
+      type: Array,
+      default: () => []
+    },
   },
 
+
+  data: function () {
+    return {
+      showError: (this.initialErrors && this.initialErrors.length) || false,
+      showValid: false,
```
patricksmithdds: It'd be nice to set `showValid` to show the green checkmark if the input is rendered with a valid value. Currently, we can get in a state like this:

image

Since I didn't change the second input, it doesn't show the valid icons, but the first does because I toggled the value. I think this could be misleading to the user.

dandds: Okay I added a fix by measuring for an initial value. One quirk is that WTForms seems to always send the string "None" as the data for radio fields, so we have to have an extra guard for that.

patricksmithdds: Hmm, that's an annoying quirk. Looks like theres a closed issue for that (https://github.com/wtforms/wtforms/issues/289), but no new releases have been made with the fix.

PR #436, opened 2018-11-07
Revoke invitation #160300315

dandds:
https://www.pivotaltracker.com/story/show/160300315

This gives workspace owners and admins the ability to revoke a pending invitation.

I also added a new macro for a confirmation button component. The button presupposes that it's being used as the D in CRUD, so it's wrapped in its own small `form` element and needs a form action specified. This should prevent any form inheritance and submission confusion in the DOM.

The confirmation popup and copy could use a UI/UX pass if you have time, @luiscielak or @andrewcroce .

screen shot 2018-11-07 at 9 46 49 am

### DIFF
--- a/atst/routes/workspaces.py
+++ b/atst/routes/workspaces.py
@@ -357,10 +357,26 @@ def update_member(workspace_id, member_id):
         )
 
 
-@bp.route("/workspaces/invitation/<token>", methods=["GET"])
+@bp.route("/workspaces/invitations/<token>", methods=["GET"])
 def accept_invitation(token):
     invite = Invitations.accept(g.current_user, token)
 
     return redirect(
         url_for("workspaces.show_workspace", workspace_id=invite.workspace.id)
+    )
+
+
+@bp.route("/workspaces/<workspace_id>/invitations/<token>/revoke", methods=["POST"])
+def revoke_invitation(workspace_id, token):
+    workspace = Workspaces.get(g.current_user, workspace_id)
+    Authorization.check_workspace_permission(
+        g.current_user,
+        workspace,
+        Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
+        "revoke member invitation",
+    )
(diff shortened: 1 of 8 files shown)


nohup.out
```diff
@@ -0,0 +1,7 @@
+3296:C 07 Nov 08:57:25.261 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
```
richard-dds: You probably didn't mean to commit this

atst/routes/workspaces.py
```diff
@@ -357,10 +357,24 @@ def update_member(workspace_id, member_id):
...
 
     return redirect(
         url_for("workspaces.show_workspace", workspace_id=invite.workspace.id)
     )
+
+
+@bp.route("/workspaces/<workspace_id>/invitations/<token>/revoke", methods=["POST"])
+def revoke_invitation(workspace_id, token):
+    workspace = Workspaces.get(g.current_user, workspace_id)
+    Authorization.check_workspace_permission(
```
richard-dds: I feel like we've done this in a few places recently, but I thought that we'd decided to do auth in the domain layer?

dandds: You're right. I was following the existing route functions in that module, which all do it directly. I think that whole module needs to be refactored.

nohup.out
dandds: sure didn't, thanks

dandds:
@richard-dds Added another Workspace domain method for getting a workspace in order to update its members. I definitely prefer having authz logic in the domain. At the same time, it seems like we're adding a lot of "get this resource for purpose X" methods, which are just thin wrappers so that the caller doesn't have to know a specific permission name. Maybe that's fine, but it doesn't sit well with me. I'd like to see if we can hit on a better, maybe more generic, way of handling this.

richard-dds:
@dandds I definitely see what you mean, and I don't really like it either. Let's think / talk about this

PR #437, opened 2018-11-07
Allow CCPO to filter requests by DOD component

richard-dds:
Filtering works just fine, but needs styling tweaks. The `flex-wrap: wrap` on `.search-bar` is causing the DOD Comopnent select field to wrap to the next line. @andrewcroce or @luiscielak, any ideas?

Edit: styles are fixed now. Thanks Andrew!

### DIFF
--- a/templates/requests/index.html
+++ b/templates/requests/index.html
@@ -37,7 +37,13 @@
     </div>
   {% endcall %}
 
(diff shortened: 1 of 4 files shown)


templates/requests/index.html
```diff
@@ -96,6 +102,15 @@ <h1>Request submitted!</h1>
             <option v-for="status in statuses" :value="status">!{ status }</option>
           </select>
         </div>
+
+        <div class='usa-input'>
+          <label for='filter-status'>Filter requests by DOD component</label>
```
patricksmithdds: The `for` should be `filter-dod-component` here.

PR #440, opened 2018-11-08
Resend a workspace invitation

richard-dds:
https://www.pivotaltracker.com/story/show/161347134

If a workspace member has a display status of "Invite expired" or "Error on invite", the workspace owner will see a button to resend the invitation. Resending will set the status of the previous invitation to "revoked".

screen shot 2018-11-08 at 5 13 11 pm

Includes a few other tweaks:
* simplified Invitations.create signature
* added full profile info for the seed users
* added some workspace members with various statuses to the seed script

### DIFF
--- a/atst/routes/dev.py
+++ b/atst/routes/dev.py
@@ -12,6 +12,8 @@ import pendulum
 from . import redirect_after_login_url
 from atst.domain.users import Users
 from atst.queue import queue
+from tests.factories import random_service_branch
+from atst.utils import pick
 
 bp = Blueprint("dev", __name__)
 
@@ -22,7 +24,7 @@ _DEV_USERS = {
         "last_name": "Stevenson",
         "atat_role_name": "ccpo",
         "email": "sam@example.com",
-        "service_branch": "Fake Service Branch",
+        "service_branch": random_service_branch(),
         "phone_number": "1234567890",
         "citizenship": "United States",
         "designation": "Military",
@@ -34,7 +36,7 @@ _DEV_USERS = {
         "last_name": "Adamson",
         "atat_role_name": "default",
         "email": "amanda@example.com",
-        "service_branch": "Fake Service Branch",
+        "service_branch": random_service_branch(),
         "phone_number": "1234567890",
         "citizenship": "United States",
         "designation": "Military",
@@ -46,7 +48,7 @@ _DEV_USERS = {
         "last_name": "Buchannan",
         "atat_role_name": "default",
         "email": "brandon@example.com",
-        "service_branch": "Fake Service Branch",
+        "service_branch": random_service_branch(),
         "phone_number": "1234567890",
         "citizenship": "United States",
         "designation": "Military",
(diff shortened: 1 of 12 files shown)


atst/routes/dev.py
```diff
@@ -21,41 +22,71 @@
...
+        "designation": "Military",
+        "date_latest_training": pendulum.date(2018, 1, 1),
     },
     "dominick": {
         "dod_id": "5678901234",
         "first_name": "Dominick",
         "last_name": "Domingo",
         "atat_role_name": "default",
         "email": "dominick@example.com",
+        "service_branch": "Fake Service Branch",
```
patricksmithdds: I think we could run into problems with this in the future -- could we use one of the real services branches? Perhaps, [choosing a random one](https://github.com/dod-ccpo/atst/blob/master/tests/factories.py#L28)?

atst/routes/workspaces.py
```diff
@@ -372,3 +373,9 @@ def revoke_invitation(workspace_id, token):
     Invitations.revoke(token)
 
     return redirect(url_for("workspaces.workspace_members", workspace_id=workspace.id))
+
+
+@bp.route("/workspaces/<workspace_id>/invitations/<token>/resend", methods=["POST"])
+def resend_invitation(workspace_id, token):
+    Invitations.resend(g.current_user, workspace_id, token)
```
patricksmithdds: The route for `create_member` explicitly send the invite email but this does not. I believe this wouldn't actually resend the invitation without calling `send_invite_email`.

templates/workspaces/members/edit.html
```diff
@@ -39,13 +39,23 @@ <h1 class='member-card__heading'>{{ member.user.full_name }}</h1>
...
               url_for("workspaces.revoke_invitation", workspace_id=workspace.id, token=member.latest_invitation.token),
               form.csrf_token
            ) }}
       {% endif %}
+      {% if member.can_resend_invitation %}
+        {{ ConfirmationButton (
+              "Resend Invitation",
+              url_for("workspaces.resend_invitation", workspace_id=workspace.id, token=member.latest_invitation.token),
+              form.csrf_token,
+              confirm_msg="Are you sure? This will invalidate the previously sent invitation."
```
patricksmithdds: Seems a bit odd to tell the user that it will invalidate the previously sent invitation, since the reason they're re-sending the invite is because the previously sent invitation is invalid. Is there some other text we can put here?

atst/routes/workspaces.py
```diff
@@ -372,3 +373,9 @@ def revoke_invitation(workspace_id, token):
     Invitations.revoke(token)
 
     return redirect(url_for("workspaces.workspace_members", workspace_id=workspace.id))
+
+
+@bp.route("/workspaces/<workspace_id>/invitations/<token>/resend", methods=["POST"])
+def resend_invitation(workspace_id, token):
```
patricksmithdds: Can you add a test for this, similar to `create_member`, that hits the endpoint and ensures that an email was queued?

templates/workspaces/members/edit.html
richard-dds: Yeah, it is a little confusing. Do you have any suggestions for copy? Is "Are you sure?" sufficient?

PR #441, opened 2018-11-09
Fix PDF clearing bug

richard-dds:
Steps to reproduce (thanks @leigh-mil):

1. Fill out a fin ver and save it as a draft
2. reopen the fin ver and remove the PDF and upload a new one
3. submit the fin ver form
4. login as CCPO and request changes to fin ver
5. log back in as the MO and reopen the fin ver form

When you open the form you will see that the original PDF that was uploaded is attached to the fin ver form!

### DIFF
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -4,6 +4,7 @@ from wtforms.fields.html5 import DateField, EmailField
 from wtforms.fields import StringField, FileField, FormField
 from wtforms.validators import InputRequired, Email, Regexp, Optional
 from flask_wtf.file import FileAllowed
+from werkzeug.datastructures import FileStorage
 
 from .fields import NewlineListField, SelectField, NumberStringField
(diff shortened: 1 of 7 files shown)


atst/forms/financial.py
```diff
@@ -237,6 +238,10 @@ def pe_id(self):
     def task_order_number(self):
         return self.task_order.number
 
+    @property
+    def has_task_order_pdf(self):
+        return isinstance(self.task_order.pdf.data, FileStorage)
```
patricksmithdds: Will this have the same problem that it may be an `Attachment` in addition to `FileStorage` -- like we saw in https://github.com/dod-ccpo/atst/pull/429/commits/9a9a6ecc9441d9bcdb0fb17d7b9757dfc5001e2a?

richard-dds: We're only calling this property to determine if the user has added an PDF to the form, in which case it would always be an instance of `FileStorage`.

PR #445, opened 2018-11-12
Filter by all member statuses

patricksmithdds:
Previously, you could only select a few of the possible member statuses (active, pending, denied), but now you can filter by any valid statuses:

image


luisgov:
@patricksmithdds not sure what happened with the alignment of the filters and the search input.. looks a bit funky.

patricksmithdds:
@luisgov Agreed, but it looks that way in master currently:

image

I'll take a look at fixing in this branch, though.

patricksmithdds:
@luisgov That alignment issue should be fixed now:

image

PR #446, opened 2018-11-13
Add ws role changes to audit log

montana-mil:
## Description
Adds more details to our audit log events and prepares these events to accept different details in the future. The audit log events for a change in a user's workspace role now reads:
> update workspace_role workspace_role id
> for User User id (User name) 
> from name of previous role to name of new role>
> in Workspace Workspace id (Workspace name)

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/161790452

## Screenshots
screen shot 2018-11-13 at 4 45 54 pm


templates/audit_log.html
```diff
@@ -27,6 +27,16 @@ <h2 class='h4 audit-log__item__name'>
 
               <br>
 
+              {% if event.event_details %}
+                for User <code>{{ event.event_details.updated_user_id }}</code> ({{ event.event_details.updated_user }})
+                <br>
+              {% endif %}
+
+              {% if event.changed_state %}
+                from {{ event.changed_state.role[0] }} to {{ event.changed_state.role[1] }}
```
patricksmithdds: This may not need to be changed yet -- but, in the future, `changed_state` will have attributes other than "role". Do you have an idea of how we could handle that?

tests/models/test_workspace_role.py
```diff
@@ -14,6 +15,50 @@
...
 
+def test_has_no_history(session):
+    owner = UserFactory.create()
+    user = UserFactory.create()
+
+    workspace = Workspaces.create(RequestFactory.create(creator=owner))
+    workspace_role = WorkspaceRoles.add(user, workspace.id, "developer")
+    audit_events = (
+        session.query(AuditEvent)
+        .filter(AuditEvent.resource_id == workspace_role.id)
```
patricksmithdds: You should be able to filter by action here as well:

```suggestion
        .filter(AuditEvent.resource_id == workspace_role.id, AuditEvent.action == "create")
```

Then, you won't need to do the list comprehension afterwards.

tests/models/test_workspace_role.py
```diff
@@ -14,6 +15,50 @@
...
+    workspace_role = WorkspaceRoles.add(user, workspace.id, "developer")
+    WorkspaceRoles.update_role(workspace_role, "admin")
+    audit_events = (
+        session.query(AuditEvent)
+        .filter(AuditEvent.resource_id == workspace_role.id)
+        .all()
+    )
+    changed_events = [event for event in audit_events if event.changed_state]
+
+    assert changed_events[0].changed_state["role"]
```
patricksmithdds: I think it would also be useful to assert that the value of the role is a sequence of both the "from" and "to" values. Another test that exercises the status history would not go unappreciated either :)

templates/audit_log.html
montana-mil: we were thinking of just handling that with more if statements until we know more concretely how we will be handling the audit log events. did you have an idea?

PR #448, opened 2018-11-16
Save user's form state for later when their session expires before submission

richard-dds:
https://www.pivotaltracker.com/story/show/161820247

### DIFF
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -97,8 +97,10 @@ def _make_authentication_context():
 def redirect_after_login_url():
     if request.args.get("next"):
         returl = request.args.get("next")
-        if request.args.get("formCache"):
-            returl += "?" + url.urlencode({"formCache": request.args.get("formCache")})
+        if request.args.get(app.form_cache.PARAM_NAME):
+            returl += "?" + url.urlencode(
+                {app.form_cache.PARAM_NAME: request.args.get(app.form_cache.PARAM_NAME)}
+            )
         return returl
     else:
         return url_for("atst.home")
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -96,7 +96,10 @@ def _make_authentication_context():
 
 def redirect_after_login_url():
     if request.args.get("next"):
-        return request.args.get("next")
+        returl = request.args.get("next")
+        if request.args.get("formCache"):
+            returl += "?" + url.urlencode({"formCache": request.args.get("formCache")})
+        return returl
     else:
         return url_for("atst.home")
 
--- a/atst/routes/requests/financial_verification.py
+++ b/atst/routes/requests/financial_verification.py
@@ -185,6 +185,7 @@ class SaveFinancialVerificationDraft(FinancialVerificationBase):
         return updated_request
 
 
+@requests_bp.route("/requests/verify/<string:request_id>/draft", methods=["GET"])
 @requests_bp.route("/requests/verify/<string:request_id>", methods=["GET"])
 def financial_verification(request_id):
     request = Requests.get(g.current_user, request_id)
--- a/atst/routes/requests/financial_verification.py
+++ b/atst/routes/requests/financial_verification.py
@@ -92,14 +92,13 @@ class FinancialVerificationBase(object):
 
 
 class GetFinancialVerificationForm(FinancialVerificationBase):
-    def __init__(self, user, request, cached_data=None, is_extended=False):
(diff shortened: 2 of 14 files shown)


atst/routes/requests/requests_form.py
```diff
@@ -29,7 +36,10 @@ def option_data():
 
 @requests_bp.route("/requests/new/<int:screen>", methods=["GET"])
 def requests_form_new(screen):
-    jedi_flow = JEDIRequestFlow(screen, request=None, current_user=g.current_user)
+    cached_data = current_app.form_cache.from_request(http_request)
```
patricksmithdds: I tested this out on the first screen of the new request flow with a couple fields filled out (DoD components, JEDI usage, and start date) and got an error: `AttributeError: 'dict' object has no attribute 'getlist'`. The traceback isn't very helpful, but it occurred when creating the form with the cached data. Here's what was cached:

```
127.0.0.1:6379> get formcache:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
"{\"csrf_token\": \"ImVmOGYwNzJjOWY0NmQyNzY1NmUxYzNkZTQ1YzI5N2MxZTY5ZDQxMzQi.DtRkhg.M7sfK5gqktdPhFg0sUuluHUefrA\", \"dod_component\": \"Defense Advanced Research Projects Agency\", \"jedi_usage\": \"this is my jedi usage.\", \"num_software_systems\": \"\", \"estimated_monthly_spend\": \"\", \"dollar_value\": \"\", \"start_date\": \"12/20/2018\", \"name\": \"\"}"
```

atst/utils/form_cache.py
```diff
@@ -0,0 +1,31 @@
...
+class FormCache(object):
+    def __init__(self, redis):
+        self.redis = redis
+
+    def from_request(self, http_request):
+        cache_key = http_request.args.get("formCache")
+        if cache_key:
+            return self.read(cache_key)
+
+    def write(self, formdata, expiry_seconds=3600, key_prefix=DEFAULT_CACHE_NAME):
```
patricksmithdds: I like that it'll expire the data -- I assume that'll purge it from redis after the expired time passes?

atst/routes/__init__.py
```diff
@@ -96,7 +96,10 @@ def _make_authentication_context():
 
 def redirect_after_login_url():
     if request.args.get("next"):
-        return request.args.get("next")
+        returl = request.args.get("next")
+        if request.args.get("formCache"):
```
patricksmithdds: Can we extract `"formCache"` into a setting on the cache, or some constant somewhere, since it's repeated a couple times in different places?

atst/routes/requests/financial_verification.py
```diff
@@ -92,13 +92,15 @@ def _raise(self, form):
...
-    def __init__(self, user, request, is_extended=False):
+    def __init__(self, user, request, cached_data=None, is_extended=False):
         self.user = user
         self.request = request
+        self.cached_data = cached_data or {}
         self.is_extended = is_extended
 
     def execute(self):
-        form = self._get_form(self.request, self.is_extended)
+        form = self._get_form(self.request, self.is_extended, formdata=self.cached_data)
```
patricksmithdds: This comment may be out of scope, but curious what you think about the following. I initially did not see that the Pivotal issue was limited to the request flow and thought that this would apply to _any_ form on the site. That may be a bit unrealistic, but I'm curious if we could make it easier for forms utilize the cache data by creating some custom `Form` class rather than passing in the data explicitly. I was thinking something like:

```python

from flask import current_app, request as http_request


class CachedForm(ValidatedForm):

    def __init__(self, form_data=None, **kwargs):
        cached_data = current_app.form_cache.from_request(http_request)
        # merge cached_data with incoming form data
        cached_data.update(form_data)
        super(self).__init__(form_data=cached_data, **kwargs)
```

I haven't tested the code, but I was hoping that any form we want to take advantage of the cache could then just extend the `CachedForm` class without any further modifications.

atst/utils/form_cache.py
richard-dds: Yep, redis will automatically expire and delete the key after `expiry_seconds`.

https://redis.io/commands/setex

atst/routes/requests/financial_verification.py
richard-dds: Done! I had a very similar idea in mind, but had decided to push it off for another time. Now that you mentioned it here I realized that it's actually less work than updating each form instantion separately.

PR #449, opened 2018-11-19
Remove workspace access

richard-dds:
https://www.pivotaltracker.com/story/show/161795061

Allow a workspace owner to revoke a user's workspace role by clicking a button on the member edit page. A workspace owner's role cannot be revoked.

screen shot 2018-11-26 at 1 28 08 pm

### DIFF
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -88,7 +88,10 @@ class Environments(object):
 
     @classmethod
     def revoke_access(cls, user, environment, target_user):
-        Authorization.check_atat_permission(
-            user, Permissions.REMOVE_CSP_ROLES, "revoke environment access"
+        Authorization.check_workspace_permission(
+            user,
+            environment.workspace,
+            Permissions.REMOVE_CSP_ROLES,
+            "revoke environment access",
         )
         EnvironmentRoles.delete(environment.id, target_user.id)
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -85,3 +85,7 @@ class Environments(object):
                 db.session.add(env_role)
 
         db.session.commit()
+
+    @classmethod
+    def revoke_access(cls, user, environment, target_user):
+        pass
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -88,4 +88,7 @@ class Environments(object):
 
     @classmethod
     def revoke_access(cls, user, environment, target_user):
-        pass
+        Authorization.check_atat_permission(
+            user, Permissions.REMOVE_CSP_ROLES, "revoke environment access"
+        )
+        EnvironmentRoles.delete(environment.id, target_user.id)
--- a/atst/domain/workspace_roles.py
+++ b/atst/domain/workspace_roles.py
@@ -36,16 +36,9 @@ class WorkspaceRoles(object):
     @classmethod
     def get_by_id(cls, id_):
         try:
-            workspace_role = (
-                db.session.query(WorkspaceRole)
-                .join(User)
-                .filter(WorkspaceRole.id == id_)
-                .one()
-            )
+            return db.session.query(WorkspaceRole).filter(WorkspaceRole.id == id_).one()
         except NoResultFound:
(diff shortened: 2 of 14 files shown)


atst/domain/environments.py
```diff
@@ -85,3 +85,7 @@ def update_environment_roles(cls, user, workspace, workspace_role, ids_and_roles
                 db.session.add(env_role)
 
         db.session.commit()
+
+    @classmethod
+    def revoke_access(cls, user, environment, target_user):
+        pass
```
patricksmithdds: Should this actually be doing something?

atst/domain/workspace_roles.py
```diff
@@ -34,6 +34,20 @@ def get(cls, workspace_id, user_id):
...
         return workspace_role
 
+    @classmethod
+    def get_by_id(cls, id_):
+        try:
+            workspace_role = (
+                db.session.query(WorkspaceRole)
+                .join(User)
+                .filter(WorkspaceRole.id == id_)
+                .one()
```
patricksmithdds: Just curious -- could we use [`get`](https://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.get) instead of `find().one()`?

atst/domain/workspaces/workspaces.py
```diff
@@ -138,3 +143,29 @@ def update(cls, workspace, new_data):
             workspace.name = new_data["name"]
 
         WorkspacesQuery.add_and_commit(workspace)
+
+    @classmethod
+    def can_revoke_access(cls, workspace, workspace_role):
```
patricksmithdds: Should this also do the `Authorization.check_workspace_permission` check? It seems like both of them would factor into "can revoke access"

atst/domain/workspaces/workspaces.py
```diff
@@ -138,3 +143,29 @@ def update(cls, workspace, new_data):
...
+            "revoke workspace access",
+        )
+        workspace_role = WorkspaceRoles.get_by_id(workspace_role_id)
+
+        if not Workspaces.can_revoke_access(workspace, workspace_role):
+            raise WorkspaceError()
+
+        workspace_role.status = WorkspaceRoleStatus.DISABLED
+        for environment in workspace.all_environments:
+            # TODO: Implement Environments.revoke_access
```
patricksmithdds: Can this TODO be tackled in this branch? If not, is there a Pivotal story to track it?

atst/domain/workspaces/workspaces.py
```diff
@@ -138,3 +143,29 @@ def update(cls, workspace, new_data):
...
+        Authorization.check_workspace_permission(
+            user,
+            workspace,
+            Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
+            "revoke workspace access",
+        )
+        workspace_role = WorkspaceRoles.get_by_id(workspace_role_id)
+
+        if not Workspaces.can_revoke_access(workspace, workspace_role):
+            raise WorkspaceError()
```
patricksmithdds: Can we put a message in this error so we'll be able to track it down in the future?

atst/domain/environments.py
richard-dds: In the future it should be removing (or deactivating) an environment role, taking whatever action is necessary in the CSP. Right now I'm not sure if there's anything useful to do, which is why I made this a placeholder.

patricksmithdds:
@richard-dds Looks like there's some conflicts with recent changes to master 😱

atst/models/workspace_role.py
```diff
@@ -83,6 +84,8 @@ def latest_invitation(self):
     def display_status(self):
         if self.status == Status.ACTIVE:
             return MEMBER_STATUSES["active"]
+        elif self.status == Status.DISABLED:
```
patricksmithdds: Looks like we need to add `DISABLED` to the `Status` enum above.

patricksmithdds: it's there already! reading is hard 😵

PR #450, opened 2018-11-19
Send invite to email provided ##161961059

leigh-mil:
## Description
This PR fixes a bug where if a workspace admin/owner invited an existing AT-AT user to a workspace, the invite email was sent to the email in the existing user's profile. 
Now the invite email is sent to the email that the inviter inputs into the add new member form. 

I need a little help with the alembic migration. I wanted to make the email field required, so I had to do two queries to add values email fields of the existing invitations. The first adds the invitee's email to the email column, but not all invitees have emails so I had to add in the second query. 
Any suggestions on making this better? 

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/161961059](https://www.pivotaltracker.com/story/show/161961059)


tests/routes/test_workspaces.py
```diff
@@ -478,3 +502,31 @@ def test_resend_invitation_sends_email(client, user_session, queue):
...
+        url_for(
+            "workspaces.resend_invitation",
+            workspace_id=workspace.id,
+            token=invite.token,
+        )
+    )
+
+    assert user.email != "example@example.com"
+    assert len(queue.get_queue().jobs[0].args[0]) == 1
+    assert queue.get_queue().jobs[0].args[0][0] == "example@example.com"
```
patricksmithdds: I think this test is testing the right thing, but could we make it's intent a bit clearer? I'm thinking something like:

```
send_mail_job = queue.get_queue().jobs[0]
assert send_mail_job.func == _send_mail
assert send_mail_job.args[0] == ["example@example.com"]
```

Renaming the `queue.get_queue().jobs[0]` makes it clear what we expect the job to be and the second assertion enforces both of the existing assertion (that there's one item in the list and it is "example@example.com")

alembic/versions/02d11579a581_add_email_to_invite.py
```diff
@@ -0,0 +1,50 @@
...
+                where i.email is null;
+            """
+        )
+    )
+    conn.execute(
+        text(
+            """
+                update invitations
+                set email = 'example@example.com'
+                where email is null;
```
patricksmithdds: Is this query still necessary? It seems the first one would set the invitation's email to the workspace user's email, so we wouldn't have an `NULL` emails by this point.

leigh-mil: I was having issues with this locally when I tried to run the migration because I had users that didn't have completed profiles. Do you think we will have that issue on staging?

patricksmithdds: Ah, that makes sense -- I wasn't thinking about incomplete user profiles.

alembic/versions/02d11579a581_add_email_to_invite.py
```diff
@@ -0,0 +1,50 @@
+"""Add email to invite
+
+Revision ID: 02d11579a581
+Revises: 4c0b8263d800
```
patricksmithdds: I think this will conflict with another recent migration added, so we'll need to update that migration path before merging.

PR #454, opened 2018-11-27
Remove Styleguide From Navigation

montana-mil:
## Description
The Styleguide displayed on the side navigation was confusing to test users, so we removed it.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/161597085

## Screenshots
screen shot 2018-11-27 at 9 24 06 am

### DIFF
--- a/templates/navigation/global_navigation.html
+++ b/templates/navigation/global_navigation.html
@@ -2,17 +2,6 @@
 
 <div class="global-navigation sidenav {% if workspace %}global-navigation__context--workspace{% endif %}">
   <ul>
-    {% if g.dev %}
-      {{ SidenavItem("Styleguide",
-        href="/styleguide",
-        icon="visible",
-        active=g.matchesPath('/styleguide'),
-        subnav=[
-          {"label":"Subnav 1", "href":"/styleguide?subnav1", "icon": "plus", "active": g.matchesPath('/styleguide?subnav1')},
-          {"label":"Subnav 2", "href":"/styleguide?subnav2", "active": g.matchesPath('/styleguide?subnav2')},
-        ]) }}
-    {% endif %}
-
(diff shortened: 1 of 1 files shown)


montana-mil:
we need a test server before we can do this. we still want the stylguide on dev, just not on the testing site.

patricksmithdds:
> we need a test server before we can do this. we still want the styleguide on dev, just not on the testing site.

I don't think we need to wait for a test server -- we have the UAT server set up and we should ensure that the styleguide link is not shown there.

templates/navigation/global_navigation.html
```diff
@@ -2,17 +2,6 @@
 
 <div class="global-navigation sidenav {% if workspace %}global-navigation__context--workspace{% endif %}">
   <ul>
-    {% if g.dev %}
```
patricksmithdds: I think instead of removing this entirely, we could set an environment variable that could let us feature flag this by environment. We can add a new env variable thats [used in the app configuration](https://github.com/dod-ccpo/atst/blob/master/atst/app.py#L94) and set the env variable to hide the link in [the uat environment only](https://github.com/dod-ccpo/atst/blob/master/deploy/kubernetes/uat/atst-envvars-configmap.yml).

patricksmithdds: > Isn't `g.dev` supposed to be that env variable?

That's true, but [we've set `FLASK_ENV=dev`](https://github.com/dod-ccpo/atst/blob/master/deploy/kubernetes/atst-envvars-configmap.yml#L8) in the deployed environment, which blurs the distinction a bit.

> I think it would be sufficient to remove the link entirely and we just know to go to /styleguide, if that makes this simpler.

That does make things simpler, and I'm OK with skipping the link altogether if you are.

PR #455, opened 2018-11-27
Refactor workspace routes

montana-mil:
## Description
The route file for `workspaces` was becoming crowded, so I broke it up into several files in a `routes/workspaces` directory.


tests/routes/test_workspaces_projects.py
```diff
@@ -0,0 +1,127 @@
+from flask import url_for
```
patricksmithdds: Could we move these test files to a `workspaces` directory to match the code structure and make the tests easier to find?

`tests/routes/test_workspaces_projects.py` => `tests/routes/workspaces/test_projects.py`

PR #456, opened 2018-11-27
Fix revoke/resend invitations in IE

patricksmithdds:
In IE, the revoke or resend invitation buttons were previously not working correctly. The buttons work by submitting a form to the appropriate URL. However, this form is nested inside another form (the edit member form), which is not valid HTML. 

The problem is actually a bit more subtle: when the popover element is rendered (when the relevant button is clicked), the displayed popover is added to the end of the DOM. This is outside the enclosing `form` element, so everything works as expected in Chrome/Firefox. IE, however, will strip the inner `form` element before the tooltip is shown. When the tooltip is shown, the "revoke" or "resend" buttons are not enclosed in any form, so they submit nothing.

The solution here is to move the inner `form` into a Vue template -- now, the form is never enclosed in the outer `form`, so no browsers will strip the `form` tag.

### DIFF
--- a/js/components/__tests__/__snapshots__/confirmation_popover.test.js.snap
+++ b/js/components/__tests__/__snapshots__/confirmation_popover.test.js.snap
@@ -1,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ConfirmationPopover matches snapshot 1`] = `<v-popover placement="top-start"><template></template> <button type="button" class="tooltip-target">Do something dangerous</button></v-popover>`;
--- a/js/components/__tests__/__snapshots__/confirmation_popover.test.js.snap
+++ b/js/components/__tests__/__snapshots__/confirmation_popover.test.js.snap
@@ -1,3 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ConfirmationPopover matches snapshot 1`] = `<v-popover placement="top-start"><template></template> <button type="button" class="tooltip-target">Do something dangerous</button></v-popover>`;
+exports[`ConfirmationPopover matches snapshot 1`] = `<v-popover-stub placement="top-start" delay="0" offset="0" trigger="click" container="body" popperoptions="[object Object]" popoverclass="vue-popover-theme" popoverbaseclass="tooltip popover" popoverinnerclass="tooltip-inner popover-inner" popoverwrapperclass="wrapper" popoverarrowclass="tooltip-arrow popover-arrow" autohide="true" handleresize="true"><template></template> <button type="button" class="tooltip-target">Do something dangerous</button></v-popover-stub>`;
--- a/js/components/__tests__/confirmation_popover.test.js
+++ b/js/components/__tests__/confirmation_popover.test.js
@@ -1,11 +1,15 @@
-import { shallowMount } from '@vue/test-utils'
+import { createLocalVue, shallowMount } from '@vue/test-utils'
+import VTooltip from 'v-tooltip'
 
 import ConfirmationPopover from '../confirmation_popover'
 
+const localVue = createLocalVue()
+localVue.use(VTooltip)
 
 describe('ConfirmationPopover', () => {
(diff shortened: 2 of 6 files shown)


montana-mil:
so even though there's still an inner `form`, it doesn't get stripped by IE because it's in a Vue template?

montana-mil:
oops, clicked the wrong comment button

patricksmithdds:
> so even though there's still an inner `form`, it doesn't get stripped by IE because it's in a Vue template?

Right -- in the previous iteration, the popover template was included inline. This would result in HTML that looks like:

```html
<form method="POST" action="/workspaces/a04403cd-f327-4b4e-a401-2e29f2dd9dee/members/b98a8bbc-37fe-46e7-b003-fbf09d9c475f/member_edit" autocomplete="false">
  <input id="csrf_token" name="csrf_token" type="hidden" value="ImUyMWU4NTM0YmZkMzQzNDUxZjI0ZjExN2ZmNjhmZDJmOTQzMWRjNjMi.DuBSpQ.KaEGvWDXG5Su-GBtk52mYeS4nvA">
  <div class='panel member-card'>
    <div class='member-card__details'>
      <v-popover placement='top-start'>
        <template slot="popover">
          <p>Are you sure?</p>
          <div class='action-group'>
            <!-- this inner form is getting dropped from the HTML by IE -->
            <form method="POST" action="/workspaces/a04403cd-f327-4b4e-a401-2e29f2dd9dee/invitations/IXcmI7CeAJoW0xCX6sgC8rRFpviIU2KXGhS7mzhBjwE/revoke">
              <input id="csrf_token" name="csrf_token" type="hidden" value="ImUyMWU4NTM0YmZkMzQzNDUxZjI0ZjExN2ZmNjhmZDJmOTQzMWRjNjMi.DuBSpQ.KaEGvWDXG5Su-GBtk52mYeS4nvA">
              <button class='usa-button usa-button-primary' type='submit'>
                Confirm
              </button>
            </form>
            <button class='usa-button usa-button-secondary' v-close-popover>
              Cancel
            </button>
          </div>
        </template>
        <button class="tooltip-target" type="button">Revoke Invitation</button>
      </v-popover>
    </div>
  </div>
</form>
```

That `<template slot="popover">` part means that the form isn't actually displayed however. When the popover is toggled, that template is used to add an element to the end of the DOM, which is the actual form that gets submitted. I believe IE is stripping out the `form` when the HTML is initially received, so the template that gets rendered is missing the crucial `form` element.

With this branch, the resulting HTML looks a bit different:

```html
<form method="POST" action="/workspaces/a04403cd-f327-4b4e-a401-2e29f2dd9dee/members/b98a8bbc-37fe-46e7-b003-fbf09d9c475f/member_edit" autocomplete="false">
  <input id="csrf_token" name="csrf_token" type="hidden" value="ImUyMWU4NTM0YmZkMzQzNDUxZjI0ZjExN2ZmNjhmZDJmOTQzMWRjNjMi.DuBS9g.wBi_oQzz4A-_z7c5upeSr6GG6oo">

  <div class='panel member-card'>
    <div class='member-card__details'>
      <div>
        <!-- the `form` element never gets rendered here, and is only added to the DOM when the popover is opened --> 
        <confirmation-popover
          btn_text='Revoke Invitation'
          action='/workspaces/a04403cd-f327-4b4e-a401-2e29f2dd9dee/invitations/IXcmI7CeAJoW0xCX6sgC8rRFpviIU2KXGhS7mzhBjwE/revoke'
          csrf_token='ImUyMWU4NTM0YmZkMzQzNDUxZjI0ZjExN2ZmNjhmZDJmOTQzMWRjNjMi.DuBS9g.wBi_oQzz4A-_z7c5upeSr6GG6oo'
          confirm_msg='Are you sure?'
          confirm_btn_text='Confirm'
          cancel_btn_text='Cancel'>
        </confirmation-popover>
      </div>
    </div>
  </div>
</form>
```

The inner `form` is in a template string in the `confirmation-popover` component, so nothing gets stripped out. Similar to before, the template is used when the popover is added, however, now the template correctly has the `form` and everything works as expected.

PR #458, opened 2018-11-27
Remove option to revoke from expired workspace invitation

leigh-mil:
## Description
Prior to this PR, the Revoke Invitation button would appear if an invite is in the pending state. Since the state of the invite never changes when it expires, the button was still showing for expired invites.
This issue was solved by adding the property is_revokable to Invitation and checking for that in the edit member template instead of whether or not the invite is pending. 

## Pivotal
[https://www.pivotaltracker.com/story/show/161933129](https://www.pivotaltracker.com/story/show/161933129)


leigh-mil:
@patricksmithdds tests have been added!

PR #459, opened 2018-11-27
Show sort indicators when page loads

patricksmithdds:
Addresses https://www.pivotaltracker.com/story/show/161753062

Previously, pages with sortable tables (workspace members and request listing for CCPO users) did not show any indication that the tables were sortable. This branch fixes that by setting a default sort order on the tables. For member listings, the default sort order is the member name:

image

and for request listing, the submission date:

image

In addition to the default sort order, I've added a couple Vue tests for each of the table components. These tests introduce a couple of new ideas:

* the components are used with the `inline-template`. That doesn't quite work with `vue-test-utils` (at least, I couldn't figure it out), so I've added a `template` option to each component that renders an empty div. In the application, this template string is always overridden by the inline template, so this setting is just to appease the tests.
* another implication of using inline templates is that the rendering is handled largely outside of the component. As such, the snapshot testing doesn't apply and we interact directly with the mounted component using `wrapper.vm`.


js/components/__tests__/members_list.test.js
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should reverse sort by name when updated with Name', () => {
+    const wrapper = createWrapper()
+    wrapper.vm.updateSort('Name')
+    const listedMembers = wrapper.vm.searchedList
+    const memberNames = listedMembers.map(member => member.name)
+    expect(memberNames).toEqual(['Luke Skywalker', 'Chewie'])
+  })
+
+  it('should sort by number of environments when environments selected', () => {
```
JRNPW: This is an interesting test. It's clear that we've decided to include a sort on number of environments (which feels odd to me from a UX perspective), but the tested conditions are the sequence of the members. Is there any way to test the order of the members' environment numbers?

js/components/__tests__/requests_list.test.js
```diff
@@ -0,0 +1,59 @@
+import { shallowMount } from '@vue/test-utils'
+
+import RequestsList from '../requests_list'
+
+describe('RequestsList', () => {
+
+  describe('isExtended', () => {
+    it('should disallow sorting if not extended', () => {
```
JRNPW: Is there a need to test this state/lack-of-behavior? Is sorting disallowed "automatically" by not being extended, or must we explicitly test it like this?

patricksmithdds: In this case, I think we do need to test that the component disallows the action, since the component itself has branching logic if the form is not extended.

js/components/__tests__/members_list.test.js
patricksmithdds: > Is there any way to test the order of the members' environment numbers?

If I understand correctly, you're suggesting that the assertion should be changed to something like:

```js
const memberEnvs = listedMembers.map(member => member.num_env)
expect(memberEnvs).toEqual([2, 3])
```

? I was using the names for the assertion, just to ensure that the ordering was correct, but I think it may not be clear from that that we're testing the correct thing.

JRNPW: Yes, I prefer testing against the memberEnvs to make it clear what we are testing and what the expected results are when sorting by it.

js/components/__tests__/requests_list.test.js
JRNPW: Cool. No issue.

JRNPW:
I ran the manual tests that I wanted to do. Ready for approval after above comment is resolved.

js/components/__tests__/members_list.test.js
patricksmithdds: This should be fixed by 30aa5f37.

PR #460, opened 2018-11-28
be more generous in finding PDF attachment file name when rendering f…

dandds:
…inancial verification form

This should fix https://www.pivotaltracker.com/n/projects/2160940/stories/162270400.

The problem we're seeing right now is that multiple requests reference the same task order entity (i.e., the mock "valid" one). We look up what PDF filename to display based on the request ID saved as part of the attachment record, so only one of the many requests referencing the valid TO can get the PDF filename. I fixed it here by being more generous in how we get the filename, but this is just a band-aid. We need TO numbers to be unique to the request.

### DIFF
--- a/atst/routes/requests/financial_verification.py
+++ b/atst/routes/requests/financial_verification.py
@@ -32,11 +32,16 @@ class FinancialVerificationBase(object):
         form = FinancialVerificationForm(obj=fv, formdata=_formdata)
 
         if not form.has_task_order_pdf:
-            try:
-                attachment = Attachment.get_for_resource("task_order", self.request.id)
-                form.task_order.pdf.data = attachment.filename
-            except NotFoundError:
-                pass
+            if isinstance(form.task_order.pdf.data, Attachment):
+                form.task_order.pdf.data = form.task_order.pdf.data.filename
+            else:
+                try:
+                    attachment = Attachment.get_for_resource(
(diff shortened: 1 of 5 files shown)


atst/routes/requests/financial_verification.py
```diff
@@ -32,11 +32,16 @@ def _get_form(self, request, is_extended, formdata=None):
         form = FinancialVerificationForm(obj=fv, formdata=_formdata)
 
         if not form.has_task_order_pdf:
-            try:
-                attachment = Attachment.get_for_resource("task_order", self.request.id)
-                form.task_order.pdf.data = attachment.filename
-            except NotFoundError:
-                pass
+            if isinstance(form.task_order.pdf.data, Attachment):
```
patricksmithdds: I'm curious if we can reword something here -- the two `if` statements seem contradictory:

```
if not form.has_task_order_pdf:
    if isinstance(form.task_order.pdf.data, Attachment):
```

It seems odd that the first thing we do if the form has no task order pdf is see if the task order pdf is an attachment.

dandds: Yeah, I think `has_task_order_pdf` is not quite the right name. We really check if it's a file upload. I'll fix.

dandds:
@patricksmithdds Had to rebase. Can I get another look when you have a minute?

PR #469, opened 2018-11-30
do not show workspace role update alert when only environment role ha…

dandds:
…s changed

https://www.pivotaltracker.com/story/show/162345640


patricksmithdds:
I think the code here looks fine, but I'd think we'd still want to show _something_ after a successful update.

dandds:
I think I'm going to close this and re-submit after we've reworked things to use `flash`. I don't want to add more url query params that will get deleted shortly after.

PR #470, opened 2018-12-03
Add validators to fin ver form

leigh-mil:
## Description
Adds validators to required form fields in the Financial Verification form so the validation throughout the form is consistent.

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/162289648](https://www.pivotaltracker.com/story/show/162289648)

### DIFF
--- a/styles/elements/_inputs.scss
+++ b/styles/elements/_inputs.scss
@@ -199,6 +199,7 @@
 
     &--anything,
     &--workspaceName,
+    &--requiredField,
     &--email {
       input {
         max-width: 30em;
--- a/templates/requests/financial_verification.html
+++ b/templates/requests/financial_verification.html
@@ -133,7 +133,7 @@
           f.task_order.number,
           placeholder="e.g.: 1234567899C0001",
           tooltip="A Contracting Officer will likely be the best source for this number.",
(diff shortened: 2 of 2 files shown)


js/lib/input_validations.js
```diff
@@ -5,7 +5,7 @@ import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrect
 export default {
   anything: {
     mask: false,
-    match: /\s*/,
+    match: /\S+/,
```
dandds: Curious why this needed to change?

leigh-mil: I may be wrong, but I read that `\s*` meant 0 or more occurrences of a white space character, and `\S+` means at least one of any non-white space character. 
Did you have something different in mind for this?

dandds: You're right about the regex, but I interpret "anything" to mean "blank is fine". Are you making the change to address the task order number on the fin ver form? I think that one should probably just use your new `inputRequired` validator.

leigh-mil: Gotcha! Interestingly, the only form field that uses the anything validator is the TO Number

PR #472, opened 2018-12-03
Update DOD ID validator to allow leading zeros

leigh-mil:
## Description
Updated validator for DOD ID so it will now allow DOD IDs that begin with zero.

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/162345371](https://www.pivotaltracker.com/story/show/162345371)

### DIFF
--- a/js/lib/input_validations.js
+++ b/js/lib/input_validations.js
@@ -60,7 +60,7 @@ export default {
     validationError: 'Optional: Please enter up to 10 digits'
   },
   dodId: {
-    mask: createNumberMask({ prefix: '', allowDecimal: false, includeThousandsSeparator: false }),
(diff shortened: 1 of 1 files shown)


js/lib/input_validations.js
```diff
@@ -60,7 +60,7 @@ export default {
     validationError: 'Optional: Please enter up to 10 digits'
   },
   dodId: {
-    mask: createNumberMask({ prefix: '', allowDecimal: false, includeThousandsSeparator: false }),
+    mask: createNumberMask({ prefix: '', allowDecimal: false, allowLeadingZeroes: true, includeThousandsSeparator: false }),
```
dandds: 😂

PR #473, opened 2018-12-03
Better error handling for Auditable class

montana-mil:
## Description
Because the `getattr()` builtin relies on `AttributeError`s to determine if an attribute exists on an object, properties that exist but throw an `AttributeError` are treated the same as properties that don't exist. This has led to some hard to track-down bugs, some of which I think may still be in the codebase.
-- @richard-dds
I just fixed the test failures by adding a few `None` assignments when `NoneType` was trying to be accessed.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/162289419

### DIFF
--- a/atst/models/invitation.py
+++ b/atst/models/invitation.py
@@ -112,4 +112,4 @@ class Invitation(Base, TimestampsMixin, AuditableMixin):
 
     @property
     def workspace_id(self):
-        return self.workspace_role.workspace_id
+        return self.workspace_role.workspace_id if self.workspace_role else None
--- a/atst/models/invitation.py
+++ b/atst/models/invitation.py
@@ -112,4 +112,4 @@ class Invitation(Base, TimestampsMixin, AuditableMixin):
 
     @property
     def workspace_id(self):
-        return self.workspace_role.workspace_id if self.workspace_role else None
+        return self.workspace_role.workspace_id
--- a/tests/models/test_workspace_role.py
+++ b/tests/models/test_workspace_role.py
@@ -199,67 +199,62 @@ def test_status_when_member_is_active():
 def test_status_when_invitation_has_been_rejected_for_expirations():
     workspace = WorkspaceFactory.create()
     user = UserFactory.create()
-    ws_role = WorkspaceRoleFactory.create(
+    workspace_role = WorkspaceRoleFactory.create(
         workspace=workspace, user=user, status=WorkspaceRoleStatus.PENDING
     )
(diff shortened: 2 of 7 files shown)


atst/models/invitation.py
```diff
@@ -112,4 +112,4 @@ def history(self):
 
     @property
     def workspace_id(self):
-        return self.workspace_role.workspace_id
+        return self.workspace_role.workspace_id if self.workspace_role else None
```
patricksmithdds: Is there any case where it's possible that the workspace_role is `None`? If we're creating an invitation that has no workspace role, I think we'd want to change that rather than handling a non-existent workspace role here.

montana-mil:
@patricksmithdds the tests on `models/test_workspace_role.py` look a little bit circular now, but I think they are ok. Would appreciate more eyes on them though.

tests/models/test_workspace_role.py
```diff
@@ -196,44 +197,69 @@ def test_status_when_member_is_active():
...
-        invitations=[InvitationFactory.create(status=InvitationStatus.REJECTED_EXPIRED)]
+    workspace = WorkspaceFactory.create()
+    user = UserFactory.create()
+    ws_role = WorkspaceRoleFactory.create(
+        workspace=workspace, user=user, status=WorkspaceRoleStatus.PENDING
+    )
+    invitation = InvitationFactory.create(
+        workspace_role=ws_role, status=InvitationStatus.REJECTED_EXPIRED
     )
+    workspace_role = WorkspaceRoleFactory.create(invitations=[invitation])
```
patricksmithdds: It shouldn't be necessary to create a second workspace role from the invitation -- you should be able to use the first `ws_role` directly in the assertion.

tests/routes/workspaces/test_invitations.py
```diff
@@ -109,7 +112,9 @@ def test_user_accepts_invite_with_wrong_dod_id(client, user_session):
     ws_role = WorkspaceRoleFactory.create(
         user=user, workspace=workspace, status=WorkspaceRoleStatus.PENDING
     )
-    invite = InvitationFactory.create(user_id=user.id, workspace_role_id=ws_role.id)
+    invite = InvitationFactory.create(
+        user_id=user.id, workspace_role=ws_role, workspace_role_id=ws_role.id
```
patricksmithdds: It's not necessary to set both the `workspace_role` and `workspace_role_id` -- the relationship should handle setting `workspace_role_id` if you assign `workspace_role`.

PR #474, opened 2018-12-03
Set focus when no workspace role is selected

leigh-mil:
## Description
Fixes a bug where there was an error appearing in the console because no workspace role was selected, so it was throwing an error when setting the focus. I fixed this by adding in conditional logic so now if there is a role selected, that has focus, and if none is selected, the first option in the list has focus.

## Pivotal Tracker
[https://www.pivotaltracker.com/story/show/161560976](https://www.pivotaltracker.com/story/show/161560976)

## Screenshots
screen shot 2018-12-03 at 4 11 01 pm

_Look ma, no errors!_

### DIFF
--- a/js/components/selector.js
+++ b/js/components/selector.js
@@ -84,7 +84,11 @@ export default {
 
     onShow: function () {
       setTimeout(() => { // timeout is a hack to make focus work in Chrome
-        this.$refs.choices.find(choice => choice.selected).$refs.input[0].focus()
+        if (this.$refs.choices.find(choice => choice.selected)) {
+          this.$refs.choices.find(choice => choice.selected).$refs.input[0].focus()
+        } else {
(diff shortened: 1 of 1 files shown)


js/components/selector.js
```diff
@@ -84,7 +84,11 @@ export default {
 
     onShow: function () {
       setTimeout(() => { // timeout is a hack to make focus work in Chrome
-        this.$refs.choices.find(choice => choice.selected).$refs.input[0].focus()
+        if (this.$refs.choices.find(choice => choice.selected)) {
+          this.$refs.choices.find(choice => choice.selected).$refs.input[0].focus()
```
patricksmithdds: Can we pull out the `find` call to reduce repetition:

```suggestion
       const selected = this.$refs.choices.find(choice => choice.selected))
        if (selected) {
          selected.$refs.input[0].focus()
```

PR #475, opened 2018-12-03
Financial form unused property

JRNPW:
I'm working on the "easy" missing tests as enumerated in https://www.pivotaltracker.com/story/show/162280003 

I removed an unused property from forms/financial.py

I have run the unit tests, and I ran manual functional tests of the FinVer form (multiple iterations of adding, saving, etc.) and approvals by CCPO. No adverse effects resulted from this change.

### DIFF
--- a/atst/forms/financial.py
+++ b/atst/forms/financial.py
@@ -215,10 +215,6 @@ class FinancialVerificationForm(CacheableForm):
         return self.request.pe_id
 
     @property
-    def task_order_number(self):
-        return self.task_order.number
-
-    @property
     def has_pdf_upload(self):
         return isinstance(self.task_order.pdf.data, FileStorage)
 


montana-mil:
can we also take it out of `atst/models/request` line 93 now?

dandds:
@montana-mil You could look into it, but off the top of my head I think we're expecting it to be there in the request data.

JRNPW:
@dandds / @montana-mil I'm approaching these test coverage items very conservatively. I'm not going much beyond the file reporting missing statement or branch coverage.

montana-mil:
i can't really tell if it's being used at all. i have a feeling it's not, but I don't have a great grasp of that part of the code.

patricksmithdds:
> can we also take it out of `atst/models/request` line 93 now?

We need to keep that in there, since it's referencing the `task_order_number` attribute on the request revision: https://github.com/dod-ccpo/atst/blob/master/atst/models/request_revision.py#L69

PR #477, opened 2018-12-04
Fix arrow for sort by role on Members List 

montana-mil:
## Description
When sorting by `Workspace Role`, the arrow should display in-line with the title of the column. The arrow was spilling over to a second line before this PR.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/161753255

## Screenshots
screen shot 2018-12-04 at 10 34 58 am
screen shot 2018-12-04 at 10 35 06 am

### DIFF
--- a/js/components/members_list.js
+++ b/js/components/members_list.js
@@ -86,7 +86,8 @@ export default {
       {
         displayName: 'Workspace Role',
         attr: 'role',
-        sortFunc: alphabeticalSort
+        sortFunc: alphabeticalSort,
+        width: "20%"
       },
     ]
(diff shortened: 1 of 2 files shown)


patricksmithdds:
I believe this is dependent on the size of the window -- it still happens for me locally:

montana-mil:
😳

montana-mil:
Question: do we really need it to say "Workspace Role" here? Could we call it "Role" instead? Asking because that would also solve this problem

leigh-mil:
It works!

PR #479, opened 2018-12-05
Tests ignore object representation

JRNPW:
This PR is the initial work to improve statement coverage. After some conversation with @patricksmithdds, we decided that we could ignore test coverage of the `__repr__` methods which do not have coverage currently.

Pivotal card https://www.pivotaltracker.com/story/show/162482514


JRNPW:
FWIW: The overall coverage before this branch is 90.54%. It increases to 90.75% with these changes. As I said: incremental, but it's progress!

PR #480, opened 2018-12-06
Flash refactor

dandds:
This refactors our alert notifications for user to rely on Flask's built-in [`flash`](http://flask.pocoo.org/docs/1.0/patterns/flashing/). This should clean things up a bit.

**note**
I left some places alone for different reasons:
- Financial verification has complicated alerts in a couple different places on the page. Since it sounds like FV will go away in its current form, I didn't want to waste an hour on those.
- The request details page shows the user status information in an alert. It's complicated to disentangle these messages from the templates that are used to render modal alerts elsewhere, so I'm leaving them alone until it's clear what to do with these in the new workflow.
- There are a few places (cert selection on the login page) where we always render the same alert. Since there's no condition associated with these, I left them hard-coded in the templates.


dandds:
I might squash these commits before merging. They're kind of gross, but I wanted to be able to backtrack if I introduced regressions.

montana-mil:
this looks good to me. do we need any tests for flash things?

dandds:
@montana-mil Good q. Maybe one to make sure the `formatted_flash` function works, otherwise we should (I hope) have tests for specific notifications.

tests/utils/test_flash.py
```diff
@@ -0,0 +1,10 @@
+from atst.utils.flash import formatted_flash as flash
+from flask import get_flashed_messages
+
+
+def test_flash_message():
+    flash("revoked_workspace_access", member_name="Lando")
```
montana-mil: lol yes 🌟

atst/routes/workspaces/index.py
```diff
@@ -33,6 +34,7 @@ def edit_workspace(workspace_id):
             url_for("workspaces.workspace_projects", workspace_id=workspace.id)
         )
     else:
+        flash("form_errors")
```
patricksmithdds: This may go a step too far -- but what do you think about doing this in `ValidatedForm` for each form?

```python
class ValidatedForm(FlaskForm):
   # ... rest of class ...
   def validate(self, *args, **kwargs):
       valid = super().validate(*args, **kwargs)
       if not valid:
           flash("form_errors")
       return valid
```

templates/login.html
```diff
@@ -11,29 +11,14 @@
       <div class='col'>
 
         <div class='login-banner'>
-          {% if request.args.get("sessionExpired") %}
-            {{ Alert('Session Expired',
-                message='Your session expired due to inactivity. Please log in again to continue.',
-                level='error'
-            ) }}
-          {% endif %}
```
patricksmithdds: Is there an easy to flash a single message separate from the others? For example, I believe this session expired message is at the top of the page so that it's more easily seen by the user while the rest can go below the logo/buttons.

atst/utils/flash.py
```diff
@@ -0,0 +1,106 @@
...
+    },
+    "revoked_workspace_access": {
+        "title_template": "Removed workspace access",
+        "message_template": """
+          <p>Removed {{ member_name }} from this workspace.</p>
+        """,
+        "category": "success",
+    },
+    "resent_workspace_invitation": {
+        "title_template": "Invitation resent",
```
leigh-mil: NBD, but I first read this as resent (angry or upset), not that the invite was re-sent!
Is this worth fixing or do we expect the copy to change here?

dandds: Ha, interesting. I can change it. Who knows what will happen to the copy!

atst/routes/workspaces/index.py
dandds: Yeah, I like it.

templates/login.html
dandds: I'll have to think about it. In the Flask implementation, they are only organized by category type, making it hard to pluck one for individual display. You're right that it should stay at the top, though.

PR #486, opened 2018-12-11
Members List Showing Markup

montana-mil:
## Description
Change members list table to fixed in order to prevent code from showing when the table loads

## Pivotal Tracker

### DIFF
--- a/templates/workspaces/members/index.html
+++ b/templates/workspaces/members/index.html
@@ -108,7 +108,7 @@
     </div>
   </form>
 
-  <div class='responsive-table-wrapper'>
+  <div class='fixed-table-wrapper'>
     <table v-if='searchedList && searchedList.length'>
       <thead>
         <tr>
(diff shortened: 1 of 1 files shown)


leigh-mil:
Works for me!

montana-mil:
blast

templates/workspaces/members/index.html
```diff
@@ -112,7 +112,7 @@
     <table v-if='searchedList && searchedList.length'>
       <thead>
         <tr>
-          <th v-for="col in getColumns()" @click="updateSort(col.displayName)" :width="col.width" :class="col.class" scope="col">
+          <th v-cloak v-for="col in getColumns()" @click="updateSort(col.displayName)" :width="col.width" :class="col.class" scope="col">
```
patricksmithdds: I think we should apply the `v-cloak` attribute to the whole table (on line 112). 

With cloak on just the `th` an empty row is displayed:

image


With cloak on the table, the display is a bit cleaner:

image

PR #489, opened 2018-12-12
change current task order name

dandds:
PT chore: https://www.pivotaltracker.com/story/show/162589090

Because we're rethinking things in a way that will make task orders central, we need to slowly deprecate our existing workflow. This PR renames the existing task order table and model so that we can add a fresh task order representation for the new workflow.


atst/domain/task_orders.py
```diff
@@ -2,19 +2,21 @@
 from flask import current_app as app
 
 from atst.database import db
-from atst.models.task_order import TaskOrder, Source, FundingType
+from atst.models.legacy_task_order import LegacyTaskOrder, Source, FundingType
 from .exceptions import NotFoundError
 from atst.utils import update_obj
 
 
 class TaskOrders(object):
```
patricksmithdds: Should we rename this domain object as well? If not, how will we handle getting a "new" task order vs getting a legacy task order?

dandds: Updated the domain class. I also changed the name of the request relation to the task order, too, to avoid confusion. I wound up doing a big S&R to make sure I caught everything, which means this is a big PR :/

PR #491, opened 2018-12-13
Suggestions from http://flask.pocoo.org/docs/1.0/security/

drummond-dds:
PT Link: https://www.pivotaltracker.com/story/show/160855838

Added some headers recommended by the Flask Security guidelines.

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -102,10 +102,14 @@ def set_default_headers(app):
         response.headers["X-Frame-Options"] = "SAMEORIGIN"
         response.headers["X-XSS-Protection"] = "1; mode=block"
 
-        if ENV == 'dev':
-            response.headers["Content-Security-Policy"] = "default-src 'self' 'unsafe-eval'; connect-src *"
+        if ENV == "dev":
+            response.headers[
+                "Content-Security-Policy"
+            ] = "default-src 'self' 'unsafe-eval'; connect-src *"
         else:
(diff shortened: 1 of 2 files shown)


templates/components/alert.html
```diff
@@ -54,7 +54,7 @@ <h2 class='alert__title'>{{title}}</h2>
           {% elif actions is iterable %}
 
             {% for action in actions %}
-              <a href={{ action["href"] }} class='icon-link'>
+              <a href='{{ action["href"] }}' class='icon-link'>
```
dandds: good catch

drummond-dds:
@patricksmithdds I made the following changes to address this problem.

1. Added `unsafe-eval` to the header since this feature is used by a certain NPM package (`svg-innerhtml`)

2. I added a conditional to return different headers in the development environment. In development we connect to a websocket which is used for code reloading. In production we can prevent this happening.

atst/app.py
```diff
@@ -91,6 +92,28 @@ def _cleanup(response):
         return response
 
 
+def set_default_headers(app):
```
patricksmithdds: Since this is just setting HTTP headers, I think this is code that we could exclude from the coverage reports. @JRNPW do you have an opinion on skipping coverage for this?

drummond-dds: I added that in with 15a32a1

PR #492, opened 2018-12-13
Change copy when workspace access is revoked

leigh-mil:
## Description
This PR updates the copy for the alert when workspace access is revoked. Adding in the alert was covered in the [Flash Refactor PR by @dandds ](https://github.com/dod-ccpo/atst/pull/480/files#diff-48c827ee79e92be838e36421df80c4a8R183)

## Pivotal
[https://www.pivotaltracker.com/story/show/162348648](https://www.pivotaltracker.com/story/show/162348648)

### DIFF
--- a/atst/utils/flash.py
+++ b/atst/utils/flash.py
@@ -12,7 +12,7 @@ MESSAGES = {
     "revoked_workspace_access": {
         "title_template": "Removed workspace access",
         "message_template": """
-          <p>Removed {{ member_name }} from this workspace.</p>
+          <p>Workspace access successfully removed from {{ member_name }}.</p>
         """,
         "category": "success",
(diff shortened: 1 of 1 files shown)


montana-mil:
this works for the banner when I press `Remove Workspace Access`. Did we want to change the banner message when `revoke all access` is clicked? Now it emits a banner that says the environment access was updated. 

I think we would just have to check if the user has no access, and emit a banner with different copy if they have no env access (vs just a change to their access).

leigh-mil:
@montana-mil I was checking out the other PRs that are up and Dan covered the issue of revoking env access in [this PR](https://github.com/dod-ccpo/atst/pull/488)!

PR #494, opened 2018-12-13
Fix conflicting pipenv dependencies

patricksmithdds:
The latest version of `flask-rq2` depends on:

* rq-scheduler>=0.8.3
* rq>=0.12,<0.13

Previously, `rq-scheduler` (0.8.3) required:

* rq>=0.8

A recent release of `rq-scheduler`, 0.9, requires:

* rq>=0.13

`rq-scheduler`'s requirement conflicts directly with `flask-rq2`'s version requirement of `rq`.

Manully pinning `rq-scheduler` to `>=0.8.3,<0.9` causes `0.8.3` to be installed, so both dependencies can be satisfied.

This PR also generates a new `Pipfile.lock`, which updates a couple of dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ defaults:
     PGUSER: root
     PGDATABASE: circle_test
     REDIS_URI: redis://localhost:6379
+    PIP_VERSION: 18.*
   dockerCmdEnvironment: &dockerCmdEnvironment
     APP_USER: atst
     APP_GROUP: atat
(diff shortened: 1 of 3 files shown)


dandds:
I tried a couple times to reproduce the issue CI is having, but even `./script/setup` in a fresh clone of the repo works for me.

patricksmithdds:
> I tried a couple times to reproduce the issue CI is having, but even `./script/setup` in a fresh clone of the repo works for me.

Thanks for giving it a shot. I was curious to try CircleCI rebuilding without cache, but they've made that more difficult with 2.0 :(

PR #495, opened 2018-12-14
Put app strings into a YAML file

drummond-dds:
This is an implementation of the editing/translation feature.

App strings are moved into a `translations.yaml` file. You can then reference a string from a template file using the syntax `{{ "login.title" | translate }}`. You can optionally pass in a variable which will be replaced in the string.

## Example

```yaml
# translations.yaml
login
    title: Welcome to the app!
    logged_in_as: You are logged in as {name}
```

```html
<!-- template -->
<h1>{{ "app.login" | translate }}</h1>
<p>{{ "app.logged_in_as" | translate({ "name": "Bob" }) }}</p>
```

```html
<!-- rendered -->
<h1>Welcome to the app!</h1>
<p>You are logged in as Bob</p>
```

## Todo

- [x] Only load json file once, not each time we render a string
- [x] Move existing app text into `strings.json`
- [x] Give `strings.json` a better place to live
- [x] Use existing [nested lookup](https://github.com/dod-ccpo/atst/blob/master/atst/utils/__init__.py#L26) method

### DIFF
--- a/atst/forms/ccpo_review.py
+++ b/atst/forms/ccpo_review.py
@@ -11,7 +11,7 @@ from atst.utils.localization import translate
 class CCPOReviewForm(CacheableForm):
     comment = TextAreaField(
         translate("forms.ccpo_review.comment_label"),
-        description=("forms.ccpo_review.comment_description"),
+        description=translate("forms.ccpo_review.comment_description"),
     )
     fname_mao = StringField(
         translate("forms.ccpo_review.fname_mao_label"), validators=[Optional(), Name()]
--- a/atst/forms/ccpo_review.py
+++ b/atst/forms/ccpo_review.py
@@ -5,20 +5,31 @@ from wtforms.validators import Email, Optional
 from .forms import CacheableForm
 from .validators import Name, PhoneNumber
 
+from atst.utils.localization import translate
+
 
 class CCPOReviewForm(CacheableForm):
     comment = TextAreaField(
-        "Instructions or comments",
-        description="Provide instructions or notes for additional information that is necessary to approve the request here. The requestor may then re-submit the updated request or initiate contact outside of AT-AT if further discussion is required. <strong>This message will be shared with the person making the JEDI request.</strong>.",
+        translate("forms.ccpo_review.comment_label"),
+        description=("forms.ccpo_review.comment_description"),
     )
-    fname_mao = StringField("First Name (optional)", validators=[Optional(), Name()])
-    lname_mao = StringField("Last Name (optional)", validators=[Optional(), Name()])
+    fname_mao = StringField(
+        translate("forms.ccpo_review.fname_mao_label"), validators=[Optional(), Name()]
+    )
+    lname_mao = StringField(
+        translate("forms.ccpo_review.lname_mao_label"), validators=[Optional(), Name()]
+    )
     email_mao = EmailField(
-        "Mission Owner e-mail (optional)", validators=[Optional(), Email()]
+        translate("forms.ccpo_review.email_mao_label"), validators=[Optional(), Email()]
     )
     phone_mao = TelField(
-        "Mission Owner phone number (optional)", validators=[Optional(), PhoneNumber()]
+        translate("forms.ccpo_review.phone_mao_label"),
+        validators=[Optional(), PhoneNumber()],
     )
-    phone_ext_mao = StringField("Extension (optional)")
-    fname_ccpo = StringField("First Name (optional)", validators=[Optional(), Name()])
-    lname_ccpo = StringField("Last Name (optional)", validators=[Optional(), Name()])
+    phone_ext_mao = StringField(translate("forms.ccpo_review.phone_ext_mao_label"))
+    fname_ccpo = StringField(
+        translate("forms.ccpo_review.fname_ccpo_label"), validators=[Optional(), Name()]
+    )
+    lname_ccpo = StringField(
+        translate("forms.ccpo_review.lname_ccpo_label"), validators=[Optional(), Name()]
+    )
--- a/atst/utils/localization.py
+++ b/atst/utils/localization.py
@@ -28,9 +28,9 @@ def _translations_file():
     if app:
         file_name = app.config.get("DEFAULT_TRANSLATIONS_FILE", file_name)
 
-    file = open(file_name)
+    f = open(file_name)
 
-    return yaml.safe_load(file)
+    return yaml.safe_load(f)
 
 
 def translate(key, variables=None):
--- a/atst/utils/localization.py
+++ b/atst/utils/localization.py
@@ -1,10 +1,7 @@
 import yaml
 import os
 from flask import current_app as app
-from functools import lru_cache
 from atst.utils import getattr_path
-
-ENV = os.getenv("FLASK_ENV", "dev")
 
 
 class LocalizationInvalidKeyError(Exception):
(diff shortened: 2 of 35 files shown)


atst/forms/ccpo_review.py
```diff
@@ -5,20 +5,31 @@
...
 
+from atst.utils.localization import translate
+
 
 class CCPOReviewForm(CacheableForm):
     comment = TextAreaField(
-        "Instructions or comments",
-        description="Provide instructions or notes for additional information that is necessary to approve the request here. The requestor may then re-submit the updated request or initiate contact outside of AT-AT if further discussion is required. <strong>This message will be shared with the person making the JEDI request.</strong>.",
+        translate("forms.ccpo_review.comment_label"),
+        description=("forms.ccpo_review.comment_description"),
```
dandds: Looks like this one is missing the `translate` function call.

translations.yaml
```diff
@@ -0,0 +1,299 @@
...
+    add_new_member_label: Add New Member
+    add_new_project_label: Add New Project
+    budget_report: Budget Report
+    members: Members
+    projects: Projects
+    workspace_settings: Workspace Settings
+requests:
+  _new:
+    new_request: New Request
+    save_and_continue: Save & Continue
```
dandds: Things like this should probably go under a more general namespace, since they might get used across the site.

atst/utils/localization.py
```diff
@@ -0,0 +1,35 @@
...
+localizations_cache_size = 1 if ENV == "dev" else None
+
+
+@lru_cache(maxsize=localizations_cache_size)
+def load_cached_translations_file(file_name):
+    return open(file_name).read()
+
+
+def translate(key, variables={}):
+    translations = yaml.safe_load(load_cached_translations_file("translations.yaml"))
```
dandds: This is an interesting approach; I hadn't thought to use the LRU cache. Could we make it more flexible? Like, what if we want to specify a path to a specific translations file when the app is deployed? (For instance, we could let the designers et. al. edit a copy hosted remotely some place that we sync for one environment when the app is rebuilt during CD.) This module could import `current_app` and read config values off of that:

```
[...]
def translate(key, variables={}):
    config_file = app.config.get("TRANSLATIONS_FILE") or "translations.yml"
    translations = yaml.safe_load(load_cached_translations_file(config_file))
    [...]
```
That seems like a weird pattern, admittedly (a function accessing mutable state from somwhere else), but Flask extensions tend to use patterns like that: http://flask.pocoo.org/docs/1.0/extensiondev/

atst/forms/ccpo_review.py
drummond-dds: Pushed up a fix for this

atst/utils/localization.py
```diff
@@ -0,0 +1,35 @@
...
+    value = getattr_path(translations, key)
+
+    if value == None:
```
patricksmithdds: This doesn't really change anything, but when comparing to `None`, [it's preferred to use `is None` instead of `== None`.](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

atst/utils/localization.py
```diff
@@ -0,0 +1,35 @@
...
+
+localizations_cache_size = 1 if ENV == "dev" else None
+
+
+@lru_cache(maxsize=localizations_cache_size)
+def load_cached_translations_file(file_name):
+    return open(file_name).read()
+
+
+def translate(key, variables={}):
```
patricksmithdds: `{}` and `[]` shouldn't be used for default argument values in Python: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

atst/utils/localization.py
```diff
@@ -0,0 +1,35 @@
...
+    def __str__(self):
+        return "Requested {key} and variables {variables} with but an error occured".format(
+            key=self.key, variables=self.variables
+        )
+
+
+localizations_cache_size = 1 if ENV == "dev" else None
+
+
+@lru_cache(maxsize=localizations_cache_size)
```
patricksmithdds: I'm curious if we want to cache the actual look up of a translation string rather than the loading of the file? I don't believe that we won't be loading more than one file, so this will always be called with the same filename.

atst/utils/localization.py
drummond-dds: Fixed

atst/utils/localization.py
drummond-dds: Fixed

translations.yaml
drummond-dds: @dandds I introduced a `common` namespace. Since there are so many strings in the app and things are changing now I think we should gradually move things to this as the app evolves rather than hit it all in the first attempt.

atst/utils/localization.py
```diff
@@ -25,8 +26,18 @@ def load_cached_translations_file(file_name):
     return open(file_name).read()
 
 
+def _translations_file():
+    file_name = "translations.yaml"
+
+    if app:
+        file_name = app.config.get("TRANSLATIONS_FILE") or file_name
```
dandds: Nitpicky, but the second arg to `get` is used as a default. I'd put the file path in something like `DEFAULT_TRANSLATION_FILE` outside the function and then do:

```
return app.config.get("TRANSLATIONS_FILE", DEFAULT_TRANSLATION_FILE)
```

(I don't think you need the condition for `app`, since it's an import.)

drummond-dds: @dandds I fixed the second argument instead of the `or`.

I had to add in the `app` or things break with an `outside of app context` error when importing `atst.routes.workspaces` on line 15 of `atst/app.py`. I think this is because we run the import before the `make_app()` on line 35

dandds: That makes sense. Importing mutable object instances in Python is always very confusing to me.

atst/utils/localization.py
drummond-dds: @patricksmithdds I removed the LRU cache

drummond-dds:
@dandds @patricksmithdds I made a change to the filters which was preventing the views from reloading the `translations.yaml` in development when editing a string.

atst/utils/localization.py
```diff
@@ -0,0 +1,37 @@
...
+    file_name = "translations.yaml"
+
+    if app:
+        file_name = app.config.get("DEFAULT_TRANSLATIONS_FILE", file_name)
+
+    return open(file_name)
+
+
+def translate(key, variables=None):
+    translations_file = _translations_file()
```
patricksmithdds: Is this loading the translations file every time something is translated? I was hoping it was possible that the file would only be loaded once and then the `safe_load`ed translations could be used each subsequent time.

drummond-dds: @patricksmithdds In production jinja2 caches the results of `{{ "something" | translate }}` - so once the cache is warmed it won't make any more translation calls.

In development we force jinja2 to not cache calls to `translate`, so developers can edit `translations.yaml` and instantly see the change.

I wonder if LRU cache would be useful for the production case. We could open + read the YAML file once. Then we would read the YAML file one time per deploy and still cache calls to `translate`? I added this in e630a7868fc9cc6d4819b406754040cfaf77dc02

atst/utils/localization.py
```diff
@@ -0,0 +1,46 @@
...
+    def __str__(self):
+        return "Requested {key} and variables {variables} with but an error occured".format(
+            key=self.key, variables=self.variables
+        )
+
+
+translations_yaml_max_cache = 0 if ENV == "dev" else None
+
+
+@lru_cache(maxsize=translations_yaml_max_cache)
```
patricksmithdds: In dev mode, this won't cache anything and that's fine. In production mode, though, since there's no arguments passed to `_translations_file`, the first call will be cached and used thereafter. Since we're doing translations in the form fields, `translate` is called when the forms are imported and therefore before the app is initialized. So, the first time this is called `app` will be falsy and the `translations.yml` file will always be used for translations, with the `DEFAULT_TRANSLATIONS_FILE` setting being ignored in production.

I think this is fine for now, but in the future, that may produce unintended consequences.

PR #497, opened 2018-12-14
Prevent Vue from Loading Variables Before Evaluation

montana-mil:
## Description
When loading tables, charts, or longer lists, some Vue markup was flashing on screen. Using `v-cloak` on these Vue components prevents that behavior so that nothing loads until all the variables have been evaluated. 

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/162622309

### DIFF
--- a/templates/audit_log/audit_log.html
+++ b/templates/audit_log/audit_log.html
@@ -3,23 +3,23 @@
 
 {% block content %}
 
+  <div v-cloak>
+    <section class="block-list">
+      <header class="block-list__header">
+        <h1 class="block-list__title">Activity History</h1>
+      </header>
 
-  <section v-cloak class="block-list">
-    <header class="block-list__header">
-      <h1 class="block-list__title">Activity History</h1>
-    </header>
-
-    <ul>
-      {% for event in audit_events %}
-        <li class="block-list__item">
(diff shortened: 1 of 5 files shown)


templates/audit_log/audit_log.html
```diff
@@ -19,7 +19,7 @@ <h1 class="block-list__title">Activity History</h1>
       {% endfor %}
     </ul>
 
-  </section>
+    {{ Pagination(audit_events, 'atst.activity_history') }}
 
-  {{ Pagination(audit_events, 'atst.activity_history') }}
+  </section>
```
dandds: It looks like moving this changes the spacing above it. Does it render badly before the Vue component is ready if it's outside the `section`?

montana-mil: it just shows the pagination before anything else, so it looks funny. do you think theres another way?

dandds: You could wrap the pagination in a blank div where it is and put `v-cloak` on the div, maybe?

montana-mil: ok, can do!

leigh-mil: Caaaaan doooooo!

PR #499, opened 2018-12-17
Workspace member buttons #162641480

leigh-mil:
## Description
This fixes a bug where the buttons on the workspace member pages were appearing when the should not. Now the buttons should only appear as follows:

**Remove Workspace Access** - Active Users
**Revoke Invitation** - Pending Invites
**Resend Invitation** - Expired and Revoked Invites

## Pivotal
[https://www.pivotaltracker.com/story/show/162641480](https://www.pivotaltracker.com/story/show/162641480)

### DIFF
--- a/atst/models/workspace_role.py
+++ b/atst/models/workspace_role.py
@@ -152,12 +152,7 @@ class WorkspaceRole(Base, mixins.TimestampsMixin, mixins.AuditableMixin):
     @property
     def can_resend_invitation(self):
         return not self.is_active and (
-            self.latest_invitation
-            and (
-                self.latest_invitation.is_rejected
-                or self.latest_invitation.is_expired
-                or self.latest_invitation.is_revoked
-            )
+            self.latest_invitation and self.latest_invitation.is_inactive
         )
 
 
--- a/atst/models/workspace_role.py
+++ b/atst/models/workspace_role.py
@@ -118,6 +118,10 @@ class WorkspaceRole(Base, mixins.TimestampsMixin, mixins.AuditableMixin):
         return self.role.display_name
 
     @property
+    def is_active(self):
+        return self.status == Status.ACTIVE
+
+    @property
     def num_environment_roles(self):
         return (
             db.session.query(EnvironmentRole)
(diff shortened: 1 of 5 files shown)


atst/models/workspace_role.py
```diff
@@ -147,8 +151,13 @@ def has_environment_roles(self):
...
     @property
     def can_resend_invitation(self):
-        return self.latest_invitation and (
-            self.latest_invitation.is_rejected or self.latest_invitation.is_expired
+        return not self.is_active and (
+            self.latest_invitation
+            and (
+                self.latest_invitation.is_rejected
+                or self.latest_invitation.is_expired
+                or self.latest_invitation.is_revoked
```
dandds: It seems like this requires the `WorkspaceRole` model to know a little too much about the `Invitation` model. Maybe invitation should have a property like `is_disabled` that's a combo of these three?

tests/domain/test_workspaces.py
```diff
@@ -304,17 +304,21 @@ def test_can_create_workspaces_with_matching_names():
     Workspaces.create(RequestFactory.create(), name=workspace_name)
 
 
-def test_can_revoke_workspace_access():
+def test_can_revoke_workspace_access_for_active_member():
```
dandds: Very nitpicky, but the name of this test and the one below confused me because this one is testing the actual revocation while the one below is testing our revocation check.

tests/routes/workspaces/test_members.py
```diff
@@ -208,16 +240,70 @@ def test_shows_revoke_button(client, user_session):
...
+        status=InvitationStatus.REJECTED_EXPIRED,
+    )
+    user_session(workspace.owner)
+    response = client.get(
+        url_for(
+            "workspaces.view_member",
+            workspace_id=workspace.id,
+            member_id=member.user.id,
+        )
+    )
```
dandds: Also, nitpicky, but a lot of these tests do the same setup work with just a couple differences (like the invitation status). Maybe pull that functionality into some kind of test utility (helper function, pytest fixture, whatever).

PR #503, opened 2018-12-18
Task Order Form

dandds:
This adds a very rough version of the new task order form process.

Not included on this PR:
- Final submission
- field validation
- new date selector
- probably other things

TODO:
- [x] squash migrations
- [x] add tests for task order route utility classes


dandds:
@patricksmithdds I switched the CLINs to be [`Numeric`](https://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.Numeric), which works out to be `decimal` in Postgres. Postgres does have monetary types, but since we only need to support USD a decimal with two places should be fine.

atst/forms/task_order.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class AppInfoForm(CacheableForm):
+    portfolio_name = StringField(
+        "Organization Portfolio Name",
+        description="The name of your office or organization. You can add multiple applications to your portfolio. Your task orders are used to pay for these applications and their environments",
+    )
+    scope = TextAreaField(
+        "Cloud Project Scope",
+        description="The name of your office or organization. You can add multiple applications to your portfolio. Your task orders are used to pay for these applications and their environments",
```
patricksmithdds: Looks like this description is the same as the one above.

templates/task_orders/new/menu.html
```diff
@@ -0,0 +1,21 @@
+<div class="progress-menu progress-menu--four">
+  <ul>
+    {% for s in screens %}
+      {% if s.complete %}
+        {% set step_indicator = 'complete' %}
+      {% elif loop.index == current %}
```
patricksmithdds: Looks like the active highlight is off by one:

image

dandds:
@patricksmithdds I addressed both of your comments. I updated the downgrade function for that migration so that it will delete any new workspaces, etc., that don't match the old request flow (i.e., they don't have a relationship to a request).

dandds:
Aaah I dismissed your review :/

PR #504, opened 2018-12-18
JRN no branch coverage

JRNPW:
This PR addresses card https://www.pivotaltracker.com/story/show/162670922

In three of these instances, the FALSE condition is practically impossible to achieve, so they are considered "intentional partial branches." For domain/invitations, all other conditions have been addressed first (and tested), so the final case becomes an "intentional partial branch."

### DIFF
--- a/atst/domain/invitations.py
+++ b/atst/domain/invitations.py
@@ -84,7 +84,7 @@ class Invitations(object):
         elif invite.is_accepted or invite.is_revoked or invite.is_rejected:
             raise InvitationError(invite)
 
-        elif invite.is_pending:
+        elif invite.is_pending:  # pragma: no branch
             Invitations._update_status(invite, InvitationStatus.ACCEPTED)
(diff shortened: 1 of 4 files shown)


atst/domain/invitations.py
```diff
@@ -84,7 +84,7 @@ def accept(cls, user, token):
         elif invite.is_accepted or invite.is_revoked or invite.is_rejected:
             raise InvitationError(invite)
 
-        elif invite.is_pending:
+        elif invite.is_pending:  # pragma: no branch
```
dandds: I think in an ideal world we would have an `else` here, but here in the real world...

JRNPW: No argument. Montana, Leigh, and I talked about this for several minutes, and we agree this is one of those, "The test would mock something that probably doesn't every really happen," so we agreed this was the alternative.

PR #508, opened 2018-12-20
Task Order review and download

dandds:
This PR allows users to complete the full task order form. It does three related things:

- Adds a basic version of the review page in our [wireframes](https://projects.invisionapp.com/share/UVPNV50WJ3P#/screens/337413977)
- Adds functionality for generating and downloading a docx summary of the task order in its current state
- Adds a route for inviting users (KO, etc.) attached to the task order (but not real invitation functionality)

**notes**
- The "Task Order Summary" link is the only link that does anything.
- As far as making or refining new task order summary documents, we can make our own *.docx files (with custom styles and whatever else). For custom values, we should use Jinja syntax directly into the docx file. When we're ready, we can extract the custom XML from the docx and check it to make sure that it's usable as a Jinja template. (This might involve correcting places where our template syntax got split across OOXML runs, like `<w:r>TO Number: {{ task_ord</w:r><w:r>er.number }}</w:r>`). Then we just have to replace the `document.xml` and `template.docx` in the repo with the new ones.

screenshot_2018-12-20 jedi cloud


atst/utils/docx.py
```diff
@@ -0,0 +1,45 @@
+import os
+from io import BytesIO
+from zipfile import ZipFile
```
patricksmithdds: TIL [`zipfile` is in the standard lib](https://docs.python.org/3/library/zipfile.html)

atst/utils/docx.py
```diff
@@ -0,0 +1,45 @@
...
+    @classmethod
+    def render(
+        cls,
+        doc_template="docx/document.xml",
+        file_template="docx/template.docx",
+        **args,
+    ):
+        document = render_template(doc_template, **args)
+        byte_str = BytesIO()
+        docx_file = ZipFile(byte_str, mode="w")
```
patricksmithdds: Can this be in a `with` statement so we don't need to remember to `close` at the end?

atst/utils/flash.py
```diff
@@ -101,6 +101,13 @@
         "message_template": "",
         "category": "success",
     },
+    "task_order_complete": {
+        "title_template": "Task Order Complete",
```
patricksmithdds: Is the "complete" title taken from anywhere? I think the task order wouldn't be complete until it's signed later by another user.

atst/utils/docx.py
```diff
@@ -0,0 +1,45 @@
...
+    @classmethod
+    def _template_path(cls, docx_file):
+        return os.path.join(app.root_path, "..", "templates", docx_file)
+
+    @classmethod
+    def _template(cls, docx_file):
+        return ZipFile(Docx._template_path(docx_file), mode="r")
+
+    @classmethod
+    def _write(cls, docx_template, docx_file, document):
```
patricksmithdds: I'm a bit confused about what this is doing -- could you add a docstring? I (think) I understand that a .docx file is actually a zipped file, but I'm not following what files are contained within.

templates/docx/document.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<w:document xmlns:ve="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:o="urn:schemas-microsoft-com:office:office" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships" xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:wp="http://schemas.openxmlformats.org/drawingml/2006/wordprocessingDrawing" xmlns:w10="urn:schemas-microsoft-com:office:word" xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main" xmlns:wne="http://schemas.microsoft.com/office/word/2006/wordml">
+  <w:body>
+    {% for key,val in data.items() %}
+    <w:p>
+      <w:r>
+        <w:t>{{ key }}: {{ val }}</w:t>
+      </w:r>
+    </w:p>
```
patricksmithdds: Is this `w:r`, etc syntax a Word thing?

atst/utils/docx.py
```diff
@@ -0,0 +1,45 @@
...
+        **args,
+    ):
+        document = render_template(doc_template, **args)
+        byte_str = BytesIO()
+        docx_file = ZipFile(byte_str, mode="w")
+        docx_template = Docx._template(file_template)
+        Docx._write(docx_template, docx_file, document)
+        docx_file.close()
+        byte_str.seek(0)
+        return byte_str
```
patricksmithdds: Does it makes sense to move the `byte_str` part to the route? I'm under the assumption the reason a `BytesIO` object is used is because that's how we need to stream the response via flask.

atst/utils/docx.py
dandds: Yeah, I am often pleasantly surprised by what's in the standard lib.

atst/utils/flash.py
dandds: I can change it. I don't think there's any copy for this yet.

templates/docx/document.xml
dandds: It's Office Open XML (OOXML), a document format. (It's even an ECMA standard: https://www.ecma-international.org/publications/standards/Ecma-376.htm). MS created it and uses it as the serialization for all their document formats. DOCX is just a zip file of various related OOXML files. There's a manifest, a file for footnotes, style definitions, etc., and most importantly the `document.xml` file, where most of the regular file content really lives. (It's a lot like EPUB, for instance.) That latter file is the only one we'll really need to touch.

atst/utils/docx.py
dandds: Being able to stream it for the route is definitely a benefit, but I'm actually using it because `ZipFile` requires a file-like object to write to, and I don't want to write to disk here.

dandds:
@patricksmithdds I think I addressed all your comments, but let me know if not.

PR #510, opened 2018-12-21
vue component for tracking CLIN total on task order funding page

dandds:
This adds a view component for the funding form that can calculate the total value of all the CLINs.

The wireframes have the total placed outside the regular form. Rather than rearrange the page, I'm having the Vue component update an element outside of itself. It's a little gross, so let me know if you see a better way to do it.

### DIFF
--- a/js/components/forms/funding.js
+++ b/js/components/forms/funding.js
@@ -65,5 +65,9 @@ export default {
     clin_02: 'updateBudget',
     clin_03: 'updateBudget',
     clin_04: 'updateBudget',
+  },
+
+  mounted: function() {
+    this.updateBudget()
   }
 }
--- a/js/components/forms/funding.js
+++ b/js/components/forms/funding.js
@@ -1,0 +1,69 @@
+import createNumberMask from 'text-mask-addons/dist/createNumberMask'
+import { conformToMask } from 'vue-text-mask'
+
+import FormMixin from '../../mixins/form'
+import textinput from '../text_input'
+
+export default {
+  name: 'funding',
+
+  mixins: [FormMixin],
+
+  components: {
+    textinput,
+  },
+
+  props: {
+    initialData: {
+      type: Object,
+      default: () => ({})
+    }
+  },
+
+  data: function () {
+    const {
+      clin_01 = 0,
+      clin_02 = 0,
+      clin_03 = 0,
+      clin_04 = 0,
+    } = this.initialData
+
+    return {
+      clin_01,
+      clin_02,
+      clin_03,
+      clin_04,
+    }
+  },
+
+  computed: {
+    totalBudget: function () {
+      return [this.clin_01, this.clin_02, this.clin_03, this.clin_04].reduce(
+        function(acc, curr) {
+          curr = curr === null ? 0 : parseInt(curr)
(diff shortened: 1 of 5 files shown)


dandds:
Because this introduces a Vue component for the funding page and we call `tojson` on the funding data, this PR is introduces a small bug. If you try to view the funding screen after filling it in, errors are raised because the Python `Decimal` objects we use for CLINs are not JSON serializable by default.

patricksmithdds:
> the Python `Decimal` objects we use for CLINs are not JSON serializable by default.

If we install [`simplejson`](https://github.com/simplejson/simplejson), `flask` will use that as the encoder and it properly handles `Decimal`s. Or, we could define our own encoder that handles decimals.

dandds:
> Or, we could define our own encoder that handles decimals.

I tried taking a lightweight approach where we define a [default function](https://docs.python.org/3/library/json.html#json.dump) for `json.dumps` that knows how to handle our edge cases. For Jinja, you should be able to define the kwargs it gives to `json.dumps` with the [`json.dumps_kwargs` policy](http://jinja.pocoo.org/docs/2.10/api/#policies). However, Flask defines its own `tojson` and obviates that policy functionality with documenting it anywhere! I went down a little rabbit hole to figure that out. 

Anyway, yeah, I'm probably going to just add `simplejson`.

dandds:
Added `simplejson`. I also removed our `rq-scheduler` constraint, since `pipenv` can resolve that sub-dependency now when it bumps to `flask-rq2` 18.3.

dandds:
For posterity, Flask expects you to set a `json_encoder` either on the app itself or on the blueprint if you need custom behavior: https://github.com/pallets/flask/blob/827536e20ad2079f3f50db11aac03e037e98503e/flask/json/__init__.py#L94-L100

js/components/forms/funding.js
```diff
@@ -0,0 +1,69 @@
...
+      return this.formatDollars(this.totalBudget);
+    },
+  },
+
+  methods: {
+    formatDollars: function (intValue) {
+      const mask = createNumberMask({ prefix: '$', allowDecimal: true })
+      return conformToMask(intValue.toString(), mask).conformedValue
+    },
+    updateBudget: function() {
```
patricksmithdds: Should we call this when the component is mounted? Currently, no value is displayed until a number is entered, but if we call `updateBudget` on mount, it'll show $0.

js/components/forms/funding.js
```diff
@@ -0,0 +1,69 @@
...
+      document.querySelector('#to-target').innerText = this.totalBudgetStr
```
patricksmithdds: I assume the `querySelector` is used because the total is shown in a different template block form the rest of the form so that the `funding` component can't be used to display the value? Ideally, I think we wouldn't have to set inner text on anything and just use Vue components instead, but I think this is small that it's OK. (Another alternative would be to `emit` an event, similar to what we're doing for revoking project roles, to communicate between components)

dandds: Talked about this in-person: I agree, but maybe not worth extra effort at the moment. We'll see where the budget total winds up in the final design.

PR #511, opened 2018-12-21
Refactor CSP integration

patricksmithdds:
This begins a refactoring of the touchpoints with CSP to be more explicit about the interactions with the CSP. This branch touches just the interaction with file storage and reporting. 

A new module in `atst/domain/csp` will attach a `csp` object to the flask app. Currently, the CSP provider is build using mock providers for the file interface and the reporting interface. The `FileProviderInterface` and `ReportingInterface` define the methods that we'll need to implement for the chosen CSP.


atst/uploader.py
```diff
@@ -35,7 +35,7 @@ def upload(self, fyle):
                 object_name=object_name,
                 extra={"acl": "private"},
             )
-        return (fyle.filename, object_name)
+        return object_name
```
dandds: We might want to move most of this implementation into the `RackspaceFileProvider`. It relies on Libcloud, which is theoretically usable across multiple CSPs, but has a couple Rackspace-specific quirks because of those open bugs we've discussed elsewhere.

There are a couple application-specific aspects of uploading we'd want to generalize across file providers, like mimetype checking. One option might be to leave those aspects here and have this module be responsible for using the file provider, but then we'd still be hiding the CSP interaction and initializing a separate uploader. What do you think?

patricksmithdds: @dandds I took a stab at this in e432da0 -- lemme know what you think!

dandds: @patricksmithdds looks great! One small thing: you still have an empty `atst/uploader.py` file.

patricksmithdds: Oops, good catch. Removed that in 1517de9.

PR #513, opened 2019-01-03
Disable CLIN fields via a config option

drummond-dds:
[Grey out and disable CLIN fields when a feature flag is set](https://www.pivotaltracker.com/story/show/162772524).

- [x] CLIN fields grey out and disable when feature flag is set
- [x] Available if feature flags are disabled
- [x] Add label to the field that says "Available soon"

---

screen shot 2019-01-07 at 10 52 38 am

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -121,6 +121,7 @@ def map_config(config):
         **config["default"],
         "ENV": config["default"]["ENVIRONMENT"],
         "DEBUG": config["default"].getboolean("DEBUG"),
+        "CLASSIFIED": config["default"].getboolean("CLASSIFIED"),
         "PORT": int(config["default"]["PORT"]),
         "SQLALCHEMY_DATABASE_URI": config["default"]["DATABASE_URI"],
         "SQLALCHEMY_TRACK_MODIFICATIONS": False,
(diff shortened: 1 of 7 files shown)


atst/app.py
```diff
@@ -46,6 +46,7 @@ def make_app(config):
 
     app.config.update(config)
     app.config.update({"SESSION_REDIS": app.redis})
+    app.config.update({"CLIN_DISABLED": True})
```
dandds: I think this should be a regular config value in the base INI file (probably just `CLASSIFIED` is fine). Since values from INI files are mostly strings by default, you'll need to adjust it in the `map_config` function in this file, like we do for `REQUIRE_CRLS`: https://github.com/dod-ccpo/atst/blob/master/atst/app.py#L131

templates/components/text_input.html
```diff
@@ -32,6 +33,10 @@
           <span class='usa-input__help'>{{ description | safe }}</span>
         {% endif %}
 
+        {% if disabled_available_soon %}
+        <span class="usa-input__coming-soon">Available soon</span>
```
dandds: Hmm, this tightly couples this component to the specifics we need for this feature. I think `disabled` as a keyword argument to the macro is fine, but the extra text should be handled differently. In the current wireframes, the "available soon" text is part of the form label, so I'd actually just update the field label in the `wtforms` form object based on the the `CLASSIFIED` config value.

drummond-dds:
@dandds I updated the PR and screenshot. Can you take another look?

PR #515, opened 2019-01-04
Task order invites #162554089

dandds:
https://www.pivotaltracker.com/story/show/162554089

This adds "invite" checkboxes to the "Oversight" section of the TO form. If checked, the selected officer is invited to the workspace/task order. This implementation relies on our existing invitation system and adds additional roles for the Task Order officers (KO, etc.).

**notes**
- For now, TO officers are redirected to the "review" page of the TO form as it stands. This will change when we have a KO review and sign page, a page for security questions for the SO to answer, etc.
- This doesn't cover authorization for the TO officers.

screenshot_2019-01-04 jedi cloud


atst/domain/roles.py
```diff
@@ -140,6 +140,24 @@
...
+        "description": "Assists in submitting task order documents. Can update Task Order.",
+        "display_name": "Contracting Officer Representative",
+        "permissions": [Permissions.VIEW_WORKSPACE],
+    },
+    {
+        "name": "security_officer",
+        "description": "Can edit security questions for a task order.",
+        "display_name": "Contracting Officer",
+        "permissions": [Permissions.VIEW_WORKSPACE, Permissions.SO_SIGN_TASK_ORDER],
+    },
```
patricksmithdds: Not sure if it's desirable, but these roles also show up in the workspace role dropdown on the add a member page: 

image

Not sure what's going on with the formatting there, either.

atst/routes/task_orders/new.py
```diff
@@ -102,6 +109,59 @@ def update(self):
...
+                )
+                ws_officer_member = WorkspaceRoles.get(self.workspace.id, officer.id)
+                invite_service = InvitationService(
+                    self.user,
+                    ws_officer_member,
+                    officer_data["email"],
+                    subject=officer_type["subject"],
+                    email_template=officer_type["template"],
+                )
+                invite_service.invite()
```
patricksmithdds: I'm curious if we need to add the user with the role without sending an invite. For example,  if I fill out the KO section of the form, but don't select the option to invite the KO, should the KO be added to the workspace, just not sent the invite?

atst/domain/task_orders.py
```diff
@@ -89,3 +94,26 @@ def all_sections_complete(cls, task_order):
...
+        "contracting_officer_representative",
+        "security_officer",
+    ]
+
+    @classmethod
+    def add_officer(cls, user, task_order, role, officer_data):
+        if role in TaskOrders.OFFICERS:
+            member = Workspaces.create_member(
+                user, task_order.workspace, {**officer_data, "workspace_role": role}
+            )
```
patricksmithdds: If I'm filling out the new task order, I'm the workspace owner. But, if I also add myself as a KO/COR/SO, then my role is replaced with that role instead. This could create the case of a workspace with no owner (and no way to add any apps or members).

alembic/versions/71cbe76c3b87_add_officers_to_task_order.py
```diff
@@ -0,0 +1,38 @@
...
+    op.create_foreign_key(None, 'task_orders', 'users', ['cor_id'], ['id'])
+    op.create_foreign_key(None, 'task_orders', 'users', ['ko_id'], ['id'])
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint(None, 'task_orders', type_='foreignkey')
+    op.drop_constraint(None, 'task_orders', type_='foreignkey')
+    op.drop_constraint(None, 'task_orders', type_='foreignkey')
```
patricksmithdds: I get an error trying to downgrade with this migration:

```
sqlalchemy.exc.CompileError: Can't emit DROP CONSTRAINT for constraint 
ForeignKeyConstraint(<sqlalchemy.sql.base.ColumnCollection object at 0x112b34cf8>, None,
table=Table('task_orders', MetaData(bind=None), schema=None)); it has no name
```

atst/domain/roles.py
dandds: Maybe we should add some kind of flag to the roles so that we know if they can be displayed as regular workspace options or not.

dandds: Since this is mostly a display concern at the moment, I added a list of exceptions to `form/data.py`.

atst/routes/task_orders/new.py
patricksmithdds: > Also, there's nothing for the KO to do until they're invited to the task order/workspace

I had forgotten that they wouldn't be able to do anything without accepting an invitation. In that case, I think it makes sense to not create the workspace role until the MO wants to invite them.

atst/domain/task_orders.py
dandds: Okay, so:

- There is now a single role for task order officers.
- If an existing workspace member is added as an officer for a TO, we will not overwrite their existing role. Users are only given the "officer" role if they don't have any existing permissions.

atst/domain/task_orders.py
```diff
@@ -89,3 +94,42 @@ def all_sections_complete(cls, task_order):
...
+                None,
+            )
+
+            if existing_member:
+                user = existing_member.user
+            else:
+                member = Workspaces.create_member(
+                    user, workspace, {**officer_data, "workspace_role": "officer"}
+                )
+                user = member.user
```
patricksmithdds: Is there any reason to set `user` here rather than using the incoming argument?

atst/forms/data.py
```diff
@@ -108,6 +108,7 @@
 WORKSPACE_ROLES = [
     (role["name"], {"name": role["display_name"], "description": role["description"]})
     for role in WORKSPACE_ROLE_DEFINITIONS
+    if role["name"] is not "officer"
```
patricksmithdds: Just a general Python thing, [string comparison should use `==` instead of `is`](https://stackoverflow.com/a/1504848).

atst/domain/task_orders.py
dandds: I don't know if I'm following. The variable name here is bad (shouldn't be the same as the incoming `user` arg), so I've changed it. But this user is the (potentially) newly created one and distinct from the incoming `user` arg.

PR #516, opened 2019-01-04
Multicheckbox with 'other' option

montana-mil:
## Description
For certain sections of the task order, we needed an option to allow the user to select multiple answers. This new Multicheckbox component also allows a user to select "other" and type a custom answer into a text input. 

The "other" checkbox is not required, but if it exists in the array of `choices` given in the form, it will work as expected, ie when "other" is checked, a text input appears.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/162667897

## Screenshots
screen shot 2019-01-04 at 2 01 18 pm
screen shot 2019-01-04 at 2 01 35 pm


atst/routes/task_orders/new.py
```diff
@@ -49,12 +49,20 @@ def task_order(self):
 
         return self._task_order
 
+    @property
+    def task_order_formdata(self):
+        task_order_dict = self.task_order.to_dictionary()
+        for field in task_order_dict:
+            if task_order_dict[field] is None:
```
patricksmithdds: Just curious -- was there a particular field that was not handling `None` correctly? Should the field use a different `coerce` method?

js/components/multi_checkbox_input.js
```diff
@@ -0,0 +1,60 @@
...
+      showError: showError,
+      showValid: !showError && !!this.initialValue,
+      validationError: this.initialErrors.join(' '),
+      otherChecked: this.initialValue.includes("other") ? true : this.otherChecked,
+      otherText: this.initialOtherValue,
+      selections: []
+    }
+  },
+
+  mounted: function () {
```
patricksmithdds: Is there any reason this can't be done by setting `selections` in the `data` method?

templates/components/multi_checkbox_input.html
```diff
@@ -0,0 +1,60 @@
...
+          <span v-show='showError'>{{ Icon('alert',classes="icon-validation") }}</span>
+          <span v-show='showValid'>{{ Icon('ok',classes="icon-validation") }}</span>
+        </legend>
+
+        <ul>
+          {% for choice in field.choices %}
+            <li>
+              {% if choice[0] != 'other' %}
+              <input type='checkbox' name='{{ field.name }}' id='{{ field.name }}-{{ field.choices.index(choice) }}' value='{{ choice[0] }}' v-model="selections"/>
+              <label for='{{ field.name }}-{{ field.choices.index(choice) }}'>{{ choice[1] }}</label>
```
patricksmithdds: Minor note -- you can use `{{ loop.index }}` (or `{{ loop.index0 }}` to start at 0) to get the index of the current element rather than finding its index again with `field.choices.index(choice)`.

js/components/other_input.js
```diff
@@ -0,0 +1,29 @@
+import FormMixin from '../mixins/form'
+import textinput from '../components/text_input'
+
+export default {
+  name: 'otherinput',
```
patricksmithdds: Is this component used anywhere?

atst/routes/task_orders/new.py
```diff
@@ -49,12 +49,20 @@ def task_order(self):
...
+
     @property
     def form(self):
         if self._form:
             pass
         elif self.task_order:
-            self._form = self._section["form"](data=self.task_order.to_dictionary())
+            self._form = self._section["form"](formdata=self.task_order_formdata)
         else:
             self._form = self._section["form"]()
```
patricksmithdds: I got an error `AttributeError: can't set attribute` when I do the following:

1. start a new task order 
1. enter a value in the "Other" choice for dev team (and other form values) and submit
1. go back to the first task order screen
1. uncheck the "Other" choice for dev team and submit

js/components/multi_checkbox_input.js
```diff
@@ -0,0 +1,60 @@
...
+    },
+    initialOtherValue: String,
+  },
+
+
+  data: function () {
+    const showError = (this.initialErrors && this.initialErrors.length) || false
+    return {
+      showError: showError,
+      showValid: !showError && !!this.initialValue,
```
patricksmithdds: I believe this causes the field to be shown as valid before any input because `initialValue` is a list (elsewhere, we often use strings) and an empty list is truthy (whereas an empty string is falsy). We may want to check whether `initialValue.length > 0` to see if the initial input is valid.

templates/components/multi_checkbox_input.html
montana-mil: this worked, but i feel like i dont understand it. is is a normal, accepted python thing?

atst/routes/task_orders/new.py
montana-mil: im couldnt replicate this. @leigh-mil can you try?

templates/components/multi_checkbox_input.html
patricksmithdds: Not sure about a "normal accepted python" thing, but I believe thats the best way to do it in Jinja. In normal python, you would usually do something like `for index, choice in enumarate(choices):`, but `enumerate` doesn't work in Jinja templates.

montana-mil: oh i see

PR #517, opened 2019-01-07
hide requests workflow

dandds:
## Description
This hides the requests workflow. We can remove the requests code entirely when we're fully confident that we don't need it anymore.

Note that this leaves the requests workflow enabled for the testing environment. A few of the tests have been skipped since they test for the request link. Some tests have been modified to reflect that changes.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/162589187

## Screenshots
screen shot 2019-01-09 at 2 26 20 pm


atst/app.py
```diff
@@ -64,9 +64,12 @@ def make_app(config):
     make_error_pages(app)
     app.register_blueprint(bp)
     app.register_blueprint(workspace_routes)
-    app.register_blueprint(requests_bp)
     app.register_blueprint(task_orders_bp)
     app.register_blueprint(user_routes)
+
+    if app.config.get("REQUESTS_WORKFLOW"):
+        app.register_blueprint(requests_bp)
```
patricksmithdds: This causes an error when in the context of a workspace. The workspace dropdown has a link to the new request screen that causes issues: https://github.com/dod-ccpo/atst/blob/7a7f8914c8012281d7c87bacec29fc76cad217b6/templates/navigation/topbar.html#L57

config/test.ini
```diff
@@ -4,3 +4,4 @@ PGDATABASE = atat_test
 CRL_DIRECTORY = tests/fixtures/crl
 WTF_CSRF_ENABLED = false
 STORAGE_PROVIDER=LOCAL
+REQUESTS_WORKFLOW=true
```
patricksmithdds: When I set this to false, a lot of the tests fail (134 failed, 289 passed). Can we make sure that tests pass in both configurations?

I'm also a little concerned with our development and deployment environments being configured differently from our test/ci environment.

dandds: I tried to dynamically change which [files pytest collects](https://docs.pytest.org/en/latest/example/pythoncollection.html#customizing-test-collection), but it didn't work like I expected. Instead, I just added a conditional skip based on that config value. You should be able to toggle it on and off now, and the whole suite will keep passing.

patricksmithdds: Looks like that didn't affect the coverage reporting -- now the build is failing because coverage is too low.

dandds: Yeah, I'm working on that atm. Dynamically changing the files we measure is hard, turns out.

dandds:
@montana-mil If you want to pick this up, I think we could delete the existing request code and just remove the toggles everywhere I've added them. Goodbye, requests.

montana-mil:
this should be good to go. @dandds should i update the PR description?

dandds:
@montana-mil Yes, please update, thanks!

PR #521, opened 2019-01-08
View workspace activity log

leigh-mil:
## Description
Adds an audit log for an individual workspace that only shows audit events related to the current workspace. We created a new permission to view the workspace audit log, which allows the workspace owner, workspace admin, and CCPO to view the activity log and have a link to it in the workspace navigation. 

## Pivotal
[https://www.pivotaltracker.com/story/show/162391501](https://www.pivotaltracker.com/story/show/162391501)

## Screenshots 
screen shot 2019-01-08 at 10 52 42 am
screen shot 2019-01-08 at 10 52 53 am
screen shot 2019-01-08 at 10 53 11 am


templates/workspaces/activity/index.html
```diff
@@ -0,0 +1,25 @@
...
+            {% autoescape false %}
+              {{ event | renderAuditEvent }}
+            {% endautoescape %}
+          </li>
+        {% endfor %}
+      </ul>
+    </section>
+
+    {{ Pagination(audit_events, 'workspaces.workspace_activity', workspace_id=workspace_id) }}
+  </div>
```
dandds: This div mostly duplicates `templates/audit_log/audit_log.html`. Can we pull out the relevant parts into a shared partial or something?

tests/domain/test_audit_log.py
```diff
@@ -31,3 +38,65 @@ def test_paginate_audit_log(ccpo):
...
+        user=admin,
+        role=Roles.get("admin"),
+        status=WorkspaceRoleStatus.ACTIVE,
+    )
+    AuditLog.get_workspace_events(admin, workspace)
+
+
+def test_ws_owner_can_view_ws_audit_log():
+    workspace = WorkspaceFactory.create()
+    AuditLog.get_workspace_events(workspace.owner, workspace)
```
dandds: A few of these tests don't have assertions. I know they'd raise errors if the user didn't have permissions, but I think the tests should have some kind of assertion, even if it's just that `get_workspace_events` returns a non-empty list in these cases.

atst/routes/workspaces/index.py
```diff
@@ -80,3 +82,25 @@ def workspace_reports(workspace_id):
...
+
+@workspaces_bp.route("/workspaces/<workspace_id>/activity")
+def workspace_activity(workspace_id):
+    workspace = Workspaces.get(g.current_user, workspace_id)
+    Authorization.check_workspace_permission(
+        g.current_user,
+        workspace,
+        Permissions.VIEW_WORKSPACE_AUDIT_LOG,
+        "view workspace reports",
+    )
```
dandds: `get_workspace_events` performs the same check, right? If that's true, I don't think we need it in the route. (It doesn't happen everywhere in practice, but its generally preferable to do perms checks in domain methods.)

PR #522, opened 2019-01-08
CSP Interface: Environments

patricksmithdds:
This PR completes the refactoring to define the boundaries with the CSP integration, adding a `cloud` module to handle creating users/environments in the CSP and generating an access token. 

I made some assumptions for the access token generation; specifically, that in order to generate an access token, we'll need (in addition to some sort of application configuration secrets) the id of the object being accessed, the id of the user doing the accessing and the permissions. 

Previously, we were not concerning ourselves with any sort of CSP id, so I've added two columns (`Environment.cloud_id` and `User.cloud_id`) to track these values. The environment ID is created when an environment is created and the user ID is created when they are given a role in any environment.


atst/domain/environments.py
```diff
@@ -15,23 +16,22 @@ class Environments(object):
     @classmethod
     def create(cls, project, name):
         environment = Environment(project=project, name=name)
-        db.session.add(environment)
-        db.session.commit()
```
dandds: I don't think you meant to drop the commit here.

atst/domain/environment_roles.py
```diff
@@ -1,8 +1,18 @@
...
 
 
 class EnvironmentRoles(object):
+    @classmethod
+    def create(cls, user, environment, role):
+        env_role = EnvironmentRole(user=user, environment=environment, role=role)
+        if not user.cloud_id:
+            user.cloud_id = app.csp.cloud.create_user(user)
+        app.csp.cloud.create_role(env_role)
+        return env_role
```
dandds: Does this need a manual commit?

dandds: This is a more general question, but have you thought about what we should do in cases where the corresponding CSP action is unsuccessful? Is it sufficient to raise an exception in the `csp` module and role back the db transaction?

patricksmithdds: > Does this need a manual commit?

This doesn't need a manual commit since the user is associated with env_role, so it'll be committed to the session with that object.

> have you thought about what we should do in cases where the corresponding CSP action is unsuccessful?

That's a good question. I think it would depend on what sort of work is required in each integration point. For example, if adding the user to the CSP is more involved, we may want to use RQ to process it asynchronously, which would give us different options as well. 

Since there's no way these mock CSP actions will fail and we don't know what each integration point will require, I'd lean towards addressing that when we have more concrete info.

PR #527, opened 2019-01-09
Simple task order view

patricksmithdds:
This PR adds a very simplified version of the task order listing and viewing page. These pages don't really cover any requirements of the final pages, but I wanted to lay the groundwork for the routes/templates as early as possible to prevent conflicts in multiple branches as the work moves on.

Listing of task orders for a workspace shows each task order with the start/end date:

image

Clicking the link takes one to a placeholder page:

image


leigh-mil:
Looks good to me!

PR #530, opened 2019-01-09
handle empty strings when totalling clin value

dandds:
Small JS fix for CLIN totals to address the issue noted here:

https://www.pivotaltracker.com/story/show/162772340/comments/198247235

### DIFF
--- a/js/components/forms/funding.js
+++ b/js/components/forms/funding.js
@@ -40,7 +40,7 @@ export default {
     totalBudget: function () {
       return [this.clin_01, this.clin_02, this.clin_03, this.clin_04].reduce(
         function(acc, curr) {
-          curr = curr === null ? 0 : parseInt(curr)
+          curr = !curr ? 0 : parseInt(curr)
           return acc + curr;
         }, 0
       )


js/components/forms/funding.js
```diff
@@ -40,7 +40,7 @@ export default {
     totalBudget: function () {
       return [this.clin_01, this.clin_02, this.clin_03, this.clin_04].reduce(
         function(acc, curr) {
-          curr = curr === null ? 0 : parseInt(curr)
+          curr = curr === null || curr === '' ? 0 : parseInt(curr)
```
montana-mil: could do `curr === (null || '')`, right?

montana-mil: or even `curr = curr ? 0 : parseInt(curr)`?

patricksmithdds: @montana-mil `curr === null || curr === ''` and `curr === (null || '')` are not equivalent. The second conditional would evaluate the `null || ''` part first so be equivalent to just `curr === ''`.

montana-mil: ok but just checking `curr` is ok, right?

dandds: @montana-mil yeah, I updated the PR. Just have to bang-invert it.

PR #531, opened 2019-01-09
Fix checkbox

montana-mil:
## Description
This ticket had been kicked back due to a bug and some functionality concerns. 

- If a user clicked `other`, enters text, then clicks another box, the text was erasing. That is now fixed.
- Whenever a user unclicked `other`, the corresponding text would delete. Now the text persists until the user submits the page. 

In addition to those issues, this PR addresses a problem that was happening when a user had completed the first page of the TO form, and then tried to go back and change one of the fields. There was a set Attribute server error occurring in that case. Now, the data being sent for update is in a form that will allow for the user to make changes on information that was already saved to the DB.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/162667897

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -63,21 +63,6 @@ class ShowTaskOrderWorkflow:
         return self._task_order
 
     @property
-    def task_order_formdata(self):
-        task_order_dict = self.task_order.to_dictionary()
-        for field in task_order_dict:
-            if task_order_dict[field] is None:
-                task_order_dict[field] = ""
-
-        # don't save other text in DB unless "other" is checked
-        if "other" not in task_order_dict["complexity"]:
-            task_order_dict["complexity_other"] = ""
-        if "other" not in task_order_dict["dev_team"]:
-            task_order_dict["dev_team_other"] = ""
-
-        return task_order_dict
-
-    @property
     def form(self):
         form_type = (
             "unclassified_form"
@@ -137,11 +122,13 @@ class UpdateTaskOrderWorkflow(ShowTaskOrderWorkflow):
     def task_order_form_data(self):
         to_data = self.form.data.copy()
         if "portfolio_name" in to_data:
-            new_name = self.form.data["portfolio_name"]
-            old_name = self.task_order.to_dictionary()["portfolio_name"]
-            if not new_name is old_name:
-                Portfolios.update(self.task_order.portfolio, {"name": new_name})
             to_data.pop("portfolio_name")
+
+        # don't save other text in DB unless "other" is checked
+        if "complexity" in to_data and "other" not in to_data["complexity"]:
+            to_data["complexity_other"] = ""
+        if "dev_team" in to_data and "other" not in to_data["dev_team"]:
+            to_data["dev_team_other"] = ""
 
         return to_data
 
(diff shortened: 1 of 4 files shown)


patricksmithdds:
> This works, but It saves anything entered in the "other" field to the database, even after a user has unchecked the "other" box

I agree with this concern -- it seems like it should be possible to check if the "other" checkbox is checked before saving that data when processing the form.

montana-mil:
i updated this so that "otherText" won't save unless "other" is checked

atst/routes/task_orders/new.py
```diff
@@ -118,18 +133,28 @@ def workspace(self):
...
             return self.task_order.workspace
 
+    @property
+    def task_order_form_data(self):
+        to_data = self.form.data.copy()
+        if "portfolio_name" in to_data:
+            new_name = self.form.data["portfolio_name"]
+            old_name = self.task_order.to_dictionary()["portfolio_name"]
+            if not new_name is old_name:
+                Workspaces.update(self.task_order.workspace, {"name": new_name})
```
dandds: I wouldn't do this kind of update to the workspace inside a property. A property should just return derived data with no side effects. I'd handle updating the workspace name in the first conditional branch of `_update_task_order`.

atst/routes/task_orders/new.py
```diff
@@ -62,6 +62,21 @@ def task_order(self):
 
         return self._task_order
 
+    @property
+    def task_order_formdata(self):
```
dandds: This never gets called. It seems like it should be combined with the property below?

atst/routes/task_orders/new.py
```diff
@@ -118,18 +118,35 @@ def workspace(self):
...
 
+    @property
+    def task_order_form_data(self):
+        to_data = self.form.data.copy()
+        if "portfolio_name" in to_data:
+            to_data.pop("portfolio_name")
+
+        # don't save other text in DB unless "other" is checked
+        if "complexity" in to_data and "other" not in to_data["complexity"]:
+            to_data["complexity_other"] = ""
```
dandds: I'd use `None` instead of empty string. That way we don't write the empty string to the db.

atst/routes/task_orders/new.py
```diff
@@ -118,18 +118,35 @@ def workspace(self):
...
     def validate(self):
         return self.form.validate()
 
     def _update_task_order(self):
         if self.task_order:
-            TaskOrders.update(self.user, self.task_order, **self.form.data)
+            if "portfolio_name" in self.form.data:
+                new_name = self.form.data["portfolio_name"]
+                old_name = self.task_order.to_dictionary()["portfolio_name"]
+                if not new_name is old_name:
```

- you can just use `self.task_order.portfolio_name` instead of making it into a dictionary
- I didn't know this until @patricksmithdds told me, but you should use `==` and `!=` for Python string comparison: https://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce/1504848#1504848

tests/routes/task_orders/test_new_task_order.py
```diff
@@ -162,6 +162,18 @@ def test_update_task_order_with_existing_task_order(task_order):
...
+def test_other_text_not_saved_if_other_not_checked(task_order):
+    to_data = {
+        **TaskOrderFactory.dictionary(),
+        "complexity": ["conus", "other"],
+        "complexity_other": "quite complex",
+    }
+    workflow = UpdateTaskOrderWorkflow(
+        task_order.creator, to_data, task_order_id=task_order.id
+    )
+    workflow.update()
```
dandds: I don't think this is doing what the test name says. It shouldn't have `other` in the `complexity` list, and then it needs to check that `complexity_other` isn't checked.

montana-mil: whoops!

PR #534, opened 2019-01-10
View task order page

patricksmithdds:
This PR fills out the page for viewing a task order:

image

A couple notes:

* I've added a `status` column to task orders whose only possible value is currently "pending"
* The next steps are all incomplete since they cannot yet be completed. In the future, after the security officer signs the thing, we'll need to update the next step here accordingly
* The links to the cloud estimate and market research don't go anywhere since neither documents exist yet
* An invitation is green if an invitation was sent and red if that officer was skipped

All links not mentioned above should work appropriately.


atst/models/task_order.py
```diff
@@ -27,6 +33,8 @@ class TaskOrder(Base, mixins.TimestampsMixin):
     so_id = Column(ForeignKey("users.id"))
     security_officer = relationship("User", foreign_keys="TaskOrder.so_id")
 
+    status = Column(SQLAEnum(Status, native_enum=False, default=Status.PENDING))
```
dandds: The `default` is not working for me when I fill in a TO. Even when I generate one by hand in the console, it doesn't get applied. 😕

styles/sections/_task_order.scss
```diff
@@ -0,0 +1,131 @@
+.task-order-summary {
+
+  @include media($medium-screen) {
+    @include grid-row;
+    flex-wrap: wrap;
+  }
+
+  .panel {
+    padding: 0 $gap * 2;
```
andrewdds: I'd take a look at `_panel.scss` https://github.com/dod-ccpo/atst/blob/63b354a2e4bdbe9441aaea8928ccccba6bea4d95/styles/elements/_panels.scss#L54

I think the `.panel-content` class already handles padded containers inside a panel.

andrewdds: In fact, I think a lot of the blocks should probably have a `.panel__content` to ensure the paddings are uniform:
atat_v2_preview_mode_-_invision

templates/workspaces/task_orders/show.html
```diff
@@ -1,7 +1,157 @@
...
+        {% endif %}
+      </div>
+      <div class="task-order-next-steps__description">
+        {{ description }}
+      </div>
+    </div>
+  </div>
+{% endmacro %}
+
+{% macro DocumentLink(title="", link_url="", description="") %}
```
andrewdds: The `a` should really encompass the whole thing, including the icon.
Also, I think the proper way to make this disabled is to make this an `a` in all cases, and:
- Add some generic styling for `.icon-link--disabled`, say `opacity: 0.3;`, `pointer-events: none;`
- Add `aria-disabled="true"` html attribute

andrewdds: Also check out https://github.com/dod-ccpo/atst/blob/master/styles/elements/_icon_link.scss and see what visual options we already accounted for. By all means, add things here (such as the `.disabled` variant).

atst/models/task_order.py
patricksmithdds: Interesting. It worked for me after I ran the migration, but I'll try again.

patricksmithdds: Nevermind, I'm seeing the same thing now 🤔

patricksmithdds:
@andrewdds Thanks for the feedback! I believe I incorporated it all:

image

atst/models/task_order.py
patricksmithdds: After futzing with this a bit, I ended up overriding the constructor for the model to set the status there rather than relying on a `default` or `server_default` setting. This is the only thing I could get to work and I think it kinda makes sense to do it there.

atst/models/task_order.py
```diff
@@ -1,10 +1,16 @@
...
+
+from sqlalchemy import Column, Enum as SQLAEnum, Numeric, String, ForeignKey, Date
 from sqlalchemy.types import ARRAY
 from sqlalchemy.orm import relationship
 
 from atst.models import Base, types, mixins
 
 
+class Status(Enum):
+    PENDING = "pending"
```
dandds: I think "pending" should be title case.

patricksmithdds: Done in 809ad71

PR #540, opened 2019-01-14
catch additional CRL download exception

dandds:
Hotfix for our current CRL issue. This catches one additional error and adds a test that bubbles a low-level `TimeoutError` the same way we're seeing it happen in CI.

### DIFF
--- a/atst/domain/authnid/crl/util.py
+++ b/atst/domain/authnid/crl/util.py
@@ -72,7 +72,8 @@ def write_crl(out_dir, target_dir, crl_location):
 
 def remove_bad_crl(out_dir, crl_location):
     crl = crl_local_path(out_dir, crl_location)
-    os.remove(crl)
+    if os.path.isfile(crl):
+        os.remove(crl)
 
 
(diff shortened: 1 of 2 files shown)


atst/domain/authnid/crl/util.py
```diff
@@ -85,7 +86,10 @@ def refresh_crls(out_dir, target_dir, logger):
                 logger.info("successfully synced CRL from {}".format(crl_location))
             else:
                 logger.info("no updates for CRL from {}".format(crl_location))
-        except requests.exceptions.ChunkedEncodingError:
+        except (
+            requests.exceptions.ChunkedEncodingError,
+            requests.exceptions.ConnectionError,
```
patricksmithdds: Do we want to get more general with it and catch any `requests.exceptions.RequestException`?

dandds: Probably a good idea. I thought it would be good to keep a handle on the _ways_ CRL downloading can be flaky and catalogue them one-by-one, but it's probably not worth the effort.

PR #543, opened 2019-01-15
Update Task Order Form - Part 2

leigh-mil:
## Description
This is the second part of revamping the TO form to look like the new design. Most of the changes are to the review page with a few formatting updates throughout the form. 

## Pivotal
[https://www.pivotaltracker.com/story/show/163066434](https://www.pivotaltracker.com/story/show/163066434)

## Screenshots
screen shot 2019-01-15 at 4 21 28 pm
screen shot 2019-01-15 at 4 21 36 pm
screen shot 2019-01-15 at 4 21 42 pm
screen shot 2019-01-15 at 4 24 07 pm
screen shot 2019-01-15 at 4 24 13 pm
screen shot 2019-01-15 at 4 24 43 pm
screen shot 2019-01-15 at 4 24 49 pm
screen shot 2019-01-15 at 4 24 53 pm
screen shot 2019-01-15 at 4 45 38 pm
screen shot 2019-01-15 at 4 45 46 pm


atst/forms/task_order.py
```diff
@@ -50,34 +50,34 @@ class AppInfoForm(CacheableForm):
         choices=[("yes", "Yes"), ("no", "No"), ("not_sure", "Not Sure")],
     )
     complexity = SelectMultipleField(
-        translate("forms.task_order.complexity_label"),
-        description=translate("forms.task_order.complexity_description"),
+        translate("forms.task_order.complexity.label"),
+        description=translate("forms.task_order.complexity.description"),
```
patricksmithdds: I like this refactored organization of the translation strings!

atst/models/task_order.py
```diff
@@ -102,6 +105,62 @@ def portfolio_name(self):
...
 
+    @property
+    def defense_component_description(self):
+        if self.defense_component:
+            return self.normalize_order(self.defense_component)
+        else:
+            return None
+
+    @property
+    def app_migration_description(self):
```
patricksmithdds: I think since these `*_description` methods are concerned with the display of particular properties only, that they are better suited to live in the template instead of here with the model.

atst/models/task_order.py
```diff
@@ -102,6 +105,62 @@ def portfolio_name(self):
...
+        if self.performance_length:
+            return translate(
+                "forms.task_order.performance_length.{}".format(self.performance_length)
+            )
+        else:
+            return None
+
+    @property
+    def ko_phone(self):
+        return usPhone(self.ko_phone_number)
```
patricksmithdds: Same with the `description` comment above -- I think this is something that should live in the template. I think the name is going to be a little confusing as well. How would one know when to use `ko_phone` instead of `ko_phone_number`?

atst/forms/data.py
```diff
@@ -176,67 +177,64 @@
...
+    ("15", translate("forms.task_order.performance_length.15")),
+    ("16", translate("forms.task_order.performance_length.16")),
+    ("17", translate("forms.task_order.performance_length.17")),
+    ("18", translate("forms.task_order.performance_length.18")),
+    ("19", translate("forms.task_order.performance_length.19")),
+    ("20", translate("forms.task_order.performance_length.20")),
+    ("21", translate("forms.task_order.performance_length.21")),
+    ("22", translate("forms.task_order.performance_length.22")),
+    ("23", translate("forms.task_order.performance_length.23")),
+    ("24", translate("forms.task_order.performance_length.24")),
```
dandds: Hard-coding these into the translations file feels a little excessive. You _could_ be excessive in the opposite direction and write a function that can return the correct duration string for a given integer (i.e., given 13, returns "1 year, 1 month"), then write a list comprehension for `range(24)` to build the list we need. That way we only ever need to change one number if the options change. But... up to you.

atst/models/task_order.py
dandds: Agreed

atst/models/task_order.py
```diff
@@ -102,6 +105,62 @@ def portfolio_name(self):
...
+        if self.app_migration:
+            text = translate(
+                "forms.task_order.app_migration.{}".format(self.app_migration)
+            )
+            return self.remove_html(text)
+        else:
+            return None
+
+    @property
+    def native_apps_description(self):
```
dandds: This one throws errors for me when I try to navigate back to screen 4 for a pre-existing task order. You're probably moving it out of the model, but make sure you can navigate back to the review screen for any existing TO.

templates/task_orders/new/review.html
```diff
@@ -18,132 +18,189 @@
...
+<h4 class='task-order-form__heading'>{{ "task_orders.new.review.scope"| translate }}</h4>
+<p>
+{{ task_order.scope or RequiredLabel() }}
+</p>
 
 <hr>
 
-<section>
-  <h2>Generated Documents</h2>
+<h3 class="subheading">{{ "task_orders.new.review.reporting"| translate }} {{ TOEditLink(screen=1) }}</h3>
```
dandds: Nitpicky, but I think we should add an ID to the "about your project" header on the first screen and link directly to that from here. That way the link will drop you right there when you click edit. (I'm also confused as to why this header is "Reporting", which doesn't appear on the first screen at all.)

leigh-mil: ¯\_(ツ)_/¯ Not sure why the header here is "Reporting" but that is what is in the wireframes!

dandds: Right, maybe bring it up with the designers.

atst/forms/data.py
```diff
@@ -1,4 +1,20 @@
+import math
+from gettext import ngettext
+
 from atst.domain.roles import PORTFOLIO_ROLES as PORTFOLIO_ROLE_DEFINITIONS
+from atst.utils.localization import translate
+
+
+def translate_duration(duration_in_months):
```
drummond-dds: I wonder if this should live in the `atst/utils/localization.py` file?

atst/utils/localization.py
```diff
@@ -41,3 +43,14 @@ def translate(key, variables=None):
         raise LocalizationInvalidKeyError(key, variables)
 
     return value.format(**variables).replace("\n", "")
+
+
+def translate_duration(duration_in_months):
+    duration = []
+    years = math.floor(duration_in_months / 12)
```
dandds: no need to change, but you can also use the [`//` operator in python](https://docs.python.org/3.6/library/stdtypes.html#numeric-types-int-float-complex)

PR #544, opened 2019-01-16
Portfolio funding screen

patricksmithdds:
This PR changes the "Portfolio Task Orders" page to a more general "Portfolio Funding" page. The page will show the end funding date, pending task orders, active task orders, and then expired task orders.

The tables of active and expired task orders are sortable by period of performance, initial value, and remaining budget.

A property for "remaining balance" is added to task orders, but just returns the budget since we don't have a way to charge against a task order and will have to figure that out in the future.

For easier testing, I've update the `seed_sample` script to seed a portfolio with two active and two expired task orders.

One other comment: I wanted to test the context sent to a template and found [a recipe from Flask's documentation](http://flask.pocoo.org/docs/1.0/signals/?highlight=template_rendered#subscribing-to-signals) to capture the template context. This required the `blinker` dependency, which I've added as a dev dependency.

localhost_8000_portfolios_b3bf3a60-cc8d-4a14-95d8-cb5f0dd1c04e_task_orders 1

localhost_8000_portfolios_b3bf3a60-cc8d-4a14-95d8-cb5f0dd1c04e_task_orders 2

### DIFF
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -41,6 +41,7 @@ def portfolio_task_orders(portfolio_id):
         if active_task_orders
         else None
     )
+    total_balance = sum([task_order["balance"] for task_order in active_task_orders])
 
     return render_template(
         "portfolios/task_orders/index.html",
@@ -49,6 +50,7 @@ def portfolio_task_orders(portfolio_id):
         active_task_orders=active_task_orders,
         expired_task_orders=task_orders_by_status.get(TaskOrderStatus.EXPIRED, []),
         funding_end_date=funding_end_date,
+        total_balance=total_balance,
     )
 
 
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from operator import itemgetter
 
 from flask import g, render_template, url_for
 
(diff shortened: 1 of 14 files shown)


tests/routes/portfolios/test_task_orders.py
```diff
@@ -0,0 +1,62 @@
...
+
+class TestPortfolioFunding:
+    def test_unfunded_portfolio(self, app, user_session):
+        portfolio = PortfolioFactory.create()
+        user_session(portfolio.owner)
+
+        with captured_templates(app) as templates:
+            response = app.test_client().get(
+                url_for("portfolios.portfolio_funding", portfolio_id=portfolio.id)
+            )
```
dandds: Just curious: why `app.test_client()` instead of the `client` fixture?

patricksmithdds: We have to pass the app to `captured_templates` and I didn't immediately see a way of getting the app from the `client` fixture. I could try poking at it a bit more, though.

atst/routes/portfolios/task_orders.py
```diff
@@ -1,14 +1,57 @@
...
+from atst.models.task_order import Status as TaskOrderStatus
 
 
 @portfolios_bp.route("/portfolios/<portfolio_id>/task_orders")
-def portfolio_task_orders(portfolio_id):
+def portfolio_funding(portfolio_id):
     portfolio = Portfolios.get(g.current_user, portfolio_id)
-    return render_template("portfolios/task_orders/index.html", portfolio=portfolio)
+    task_orders_by_status = defaultdict(list)
+    serialize_task_order = lambda task_order: {
```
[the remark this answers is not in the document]
patricksmithdds: No reason other than I'm not sure where a better place would be. It seems that the serializer might be specific to a particular route/template, but that may not be true?

PR #545, opened 2019-01-16
Allow creating a task order for an existing portfolio

patricksmithdds:
This PR adds a new route (`/portfolios/<portfolio_id>/task_orders/new/<int:screen>`) that allows creating a new Task Order for an existing Portfolio. The portfolio id is only necessary on the first screen of the Task Order process. After the first screen, the task order is updated as it was before.

This is most easily tested on top of #544, since that has a "New Task Order" button on the top of the portfolio funding screen.

Note, the task order forms could be pre-populated with some of the existing portfolio's information (for example, the `portfolio_name` field, which is ignored for anything but the first task order for a portfolio), but I wanted to throw this PR up here to make it easier for us to test portfolios with multiple associated task orders manually.

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -100,11 +100,14 @@ class ShowTaskOrderWorkflow:
 
 
 class UpdateTaskOrderWorkflow(ShowTaskOrderWorkflow):
-    def __init__(self, user, form_data, screen=1, task_order_id=None):
+    def __init__(
+        self, user, form_data, screen=1, task_order_id=None, portfolio_id=None
+    ):
         self.user = user
         self.form_data = form_data
         self.screen = screen
         self.task_order_id = task_order_id
+        self.portfolio_id = portfolio_id
         self._task_order = None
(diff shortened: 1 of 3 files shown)


atst/routes/task_orders/new.py
```diff
@@ -144,7 +147,10 @@ def _update_task_order(self):
                     Portfolios.update(self.task_order.portfolio, {"name": new_name})
             TaskOrders.update(self.user, self.task_order, **self.task_order_form_data)
         else:
-            pf = Portfolios.create(self.user, self.form.portfolio_name.data)
+            if self.portfolio_id:
+                pf = Portfolios.get(self.user, self.portfolio_id)
+            else:
+                pf = Portfolios.create(self.user, self.form.portfolio_name.data)
```
dandds: maybe move this into a `portfolio` property?

patricksmithdds: I'd be a bit wary of having `Portfolios.create` in a property and I think I'd like to keep it explicit in the `_update` method.

PR #546, opened 2019-01-16
am COR checkbox

montana-mil:
## Description
When a user clicks the checkbox labeled "I am the COR for this Task Order," the COR section of the form collapses and the user's information is saved for the COR form data when the form is submitted.

If the user goes back to the oversight page after clicking this checkbox, the checkbox is still checked. 

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163067281

## Screenshots
screen shot 2019-01-16 at 11 47 14 am
screen shot 2019-01-16 at 11 47 07 am


atst/routes/task_orders/new.py
```diff
@@ -140,7 +154,7 @@ def _update_task_order(self):
             if "portfolio_name" in self.form.data:
                 new_name = self.form.data["portfolio_name"]
                 old_name = self.task_order.portfolio_name
-                if not new_name == old_name:
+                if not new_name is old_name:
```
patricksmithdds: This should stay as `==` -- the `is` operator should really only be used when comparing to `None`.

atst/forms/validators.py
```diff
@@ -78,3 +78,18 @@ def _list_items_unique(form, field):
             raise ValidationError(message)
 
     return _list_items_unique
+
+
+def RequiredIfNot(other_field_name, message=translate("forms.validators.is_required")):
```
patricksmithdds: This is really nifty! 🥇 Could you add a quick docstring on here to explain what the validator does?

atst/routes/task_orders/new.py
montana-mil: oo yes, that was a merge conflict mistake

PR #547, opened 2019-01-16
Fix TO Get Started Page in IE10

patricksmithdds:
Fixes a rendering issue on IE10 on the TO get started page. When on a large screen, ensure the next steps don't take up more than 1/3 of the container width.


montana-mil:
is this a good place to put in a fix for small screens?
screen shot 2019-01-16 at 2 38 28 pm

patricksmithdds:
@montana-mil Good eye! Should be fixed in ce02ced.

PR #556, opened 2019-01-17
Navigate to Task Order Form review page

leigh-mil:
## Description
The TO Form review page was very buggy, if you tried to navigate to it before going through each of the previous steps it would give you an error. 
This fix assigns all of the TO attributes to variables inside if statements, which prevents the errors from happening. 

## Pivotal
[https://www.pivotaltracker.com/story/show/163305480](https://www.pivotaltracker.com/story/show/163305480)

## Screenshots
screen shot 2019-01-17 at 3 58 09 pm
screen shot 2019-01-17 at 3 58 13 pm


templates/task_orders/new/review.html
```diff
@@ -19,19 +19,55 @@
...
-  {% set team_experience_description = "forms.task_order.team_experience.{}".format(task_order.team_experience) | translate %}
+  {% set team_experience = "forms.task_order.team_experience.{}".format(task_order.team_experience) | translate %}
+{% endif %}
+
+{% if task_order.performance_period %}
+  {% set performance_length = task_order.performance_length | translateDuration %}
+{% endif %}
+
+{% if task_order.budget %}
+  {% set budget = '${:,.2f}'.format(task_order.budget) %}
```
dandds: Is this the same as the `dollars` filter? If it's not exactly the same and we're not using that filter anywhere else, we should update it and use it here.

PR #557, opened 2019-01-17
handle exceptions when loading CRLs

dandds:
This allows the app to continue setup if it encounters a corrupt or missing CRL. It also improves log output when we authenticate a user and don't have a CRL for their cert.

### DIFF
--- a/atst/domain/authnid/crl/__init__.py
+++ b/atst/domain/authnid/crl/__init__.py
@@ -5,6 +5,12 @@ import hashlib
 from OpenSSL import crypto, SSL
 
 
+def get_common_name(x509_name_object):
+    for comp in x509_name_object.get_components():
+        if comp[0] == b"CN":
+            return comp[1].decode()
+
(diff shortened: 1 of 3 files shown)


dandds:
@patricksmithdds 

> If a CRL cannot be parsed and a client cert has a CA in the chain that matches the missing CRL's CA, do we then reject the client cert?

Right, that's the behavior at the moment. (Letting them in without the CRL check would be "fail open".) I think a future improvement would be to make the Python app aware of the CA names in the chain and keep track of when a CRL for that CA is missing. Then, if a user logs in with a valid cert for a CA in our chain but we're missing the CRL, we can at least reject them with a useful flash message. 

> Is that covered by the test_throws_error_for_missing_issuer test?

Yup, that's the case covered by that test. I expanded the test and added some comments to clarify it. I was tempted to add error codes so that we're not always checking error message strings in these tests, but life is short.

PR #558, opened 2019-01-17
Checkbox is inline for select box validation checkmark

drummond-dds:
Fix the validation checkmark being pushed onto the second line on the `Department of Defense Component`

---

## Before

screen shot 2019-01-17 at 16 20 58

---

## After

screen shot 2019-01-17 at 16 18 29

### DIFF
--- a/templates/components/options_input.html
+++ b/templates/components/options_input.html
@@ -13,7 +13,7 @@
 
       <fieldset v-on:change="onInput" class="usa-input__choices {% if inline %}usa-input__choices--inline{% endif %}">
         <legend>
-          <div class="usa-input__title">
+          <div class="usa-input__title{% if not field.description %}-inline{% endif %}">
             {{ field.label | striptags}}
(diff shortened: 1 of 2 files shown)


templates/components/options_input.html
```diff
@@ -1,7 +1,7 @@
 {% from "components/icon.html" import Icon %}
 {% from "components/tooltip.html" import Tooltip %}
 
-{% macro OptionsInput(field, tooltip, inline=False) -%}
+{% macro OptionsInput(field, tooltip, inline=False, select=False) -%}
```
dandds: I'm not sure about the word "select" here to signify that it needs to be an inline block. Did you choose that because `defense_component` is a dropdown selector? The checkbox is on the wrong line at the moment because that field doesn't have a description, not because of the type of field.

drummond-dds:
@dandds I removed the `select` parameter and just check if there is a description present. This also works on the funding page now.

PR #559, opened 2019-01-18
Disable crls in scripts

dandds:
This should fix a problem we're seeing in the nightly build. I've introduced two changes:

1. We can now pass a dictionary directly to `make_config`. Because we often have to initialize an application context for the scripts (like `script/remove_sample_data.py`, which broke last night) they should be able to modify settings in a way relevant to the scripts need. In this case, they don't need an application instance with CRLs loaded, so I've disabled it for all of them.
2. I deprecated the `REQUIRE_CRLS` setting in favor of `DISABLE_CRL_CHECK`. In both cases, the application won't load CRLs. If this looks okay, I'll apply the updated k8s config.

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -139,7 +139,7 @@ def map_config(config):
     }
 
 
-def make_config():
+def make_config(direct_config=None):
     BASE_CONFIG_FILENAME = os.path.join(os.path.dirname(__file__), "../config/base.ini")
     ENV_CONFIG_FILENAME = os.path.join(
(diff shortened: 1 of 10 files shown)


patricksmithdds:
> like script/remove_sample_data.py, which broke last night

If the script is breaking because the CRLs aren't right, does this change mean that the scripts will work, but then the app will still have issues with CRLs?

dandds:
@patricksmithdds That's right. Technically #559 also fixes this problem, but it seemed like the scripts should just not bother with the CRLs in the first place.

PR #560, opened 2019-01-18
Toggle DOD ID Input For Officers

montana-mil:
## Description
Only require the user to provide the KO's DOD ID if the user is inviting the KO. If the "Invite KO" checkbox is left unchecked, the section to input the DOD ID remains hidden. Only when they check this box is the input for the KO's DOD ID revealed. Similarly for the COR and SO Invite checkboxes.

This PR also refactors a recent custom validator to be more dynamic. 

If the user navigates away from the Oversight page, and then back to the Oversight Page, their choice of invite checkbox values will be preserved.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163067040

## Screenshots
screen shot 2019-01-18 at 9 59 06 am
screen shot 2019-01-18 at 9 59 14 am


templates/task_orders/new/oversight.html
```diff
@@ -13,24 +13,26 @@
...
 <h3 class="subheading">{{ "task_orders.new.oversight.ko_info_title" | translate }}</h3>
 <p>{{ "task_orders.new.oversight.ko_info_paragraph" | translate }}</p>
-{{ UserInfo(form.ko_first_name, form.ko_last_name, form.ko_email, form.ko_phone_number) }}
-{{ CheckboxInput(form.ko_invite) }}
-{{ TextInput(form.ko_dod_id, placeholder="1234567890", tooltip="Why", tooltip_title='Why', validation='dodId')}}
+<oversight inline-template v-bind:initial-data='{{ form.data|tojson }}'>
+    <div class='usa-input'>
+        {{ UserInfo(form.ko_first_name, form.ko_last_name, form.ko_email, form.ko_phone_number) }}
+        {{ CheckboxInput(form.ko_invite) }}
+        <template v-if="ko_invite">
```
patricksmithdds: Could we do the same thing for the COR and SO invites as well?

atst/forms/task_order.py
```diff
@@ -128,11 +128,11 @@ class OversightForm(CacheableForm):
     cor_email = StringField(translate("forms.task_order.oversight_email_label"))
     cor_phone_number = TelField(
         translate("forms.task_order.oversight_phone_label"),
-        validators=[RequiredIfNot("am_cor"), PhoneNumber()],
+        validators=[RequiredIf("am_cor", False), PhoneNumber()],
     )
     cor_dod_id = StringField(
         translate("forms.task_order.oversight_dod_id_label"),
-        validators=[RequiredIfNot("am_cor"), Length(min=10), IsNumber()],
+        validators=[RequiredIf("am_cor", False), Length(min=10), IsNumber()],
```
patricksmithdds: I think this is a bit harder to read with the second parameter. Personally, I like to use keyword args for `True`/`False` parameters since it's rarely immediately clear what True/False means in a parameter list: `RequiredIf("am_cor", checked=False)`.

What do you think of having the `RequiredIf` validator and then a second one, `RequiredIfNot`, that abstracts away the `checked` parameter? `RequiredIfNot("am_cor")` reads better than `RequiredIf("am_cor", False)` to me.

Or, another idea: the `RequiredIf` validator could take a function thats given the form so that one validator could be used. Then, we could use it like:

```python
cor_dod_id = StringField(validators=[RequiredIf(lambda form: not form._fields.get("am_cor").data)])
ko_dod_id = StringField(validators=[RequiredIf(lambda form: form._fields.get("ko_invite").data)])
```

dandds:
ah, Patrick beat me by a minute

templates/task_orders/new/oversight.html
montana-mil: easy

atst/forms/task_order.py
montana-mil: i like the lambda solution better

atst/forms/validators.py
```diff
@@ -80,26 +80,22 @@ def _list_items_unique(form, field):
     return _list_items_unique
 
 
-def RequiredIfNot(other_field_name, message=translate("forms.validators.is_required")):
+def RequiredIf(other_field, message=translate("forms.validators.is_required")):
```
dandds: Can we change the `other_field` arg name? `other_field` implies that it's still a field name that gets passed.

atst/forms/task_order.py
```diff
@@ -128,11 +132,19 @@ class OversightForm(CacheableForm):
...
+            RequiredIf(lambda form: not form._fields.get("am_cor").data),
+            PhoneNumber(),
+        ],
     )
     cor_dod_id = StringField(
         translate("forms.task_order.oversight_dod_id_label"),
-        validators=[RequiredIfNot("am_cor"), Length(min=10), IsNumber()],
+        validators=[
+            RequiredIf(lambda form: not form._fields.get("am_cor").data),
+            RequiredIf(lambda form: form._fields.get("cor_invite").data),
```
patricksmithdds: I think these two don't play well together. I started a new task order, got to the oversight page and didn't check either the "I am the COR" or "Invite COR" checkboxes and the form will not submit because the COR DOD ID (that is hidden) is still required. Instead, I think we can combine these two like `RequiredIf(lambda form: (not form._fields.get("am_cor").data and form.fields.get("cor_invite").data)`

atst/forms/validators.py
```diff
@@ -80,26 +80,22 @@ def _list_items_unique(form, field):
...
     """ A validator which makes a field required only if another field
-        has a falsy value
+        has a truthy value
         Args:
-            other_field_name (str): the name of the field we check before
-                determining if this field is required; if this other field is falsy,
-                the field will be required
+            other_field_value (function): calling this on form results in
```
patricksmithdds: The arg name here doesn't match the name in the argument list.

atst/forms/validators.py
patricksmithdds: I think `other_field` is no longer accurate since this is now just a callable that can do _anything_ with the form.

atst/routes/task_orders/new.py
```diff
@@ -81,6 +81,10 @@ def form(self):
             elif self._section["section"] == "oversight":
                 if self.user.dod_id == self.task_order.cor_dod_id:
                     self._form.am_cor.data = True
+                if self.task_order.ko_dod_id:
+                    self._form.ko_invite.data = True
+                if self.task_order.so_dod_id:
+                    self._form.so_invite.data = True
```
- I think we need to check if the COR has been invited, too.
- I think it would be better to check if the TO has the relationship to each officer populated. (It has relationships like `contracting_officer`, etc., that are populated when the officer is really invited to the TO.)

atst/forms/task_order.py
montana-mil: 👍

atst/forms/validators.py
montana-mil: i dont really know what to call it. i was reading about lambda functions over the weekend, but i feel like i dont really have a great grasp on how this one works. I was more worried about using `form` as a keyword in the lambda function, thinking that might be confusing to read within the validator

PR #564, opened 2019-01-21
Hide tooltip when CLIN field is disabled

drummond-dds:


dandds:
@patricksmithdds In this specific case, since the disabled CLIN fields have additional text ("available soon") I think removing them is okay. They aren't on that page for the current wireframes. Probably worth asking for guidance in the design sync, tho.

PR #565, opened 2019-01-22
Require First TO Screen to be filled out

montana-mil:
## Description
We were running into some bugs when trying to begin a task order form without first providing a Portfolio Name. Since the Portfolio Name is given in the first screen, we now require that screen to be submitted before being able to navigate ahead to other pages. 

- The other pages are grayed out in the top navigation until screen one is complete.
- The user is able to navigate to Funding, Oversight, or Review once screen one is complete.
- The `Done` button at the bottom of the Review screen is disabled until all screens are completed.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163300624
https://www.pivotaltracker.com/story/show/163268456

## Screenshots
screen shot 2019-01-22 at 11 19 49 am
screen shot 2019-01-22 at 11 19 35 am
screen shot 2019-01-22 at 11 20 31 am

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -108,22 +108,13 @@ class ShowTaskOrderWorkflow:
 
         return screen_info
 
-    def completed(self):
-        screen_info = deepcopy(TASK_ORDER_SECTIONS)
-
+    @property
+    def is_complete(self):
         if self.task_order:
-            for section in screen_info:
-                if (
-                    not TaskOrders.is_section_complete(
-                        self.task_order, section["section"]
-                    )
-                    and section["section"] != "review"
-                ):
-                    return False
+            if TaskOrders.all_sections_complete(self.task_order):
+                return True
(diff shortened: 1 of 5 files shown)


atst/routes/task_orders/new.py
```diff
@@ -102,6 +102,23 @@ def display_screens(self):
...
+                    not TaskOrders.is_section_complete(
+                        self.task_order, section["section"]
+                    )
+                    and section["section"] != "review"
+                ):
+                    return False
+        else:
+            return False
+
+        return True
```
dandds: I think this is the same implementation as what's in `TaskOrders.all_sections_complete` (which we aren't using yet). Maybe try that in this method?

dandds: I'd also implement this as a property, since it acts like an attribute of the workflow. (I miss being able to add question marks to method names like in Ruby. Maybe `is_complete`.)

PR #568, opened 2019-01-23
Add v-cloak to Portfolio edit page

montana-mil:
## Description
A `!` was briefly showing instead of  `✓` on the name and description of a Portfolio in the Portfolio's edit page.  This PR adds a v-cloak to the portfolio edit form to prevent that render glitch.

## Pivotal Tracker

### DIFF
--- a/templates/components/text_input.html
+++ b/templates/components/text_input.html
@@ -15,6 +15,7 @@
   noMaxWidth=False) -%}
 
   <textinput
+    v-cloak
     name='{{ field.name }}'
     validation='{{ validation }}'
     {% if paragraph %}paragraph='true'{% endif %}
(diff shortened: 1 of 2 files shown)


patricksmithdds:
Should this be applied in the text input component? I'm curious if this happens everywhere we're using a text input with validation.

montana-mil:
@patricksmithdds it might even be everywhere we are using validation on an input

patricksmithdds:
> it might even be everywhere we are using validation on an input

Is it possible to add the `v-cloak` somewhere that applies to all of those validated inputs?

montana-mil:
@patricksmithdds i tried to figure out a way to do that, but I'm not coming up with anything. Did you have something in mind? Is it possible?

patricksmithdds:
> i tried to figure out a way to do that, but I'm not coming up with anything. Did you have something in mind? Is it possible?

What you did in 41e6500 is what I had in mind.

PR #570, opened 2019-01-23
Allow CCPO to create TO on anyone's portfolio

montana-mil:
## Description
A CCPO user is now able to create a new Task Order from any other user's Portfolio > Funding page. There was a bug in permissions that is now fixed.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163268456

### DIFF
--- a/tests/domain/test_portfolios.py
+++ b/tests/domain/test_portfolios.py
@@ -300,6 +300,10 @@ def test_get_for_update_information():
     ccpo = UserFactory.from_atat_role("ccpo")
     assert Portfolios.get_for_update_information(ccpo, portfolio.id)
 
+    developer = UserFactory.from_atat_role("developer")
+    with pytest.raises(UnauthorizedError):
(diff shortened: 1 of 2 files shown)


tests/domain/test_portfolios.py
```diff
@@ -298,8 +298,7 @@ def test_get_for_update_information():
     assert portfolio == admin_ws
 
     ccpo = UserFactory.from_atat_role("ccpo")
-    with pytest.raises(UnauthorizedError):
-        Portfolios.get_for_update_information(ccpo, portfolio.id)
+    assert Portfolios.get_for_update_information(ccpo, portfolio.id)
```
dandds: Very small thing, but now this test doesn't have any case where a user _can't_ get the portfolio in order to update it. Could you add one? Like, rando calrissian can't access it.

montana-mil: done!

PR #571, opened 2019-01-23
Use a custom JSON encoder

patricksmithdds:
This PR adds a custom JSON encoder for the flask app. Previously, we were using a special filter, `mixedContentToJson`, to serialize file objects (`Attachment` instances) in a specific way. Now, we can use the `tojson` filter everywhere and the custom JSON encoder can be updated as necessary in the future.

### DIFF
--- a/tests/utils/test_json.py
+++ b/tests/utils/test_json.py
@@ -1,0 +1,15 @@
+import json
+from atst.utils.json import CustomJSONEncoder
+
+from tests.factories import AttachmentFactory
+
+
+encoder = CustomJSONEncoder()
+
+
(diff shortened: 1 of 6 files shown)


tests/utils/test_json.py
```diff
@@ -0,0 +1,15 @@
...
+from atst.utils.json import CustomJSONEncoder
+
+from tests.factories import AttachmentFactory
+
+
+encoder = CustomJSONEncoder()
+
+
+def test_custom_encoder_serializes_attachments():
+    filename = "jar_jar_is_secretly_a_sith_lord.pdf"
```
dandds: 👺

PR #573, opened 2019-01-23
Show COR as invited if am_cor is checked

montana-mil:
## Description
After the TO has been completed and submitted, the user is brought to a "Next Steps" page. If the user had checked `am COR` on the Oversight screen, the COR shows as "invited."

COR also shows as "invited" at the bottom of the Review screen.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163067281

## Screenshots
screen shot 2019-01-23 at 3 34 10 pm
screen shot 2019-01-23 at 3 34 19 pm

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -264,7 +264,6 @@ def new(screen, task_order_id=None, portfolio_id=None):
         screens=workflow.display_screens,
         form=workflow.form,
         complete=workflow.is_complete,
-        user=g.current_user,
     )
 
 
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -158,6 +158,7 @@ class UpdateTaskOrderWorkflow(ShowTaskOrderWorkflow):
                 "cor_email": self.user.email,
                 "cor_phone_number": self.user.phone_number,
                 "cor_dod_id": self.user.dod_id,
+                "cor_id": self.user.id,
             }
             to_data = {**to_data, **cor_data}
 
@@ -263,6 +264,7 @@ def new(screen, task_order_id=None, portfolio_id=None):
         screens=workflow.display_screens,
         form=workflow.form,
         complete=workflow.is_complete,
+        user=g.current_user,
     )
 
 
--- a/templates/task_orders/new/review.html
+++ b/templates/task_orders/new/review.html
@@ -32,7 +32,7 @@
   </div>
 {% endmacro %}
 
-{% macro ReviewOfficerInfo(heading, first_name, last_name, email, phone_number, dod_id, invite, am_cor) %}
+{% macro ReviewOfficerInfo(heading, first_name, last_name, email, phone_number, dod_id, officer) %}
   <div class="col col--grow">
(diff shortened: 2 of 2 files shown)


templates/task_orders/new/review.html
```diff
@@ -184,7 +184,10 @@ <h3 class="subheading">{{ "task_orders.new.review.oversight"| translate }} {{ TO
 
 <div class="row">
   {{ ReviewOfficerInfo("task_orders.new.review.ko", task_order.ko_first_name, task_order.ko_last_name, task_order.ko_email, task_order.ko_phone_number, task_order.ko_dod_id, task_order.ko_invite) }}
-  {{ ReviewOfficerInfo("task_orders.new.review.cor", task_order.cor_first_name, task_order.cor_last_name, task_order.cor_email, task_order.cor_phone_number, task_order.cor_dod_id, task_order.cor_invite) }}
+  {% if task_order.cor_dod_id == user.dod_id%}
+    {% set am_cor = True%}
+  {% endif %}
+  {{ ReviewOfficerInfo("task_orders.new.review.cor", task_order.cor_first_name, task_order.cor_last_name, task_order.cor_email, task_order.cor_phone_number, task_order.cor_dod_id, task_order.cor_invite, am_cor) }}
```
dandds: There's actually a bug in the way we're using this macro that I didn't catch the first time around. `task_order.cor_invite` isn't a valid property on the task order model. Jinja evaluates it to `None` every time here. (It is a field on the TO overview form, but not a property on the model.) What we should check is whether the task order has the relationship to that officer: `task_order.contracting_officer` in this case.

I don't think we need any special handling for `am_cor` once we change this, because the COR relationship on the task order (`task_order.contracting_officer_representative`) will  be defined after the user submits the Overview form with that box checked.

atst/routes/task_orders/new.py
```diff
@@ -158,6 +158,7 @@ def task_order_form_data(self):
                 "cor_email": self.user.email,
                 "cor_phone_number": self.user.phone_number,
                 "cor_dod_id": self.user.dod_id,
+                "cor_id": self.user.id,
```
dandds: I doubt adding this does anything for us, but doublecheck me. I think the code is going to look the user up by the DOD ID either way.

templates/task_orders/new/review.html
montana-mil: makes sense

atst/routes/task_orders/new.py
montana-mil: i added this so that the `contracting_officer_representative` relationship is created

montana-mil: not sure if that's the way to do that!

dandds: nope, you're right

atst/routes/task_orders/new.py
```diff
@@ -263,6 +264,7 @@ def new(screen, task_order_id=None, portfolio_id=None):
         screens=workflow.display_screens,
         form=workflow.form,
         complete=workflow.is_complete,
+        user=g.current_user,
```
dandds: I don't think we need to pass this anymore, right?

PR #578, opened 2019-01-24
DOD ID Toggle Fixes

montana-mil:
## Description
This PR allows the user to toggle checkboxes on the Oversight screen without losing text that they had inputted. It also fixes a bug where if `Invite COR` was clicked, then `Am COR` was toggled, the text input field for the COR DOD ID would appear, but the checkbox for `Invite COR` was unchecked.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163436529
https://www.pivotaltracker.com/story/show/163067040

### DIFF
--- a/js/components/forms/oversight.js
+++ b/js/components/forms/oversight.js
@@ -2,15 +2,37 @@ import FormMixin from '../../mixins/form'
 import textinput from '../text_input'
 import checkboxinput from '../checkbox_input'
 
+const dodid = {
+  name: 'dodid',
+
+  mixins: [FormMixin],
+
+  components: {
+    textinput,
+  },
+
+  props: {
+    initialKoInvite: Boolean,
+    initialSoInvite: Boolean,
+  },
+
+  data: function() {
+    return {
+      ko_invite: this.initialKoInvite,
+      so_invite: this.initialSoInvite,
+    }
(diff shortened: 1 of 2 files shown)


js/components/forms/oversight.js
```diff
@@ -2,6 +2,49 @@ import FormMixin from '../../mixins/form'
...
+
+  props: {
+    initialKoInvite: Boolean,
+    initialSoInvite: Boolean,
+  },
+
+  data: function() {
+    return {
+      ko_invite: this.initialKoInvite,
+      so_invite: this.initialSoInvite,
```
dandds: I tinkered with this for a little while. I don't think you need separate invite types on the `dodid` component: just seed each instance with the right type of invite from the parent, like:

```
<dodid v-bind:initial-invite="ko_invite" ...>
```

It seems like we shouldn't need a separate `cordata` component, either, but I couldn't figure out a way to make that work. Maybe someone with more Vue experience has an idea. @patricksmithdds ?

montana-mil: It totally feels like we should be able to do it all in one component, but `<keep-alive>` only works when it's directly above the component you are going to be hiding.

I'll update the `dodid` component!

PR #581, opened 2019-01-25
Fix oversight component mistake

montana-mil:
## Description
Fixes a mistake I made on #578

### DIFF
--- a/js/components/forms/oversight.js
+++ b/js/components/forms/oversight.js
@@ -17,7 +17,8 @@ const dodid = {
 
   data: function() {
     return {
-      invite: this.initialInvite,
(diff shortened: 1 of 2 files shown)


js/components/forms/oversight.js
```diff
@@ -17,7 +17,8 @@ const dodid = {
 
   data: function() {
     return {
-      invite: this.initialInvite,
+      ko_invite: this.initialInvite,
+      so_invite: this.initialInvite
```
dandds: I think this line was fine as it was; an individual instance of the `dodid` component just has to know it has an invite, it doesn't have to care what kind.

PR #582, opened 2019-01-28
Accept decimals in CLIN fields

leigh-mil:
## Description
Fixes a bug where users could not input CLIN values that included cents. 
In `funding.js` I used `toLocalString()` instead of the mask because the mask did not have any settings that properly formatted values that included trailing zeros for cents values (ie: $1.00 or $1.10). 

## Pivotal
[https://www.pivotaltracker.com/story/show/163465960](https://www.pivotaltracker.com/story/show/163465960)

## Screenshots
screen shot 2019-01-28 at 10 14 26 am

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -109,10 +109,7 @@ class TaskOrders(object):
 
     @classmethod
     def all_sections_complete(cls, task_order):
-        sections = TaskOrders.SECTIONS
-        if not app.config.get("CLASSIFIED"):
-            sections["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
-        for section in sections:
+        for section in TaskOrders.SECTIONS.keys():
             if not TaskOrders.is_section_complete(task_order, section):
                 return False
 
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -94,10 +94,7 @@ class TaskOrders(object):
 
     @classmethod
     def is_section_complete(cls, task_order, section):
-        sections = TaskOrders.SECTIONS
-        if not app.config.get("CLASSIFIED"):
-            sections["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
-        if section in sections:
+        if section in TaskOrders.sections():
             for attr in TaskOrders.SECTIONS[section]:
                 if getattr(task_order, attr) is None:
                     return False
@@ -114,6 +111,13 @@ class TaskOrders(object):
                 return False
 
         return True
+
+    @classmethod
+    def sections(cls):
+        section_list = TaskOrders.SECTIONS
+        if not app.config.get("CLASSIFIED"):
+            section_list["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
+        return section_list
 
     OFFICERS = [
         "contracting_officer",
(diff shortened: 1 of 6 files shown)


atst/forms/task_order.py
```diff
@@ -97,20 +97,44 @@ class FundingForm(CacheableForm):
...
         render_kw={"accept": ".pdf,.png,application/pdf,image/png"},
     )
-    clin_01 = IntegerField(translate("forms.task_order.clin_01_label"))
-    clin_02 = IntegerField(translate("forms.task_order.clin_02_label"))
-    clin_03 = IntegerField(translate("forms.task_order.clin_03_label"))
-    clin_04 = IntegerField(translate("forms.task_order.clin_04_label"))
+    clin_01 = DecimalField(
+        translate("forms.task_order.clin_01_label"),
+        places=2,
+        validators=[InputRequired(message="Please enter a dollar amount")],
```
dandds: Any reason not to put this in the translations file? (I know the JS can't use the translations yet.)

atst/forms/task_order.py
```diff
@@ -97,20 +97,44 @@ class FundingForm(CacheableForm):
...
+    clin_02 = DecimalField(
+        translate("forms.task_order.unclassified_clin_02_label"),
+        places=2,
+        validators=[InputRequired(message="Please enter a dollar amount")],
+    )
+    clin_04 = DecimalField(
+        translate("forms.task_order.unclassified_clin_04_label"),
+        places=2,
+        validators=[InputRequired(message="Please enter a dollar amount")],
+    )
```
dandds: These shouldn't be required in an unclassified context.

atst/forms/task_order.py
```diff
@@ -97,20 +97,44 @@ class FundingForm(CacheableForm):
...
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.clin_02.data = "0"
-        self.clin_04.data = "0"
+        self.clin_02.data = "0.00"
+        self.clin_04.data = "0.00"
```
dandds: The problem doesn't originate with this PR, but I just noticed it and think we should fix it here. What happens here is that we write zero values for the classified fields to the database, which is wrong. If a column doesn't have a value it should be `NULL`. What happens if we just drop the `InputRequired` validators and this `init` override?

atst/domain/task_orders.py
```diff
@@ -102,7 +109,10 @@ def is_section_complete(cls, task_order, section):
 
     @classmethod
     def all_sections_complete(cls, task_order):
-        for section in TaskOrders.SECTIONS.keys():
+        sections = TaskOrders.SECTIONS
```
drummond-dds: I think we can leave this method as it was originally. All the actual logic for if the section is valid is in `is_section_complete` and we do the classified/unclassified logic there.

leigh-mil: good catch!

atst/domain/task_orders.py
```diff
@@ -90,7 +94,10 @@ def update(cls, user, task_order, **kwargs):
 
     @classmethod
     def is_section_complete(cls, task_order, section):
-        if section in TaskOrders.SECTIONS:
+        sections = TaskOrders.SECTIONS
+        if not app.config.get("CLASSIFIED"):
+            sections["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
+        if section in sections:
```
dandds: Nitpicky, but I think there should be a `TaskOrders` domain method that returns the correct section dictionary, instead of building it here.

PR #520, opened 2019-01-08
Task order validations

drummond-dds:
This PR implements the ["Save as Draft" feature](https://www.pivotaltracker.com/story/show/162772188).

When clicking "Save as Draft" we set a hidden field via the vue component called `is_draft` to `True`. This field is only an attribute and it does not get saved to the database. When running the `validate()` on a form when the `is_draft` attribute is set to `True` we run through the validations we run the first validation (`OptionalAsDraft()`) which returns a status as `valid=True` for that field and halts all further validations down the chain.

**NOTE**: `OptionalAsDraft()` must be the first validation to run.

When we want to check the completion status we run the method `section_completion_status()` which replaces `is_section_complete()` and this method now returns the form steps completion state of either `["complete", "draft", "incomplete"]`.

The form steps component now uses the `section_completion_status` to display either complete, draft, incomplete in the timeline at the top.

The method `all_sections_complete()` has also been changed to make use of `section_completion_status()`.

PR #585, opened 2019-01-29
Save as draft without draft button

drummond-dds:
This PR replaces the old "Save as Draft" feature with the "Save as Draft" button - https://github.com/dod-ccpo/atst/pull/520

This implementation allows saving of the form without all the data required, but validation rules such as phone number/email/dod format are enforced when data is present. I have also renamed and reimplemented the `is_section_complete` to `section_completion_status` which now checks that all the required fields are present before marking that section as complete, draft, or incomplete.

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/162772188)

---

screen shot 2019-01-29 at 14 04 17

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -16,6 +16,7 @@ class TaskOrderError(Exception):
 class TaskOrders(object):
     SECTIONS = {
         "app_info": [
+            "portfolio_name",
             "scope",
             "defense_component",
             "app_migration",
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -1,5 +1,4 @@
 from sqlalchemy.orm.exc import NoResultFound
-
 from flask import current_app as app
 
 from atst.database import db
@@ -93,21 +92,33 @@ class TaskOrders(object):
         return task_order
 
     @classmethod
-    def is_section_complete(cls, task_order, section):
-        if section in TaskOrders.sections():
+    def section_completion_status(cls, task_order, section):
+        if section in TaskOrders.SECTIONS:
+            passed = []
+            failed = []
+
             for attr in TaskOrders.SECTIONS[section]:
-                if getattr(task_order, attr) is None:
-                    return False
+                if not app.config.get("CLASSIFIED") and attr in ["clin_02", "clin_04"]:
+                    pass
+                elif not getattr(task_order, attr):
+                    failed.append(attr)
+                else:
+                    passed.append(attr)
 
-            return True
-
-        else:
-            return False
+            if not failed:
+                return "complete"
+            elif passed and failed:
+                return "draft"
+            else:
+                return "incomplete"
 
     @classmethod
     def all_sections_complete(cls, task_order):
         for section in TaskOrders.SECTIONS.keys():
-            if not TaskOrders.is_section_complete(task_order, section):
+            if (
+                TaskOrders.section_completion_status(task_order, section)
+                is not "complete"
+            ):
                 return False
 
         return True
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -93,7 +93,7 @@ class TaskOrders(object):
 
     @classmethod
     def section_completion_status(cls, task_order, section):
-        if section in TaskOrders.SECTIONS:
+        if section in TaskOrders.mission_owner_sections():
             passed = []
             failed = []
 
@@ -124,7 +124,7 @@ class TaskOrders(object):
         return True
 
     @classmethod
-    def sections(cls):
+    def mission_owner_sections(cls):
         section_list = TaskOrders.SECTIONS
         if not app.config.get("CLASSIFIED"):
             section_list["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -99,12 +99,10 @@ class TaskOrders(object):
             failed = []
 
             for attr in TaskOrders.SECTIONS[section]:
-                if not app.config.get("CLASSIFIED") and attr in ["clin_02", "clin_04"]:
-                    pass
-                elif not getattr(task_order, attr):
-                    failed.append(attr)
-                else:
+                if getattr(task_order, attr):
                     passed.append(attr)
+                else:
+                    failed.append(attr)
 
             if not failed:
                 return "complete"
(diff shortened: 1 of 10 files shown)


drummond-dds:
@dandds I rebased

atst/domain/task_orders.py
```diff
@@ -93,21 +92,33 @@ def update(cls, user, task_order, **kwargs):
...
     @classmethod
-    def is_section_complete(cls, task_order, section):
-        if section in TaskOrders.sections():
-            for attr in TaskOrders.SECTIONS[section]:
-                if getattr(task_order, attr) is None:
-                    return False
-
-            return True
+    def section_completion_status(cls, task_order, section):
+        if section in TaskOrders.SECTIONS:
```
dandds: I think that this method should just use the `TaskOrders.sections()` method, which returns the correct list based on the app config. (Incidentally, I think that method should be a property, like `TaskOrders.mission_owner_sections`, if you don't mind making that change while you're at it.)

atst/forms/task_order.py
```diff
@@ -10,7 +10,7 @@
 )
 from wtforms.fields.html5 import DateField, TelField
 from wtforms.widgets import ListWidget, CheckboxInput
-from wtforms.validators import Length, InputRequired
+from wtforms.validators import Length, Required, InputRequired, Optional
```
dandds: I think that any radio button field is going to need the `Optional` validator. I can't get past the first screen without selecting something to "App Migration", etc.

dandds: Also note that when you do this, it will cause problems on the review page that have to be resolved. The review page template uses the app migration, etc., choice to look up the correct text in the translations file, and will try to look up an empty string.

tests/routes/task_orders/test_new_task_order.py
```diff
@@ -170,9 +170,9 @@ def test_show_task_order_display_screen(task_order):
     screens = workflow.display_screens
     # every form section is complete
     for i in range(2):
-        assert screens[i]["complete"]
+        assert screens[i]["completion"] == "complete"
     # the review section is not
-    assert not screens[3].get("complete")
+    assert not screens[3]["completion"]
```
dandds: Just curious: why is this `None` and not `"incomplete"`?

drummond-dds: Fixed in 88683467134208bdb3178aa74381cffb1aea660b

atst/domain/task_orders.py
drummond-dds: I modified the method to use `TaskOrders.sections()` renamed to `TaskOrders.mission_owner_sections()` 7587ba8c1da0f487a4ee27219c06bc600e0bb450

I don't know about making it a property. Wouldn't that need to be on an instance of the class? Then it might not make as much sense?

atst/forms/task_order.py
drummond-dds: Fixed in 36bb03907051470e7d4e29aff87fbead16dadd66

atst/domain/task_orders.py
dandds: Hmm I think you're right, it might not be possible in Python to have a class property. (Semantically I think that's what it is, though.)

atst/forms/task_order.py
```diff
@@ -100,25 +105,37 @@ class FundingForm(CacheableForm):
     clin_01 = DecimalField(
         translate("forms.task_order.clin_01_label"),
         validators=[
-            InputRequired(message=(translate("forms.task_order.clin_validation_error")))
+            Optional(),
+            InputRequired(
+                message=(translate("forms.task_order.clin_validation_error"))
+            ),
```
dandds: I didn't notice this before, but `InputRequired` is probably useless here. Without that, the field will still generate validation errors if you manage to POST data that can't be coerced to a decimal. (If we could figure out how to customize that message so that it's `forms.task_order.clin_validation_error`, that would be great, but glancing at the docs I don't see how.)

templates/task_orders/new/review.html
```diff
@@ -67,8 +67,21 @@ <h3 class="subheading">{{ "task_orders.new.review.app_info"| translate }} {{ TOE
 <h3 class="subheading">{{ "task_orders.new.review.reporting"| translate }} {{ TOEditLink(screen=1, anchor="reporting") }}</h3>
 
 <div class="row">
-  {{ ReviewField(("forms.task_order.app_migration.label" | translate), ("forms.task_order.app_migration.{}".format(task_order.app_migration) | translate), filter="removeHtml") }}
-  {{ ReviewField(("forms.task_order.native_apps.label" | translate), ("forms.task_order.native_apps.{}".format(task_order.native_apps))| translate) }}
+  {{
+    ReviewField(("forms.task_order.app_migration.label" | translate),
+      (
+        ("forms.task_order.app_migration.{}".format(task_order.app_migration) | translate) if task_order.app_migration else RequiredLabel()
```
dandds: The macro `ReviewField` is supposed to apply the `RequiredLabel` if no `field` is passed. I'd pass `None` for the second argument to the macro in these cases. In this specific case, the `removeHTML` filter gets called for the `RequiredLabel`:

screen shot 2019-02-01 at 1 18 36 pm

atst/domain/task_orders.py
```diff
@@ -93,27 +93,39 @@ def update(cls, user, task_order, **kwargs):
...
-        if section in TaskOrders.sections():
+    def section_completion_status(cls, task_order, section):
+        if section in TaskOrders.mission_owner_sections():
+            passed = []
+            failed = []
+
             for attr in TaskOrders.SECTIONS[section]:
-                if getattr(task_order, attr) is None:
-                    return False
+                if not app.config.get("CLASSIFIED") and attr in ["clin_02", "clin_04"]:
```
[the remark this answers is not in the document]
drummond-dds: Fixed in 4daadf59d2ec792b09730eb15a42eecc6c2b8b41

templates/task_orders/new/review.html
drummond-dds: How do the changes in 0ace907fdf7461ab890e04d596fc12640142158c look?

atst/forms/task_order.py
drummond-dds: I've removed that validation in 0ace907fdf7461ab890e04d596fc12640142158c

Do you want to work out how to translate the decimal error message in this PR or a follow up?

dandds: Actually, thinking about it, the error message there doesn't matter. It's not possible to enter non-numeric data in the form field, so you'd have to POST data some other way, in which case, you're on your own

PR #587, opened 2019-01-31
Bug fixes for date picker component

drummond-dds:
This PR fixes the following bugs

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/162771867)

- [x] Input length limits enforced
- [x] Dates in error message are in the correct format
- [x] Page doesn't shift with error message
- [x] Fixed persistence issues with invalid dates

### DIFF
--- a/atst/routes/users.py
+++ b/atst/routes/users.py
@@ -38,4 +38,11 @@ def update_user():
         if next_url:
             return redirect(next_url)
 
-    return render_template("user/edit.html", form=form, user=user, next=next_url)
+    return render_template(
+        "user/edit.html",
+        form=form,
+        user=user,
+        next=next_url,
+        mindate=(dt.datetime.now() - dt.timedelta(days=365)),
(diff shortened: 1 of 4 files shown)


drummond-dds:
@dandds I thought since we are going the ES6 way we should go full in. [I wasn't aware of the syntax before and I think its quite nice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#Description).

I will remove the old `DateInput` in a follow up commit.

patricksmithdds:
@drummond-dds 

> I thought since we are going the ES6 way we should go full in. [I wasn't aware of the syntax before and I think its quite nice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions#Description).

It looks like that's not supported in IE -- does `parcel` handle transpiling that for us?

drummond-dds:
@dandds parcel transpiles it for us

PR #591, opened 2019-02-01
update k8s config so auth traffic is only directed to web pods

dandds:
We were seeing an issue where some auth traffic was being directed to the workers pods, which are only responsible for processing background jobs. This updates our k8s config to make labels and selectors for HTTP traffic more specific.

### DIFF
--- a/deploy/kubernetes/atst.yml
+++ b/deploy/kubernetes/atst.yml
@@ -12,6 +12,9 @@ metadata:
   name: atst
   namespace: atat
 spec:
+  selector:
+    matchLabels:
+      role: web
   replicas: 2
   strategy:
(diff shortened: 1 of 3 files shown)


ddsdevon:
If the worker only does background jobs, should it be receiving requests from any of the Services (there is both a "regular" atst service and an auth one I think)?
If not, I recommend changing the label on the worker Deployment instead of trying to catch all the references for the regular atst nodes.

dandds:
@ddsdevon Yeah, that makes sense. Patrick and I talked about leaving the existing label so we can still use it for CLI queries and adding a new `role` label that's more specific to what the pod is for. I've updated the PR with that change.

PR #606, opened 2019-02-06
Upload PDF on KO Review Page

montana-mil:
## Description
Adds a PDF uploader to the KO Review page

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163368662

## Screenshots
screen shot 2019-02-06 at 10 01 35 am
screen shot 2019-02-06 at 10 02 05 am
screen shot 2019-02-06 at 10 01 30 am


templates/portfolios/task_orders/review.html
```diff
@@ -58,11 +61,28 @@ <h1 class="task-order-form__heading subheading">
...
+                </div>
+              </template>
+              <template v-else>
+                <p>Uploaded {{ form.pdf.data.filename }}</p>
+                <div>
+                  <button type="button" v-on:click="showUploadInput">Change</button>
+                </div>
+              </template>
+            </div>
+          </upload>
```
dandds: I think that we should pull this into an `UploadInput` macro that we can use here and in the Funding section of the TO form. Both could use your new JS component, and then we can delete that code from the `funding` Vue component.

atst/models/task_order.py
```diff
@@ -81,16 +83,25 @@ def csp_estimate(self):
...
+
+    def _set_attachment(self, new_attachment, attribute):
+        if isinstance(new_attachment, Attachment):
+            return new_attachment
+        elif isinstance(new_attachment, FileStorage):
+            return Attachment.attach(new_attachment, "task_order", self.id)
+        elif not new_attachment and hasattr(self, attribute):
+            return None
+        else:
+            raise TypeError("Could not set attachment with invalid type")
```
dandds: :+1: Nice. We might even want this as a utility function eventually so other models can use it.

dandds:
@montana-mil Oh, also, do we have a test where the KO submits a completed form here? That would be good to add.

alembic/versions/1f690989e38e_add_pdf_to_task_order.py
```diff
@@ -0,0 +1,36 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('task_orders', sa.Column('pdf_attachment_id', postgresql.UUID(as_uuid=True), nullable=True))
+    op.drop_constraint('task_orders_attachments_attachment_id', 'task_orders', type_='foreignkey')
+    op.alter_column('task_orders', 'attachment_id', new_column_name='csp_attachment_id')
+    op.create_foreign_key(None, 'task_orders', 'attachments', ['pdf_attachment_id'], ['id'])
+    op.create_foreign_key(None, 'task_orders', 'attachments', ['csp_attachment_id'], ['id'])
```
dandds: These foreign keys need names, btw. Didn't notice until I tried to downgrade to move to another branch.

PR #608, opened 2019-02-06
Reskin navigation, phase 1

patricksmithdds:
This PR updates the top and side navigation to match the new designs. The colors and links on the top bar hav been updated:

image

Once logged in, the user will see a side nav with all of the portfolios that they are a member of on any page (for now, ignore the portfolio-specific side nav -- that will get removed in the next branch in favor of more contextual links):

image

Hovering over a different portfolio:

image

Viewing a page that has no current portfolio just shows without the highlight:

image

The sidebar currently minimizes/maximizes based on window size. The minimized sidebar has a mock "Show" link that currently does nothing:

image

I chatted a bit with Upma and she agreed that this is a good enough start -- we can touch up some places as necessary if everyone else agrees. 

One non-critical bug I noticed is that the white part of the sidebar does not extend to the bottom of the page on IE10. That will be addressed separately.

Another change is that the page listing portfolios is no longer linked anywhere. The "home" route will now redirect to the alphabetically first portfolio that the user has access to.


atst/app.py
```diff
@@ -90,6 +91,14 @@ def _set_globals():
         g.Authorization = Authorization
         g.Permissions = Permissions
 
+    @app.context_processor
+    def _portfolios():
+        if not g.current_user:
+            return {}
+
+        portfolios = Portfolios.for_user(g.current_user)
+        return {"portfolios": portfolios}
```
dandds: moar context processors!

atst/routes/__init__.py
```diff
@@ -70,7 +71,13 @@ def home():
                 url_for("portfolios.portfolio_applications", portfolio_id=portfolio_id)
             )
     else:
-        return redirect(url_for("portfolios.portfolios"))
+        portfolios = Portfolios.for_user(g.current_user)
+        first_portfolio = sorted(portfolios, key=lambda portfolio: portfolio.name)[0]
```
dandds: Batting around in my head whether the `Portfolios.for_user` should sort these on the query. Maybe that's heavy-handed.

atst/routes/portfolios/__init__.py
```diff
@@ -15,14 +15,12 @@
...
 @portfolios_bp.context_processor
 def portfolio():
-    portfolios = Portfolios.for_user(g.current_user)
     portfolio = None
     if "portfolio_id" in http_request.view_args:
         try:
             portfolio = Portfolios.get(
                 g.current_user, http_request.view_args["portfolio_id"]
             )
-            portfolios = [ws for ws in portfolios if not ws.id == portfolio.id]
```
dandds: `ws` for `workspace` probably escaped my attention in a bunch of places. :eyes:

styles/elements/_sidenav.scss
```diff
@@ -1,22 +1,53 @@
 .sidenav {
```
dandds: File under "not critical", but the top of the side nav overlaps the topbar by about a pixel. (I did a quick inspect and didn't immediately see why that would be.)

templates/base.html
```diff
@@ -21,7 +21,9 @@
     {% include 'navigation/topbar.html' %}
 
     <div class='global-layout'>
-      {% include 'navigation/global_navigation.html' %}
+      {% block global_sidenav %}
+        {% include 'navigation/global_sidenav.html' %}
+      {% endblock %}
```
dandds: Would we override this some place?

styles/elements/_sidenav.scss
patricksmithdds: I think I remember seeing some "margin: -1px", that may be causing it.

templates/base.html
patricksmithdds: I thought we would, but I'm not sure. I'm thinking we may want to when the user has no portfolios, for example.

styles/elements/_sidenav.scss
patricksmithdds: After you pointed that 1px out, it was super annoying! I removed the negative margin in bd063b4, so that should be better now.

PR #611, opened 2019-02-07
Reskin portfolio navigation

patricksmithdds:
This PR updates the portfolio-specific navigation. On each page inside a portfolio, a header is shown that has breadcrumbs, basic information about the portfolio and links to different sections for the portfolio.

Screenshots!

Applications:

image

Reports:

image

Funding:

image

Admin:

image

The breadcrumbs currently only show one level deep. For example, looking at a specific task order still shows the "Funding" crumb:

image

Breadcrumbs are set explicitly in the specific templates via a `secondary_breadcrumb` variable. Currently, the breadcrumbs are a single unclickable text element, but in the future, I envision this being a list of elements with text and a url. 

Clicking the portfolio name in the breadcrumbs will take you back to the applications page.


atst/routes/portfolios/__init__.py
```diff
@@ -11,6 +11,7 @@
 from atst.domain.portfolios import Portfolios
 from atst.domain.authz import Authorization
 from atst.models.permissions import Permissions
+from atst.utils.localization import translate
```
dandds: This looks like an unused import, but code checker isn't squawking about it. 🤔

templates/portfolios/header.html
```diff
@@ -0,0 +1,51 @@
...
+{% endmacro %}
+
+<div class='portfolio-header row'>
+  <div class='col col--grow'>
+    <div class='portfolio-header__name'>
+      {{ portfolio.name }}
+    </div>
+    <div class='portfolio-header__budget row'>
+      <span>Available budget</span>
+      <button type="button" tabindex="0" class="icon-tooltip" v-tooltip.right="{content: 'The available budget shown includes the available budget of all active task orders', container: false}">
```
dandds: We could file this as a separate bug (I don't know when the problem started) but the tooltip looks pretty janky now.

tests/routes/portfolios/test_applications.py
```diff
@@ -38,6 +39,7 @@ def test_user_without_permission_has_no_budget_report_link(client, user_session)
     )
 
 
+@pytest.mark.skip(reason="Temporarily no add activity log link")
 def test_user_with_permission_has_activity_log_link(client, user_session):
```
dandds: I don't have a solution in mind, but situations like this are why tests that measure very specific things about the page bug me.

atst/routes/portfolios/__init__.py
patricksmithdds: Good catch!

templates/portfolios/header.html
dandds: screen shot 2019-02-07 at 10 48 55 am

patricksmithdds: I agree it looks a little wonky. I think this is because we usually use top placement for tooltips, but we're using the right placement here. Looking at the styleguide, the little arrow fella seems intentional for right tooltips, so I think it's good to file a separate bug for this.

PR #613, opened 2019-02-08
Reskin portfolio applications page

patricksmithdds:
This PR reskins the portfolio applications page to match new designs:

image

When the page loads, all applications are set to show the list of environments. The environments can be toggled via the +/- button at the top right of each application.

image

The "Team" link shows the number of users in each application (a new property on applications was added for this) and links to the _portfolio_ members page.

If a user does not have permission to edit an application settings, the link is not shown:

image


templates/portfolios/applications/index.html
```diff
@@ -3,52 +3,88 @@
...
+      action_href=url_for('portfolios.new_application', portfolio_id=portfolio.id) if can_create_applications else None,
+      icon='cloud',
+      sub_message=None if can_create_applications else 'Please contact your JEDI Cloud portfolio administrator to set up a new application.'
+    ) }}
+
+  {% else %}
+
+    <div class='application-list'>
+      {% for application in portfolio.applications|sort(attribute='name') %}
+        <div is='toggler' default-visible v-cloak class='accordian application-list-item'>
```
dandds: Did we decide visible by default? I thought it was invisible, but I could be mistaken. Either way, neat that you can do that with just inline Vue.

templates/portfolios/applications/index.html
```diff
@@ -3,52 +3,88 @@
...
+                    <div class='separator'></div>
+                  {% endif %}
+                  <a class='icon-link' href='{{ url_for("portfolios.portfolio_members", portfolio_id=portfolio.id) }}'>
+                    <span>{{ "portfolios.applications.team_text" | translate }}</span>
+                    <span class='counter'>{{ application.num_users }}</span>
+                  </a>
+                </div>
+              </div>
+              <div class='col'>
+                <a v-on:click="props.toggle" href='#' class='icon-link toggle-link'>
```
dandds: Maybe it's bad semantics, but this might looks better as a `span`. I get an outline around it in FF when I click it.

templates/portfolios/applications/index.html
patricksmithdds: You are correct. I updated that in 2374b063.

templates/portfolios/applications/index.html
patricksmithdds: Good call! Wasn't sure how to make that go away. Fixed in 0ce90133.

PR #616, opened 2019-02-08
Send officer invitations after TO form complete

dandds:
This started as a resolution for this bug:

https://www.pivotaltracker.com/story/show/163334792

Now, we'll send officer invitations only after the mission owner has completed their part of the TO process. This means that they have to fill in the entire TO form and click "done" on the review screen. As a consequence of this, any officers they intend to invite are technically not invited the first time the MO looks at the TO review page. To address this, I introduced an additional invitation state for that page, "pending TO completion".

I also added a method to the TO model so that it can provide a dictionary of data for an officer type given as an argument.


atst/routes/task_orders/invite.py
```diff
@@ -3,17 +3,67 @@
...
+        flash("task_order_congrats", portfolio=portfolio)
+        return redirect(
+            url_for(
+                "portfolios.view_task_order",
+                portfolio_id=task_order.portfolio_id,
+                task_order_id=task_order.id,
+            )
+        )
+    else:
+        return redirect(
```
patricksmithdds: It looks like this will keep the user on the page they're currently on? I think that's fine, but we should probably `flash` a message telling them why they're still seeing the same page.

atst/models/task_order.py
```diff
@@ -156,6 +159,44 @@ def portfolio_name(self):
...
+        return self.so_invite and not self.security_officer
+
+    _OFFICER_PREFIXES = {
+        "contracting_officer": "ko",
+        "contracting_officer_representative": "cor",
+        "security_officer": "so",
+    }
+    _OFFICER_PROPERTIES = ["first_name", "last_name", "phone_number", "email", "dod_id"]
+
+    def officer_dictionary(self, officer_type):
```
patricksmithdds: 😍

dandds:
@patricksmithdds Made the flash update. Rachel is going to revisit the content for the officer invites on the Overview screen.

PR #622, opened 2019-02-11
Oversight invites styling

leigh-mil:
## Description
Made the following design fixes:
1) Added title and description to edit form
2) Made field titles normal weight
3) Added DoD ID field and corresponding checkbox

## Pivotal
[https://www.pivotaltracker.com/story/show/163463465](https://www.pivotaltracker.com/story/show/163463465)

## Screenshots
screen shot 2019-02-11 at 11 47 27 am

### DIFF
--- a/translations.yaml
+++ b/translations.yaml
@@ -255,6 +255,10 @@ forms:
   portfolio:
     name_label: Portfolio Name
     name_length_validation_message: Portfolio names must be at least 4 and not more than 100 characters
+  officers:
+    contracting_officer_invite: Invite KO to Task Order Builder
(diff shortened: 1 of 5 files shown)


translations.yaml
```diff
@@ -461,21 +465,25 @@ task_orders:
       not_uploaded: Not Uploaded
   invitations:
     dod_id_label: DoD ID
+    edit_text: A new invitation e-mail will be sent when new information is saved.
```
patricksmithdds: I don't think this is true -- if you edit the name, for example, we don't send any emails.

PR #624, opened 2019-02-11
Toggle sidenav

dandds:
https://www.pivotaltracker.com/story/show/163739494

This implements most of the sidenav toggle functionality. The sidenav is open by default. When the user clicks "show" or "hide", we save their selection in a cookie that then determines whether the nav is open or closed on subsequent page loads.

screen shot 2019-02-11 at 2 43 18 pm
screen shot 2019-02-11 at 2 43 31 pm

**notes**
- This behavior is not responsive yet. We want the collapsed version of the nav to be the default for smaller screen sizes, and the nav should take up the full width when expanded. I (or someone else) can work on this on a subsequent PR.
- The cookie implementation is dead simple: a key/value, and the path. If we think we need additional attributes, we can talk about it.
- This implementation is not pixel-perfect to the designs. In particular, we need larger chevrons/angle brackets for the "show" and "hide" toggles. (I couldn't find a unicode character that fit the bill.)
- Moving between pages feels wonky. The sidenav disappears completely and then reappears in a way that seems more noticeable than other page elements. I applied `v-cloak` so that it at least doesn't appear with partial content.


templates/navigation/global_sidenav.html
```diff
@@ -2,23 +2,33 @@
...
-  <div class="sidenav__title">Show >>></div>
+<div v-cloak is="SidenavToggler" class="sidenav-container">
+  <template slot-scope='props'>
+    <div v-bind:class="{'global-navigation': true, 'sidenav': props.isVisible, 'sidenav--minimized': !props.isVisible}">
+      <a href="#" v-on:click="props.toggle" class="sidenav__toggle">
+        <template v-if="props.isVisible">
+          << Hide
+        </template>
+        <template v-else>
+          Show >>
```
patricksmithdds: I think we can use [`angle-double-right`](https://fontawesome.com/icons/angle-double-right?style=solid) instead of ` >>` here (and `angle-double-left` for the hide). We currently don't have that included in icons yet, but it should be straight forward to copy it in.

js/components/sidenav_toggler.js
```diff
@@ -0,0 +1,30 @@
...
+        }
+      },
+    },
+  },
+
+  methods: {
+    toggle: function(e) {
+      e.preventDefault()
+      this.isVisible = !this.isVisible
+      document.cookie = cookieName + '=' + this.isVisible + '; path=/'
```
patricksmithdds: Do we need to worry about clearing out this cookie when logging out? I kind of like that it remembers my setting, but that only works if we assume people are not sharing the same browser.

dandds: good idea

PR #625, opened 2019-02-11
Blank slate for no portfolios

drummond-dds:
New user landing page for when a user has no portfolios. We can ship it without the "sad icon" in the ticket, and the `TODO` text will be added in a follow up PR.

[Pivotal Tracker](https://www.pivotaltracker.com/story/show/163272393).

screen shot 2019-02-11 at 16 07 38

### DIFF
--- a/templates/portfolios/blank_slate.html
+++ b/templates/portfolios/blank_slate.html
@@ -19,7 +19,6 @@
   <div class="center">
     {{
       Tooltip(
-        message=("portfolios.index.empty.help_message" | translate),
         title=("portfolios.index.empty.help_title" | translate),
       )
     }}
--- a/templates/portfolios/blank_slate.html
+++ b/templates/portfolios/blank_slate.html
@@ -15,13 +15,5 @@
       message=("portfolios.index.empty.title" | translate),
     )
   }}
-
-  <div class="center">
-    {{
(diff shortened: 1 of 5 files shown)


templates/portfolios/blank_slate.html
```diff
@@ -0,0 +1,28 @@
...
+      icon="cloud",
+      message=("portfolios.index.empty.title" | translate),
+    )
+  }}
+
+  <div class="center">
+    {{
+      Tooltip(
+        message=("portfolios.index.empty.help_message" | translate),
+        title=("portfolios.index.empty.help_title" | translate),
```
patricksmithdds: I'm curious if it's better to either leave this out entirely if we just have a "todo" message, or come up with a message ourselves?

drummond-dds: Good idea, removed it in f22fda4e28d3833b9e4313011418d4801b1bb217

patricksmithdds: Should we remove the _whole_ tooltip? I don't think there's a reason to have it in there unless we can actually give the user some contextual info.

drummond-dds: Makes sense - I removed that in 6172133

PR #626, opened 2019-02-11
Update styles on portfolio funding page

patricksmithdds:
Update styles on the portfolio funding page to match new designs:

localhost_8000_portfolios_c6f5562a-7562-4490-825e-0209818bd215_task_orders

localhost_8000_portfolios_2e1755b0-f589-494a-976b-41d0a1749c8c_task_orders

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -100,15 +100,20 @@
   font-weight: bold;
 }
 
+@mixin shadow-panel {
+  padding: $gap / 2 0;
+  box-shadow: 0 6px 18px 0 rgba(144,164,183,0.3);
+  border-top: none;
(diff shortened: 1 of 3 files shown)


styles/components/_portfolio_layout.scss
```diff
@@ -100,15 +100,20 @@
   font-weight: bold;
 }
 
+@mixin shadow-panel {
```
leigh-mil: could we define `shadow-panel` as a variable since we are going to need to use it in a bunch of places?

patricksmithdds: That's a good idea. I moved it to `_panels.scss` because it's still kinda related to panels, but you should be able to use it anywhere you want to now!

PR #627, opened 2019-02-11
Portfolio admin reskin

leigh-mil:
## Description
Updates to the Portfolio Admin page based on the new designs. 
As per a conversation with Upma, I changed the url from `/edit` to `/admin` and removed the `Cancel` button. 

## Pivotal
[https://www.pivotaltracker.com/story/show/163834239](https://www.pivotaltracker.com/story/show/163834239)

## Screenshots
screen shot 2019-02-11 at 4 38 31 pm


atst/routes/portfolios/index.py
```diff
@@ -18,11 +18,22 @@ def portfolios():
...
-    return render_template("portfolios/edit.html", form=form, portfolio=portfolio)
+    pagination_opts = Paginator.get_pagination_opts(http_request)
+    audit_events = AuditLog.get_portfolio_events(
+        g.current_user, portfolio, pagination_opts
+    )
+    return render_template(
+        "portfolios/edit.html",
+        form=form,
+        portfolio=portfolio,
+        portfolio_name=portfolio.name,
```
patricksmithdds: I don't think these `portfolio_*` args are necessary if you're passing the `portfolio` into `render_template`.

templates/fragments/audit_events_log.html
```diff
@@ -1,10 +1,7 @@
 {% from "components/pagination.html" import Pagination %}
 
-<section class="block-list">
-  <header class="block-list__header">
-    <h1 class="block-list__title">{{ "audit_log.header_title" | translate }}</h1>
-  </header>
-
+<section class="block-list activity-log">
+  <div class='subheading'>Activity Log</div>
```
drummond-dds: Should this be a translation?

styles/elements/_block_lists.scss
```diff
@@ -15,6 +16,11 @@
   display: flex;
   flex-direction: row;
   justify-content: space-between;
+  background-color: $color-gray-lightest;
+  padding: $gap 2 * $gap;
+  border-top: none;
+  border-bottom: none;
```
patricksmithdds: I think this affected the applications list, as well -- could you update the styles on the applications to add more padding?

Before:

image

After:

image

The toggle buttons are too close to the top border now.

styles/elements/_block_lists.scss
```diff
@@ -1,5 +1,6 @@
 @mixin block-list {
   @include panel-margin;
+  box-shadow: 0 6px 18px 0 rgba(144,164,183,0.3);
```
drummond-dds: We have this `rgba(144,164,183,0.3)` in a few places now.  I wonder if we should extract it out to a variable like we have for `$color-white` etc.

styles/elements/_block_lists.scss
leigh-mil: @patricksmithdds Good catch!

styles/elements/_block_lists.scss
leigh-mil: I commented the same thing on Patrick's PR!

PR #629, opened 2019-02-12
Add Funding Through to Portfolio Header

montana-mil:
## Description
"Funding Through" now shows on all portfolio screen headers. The cents portion of the funding amount is also minimized.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163834698

## Screenshots
screen shot 2019-02-12 at 9 58 15 am

### DIFF
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -18,13 +18,16 @@ def dollars(value):
         numberValue = 0
     return "${:,.2f}".format(numberValue)
 
+
 def justDollars(value):
     raw = dollars(value)
     return raw[:-3]
 
+
 def justCents(value):
     raw = dollars(value)
     return raw[-3:]
+
 
 def usPhone(number):
     phone = re.sub(r"\D", "", number)
--- a/atst/filters.py
+++ b/atst/filters.py
@@ -21,12 +21,12 @@ def dollars(value):
 
 def justDollars(value):
     raw = dollars(value)
-    return raw[:-3]
+    return raw.split(".")[0]
 
 
 def justCents(value):
     raw = dollars(value)
-    return raw[-3:]
+    return raw.split(".")[1]
 
 
 def usPhone(number):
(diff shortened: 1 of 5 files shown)


styles/components/_portfolio_layout.scss
```diff
@@ -89,6 +89,42 @@
...
+      @include icon-color($color-blue);
+    }
+  }
+
+  .unfunded {
+    color: $color-red;
+    .icon {
+      @include icon-color($color-red);
+    }
+    }
```
patricksmithdds: Tiny thing -- this should be dedented once to match the `.unfunded` block started on line 122.

atst/routes/portfolios/__init__.py
```diff
@@ -31,4 +32,24 @@ def user_can(permission):
...
+    if not portfolio is None:
+        active_task_orders = [
+            task_order for task_order in portfolio.task_orders if task_order.is_active
+        ]
+        funding_end_date = (
+            sorted(active_task_orders, key=attrgetter("end_date"))[-1].end_date
+            if active_task_orders
+            else None
+        )
+        funded = len(active_task_orders) > 1
```
patricksmithdds: Can these be removed from the [`task_orders` route](https://github.com/dod-ccpo/atst/blob/master/atst/routes/portfolios/task_orders.py#L58-L59) now that they're unused?

atst/filters.py
```diff
@@ -19,6 +19,16 @@ def dollars(value):
...
 
 
+def justDollars(value):
+    raw = dollars(value)
+    return raw[:-3]
+
+
+def justCents(value):
+    raw = dollars(value)
+    return raw[-3:]
```
patricksmithdds: Rather than use "3" here, which I think relies on the functionality of `dollars`, could we split the `raw` string on the `.`?

styles/components/_portfolio_layout.scss
patricksmithdds: Sorry, the formatting is still a little misaligned in  [`ef4f39a`](/dod-ccpo/atst/pull/629/commits/ef4f39a53e8ca4f8e466815809f6f2d1c5047de1)

montana-mil: oops

PR #636, opened 2019-02-12
recalculate reporting mock data based on current month

dandds:
https://www.pivotaltracker.com/story/show/163877484

The date range for the mock reporting data is now calculated dynamically by the current month. Data always ends at the month previous to the current one, so that we can show what reporting might look like with monthly billing cycles.


atst/domain/csp/reports.py
```diff
@@ -31,131 +32,152 @@ def make_env(name):
         self.environments = [make_env(env_name) for env_name in envs]
 
 
+def generate_sample_dates(_max=8):
+    current = datetime.datetime.today()
+    sample_dates = []
+    for _i in range(_max):
```
patricksmithdds: I believe it's idiomatic to just use `_` for variables that aren't used anywhere else.

dandds: Ah, good to know. I was using the Ruby idiom 😬

PR #642, opened 2019-02-13
Fix margin that was causing alignment issue

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163869712)

Margin was causing alignment to be off. This PR fixes the margin around the "Let's do cloud!" button.\

## Before

screen shot 2019-02-13 at 10 49 26

## After

screen shot 2019-02-13 at 10 49 10

### DIFF
--- a/styles/sections/_task_order.scss
+++ b/styles/sections/_task_order.scss
@@ -49,6 +49,10 @@
         top: 2.5rem;
         margin-left: -23rem;
       }
+
+      .usa-button {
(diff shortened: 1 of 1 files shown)


drummond-dds:
@dandds Its also like that in the [DOD styleguide](https://designsystem.digital.gov/components/buttons/). I think its so a row with just a button doesn't have a weird `0.5em` padding on the left and is aligned correctly with the text.

PR #643, opened 2019-02-13
Trim extra whitespace on blur

leigh-mil:
## Description
Fixes a bug where users could type in 4 spaces and it would be a valid portfolio name. Updated the textInput component so it will trim beginning and trailing whitespace. 

## Pivotal
[https://www.pivotaltracker.com/story/show/163748431](https://www.pivotaltracker.com/story/show/163748431)

### DIFF
--- a/js/components/text_input.js
+++ b/js/components/text_input.js
@@ -85,7 +85,8 @@ export default {
     },
 
     onBlur: function(e) {
-      this._checkIfValid({ value: e.target.value, invalidate: true })
+      this._checkIfValid({ value: e.target.value.trim(), invalidate: true })
+      this.value = e.target.value.trim()
     },
 
     //


js/components/text_input.js
```diff
@@ -85,7 +85,8 @@ export default {
     },
 
     onBlur: function(e) {
-      this._checkIfValid({ value: e.target.value, invalidate: true })
+      this._checkIfValid({ value: e.target.value.trim(), invalidate: true })
+      this.value = e.target.value.trim()
```
dandds: Is there a reason not to just do this inside `onInput`? Doing it there makes a little more sense to me, but maybe there's a good reason to do it in `onBlur`?

leigh-mil: @dandds when I put it in `onInput` it was removing the whitespace as I typed, which I thought was a poor user experience!

dandds: Yup, you're right, that's a very good reason.

One semi-related thing I notice is that `onChange` has a comment saying it's supposed to handle blur events... I wonder why it isn't. Maybe we should refactor that functionality into `onBlur` at some point.

PR #649, opened 2019-02-14
Fix icon alert position

luisgov:
This PR fixed the alignment of the alert icon with the alert message in the funding screen.

*Before Fix:*
screen shot 2019-02-13 at 14 59 12

*After Fix:*
screen shot 2019-02-13 at 14 59 04

This PR fixes: https://www.pivotaltracker.com/n/projects/2160940/stories/163794280

### DIFF
--- a/styles/sections/_task_order.scss
+++ b/styles/sections/_task_order.scss
@@ -471,3 +471,8 @@
     }
   }
 }
+
+.to-expiration-alert .icon {
+    position: relative;
+    top: 3px;
+}


styles/sections/_task_order.scss
```diff
@@ -471,3 +471,8 @@
     }
   }
 }
+
+.to-expiration-alert .icon {
```
patricksmithdds: Should this change be made everywhere? It seems like the icon mis-alignment is in more places than just here.

montana-mil:
is this https://www.pivotaltracker.com/story/show/163794280?

styles/sections/_task_order.scss
luisgov: @patricksmithdds  it does not happen everywhere that we have a text-icon combo. do you have another example handy?

dandds: @luiscielak I've got one:

screen shot 2019-02-14 at 2 46 28 pm

The check and "funded through". I wonder if there are specific circumstances that cause it that we could track down.

patricksmithdds: yeah, that screenshot is the particular example I was thinking of.

dandds:
@luiscielak Are you still looking at this, or should we close for now? (No pressure, just cleaning up the PR list.)

luisgov:
@dandds going to close this one for now.

PR #654, opened 2019-02-15
Draft review alert

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/162772188)

We were not displaying the flash message explaining that the TO was in a draft state. I added this to match the ticket acceptance criteria.

localhost_8000_task_orders_new_4_9fd29841-953d-42b1-9486-2ca389c285e4 1

### DIFF
--- a/tests/routes/task_orders/test_new_task_order.py
+++ b/tests/routes/task_orders/test_new_task_order.py
@@ -4,6 +4,7 @@ from flask import url_for
 from atst.domain.task_orders import TaskOrders
 from atst.models.attachment import Attachment
 from atst.routes.task_orders.new import ShowTaskOrderWorkflow, UpdateTaskOrderWorkflow
+from atst.utils.localization import translate
 
 from tests.factories import UserFactory, TaskOrderFactory, PortfolioFactory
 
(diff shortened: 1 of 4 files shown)


tests/routes/task_orders/test_new_task_order.py
```diff
@@ -142,6 +143,27 @@ def test_task_order_validates_email_address(client, user_session):
     assert "Invalid email" in body
 
 
+def test_review_screen_when_all_sections_complete(client, user_session):
+    to = task_order()
```
dandds: You don't need to call pytest fixtures directly like this; you just name them as test function args, like:

```
def test_review_screen_when_all_sections_complete(client, user_session. task_order):
```

https://docs.pytest.org/en/latest/fixture.html

dandds: It looks like this happens a few times from a previous PR, too. Can you clean up all instances? We're getting warnings in the pytest output.

drummond-dds: Nice catch - fixed in 13999d7c60045c080b6392810d72dd49a7d5aa22

PR #655, opened 2019-02-15
Don't show outline on an anchor tag that doesn't have a href

drummond-dds:
On IE don't show outline if we're clicking on an anchor tag that doesn't have a `href`.

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163759711)


drummond-dds:
@dandds I got browserstack working correctly now!! (thanks @leigh-mil).

I fixed the issues in 703900b906a2e5894ed9461d482fc79f5da6e82d

PR #658, opened 2019-02-18
IE doesn't support CSS mask images

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163729626)

We were using a CSS feature [not supported in IE](https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image#Browser_compatibility).

I copied the `alert.svg` file and added in a `fill="#fdb81e"` attribute to generate a "`gold`" color version of the icon.

## Before

screen shot 2019-02-04 at 3 43 29 pm_big

## After

screen shot 2019-02-18 at 14 04 07


drummond-dds:
@dandds I replaced `alert.svg` with `alert-gold.svg` and renamed. Also did a rebase.

PR #661, opened 2019-02-19
DD-254 form

dandds:
story: https://www.pivotaltracker.com/story/show/163568954

Note that I modified the ACs a little and left some questions and comments. This is only a partial implementation of the SO user flow, since it does not include a real DD-254 PDF yet.

This implements a very basic Security Officer review screen. The form fields, labels, and descriptions are derived from existing versions of the DD-254 mentioned on that Pivotal card.

The data is saved to a `dd_254s` table and can be used to fill in a DD-254 PDF on a subsequent story. When the data is saved, we mark the SO review step as complete on the TO view page. 

screenshot_2019-02-19 jedi cloud

screen shot 2019-02-19 at 9 29 37 am

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -173,6 +173,14 @@ class TaskOrders(object):
                 "{} is not an officer role on task orders".format(officer_type)
             )
 
+    @classmethod
+    def add_dd_254(user, task_order, dd_254_data):
+        dd_254 = DD254(**dd_254_data)
+        task_order.dd_254 = dd_254
+
+        db.session.add(task_order)
+        db.session.commit()
+
 
 class DD254s:
     # TODO: standin implementation until we have a real download,
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -186,7 +186,7 @@ class DD254s:
     # TODO: standin implementation until we have a real download,
     # sign, and verify process for the DD 254 PDF
     @classmethod
-    def complete(cls, dd254):
+    def is_complete(cls, dd254):
         if dd254 is None:
             return False
 
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -187,6 +187,9 @@ class DD254s:
     # sign, and verify process for the DD 254 PDF
     @classmethod
     def complete(cls, dd254):
+        if dd254 is None:
+            return False
+
         for col in DD254.__table__.columns:
             if getattr(dd254, col.name) is None:
                 return False
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -4,6 +4,7 @@ from flask import current_app as app
 from atst.database import db
 from atst.models.task_order import TaskOrder
 from atst.models.permissions import Permissions
+from atst.models.dd_254 import DD254
 from atst.domain.portfolios import Portfolios
 from atst.domain.authz import Authorization
 from .exceptions import NotFoundError
@@ -171,3 +172,15 @@ class TaskOrders(object):
             raise TaskOrderError(
                 "{} is not an officer role on task orders".format(officer_type)
             )
+
+
+class DD254s:
+    # TODO: standin implementation until we have a real download,
+    # sign, and verify process for the DD 254 PDF
+    @classmethod
+    def complete(cls, dd254):
+        for col in DD254.__table__.columns:
+            if getattr(dd254, col.name) is None:
+                return False
+
+        return True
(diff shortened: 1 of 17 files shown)


leigh-mil:
lgtm!

atst/domain/task_orders.py
```diff
@@ -171,3 +172,26 @@ def add_officer(cls, user, task_order, officer_type, officer_data):
...
+
+        db.session.add(task_order)
+        db.session.commit()
+
+
+class DD254s:
+    # TODO: standin implementation until we have a real download,
+    # sign, and verify process for the DD 254 PDF
+    @classmethod
+    def complete(cls, dd254):
```
patricksmithdds: Should this be `is_complete`? I can't tell if this method should mark a DD254 as complete or returns True/False based on completeness.

atst/forms/dd_254.py
```diff
@@ -0,0 +1,38 @@
...
+from atst.utils.localization import translate
+
+
+class DD254Form(CacheableForm):
+    certifying_official = StringField(
+        translate("forms.dd_254.certifying_official.label"),
+        description=translate("forms.dd_254.certifying_official.description"),
+        validators=[Required()],
+    )
+    co_title = StringField(
```
patricksmithdds: I know we were avoiding use "co" previously -- I'm curious if this is going to be confusing in the future? Should we just use the full `certifying_official` prefix?

atst/routes/portfolios/task_orders.py
```diff
@@ -149,3 +152,63 @@ def edit_task_order_invitations(portfolio_id, task_order_id):
             task_order=task_order,
             form=form,
         )
+
+
+def so_review_form(task_order):
+    if task_order.dd_254:
+        dd_254 = task_order.dd_254
+        form = DD254Form(obj=dd_254)
+        form.required_distribution.data = dd_254.required_distribution
```
patricksmithdds: Is this necessary? Does the widget not properly initialize itself?

atst/routes/portfolios/task_orders.py
```diff
@@ -149,3 +152,63 @@ def edit_task_order_invitations(portfolio_id, task_order_id):
...
+        return form
+    else:
+        so = task_order.officer_dictionary("security_officer")
+        form_data = {
+            "certifying_official": "{} {}".format(
+                so.get("first_name"), so.get("last_name")
```
patricksmithdds: Should the `phone_number` below also use `.get`? I think this should provide a default of `""` also, to avoid potentially generating a name like `Randy None`.

translations.yaml
```diff
@@ -246,6 +246,20 @@ forms:
     so_invite_label: Invite Security Officer to Task Order Builder
     skip_invite_description: |
       <i>An invitation won't actually be sent until you click Done on the Review page. You can skip this for now and invite them later.</i>
+  dd_254:
+    certifying_official:
+      label: Name of Certifying Official
+      description: (First, Last, Middle)
```
patricksmithdds: The order here doesn't make any sense to me. Should it be "last, first, middle"? In that case, generating the name in the form `so_review_form` needs to be updated as well.

dandds: Yup, typo. We do need to update the order, good catch.

atst/routes/portfolios/task_orders.py
dandds: Yes, unfortunately. We do the same thing [here](https://github.com/dod-ccpo/atst/blob/master/atst/routes/task_orders/new.py#L77-L79).

patricksmithdds: Interesting. I wonder if it's worth attempting to make the widget smarter about handling that data?

dandds: I tried a couple things with no success, including an example straight from the docs: https://wtforms.readthedocs.io/en/stable/widgets.html#custom-widgets. I think it's worth digging into more: https://www.pivotaltracker.com/story/show/164117794

templates/portfolios/task_orders/so_review.html
```diff
@@ -0,0 +1,43 @@
...
+        {{ form.csrf_token }}
+        <h3 class="subheading">{{ "task_orders.so_review.certification" | translate }}</h3>
+        {{ TextInput(form.certifying_official) }}
+        {{ TextInput(form.certifying_official_title) }}
+        {{ TextInput(form.certifying_official_phone, placeholder='(123) 456-7890', validation='usPhone') }}
+        {{ TextInput(form.certifying_official_address, paragraph=True) }}
+
+        <hr>
+
+        {{ MultiCheckboxInput(form.required_distribution) }}
```
patricksmithdds: Should we have a text box for the user to enter the other options?

dandds: That's one of my open questions for Kristian. The PDF form doesn't explicitly say "if other, specify" or anything like that.

PR #662, opened 2019-02-19
Manage invitations should be inactive until user has completed TO flow

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163947440)

### Steps to reproduce
Find an incomplete TO
Navigate to its TO Summary page (Portfolio > Funding > View)
### Expected
The Manage Invitations link should be inactive until the TO is completed and invitations are actually sent.

### Actual
Currently this link is always active

### DIFF
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -5,6 +5,7 @@ from flask import g, redirect, render_template, url_for, request as http_request
 from . import portfolios_bp
 from atst.database import db
 from atst.domain.task_orders import TaskOrders
+from atst.domain.exceptions import NotFoundError
 from atst.domain.portfolios import Portfolios
 from atst.domain.authz import Authorization
 from atst.forms.officers import EditTaskOrderOfficersForm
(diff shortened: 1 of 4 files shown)


atst/routes/portfolios/task_orders.py
```diff
@@ -113,12 +113,22 @@ def task_order_invitations(portfolio_id, task_order_id):
...
     task_order = TaskOrders.get(g.current_user, task_order_id)
     form = EditTaskOrderOfficersForm(obj=task_order)
-    return render_template(
-        "portfolios/task_orders/invitations.html",
-        portfolio=portfolio,
-        task_order=task_order,
-        form=form,
-    )
+
+    if TaskOrders.all_sections_complete(task_order):
```
dandds: Good catch, block access if they try to hit the route directly.

tests/routes/portfolios/test_task_orders.py
```diff
@@ -202,6 +216,27 @@ def test_can_view_task_order_invitations(client, user_session):
...
+            task_order_id=task_order.id,
+        )
+    )
+    assert (
+        url_for(
+            "portfolios.view_task_order",
+            task_order_id=task_order.id,
+            portfolio_id=portfolio.id,
+        )
+        in response.location
```
dandds: I'm not saying you should change this, but often I like to build the real route and do an equality check in these situations:
```
url_for(
            "portfolios.view_task_order",
            task_order_id=task_order.id,
            portfolio_id=portfolio.id,
            _external=True
        )
```
That gives you a fully qualified URL you can check against `response.location`.

atst/routes/portfolios/task_orders.py
dandds: Should this return a 404, though, I wonder? The page effectively doesn't exist yet, for the user's purposes.

PR #663, opened 2019-02-19
Allow User to Edit Sections of KO Review

montana-mil:
## Description
Clicking the `edit` link next to  `Reporting`, `Funding`, or `Oversight` sections brings the user to the new TO form where they can edit and submit these changes. 

The submit button says `Save` instead of `Save & Continue`, and the user is redirected back to the KO Review page on submission.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163957691

## Screenshots
screen shot 2019-02-19 at 11 47 17 am

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -200,50 +200,32 @@ def get_started():
 @task_orders_bp.route("/task_orders/new/<int:screen>/<task_order_id>")
 @task_orders_bp.route("/portfolios/<portfolio_id>/task_orders/new/<int:screen>")
 def new(screen, task_order_id=None, portfolio_id=None):
-    workflow = ShowTaskOrderWorkflow(g.current_user, screen, task_order_id)
-
     if task_order_id and screen is 4:
         task_order = TaskOrders.get(g.current_user, task_order_id)
         if not TaskOrders.all_sections_complete(task_order):
             flash("task_order_draft")
 
-    if workflow.task_order:
-        ko_review_url = url_for(
-            "portfolios.ko_review",
-            portfolio_id=workflow.task_order.portfolio.id,
-            task_order_id=task_order_id,
-            _external=True,
-        )
-        redirect_url = url_for(
-            "portfolios.ko_review",
-            portfolio_id=workflow.task_order.portfolio.id,
-            task_order_id=task_order_id,
-        )
+    workflow = ShowTaskOrderWorkflow(g.current_user, screen, task_order_id)
+    template_args = {
+        "current": screen,
+        "task_order_id": task_order_id,
+        "portfolio_id": portfolio_id,
+        "screens": workflow.display_screens,
+        "form": workflow.form,
+        "complete": workflow.is_complete,
+    }
 
-        if http_request.referrer == ko_review_url:
-            return render_template(
-                workflow.template,
-                current=screen,
+    if workflow.task_order:
+        template_args["task_order"] = workflow.task_order
+        if http_request.args.get("ko_edit"):
+            template_args["ko_edit"] = True
+            template_args["next"] = url_for(
+                "portfolios.ko_review",
+                portfolio_id=workflow.task_order.portfolio.id,
                 task_order_id=task_order_id,
-                task_order=workflow.task_order,
-                portfolio_id=portfolio_id,
-                screens=workflow.display_screens,
-                form=workflow.form,
-                complete=workflow.is_complete,
-                from_ko_review=True,
-                next=redirect_url,
             )
 
-    return render_template(
-        workflow.template,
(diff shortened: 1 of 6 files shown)


atst/routes/task_orders/new.py
```diff
@@ -207,6 +207,33 @@ def new(screen, task_order_id=None, portfolio_id=None):
...
+                current=screen,
+                task_order_id=task_order_id,
+                task_order=workflow.task_order,
+                portfolio_id=portfolio_id,
+                screens=workflow.display_screens,
+                form=workflow.form,
+                complete=workflow.is_complete,
+                from_ko_review=True,
+                next=redirect_url,
+            )
```
- Would it work to just add a query parameter to the edit link on the KO review page, like `ko_edit=True`? You could just check for the presence of that query param, instead of checking the referrer.
- I think we could condense this a little by making an initial dictionary of keyword arguments for the template, like:
```
template_args = {
    "current": screen,
    "task_order_id": task_order_id,
    "task_order": workflow.task_order,
    "portfolio_id": portfolio_id,
    "screens": workflow.display_screens,
    "form": workflow.form,
    "complete": workflow.is_complete,
}
```
Then you can just add your new args to it if necessary (`template_args["next"] = redirect_url`, etc.), and then:

```
return render_template(workflow.template, **template_args)
```

atst/routes/task_orders/new.py
```diff
@@ -229,15 +256,19 @@ def update(screen, task_order_id=None, portfolio_id=None):
...
-                task_order_id=workflow.task_order.id,
+        if http_request.args.get("next"):
+            return redirect(http_request.args.get("next"))
+        else:
+            return redirect(
+                url_for(
+                    "task_orders.new",
+                    screen=screen + 1,
+                    task_order_id=workflow.task_order.id,
+                )
```
dandds: I think it might read a little better to set a variable, `redirect_url`, and then have just one return statement: `return redirect(redirect_url)`.

templates/task_orders/_new.html
```diff
@@ -9,6 +9,10 @@
   {% include "fragments/flash.html" %}
 
   {% block form_action %}
+    {% if from_ko_review %}
+      <form method='POST' action="{{ url_for('task_orders.update', screen=current, task_order_id=task_order_id, next=next) }}" autocomplete="off" enctype="multipart/form-data">
+    {% endif %}
+
     {% if task_order_id %}
       <form method='POST' action="{{ url_for('task_orders.new', screen=current, task_order_id=task_order_id) }}" autocomplete="off" enctype="multipart/form-data">
```
dandds: I think this one on line 17 could be `task_orders.update` like the others, and then the only difference between the three version of the `form` element are the keyword args for `url_for`. This is another case where it might be nicer to build a dictionary of kwargs and then splat it, so we only have the `form` element once in the template.

templates/task_orders/_new.html
```diff
@@ -39,7 +43,11 @@ <h1 class="task-order-form__heading subheading">
     {% block next %}
 
       <div class='action-group'>
-        <input type='submit' class='usa-button usa-button-primary' value='Save & Continue' />
+        {% if from_ko_review %}
+          <input type='submit' class='usa-button usa-button-primary' value='Save' />
+        {% else %}
+          <input type='submit' class='usa-button usa-button-primary' value='Save & Continue' />
+        {% endif %}
```
dandds: `value='{{ "Save & Continue" if from_ko_review else "Save" }}'` might be nicer here.

montana-mil: oh yeah, much better

templates/task_orders/_new.html
montana-mil: I'm confused on why we have the same routes on `new` and `update`. Are they overlapping in functionality in some way?

dandds: `new` is for GET requests and `update` is for POST. The URL is the same. Because building `task_orders.new` makes the same URL for the form action as `task_orders.update` it still works, but it's confusing and we should use `task_orders.update` for all of these.

atst/routes/task_orders/new.py
montana-mil: ok so what I ended up doing is something that I'm not super confident about, but I think it makes the `TOEditLink` macro more generalizable for the future. So if we have edit links from elsewhere in the site, we can pass extra arguments into the macro and then parse those kwargs within the route to know where the user was coming from and to be able to redirect them back after the edits are submitted.

montana-mil: we talked irl, but we decided to nix the `TOEditLink` macro altogether because it is capturing logic that we are no longer using. That simplified passing `ko_edit` along as a query parameter so we don't need to `eval` anything.

PR #665, opened 2019-02-20
Turn on Logging for SQLAlchemy

montana-mil:
## Description
When the environment is not production, SQLAlchemy logs now show on the server console. There is a lot of output, so this may need to be tweaked.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164090737

## Screenshots
screen shot 2019-02-20 at 8 48 35 am

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -72,6 +72,7 @@ def make_app(config):
 
     if ENV != "prod":
         app.register_blueprint(dev_routes)
+        app.config["SQLALCHEMY_ECHO"] = True
 
     app.form_cache = FormCache(app.redis)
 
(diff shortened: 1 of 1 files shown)


atst/app.py
```diff
@@ -72,6 +72,7 @@ def make_app(config):
 
     if ENV != "prod":
         app.register_blueprint(dev_routes)
+        app.config["SQLALCHEMY_ECHO"] = True
```
patricksmithdds: This would echo the statements on the staging & test environments as well -- is that something we want? I'm curious if it would be preferable to [use the `DEBUG` setting when configuring the app](https://github.com/dod-ccpo/atst/blob/6a7e490c20b6005a5d349d37ecfbcd7b21a12412/atst/app.py#L134-L138).

montana-mil: I played around with the DEBUG setting, but couldn't trigger any logging. 

I think it makes sense to just do this for dev environment!

PR #667, opened 2019-02-20
Fix hover state of disabled link on IE

drummond-dds:
The `DD 254` link on the task order page was displaying a very faint background on hover when using IE. This PR fixes that.

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163759711)

### DIFF
--- a/styles/elements/_icon_link.scss
+++ b/styles/elements/_icon_link.scss
@@ -67,8 +67,13 @@
 
   &.icon-link--disabled {
     opacity: 0.3;
-    pointer-events: none;
     text-decoration: none;
+    pointer-events: none;
+
+    &:hover {
+      cursor: default;
+      background: $color-white;
+    }
   }
 
   &.icon-link--left {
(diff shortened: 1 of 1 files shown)


styles/elements/_icon_link.scss
```diff
@@ -67,8 +67,13 @@
 
   &.icon-link--disabled {
     opacity: 0.3;
-    pointer-events: none;
     text-decoration: none;
+    pointer-events: none;
+
+    &:hover {
+      cursor: default;
+      background: $color-white;
```
patricksmithdds: Would this still work if the link is in an element that has a non-white background?

dandds: Ah, good q.

drummond-dds: I tried `background: inherit;` and that seems to work. Should cover this scenario 219636245b7143cbdaba1f62c84410077189719e

PR #669, opened 2019-02-20
Fix side scroll issue on Firefox

leigh-mil:
## Description
The list of TOs for a portfolio was side scrolling in Firefox. Adding padding to the div that contained all of the content fixed the issue. 

## Pivotal
[https://www.pivotaltracker.com/story/show/163794683](https://www.pivotaltracker.com/story/show/163794683)

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -278,6 +278,8 @@
 }
 
 .portfolio-funding {
+  padding: 2 * $gap;
+
   .panel {
     @include shadow-panel;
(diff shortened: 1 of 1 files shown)


styles/components/_portfolio_layout.scss
```diff
@@ -278,6 +278,8 @@
 }
 
 .portfolio-funding {
+  padding: 1.6rem;
```
patricksmithdds: Can we use `2 * $gap` rather than a specific `rem` value?

PR #671, opened 2019-02-21
Focus on checkbox/radio in fieldset passes through click event on IE

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/163990226)

## Steps to Reproduce

1. Start a new TO
2. Complete initial fields
3. Before the entire array of "App Migration" buttons is fully visible on screen, attempt to click the top one (migrating from on-premise data center)

## Actual Behavior/Result

Focus switches to the entire radio button array, but the button that the user thinks that they selected is not yet indicated as selected

## Expected Behavior/Result

The button that the user clicks should be marked as selected

Note: The same thing will occur with "Project Complexity" or "About Your Team" on the same form.

---

## Broken

broken

## Fixed

fixed

### DIFF
--- a/templates/components/checkbox_input.html
+++ b/templates/components/checkbox_input.html
@@ -6,7 +6,7 @@
   <checkboxinput name='{{ field.name }}' inline-template key='{{ field.name }}'>
     <div class='usa-input {{ classes }} {% if field.errors %}usa-input--error{% endif %}'>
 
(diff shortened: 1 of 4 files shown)


drummond-dds:
@dandds I don't think so. It [looks like IE treats `fieldset` differently](https://allyjs.io/data-tables/focusable.html#fieldset-element) but I'm not sure.

PR #673, opened 2019-02-22
Remove no-longer-used requests code

patricksmithdds:
This PR removes code related to the old requests flow. Request models, routes, and templates are removed. Related forms, utilities, and js components are also removed.

After removing a lot of the code, the coverage dropped below 90%. I added some more tests to bump that number back up, though On [master](https://circleci.com/gh/dod-ccpo/atst/5327), python test coverage is at 91.39%. On [this branch](https://circleci.com/gh/dod-ccpo/atst/5360), python test coverage is up to 91.71%.


alembic/versions/c92cec2f32d4_remove_request_related_models.py
```diff
@@ -0,0 +1,150 @@
...
+    op.drop_constraint('workspaces_request_id_fkey', 'portfolios', type_='foreignkey')
+    op.drop_table('requests')
+    op.drop_column('portfolios', 'request_id')
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('portfolios', sa.Column('request_id', postgresql.UUID(), autoincrement=False, nullable=True))
+    op.create_foreign_key('workspaces_request_id_fkey', 'portfolios', 'requests', ['request_id'], ['id'])
```
dandds: This foreign key statement might be out of order. I get an error here when trying to downgrade.

dandds: In fact, a couple things about the order don't work. I moved the foreign key and index creation to the bottom of the function and put the request table statement at the top and it worked.

atst/domain/csp/reports.py
```diff
@@ -33,10 +33,10 @@ def make_env(name):
 
 
 def generate_sample_dates(_max=8):
-    current = datetime.datetime.today()
```
dandds: just need to remove the import, i think

alembic/versions/c92cec2f32d4_remove_request_related_models.py
patricksmithdds: Sorry bout that -- it should be fixed in 9594e0256235c94d9616c7f6828996989d4b376a.

PR #674, opened 2019-02-22
WIP: Use crl archive

ddsdevon:
This PR switches the build over to using the Rackspace Cloud Files hosted CRL archive maintained by the crlupdater process.

### DIFF
--- a/script/install_crls
+++ b/script/install_crls
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# script/install_crls: Fetches and unpacks an archive of all relevant DoD CRL
+#                      files from cloud storage
+
+source "$(dirname "${0}")"/../script/include/global_header.inc.sh
+
+# Get Rackspace API token
+RACKSPACE_TOKEN=$(curl -H 'Content-type: application/json' -XPOST ${RACKSPACE_AUTH_URL} -d "{\"auth\":{\"RAX-KSKEY:apiKeyCredentials\":{\"username\":\"${RACKSPACE_USERNAME}\",\"apiKey\":\"${RACKSPACE_PASSWORD}\"}}}" | jq -r '.access.token.id')
+export RACKSPACE_TOKEN
+
+# Fetch DoD CRL archive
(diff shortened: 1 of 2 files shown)


script/install_crls
```diff
@@ -0,0 +1,19 @@
...
+export RACKSPACE_TOKEN
+
+# Ensure the ./crl director exists
+mkdir -p ./crl
+
+# Fetch DoD CRL archive
+curl -H 'Content-type: application/json' -H "X-Auth-Token: ${RACKSPACE_TOKEN}" -XGET ${RACKSPACE_STORAGE_URL}/MossoCloudFS_1095493/crls/dod_crls.tar.bz -o ./crl/crls.tar.bz
+
+# Unpack CRL archive
+tar -xjvf ./crl/crls.tar.bz -C ./crl
```
dandds: @ddsdevon I wonder if this is something the app should know how to do for itself. We have long-lived instances of the app on uat.atat.code.mil and test.atat.code.mil that don't get rebuilt very often, so they often have stale CRLs. The app should sync the CRLs during the build, like we already do. Then, I think that when the app sees that a CRL it's trying to use has expired, it should fetch them again. (Slower login for one user, about every two weeks. :) )

We already have a CSP [file provider](https://github.com/dod-ccpo/atst/blob/master/atst/domain/csp/files.py) for doing uploads and downloads. I could add a comparable CSP provider that knows how to sync copies of the CRLs correctly. That way when we move to the real CSP we just have to write a new CRL provider.

Do you mind if I push an example to this branch on Monday?

dandds: Something like this: deda3d176513a46ca407d4266ff4f3941f27c85d (branch sync-crls).

This would allow the app to grab new copies from Cloudfiles as-needed. I'd rewrite the CRL cache class we use for auth to do that next.

ddsdevon: @dandds Sure, go for it! I just made this as a quick and dirty way to start using the archive. I have no issues with it being in Python instead. Feel free to branch off this, or create an entirely new one and just PR that.

ddsdevon:
Closing PR since we are not going to use the originally purposed version of the branch

PR #677, opened 2019-02-25
Move defense_component to portfolios model

montana-mil:
## Description
`defense_component` is now an attribute on the `Portfolio` model instead of `TaskOrder`. If a new task order is created outside the context of a portfolio, it creates a portfolio that has the selected `defense_component`. If the task order is created from within the context of an existing portfolio, the defense component of the task order is that of its portfolio.

The defense component is displayed (read only) on the Portfolio Settings (admin) screen.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163954462

## Screenshots
screen shot 2019-02-26 at 11 35 19 am


templates/fragments/task_order_review/app_info.html
```diff
@@ -2,7 +2,7 @@
 
 <div class="row">
   {{ ReviewField(("task_orders.new.review.portfolio" | translate), task_order.portfolio_name) }}
-  {{ ReviewField(("task_orders.new.review.dod" | translate), task_order.defense_component, filter="normalizeOrder") }}
+  {{ ReviewField(("task_orders.new.review.dod" | translate), task_order.portfolio.defense_component, filter="normalizeOrder") }}
```
dandds: You could just use the property on `task_order`, right?

templates/portfolios/admin.html
```diff
@@ -24,6 +24,12 @@
           <button type="submit" class="usa-button usa-button-big usa-button-primary" tabindex="0">Save</button>
         </div>
       </div>
+      <div class='defense-row'>
```
dandds: lol this class name

tests/routes/task_orders/test_new_task_order.py
```diff
@@ -203,7 +209,8 @@ def test_show_task_order_form(task_order):
         task_order.creator, task_order_id=task_order.id
     )
     assert (
-        another_workflow.form.data["defense_component"] == task_order.defense_component
```
dandds: This should also still work with your property on the TO model, right?

templates/fragments/task_order_review/app_info.html
montana-mil: oh yep! this was before i added that prop. thanks for catching

PR #679, opened 2019-02-26
sync crls from CSP source

dandds:
https://www.pivotaltracker.com/story/show/163602397

This makes two major updates:

- During the build, the app will sync copies of the CRLs from the Cloudfiles location [crlupdater](https://github.com/dod-ccpo/crlupdater) is continually updating.
- The `CRLCache` will track the expirations for its CRLs. If a user tries to log in and the relevant CRL has expired, the `CRLCache` will pull the CRLs from Cloudfiles again and rebuild its cache. This way, long-running instances of the app (like UAT) can get new CRLs ad-needed without a manual refresh or a separate cron job.


ddsdevon:
@dandds You can also nuke the "Load Cache"/"Save Cache" entries for the CRLs. It doesn't actually help any at this point.

drummond-dds:
👍 Looks good thanks for explaining!

PR #685, opened 2019-02-28
Portfolio Name and Defense Component read only

montana-mil:
## Description
When a new TO is started from an existing portfolio, the portfolio name and defense component are inherited from the portfolio and are read only on the form.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/163953148

## Screenshots
screen shot 2019-02-28 at 11 19 36 am

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -50,9 +50,10 @@ class ShowTaskOrderWorkflow:
         self.user = user
         self.screen = screen
         self.task_order_id = task_order_id
-        self.portfolio_id = portfolio_id
-        self._section = TASK_ORDER_SECTIONS[screen - 1]
         self._task_order = None
+        self.portfolio_id = portfolio_id
+        self._portfolio = None
+        self._section = TASK_ORDER_SECTIONS[screen - 1]
         self._form = None
 
     @property
@@ -64,9 +65,13 @@ class ShowTaskOrderWorkflow:
 
     @property
     def portfolio(self):
-        if self.task_order:
-            return self.task_order.portfolio
-        return Portfolios.get(self.user, self.portfolio_id)
+        if not self._portfolio:
+            if self.task_order:
+                self._portfolio = self.task_order.portfolio
+            elif self.portfolio_id:
+                self._portfolio = Portfolios.get(self.user, self.portfolio_id)
+
+        return self._portfolio
 
     @property
     def form(self):
@@ -97,16 +102,12 @@ class ShowTaskOrderWorkflow:
         else:
             self._form = self._section[form_type]()
 
-        if self.pf_attributes_read_only() and self.screen == 1:
+        if self.pf_attributes_read_only and self.screen == 1:
             self._form = task_order_form.AppInfoWithExistingPortfolioForm(
                 obj=self.task_order
             )
 
         return self._form
-
-    @form.setter
-    def form(self, value):
-        self._form = value
 
     @property
     def template(self):
@@ -126,19 +127,19 @@ class ShowTaskOrderWorkflow:
 
     @property
     def is_complete(self):
-        if self.task_order:
-            if TaskOrders.all_sections_complete(self.task_order):
-                return True
+        if self.task_order and TaskOrders.all_sections_complete(self.task_order):
+            return True
         else:
             return False
 
+    @property
     def pf_attributes_read_only(self):
-        if self.task_order:
-            if self.task_order.portfolio.num_task_orders > 1:
-                return True
+        if self.task_order and self.portfolio.num_task_orders > 1:
+            return True
         elif self.portfolio_id:
             return True
-        return False
+        else:
+            return False
 
 
 class UpdateTaskOrderWorkflow(ShowTaskOrderWorkflow):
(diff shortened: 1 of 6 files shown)


atst/routes/task_orders/new.py
```diff
@@ -237,6 +237,11 @@ def new(screen, task_order_id=None, portfolio_id=None):
                 task_order_id=task_order_id,
             )
             url_args["next"] = template_args["next"]
+    elif http_request.args.get("new_to_on_portfolio"):
```
dandds: I don't think we need a special URL query arg to do this. If there is a `portfolio_id`, it means the user is making a new TO on an existing portfolio. Then, I'd just pass the portfolio to the template and check those properties on the portfolio.

templates/task_orders/new/app_info.html
```diff
@@ -14,11 +15,17 @@
...
+    {{ ReviewField(heading="forms.portfolio.name_label" | translate, field=portfolio_name) }}
+{% else %}
+    {{ TextInput(form.portfolio_name, placeholder="The name of your office or organization", validation="portfolioName") }}
+    {{ TextInput(form.scope, paragraph=True) }}
+    <p><i>{{ "task_orders.new.app_info.sample_scope" | translate | safe }}</i></p>
+{% endif %}
+
 <div class="subheading--black">
-  {{ OptionsInput(form.defense_component) }}
+    {{ ReviewField(heading="forms.task_order.defense_component_label" | translate, field=defense_component) if new_to_on_portfolio else OptionsInput(form.defense_component) }}
```
dandds: NBD, but this is a case where a regular `if/else` setup might read better, only because it takes less horizontal space.

montana-mil: oh ha! sublime wraps it so it looks fine to me. I'll switch it.

atst/routes/task_orders/new.py
montana-mil: are there a bunch of places where we use `portfolio_id` in templates? I was thinking since we are going to send `portfolio`, maybe I would take that `portfolio_id` entirely. I'm wondering if it's worth it to do that.

montana-mil: ended up taking `portfolio_id` out

dandds:
@montana-mil Like we talked about, this hides the input for "project scope" along with the one for "defense component". I'm not sure that's right; we're checking with DDS about how we should handle scope here.

leigh-mil:
Its not working for me when I fill out the form! 
If I fill out the rest of the form then click Save & Continue, I am getting errors for the name and component fields!

screen shot 2019-03-04 at 2 17 00 pm

atst/routes/task_orders/new.py
```diff
@@ -206,21 +226,27 @@ def get_started():
...
         "form": workflow.form,
         "complete": workflow.is_complete,
     }
 
+    if workflow.pf_attributes_read_only(portfolio_id):
+        template_args["portfolio"] = workflow.get_portfolio(portfolio_id=portfolio_id)
+        if screen == 1:
+            workflow.form = task_order_form.AppInfoWithExistingPortfolioForm(
+                obj=workflow.task_order
+            )
```
dandds: The workflow object should manage its own state, so I think this should get moved back into that class.

atst/routes/task_orders/new.py
```diff
@@ -206,21 +226,27 @@ def get_started():
...
+
+    if task_order_id and screen is 4:
+        task_order = TaskOrders.get(g.current_user, task_order_id)
+        if not TaskOrders.all_sections_complete(task_order):
+            flash("task_order_draft")
```
dandds: Did you move this just to make it more readable? (While you're here, the look-up for the TO is redundant. The `workflow` already has it as `workflow.task_order`.)

atst/routes/task_orders/new.py
```diff
@@ -116,6 +120,20 @@ def is_complete(self):
         else:
             return False
 
+    def pf_attributes_read_only(self, portfolio_id=None):
```
dandds: It might simplify things to make `portfolio_id` an init kwarg on `ShowTaskOrderWorkflow`. That way this method can just be a property.

atst/routes/task_orders/new.py
```diff
@@ -116,6 +120,20 @@ def is_complete(self):
...
+        if self.task_order:
+            if self.task_order.portfolio.num_task_orders > 1:
+                return True
+        elif portfolio_id:
+            if self.get_portfolio(portfolio_id).num_task_orders > 0:
```
dandds: I might be wrong, but if there is no TO and there's a `portfolio_id`, don't we know it's read-only regardless of how many TOs there are? You can't have a portfolio with zero TOs.

atst/routes/task_orders/new.py
```diff
@@ -116,6 +120,20 @@ def is_complete(self):
...
+                return True
+        return False
+
+    def get_portfolio(self, portfolio_id=None):
```
dandds: This could become a property exactly like `workflow.task_order` if we pass the `portfolio_id` as an initialization kwarg.

atst/routes/task_orders/new.py
montana-mil: i think when you fund a new portfolio, on the first screen, you don't have any TOs yet. This was a hard one to catch, and i might have been getting exhausted when i came up with this line. let me double check to make sure it's necessary.

atst/routes/task_orders/new.py
montana-mil: i like it! ty

atst/routes/task_orders/new.py
montana-mil: yep! does it actually make it more readable? and ill fix that TO lookup.

dandds: Yep, I think it's a good move. I just wanted to make sure there wasn't a technical reason I wasn't seeing.

atst/routes/task_orders/new.py
montana-mil: nope, we dont need it! the case i was thinking of doesn't get a `portfolio_id`, so it would never enter that `if` statement. Turns out the test I wrote for it was working with the wrong baseline assumption!

PR #691, opened 2019-03-05
Check if form is complete before showing the TO Review page

leigh-mil:
## Description
Add in a check to make sure that the TO is completed before the COR can view or fill out the TO Review form.

There was an existing function `all_sections_complete()` that was implemented for similar checks that I also used for this check. However, I noticed that this checks to make sure that all form fields on the first three pages have been filled out and does not consider whether or not the MO has submitted the form by clicking 'Done' on the 4th page. I noticed that when the MO is also the COR and has completed the form, but has not clicked 'Done', the COR/MO would be able to view and fill out the TO Review form. 

For this PR, I made the assumption that the current implementation of `all_sections_complete()` is correctly determining when a TO is complete, and that it is ok that the COR/MO can view the form before 'Done' is clicked. If this assumption was not correct, let me know and I will fix it!

## Pivotal
[https://www.pivotaltracker.com/story/show/164282346](https://www.pivotaltracker.com/story/show/164282346)

### DIFF
--- a/atst/domain/exceptions.py
+++ b/atst/domain/exceptions.py
@@ -34,3 +34,12 @@ class UnauthenticatedError(Exception):
 
 class UploadError(Exception):
     pass
+
+
+class NoAccessError(Exception):
+    def __init__(self, resource_name):
+        self.resource_name = resource_name
+
+    @property
+    def message(self):
+        return "Route for {} cannot be accessed".format(self.resource_name)
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -5,7 +5,7 @@ from flask import g, redirect, render_template, url_for, request as http_request
 from . import portfolios_bp
 from atst.database import db
 from atst.domain.task_orders import TaskOrders, DD254s
-from atst.domain.exceptions import NotFoundError
+from atst.domain.exceptions import NotFoundError, NoAccessError
 from atst.domain.portfolios import Portfolios
 from atst.domain.authz import Authorization
 from atst.forms.officers import EditTaskOrderOfficersForm
@@ -93,7 +93,7 @@ def ko_review(portfolio_id, task_order_id):
             form=KOReviewForm(obj=task_order),
         )
     else:
-        raise NotFoundError("task_order")
+        raise NoAccessError("task_order")
 
 
 @portfolios_bp.route(
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -85,12 +85,15 @@ def ko_review(portfolio_id, task_order_id):
 
     Authorization.check_is_ko_or_cor(g.current_user, task_order)
 
-    return render_template(
-        "/portfolios/task_orders/review.html",
-        portfolio=portfolio,
-        task_order=task_order,
-        form=KOReviewForm(obj=task_order),
-    )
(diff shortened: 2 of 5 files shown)


atst/routes/portfolios/task_orders.py
```diff
@@ -85,12 +85,15 @@ def ko_review(portfolio_id, task_order_id):
...
-    )
+    if TaskOrders.all_sections_complete(task_order):
+        return render_template(
+            "/portfolios/task_orders/review.html",
+            portfolio=portfolio,
+            task_order=task_order,
+            form=KOReviewForm(obj=task_order),
+        )
+    else:
+        raise NotFoundError("task_order")
```
dandds: This might be hairsplitting, but I'm afraid that `NotFoundError` could be misleading in the logs here. (And I think we might be using this incorrectly in other places.) The end result we want is a 404, but the logs will tell a developer `no task_order could be found`, which isn't true. I think we need something like a `NoAccessError` for these situations, where that TO is not in a state where the route can be accessed. That error should _also_ present as a 404 to the user, but will be more specific in the logs.

dandds:
@leigh-mil and maybe a bug about our use of `NotFoundError` in other places. Basically, the route functions should not find a TO and then raise `NotFoundError`, that's weird. Exceptions are for developers, HTTP status codes are for end-users.

atst/domain/exceptions.py
```diff
@@ -34,3 +34,12 @@ def message(self):
...
     pass
+
+
+class NoAccessError(Exception):
+    def __init__(self, resource_name):
+        self.resource_name = resource_name
+
+    @property
+    def message(self):
+        return "Route for {} cannot bef accessed".format(self.resource_name)
```
dandds: i feel bad noting this typo, but, there is a typo.

templates/portfolios/task_orders/show.html
```diff
@@ -32,7 +32,7 @@
         </div>
       </div>
       <div class="task-order-next-steps__action col">
-        {% if not task_order.is_active and button_text and button_url %}
+        {% if not task_order.is_active and button_active and button_text and button_url %}
```
dandds: I think this condition is getting confusing. We already use conditional logic for these button args in other places in the template (like `button_text=is_so and ("common.edit" | translate)` on line 140). I would probably refactor this for every step to be something like

```
{% set show_so_button = is_so and to_form_complete %}
Step(
            button_text=show_so_button and ("common.edit" | translate),
            button_url=show_so_button and url_for("portfolios.so_review", portfolio_id=portfolio.id, task_order_id=task_order.id),
            complete=dd_254_complete,
[...]
```

using the DD-254 step as an example.

templates/portfolios/task_orders/show.html
```diff
@@ -6,6 +6,10 @@
 
 {% block portfolio_content %}
 
+{% set show_dd_254_button = is_so and to_form_complete %}
+{% set show_to_info_button = (is_cor or is_ko) and to_form_complete %}
+{% set show_sign_to_button = is_ko and dd_254_complete and not is_to_signed %}
```
dandds: This effectively shows that these `is_so`, etc., variables are only necessary for setting the variables you're making here. A good refactor would probably be to handle more of this logic in the route function. Don't do it on this PR, but maybe keep it in the back of your head.

PR #696, opened 2019-03-11
Fix TO Form progress bar

leigh-mil:
## Description
Some fields were registering as 'completed' even though no data was added in by the user, which was making the progress indicator falsely show that a section was completed. The problem was that `None` was getting saved as `'None'` in the database for several of the form fields. Adding in default values and filters fixed this, as well as adding in a check for Radio fields so that a value that is not one of the options can get saved in the database.

## Pivotal
[https://www.pivotaltracker.com/story/show/164278538](https://www.pivotaltracker.com/story/show/164278538)

## Screenshot
Screen Shot 2019-03-11 at 3 22 12 PM

### DIFF
--- a/atst/forms/forms.py
+++ b/atst/forms/forms.py
@@ -18,11 +18,16 @@ class BaseForm(FlaskForm):
     def data(self):
         # remove 'csrf_token' key/value pair
         # remove empty strings and None from list fields
+        # prevent values that are not an option in a RadioField from being saved to the DB
         _data = super().data
         _data.pop("csrf_token", None)
         for field in _data:
             if _data[field].__class__.__name__ == "list":
                 _data[field] = [el for el in _data[field] if el not in EMPTY_LIST_FIELD]
+            if self[field].__class__.__name__ == "RadioField":
+                choices = [el[0] for el in self[field].choices]
+                if _data[field] not in choices:
+                    _data[field] = None
         return _data
 
     def validate(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


atst/forms/task_order.py
```diff
@@ -31,6 +31,13 @@ class AppInfoWithExistingPortfolioForm(BaseForm):
     scope = TextAreaField(
         translate("forms.task_order.scope_label"),
         description=translate("forms.task_order.scope_description"),
+        filters=[lambda x: x or None],
+    )
+    defense_component = SelectField(
```
dandds: Check me, but I think this change should be applied to the `defense_component` in the child class, `AppInforForm`. It only needs to appear in one. This form class is for new TOs on existing portfolios.

atst/forms/forms.py
```diff
@@ -18,11 +18,16 @@ def __init__(self, formdata=None, **kwargs):
...
+        # prevent values that are not an option in a RadioField from being saved to the DB
         _data = super().data
         _data.pop("csrf_token", None)
         for field in _data:
             if _data[field].__class__.__name__ == "list":
                 _data[field] = [el for el in _data[field] if el not in EMPTY_LIST_FIELD]
+            if self[field].__class__.__name__ == "RadioField":
+                choices = [el[0] for el in self[field].choices]
+                if _data[field] not in choices:
+                    _data[field] = None
```
dandds: Note to talk about this IRL

PR #700, opened 2019-03-13
CRL Failover Config

montana-mil:
## Description
This adds a config to `atst/config/base.ini` called `CRL_FAIL_OPEN`. This config defaults to `False`, but can be set to `True` if needed. There is also a new `CRLInvalidException` which is raised if the CRL is missing or expired.

When a CRL is expired (or missing), the user will see an error page with the message "Error 008."

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164179789


tests/domain/authnid/test_crl.py
```diff
@@ -183,7 +81,7 @@ def test_can_dynamically_update_crls(tmpdir):
     assert cache.crl_check(cert)
     # override the original CRL with one that revokes atat.mil.crt
     shutil.copyfile("tests/fixtures/test.der.crl", crl_file)
-    with pytest.raises(CRLRevocationException):
+    with pytest.raises(CRLInvalidException):
```
dandds: > override the original CRL with one that revokes atat.mil.crt

So this _should_ raise a revocation exception. If it isn't, there's either a problem with the test fixtures or the CRL code 😮

tests/domain/authnid/test_crl.py
```diff
@@ -233,7 +131,32 @@ def test_no_op_crl_cache_logs_common_name():
     assert "ART.GARFUNKEL.1234567890" in logger.messages[-1]
 
 
-def test_updates_expired_certs(ca_file, expired_crl_file, crl_file, ca_key, make_x509):
+def test_expired_crl_raises_CRLInvalidException_with_failover_config_false(
+    app, ca_file, expired_crl_file, ca_key, make_x509, rsa_key, monkeypatch
+):
+    app.config.update({"CRL_FAIL_OPEN": False})
```
dandds: I don't think you need to set this explicitly if it's the default.

tests/domain/authnid/test_crl.py
```diff
@@ -233,7 +131,32 @@ def test_no_op_crl_cache_logs_common_name():
...
+def test_expired_crl_passes_with_failover_config_true(
+    ca_file, expired_crl_file, ca_key, make_x509, rsa_key, monkeypatch, app
+):
+    app.config.update({"CRL_FAIL_OPEN": True})
+    client_cert = make_x509(rsa_key(), signer_key=ca_key, cn="chewbacca")
+    client_pem = client_cert.public_bytes(Encoding.PEM)
+    crl_cache = CRLCache(ca_file, crl_locations=[expired_crl_file])
+
+    assert crl_cache.crl_check(client_pem)
+    app.config.update({"CRL_FAIL_OPEN": False})
```
dandds: I would use the pytest fixture and `yield` setup for the config update here. If the test fails here, the config will not be reset, which could effect other tests.

tests/domain/authnid/test_crl.py
montana-mil: ok, i was wondering about that. i thought that if for some reason, someone changes the config to true and then runs tests, this one would fail if its not set explicitly

tests/domain/authnid/test_crl.py
montana-mil: oh i see. i'll make a fixture

tests/domain/authnid/test_crl.py
dandds: Yeah, I think we can rely on the config in `config/test.ini`. If you look at `pytest.ini` in the repo root, we set the env to "test" for `pytest` by default, so it should always build that config. If you modify config in-place for testing purposes, you should use a fixture that will clean up your change, like we talked about below.

tests/domain/authnid/test_crl.py
montana-mil: fixed together!

leigh-mil:
Looks really great!

atst/domain/authnid/crl/__init__.py
```diff
@@ -170,6 +180,11 @@ def crl_check(self, cert):
             return True
 
         except crypto.X509StoreContextError as err:
+            if err.args[0][0] == CRL_EXPIRED_ERROR_CODE:
+                if app.config.get("CRL_FAIL_OPEN"):
```
dandds:
```suggestion
                if app.config.get("CRL_FAIL_OPEN"):
                    self._log_info(
                        "Encountered expired CRL for certificate with CN {} and issuer CN {}, failing open.".format(
                            parsed.get_subject().CN, parsed.get_issuer().CN
                        )
                    )
```
Trying out the suggestion feature. I think you should see a toggle that allows you to just accept it? Anyway, I think that we should log here when we fail open.

PR #701, opened 2019-03-13
Portfolio perms

dandds:
**PR description**

This PR does the initial lift to add the new portfolio permissions structure. I know this is a wall of text, but please read and respond if you have comments/questions.

I made a diagram as a refresher:

Portfolio Permissions Model

- A `user` has a relationship to a `portfolio` through a `portfolio_role`.
- A `portfolio_role` has relationships to many `permission_sets` through another join table (`portfolio_roles_permission_sets`).
- `permission_sets` are basically fixture data. Each one is a bucket with discrete permissions. The list for "View Portfolio App Management" is included as an example. 

A user's permissions with a portfolio is a combination of all the permissions they have available by way of their `portfolio_role` and joined `permission_sets`. In the example, the user would have all of the permissions in the `permission_sets` the `portfolio_role` is pointing to. In practice, the many-to-many relationship between the `portfolio_role` and `permission_sets` should be an implementation detail, and all of a `portfolio_role`'s permissions should be available on the `portfolio_role` as a `permissions` attribute.

**notes**
- This doesn't apply any new access restrictions, just changes the mechanisms for the existing ones. We'll have to apply new access restrictions as we go.
- I've decided to _not_ migrate existing `portfolio_role`->`role` data to the new `portfolio_role`->multiple `permission_sets`. It would be time-consuming, it's not clear how to translate all of the old roles to the new permissions, and writing a downgrade for the change would be just as time-consuming. We'll reset the seed data on the live sites when this is merged.
- Because of the preceding point, running this branch will break user access for any existing portfolios you have.
- I tried to decompose the permissions we were given by design into common-sense permissions. All of those are subject to change, so if you'd prefer more explicit atomic permissions, let me know.
- I'm leaving CCPO in a half-baked state. Some related tests are skipped. The upcoming PR described below will restore CCPI status and the accompanying tests.
- I've made a separate card for recording permission changes in the audit log in the new model. Listening for database write events in the new setup is not straightforward. In order to move forward with the sprint, I've split that task out into a [new card](https://www.pivotaltracker.com/story/show/164648971). I have a branch from this PR with a not-really-working attempt and notes, [record-portfolio-permission-changes](https://github.com/dod-ccpo/atst/tree/record-portfolio-permission-changes).
- Minor note: The downgrade that adds `role_id` back to `portfolio_roles` makes it nullable, because adding a non-nullable column to existing data explodes (there's not default to give it without fixture `roles` in place).

**next PRs**

There will be two follow-up PRs:
- one will fix CCPO access by creating site-wide roles
- one will move our authorization handling into route decorators for simplicity

**next step notes**
- I think at every tier of access we should have a role model analogous to `portfolio_role`. This would mean adding:
  - `users_permission_sets`: This table would join users directly to many `permission_sets`. The permission sets a user has in this table would be site-wide permissions, so this is primarily for CCPO. The first cut of this will probably make CCPO users who have _every_ permission set, and we can refine it later for multiple CCPO user types.
  - `application_role`: this very analogous to `portfolio_role`, except that it joins a `user` to an `application`
- I don't think every user needs site-wide permission sets, or that every application user needs a `portfolio_role`. In the latter case, an `application_role` should be sufficient for access to the extremely restricted amount portfolio info that user can see. 
- When checking a user's ability to take a portfolio action, we'd check the perms on their `portfolio_role` and `site_role` (returning an empty list for users with no `site_role`). When checking a user's ability to take an application action, we'd check the perms on their `application_role`, `portfolio_role`, and `site_role`.


atst/domain/authz.py
```diff
@@ -1,14 +1,17 @@
...
 
 class Authorization(object):
     @classmethod
     def has_portfolio_permission(cls, user, portfolio, permission):
-        return permission in PortfolioRoles.portfolio_role_permissions(
-            portfolio, user
-        ) or Authorization.is_ccpo(user)
+        port_role = next(
+            (pr for pr in user.portfolio_roles if pr.portfolio == portfolio), None
+        )
```
patricksmithdds: Doesn't change much, but there is a [`first_or_none` utility for this](https://github.com/dod-ccpo/atst/blob/5f30b347c7713ec02aac12337e6152310544d263/atst/utils/__init__.py#L4). (Personally, I just hate writing `next`)

Is it possible for the user to have more than one portfolio role? Or, just more than one permission sets?

atst/models/portfolio_role.py
```diff
@@ -30,6 +29,14 @@ class Status(Enum):
     PENDING = "pending"
 
 
+portfolio_roles_roles = Table(
```
patricksmithdds: You may want to rename this table to match the `portfolio_roles_permission_sets` name.

atst/models/permissions.py
```diff
@@ -47,3 +46,43 @@ class Permissions(object):
     VIEW_TASK_ORDER = "view_task_order"
     UPDATE_TASK_ORDER = "update_task_order"
     ADD_TASK_ORDER_OFFICER = "add_task_order_officers"
```
patricksmithdds: Is the idea that all of these permissions will be removed in favor of the ones below?

atst/domain/task_orders.py
```diff
@@ -162,7 +162,13 @@ def add_officer(cls, user, task_order, officer_type, officer_data):
                 portfolio_user = existing_member.user
             else:
                 member = Portfolios.create_member(
-                    user, portfolio, {**officer_data, "portfolio_role": "officer"}
+                    user,
+                    portfolio,
+                    {
+                        **officer_data,
+                        "portfolio_role": "officer",
```
patricksmithdds: Is this `portfolio_role: "officer"` information still used? Or, does adding the permission sets obviate the need to have a special "officer" role?

atst/models/user.py
```diff
@@ -11,9 +11,9 @@ class User(Base, mixins.TimestampsMixin, mixins.AuditableMixin):
 
     id = types.Id()
     username = Column(String)
-    atat_role_id = Column(UUID(as_uuid=True), ForeignKey("roles.id"))
+    atat_role_id = Column(UUID(as_uuid=True), ForeignKey("permission_sets.id"))
```
patricksmithdds: This may be covered by your note on next steps, but does should this use a join table? Or: will a user have multiple site-wide permission sets? Will there be either the CCPO set or nothing?

atst/domain/permission_sets.py
```diff
@@ -0,0 +1,174 @@
...
+        "name": "default",
+        "display_name": "Default",
+        "description": "",
+        "permissions": [Permissions.REQUEST_JEDI_PORTFOLIO],
+    },
+]
+
+_VIEW_PORTFOLIO_PERMISSION_SETS = [
+    {
+        "name": "view_portfolio",
```
patricksmithdds: I'm curious if there's a way to get around using these string names elsewhere. Perhaps, similar to `Permissions`, using an enum-like class:

```
class PermissionSets(object):
    VIEW_PORTFOLIO = 'view_portfolio'
```

And then elsewhere, one could use `PermissionSets.VIEW_PORTFOLIO` instead of `"view_portfolio"`. 

I think the main help of this would be for developers -- I _know_ I'm going to typo one of the permission sets at some point, so I'd rather see an error about "PermissionSets has no attribute 'view_porfolio'" instead of receiving a 404 while testing because the perm set can't be found.

atst/domain/permission_sets.py
```diff
@@ -0,0 +1,174 @@
...
+            Permissions.VIEW_PORTFOLIO_ADMIN,
+            Permissions.VIEW_PORTFOLIO_NAME,
+            Permissions.VIEW_PORTFOLIO_USERS,
+            Permissions.VIEW_PORTFOLIO_ACTIVITY_LOG,
+            Permissions.VIEW_PORTFOLIO_POC,
+        ],
+    },
+]
+
+_EDIT_PORTFOLIO_PERMISSION_SETS = [
```
patricksmithdds: Rather than grouping between view and edit permission sets, have you considered grouping by app mgmt/funding/reporting/portfolio mgmt? I think that would make the tables when creating users easier, but I'm not sure if it would complicate matters elsewhere.

Also, if they're grouped by topic instead of level of permission, we could more easily add another level of permission -- will it always be nothing/edit/view, or could there be a view-some-edit-some set in the future?

templates/portfolios/applications/index.html
```diff
@@ -3,7 +3,7 @@
 
 {% extends "portfolios/base.html" %}
 
-{% set can_create_applications = user_can(permissions.ADD_APPLICATION_IN_PORTFOLIO) %}
+{% set can_create_applications = user_can(permissions.CREATE_APPLICATION) %}
```
patricksmithdds: It's gonna be real nice if/when this `user_can` macro can figure out the permission based on the portfolio/application/site permissions!

atst/domain/authz.py
dandds: I don't think a user should ever have more than one `portfolio_role` in a `portfolio`, no. Just multiple permission sets associated with that portfolio role.

atst/models/portfolio_role.py
dandds: Good catch!

atst/models/permissions.py
dandds: Yeah, I need to clean up the perms list.

atst/domain/task_orders.py
dandds: Yeah, good catch. It's not necessary anymore. I'm still not sure if officers need to have their authorization scoped to just the TO they're invited to (versus having portfolio-wide funding permissions), which would mean that they're a special case.

atst/models/user.py
dandds: There are potentially multiple types of CCPO users, we've found out. I think we can cover it by making a `users_permission_sets` join table. (At first I thought about having something closer to the `user` -> `portfolio_role` -> `portfolio_roles_permission_sets` setup, but I think that's needlessly complicated for site-wide permissions because the user can't have relationships to multiple sites.) And yes, that's the next PR.

atst/domain/permission_sets.py
dandds: Yeah, that's a very good idea. They could probably live in the `PermissionSets` domain class in this module.

atst/domain/permission_sets.py
dandds: Good question. I originally broke them down this way so that it was easy to get the list of "view" permission sets, because those are always applied as the base permission sets for portfolio users. I think your way would read better and be easier to update; I'd just have to come up with an easy way to still derive those defaults.

templates/portfolios/applications/index.html
dandds: Yeah, and hopefully the function underlying `user_can` will be the exact same function underlying the eventual route decorator! Thanks for taking a look through!

atst/domain/permission_sets.py
dandds: Actually, I take that back. That list of defaults should be its own constant in the `PortfolioRoles` domain class.

dandds: done in 088057f

atst/domain/permission_sets.py
dandds: done in 06c4df9

atst/domain/task_orders.py
dandds: fixed in 06c4df99a1ec269017ccb436c6c0dc0ce65f1a90

atst/domain/authz.py
dandds: done in df15995

atst/models/permissions.py
dandds: trimmed in c11b415

atst/models/portfolio_role.py
dandds: done in ee088f0

atst/domain/invitations.py
```diff
@@ -119,7 +119,7 @@ def resend(cls, user, portfolio_id, token):
         Authorization.check_portfolio_permission(
             user,
             portfolio,
-            Permissions.ASSIGN_AND_UNASSIGN_ATAT_ROLE,
+            Permissions.CREATE_PORTFOLIO_USERS,
```
drummond-dds: Question related to [the PR I merged recently](https://github.com/dod-ccpo/atst/pull/694) for resending invitations from the manage invitations page. 

[We left off permissioning](https://github.com/dod-ccpo/atst/blob/master/atst/routes/portfolios/task_orders.py#L111) in the resend method. Do you think I should add the same check in this method to my new resend method in `atst/routes/portfolios/task_orders.py`?

dandds: That's the one it needs, but I think we should wait until we have the route decorator because we'll make sure every route has an appropriate access/authorization decorator on that PR.

montana-mil:
i just checked the upgrade and downgrade for each of the 3 migrations, and they all worked fine 🎉

JRNPW:
I ran some basic tests locally on this branch. All functions working.

- Create new Portfolio/TO
- View as other, invited officers
- Add DD254 data, LOA, etc.
- Add Application, Environment, and Members

atst/domain/portfolios/portfolios.py
```diff
@@ -1,4 +1,4 @@
-from atst.domain.roles import Roles
+from atst.domain.permission_sets import PermissionSets
```
JRNPW: Putting this comment here, since I can't comment on the actual (unchanged) lines below: This PR makes Line 86 untested (a change from master).

dandds: We discussed on Slack, and this is because of the "no CCPO" problem that will be fixed on a follow-up branch.

atst/models/portfolio_role.py
```diff
@@ -1,5 +1,5 @@
 from enum import Enum
```
JRNPW: On master, the following are reported as partially covered branches: 87->88, 90->91. On this PR, it increases to three: 75->79, 97->98, 100->101. I assume that 75->79 is the new one. Can we make sure that we're back to two on this PR?

atst/models/permission_set.py
```diff
@@ -5,8 +5,8 @@
 from atst.models import Base, types, mixins
 
 
-class Role(Base, mixins.TimestampsMixin):
-    __tablename__ = "roles"
+class PermissionSet(Base, mixins.TimestampsMixin):
```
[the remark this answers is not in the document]
dandds: These methods aren't used anywhere, so I'd feel better removing them. (We overwrite the whole permission list when we need to make a change to the `PermissionSet` in `seed_roles.py`.)

atst/models/portfolio_role.py
dandds: 75->79 is sort of related to the problem I broke out into a card: https://www.pivotaltracker.com/n/projects/2160940/stories/164648971, but the test `test_has_portfolio_status_history` should cover it, to my eye. I don't know why it's not; that's exactly the code that should give the result in the assertions for that test.

In any event, I added more coverage for the status display logic to bring the total file coverage up.

PR #704, opened 2019-03-14
Clean up EditLink anchors

leigh-mil:
## Description
When I was looking into restarting a card, I noticed that some of the edit links in the KO Review Form had unnecessary anchors and that the anchor for Reporting on the TO Review page was not working properly. 

## Pivotal
[https://www.pivotaltracker.com/story/show/164669163](https://www.pivotaltracker.com/story/show/164669163)

### DIFF
--- a/templates/portfolios/task_orders/review.html
+++ b/templates/portfolios/task_orders/review.html
@@ -51,7 +51,7 @@
 
           <div class="h2">
             {{ "task_orders.new.review.funding"| translate }}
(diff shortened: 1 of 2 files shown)


montana-mil:
did you make a PT story for this? Do we need to? (genuine question)

leigh-mil:
> did you make a PT story for this? Do we need to? (genuine question)

I probably should... @dandds, should I?

PR #705, opened 2019-03-14
Have officer invite url point to the TO view page

leigh-mil:
## Description
Fix the link in the invite email so that all officers first go to the TO view page that includes the next steps. 

## Pivotal
[https://www.pivotaltracker.com/story/show/164544318](https://www.pivotaltracker.com/story/show/164544318)

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -186,17 +186,6 @@ class TaskOrders(object):
             )
 
     @classmethod
-    def is_officer_for_to(cls, task_order, user):
-        if (
-            user == task_order.contracting_officer
-            or user == task_order.contracting_officer_representative
-            or user == task_order.security_officer
-        ):
-            return True
-        else:
-            return False
-
-    @classmethod
     def add_dd_254(user, task_order, dd_254_data):
         dd_254 = DD254(**dd_254_data)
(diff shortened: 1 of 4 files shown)


atst/domain/task_orders.py
```diff
@@ -185,6 +185,17 @@ def add_officer(cls, user, task_order, officer_type, officer_data):
                 "{} is not an officer role on task orders".format(officer_type)
             )
 
+    @classmethod
+    def is_officer_for_to(cls, task_order, user):
+        if (
```
montana-mil: this might be a little more concise with something like this:
```
officers = [task_order.contracting_officer, task_order.contracting_officer_representative, task_order.security_officer]
return True if user in officers else False
```
or if that's not as readable, having the if/else but still using the list. just to collapse the `or`s a bit

leigh-mil: I like this pattern! I think I'm going to change this completely and create an `officers` property on the TO model instead that returns the list you made in the first line

leigh-mil:
> Could be a good place to add a test somewhere. This is not a fully fleshed out thought, so lmk if that is not as straightforward as im thinking.

Added in some tests for the COR and SO. There were existing tests to check if the KO or a non-officer user were directed to the correct url.

PR #708, opened 2019-03-14
Use NoAccessError instead of NotFoundError

leigh-mil:
## Description
`NotFoundError` was being used in places where a resource was found, but because of the state of the resource the user is not able to access it. This PR updates those cases to use `NoAccessError` instead.
I did a project-wide search to look at every instance that `NotFoundError` was used, but I would appreciate a second set of eyes to make sure I didn't miss any!

## Pivotal
[https://www.pivotaltracker.com/story/show/164444503](https://www.pivotaltracker.com/story/show/164444503)

### DIFF
--- a/atst/domain/invitations.py
+++ b/atst/domain/invitations.py
@@ -111,6 +111,10 @@ class Invitations(object):
     @classmethod
     def lookup_by_portfolio_and_user(cls, portfolio, user):
         portfolio_role = PortfolioRoles.get(portfolio.id, user.id)
+
+        if portfolio_role.latest_invitation is None:
+            raise NotFoundError("invitation")
+
         return portfolio_role.latest_invitation
 
     @classmethod
--- a/atst/routes/portfolios/task_orders.py
+++ b/atst/routes/portfolios/task_orders.py
@@ -126,8 +126,11 @@ def resend_invite(portfolio_id, task_order_id, form=None):
 
     invitation = Invitations.lookup_by_portfolio_and_user(portfolio, officer)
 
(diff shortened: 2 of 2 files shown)


atst/routes/portfolios/task_orders.py
```diff
@@ -126,9 +126,12 @@ def resend_invite(portfolio_id, task_order_id, form=None):
 
     invitation = Invitations.lookup_by_portfolio_and_user(portfolio, officer)
 
-    if not invitation or (invitation.status is not InvitationStatus.PENDING):
+    if not invitation:
         raise NotFoundError("invitation")
```
dandds: Could you move the `NotFoundError` into the `lookup_by_portfolio_and_user` method? I think that's more consistent with how we've been handling things. (i.e., `get`-type methods in domain classes raise `NotFoundError`).

atst/domain/invitations.py
```diff
@@ -110,7 +110,11 @@ def revoke(cls, token):
 
     @classmethod
     def lookup_by_portfolio_and_user(cls, portfolio, user):
-        portfolio_role = PortfolioRoles.get(portfolio.id, user.id)
+        try:
+            portfolio_role = PortfolioRoles.get(portfolio.id, user.id)
+        except NoResultFound:
+            raise NotFoundError("invitation")
+
         return portfolio_role.latest_invitation
```
dandds: Ah, actually `PortfolioRoles.get` will raise `NotFoundError`. I meant that we should raise it if `latest_invitation` is `None`, which is what this method is checking for.

PR #709, opened 2019-03-14
Resend invites fix

leigh-mil:
## Description
Bug fix where users could not resend an invite if it had expired. 

## Pivotal
[https://www.pivotaltracker.com/story/show/164642926](https://www.pivotaltracker.com/story/show/164642926)

### DIFF
--- a/tests/routes/portfolios/test_task_orders.py
+++ b/tests/routes/portfolios/test_task_orders.py
@@ -1,6 +1,6 @@
 from flask import url_for
 import pytest
-from datetime import timedelta, date
+from datetime import timedelta, date, datetime
 
 from atst.domain.permission_sets import PermissionSets
 from atst.domain.task_orders import TaskOrders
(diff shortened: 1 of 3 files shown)


montana-mil:
I'm going to pick this up next week when @leigh-mil is away

montana-mil:
This does not test for differences between raising `NotFoundError` and raising `NoAccessError`

tests/routes/portfolios/test_task_orders.py
```diff
@@ -694,3 +694,33 @@ def test_resend_invite_when_not_pending(app, client, user_session, portfolio, us
     assert original_invitation.status == InvitationStatus.ACCEPTED
     assert response.status_code == 404
     assert len(queue.get_queue()) == queue_length
+
+
+def test_resending_revoked_invite(app, client, user_session, portfolio, user):
```
dandds: This ones is good, but since the card is about being able to resend when the invite is expired, can you add a test for that too, please?

montana-mil: whoops, yes!

PR #711, opened 2019-03-15
Verify PDF signatures

drummond-dds:
Given a signed PDF we can perform some checks to confirm that:

* It is signed by a DOD issued certificate.
* That the DOD certificate used to sign hasn't expired or been revoked - Uses CRL checker for this.
* Hasn't been modified since it was signed
* Who the signer was (signer common name)

The method returns a dict of information relating to the PDF and signers certificates. This is useful for testing and debugging. The usage in our implementation is still to be determined so I included output I thought would be needed.

The test suite uses a bunch of signed PDFs for fixtures. I'm pretty happy with how its performing so far but I'd like to get more data to test against. You might notice some pending specs for PDFs I don't currently have.


tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+        "signatures": [
+            {
+                "cert_common_name": "WILLIAMS.SALLY.3453453453",
+                "hashed_binary_data": "d1fb3c955b57f139331586276ba4abca90ecc5d36b53fe6bbbbbd8707d7124bb",
+                "hashing_algorithm": "sha256",
+                "is_valid": False,
+                "is_valid_cert": True,
+                "is_valid_hash": False,
+                "is_valid_signature": True,
+                "signers_serial": 9_662_248_800_192_484_626,
```
drummond-dds: This is kinda funky but its how [black](https://github.com/ambv/black/commits/master) auto formats DoD IDs

dandds: I kind of like it. Ruby can do it too: https://batsov.com/articles/2013/07/02/the-elements-of-style-in-ruby-number-5-readability-of-long-numeric-literals/

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+            data.append(int(contents[i : i + 2], 16))
+
+        return cms.ContentInfo.load(bytes(data))["content"]
+
+    @property
+    def signers_cert(self):
+        """
+        This returns the certificate used to sign the PDF
+        """
+        if self._signers_cert == None:
```
dandds: `if self._signers_cert is None` is more idiomatic, and I think it's better optimized.

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+            return False
+
+    @property
+    def is_signature_valid(self):
+        """
+        Get signed PDF signature and determine if it was actually signed
+        by the certificate that it claims it was. Returns a boolean.
+        """
+        public_key = (
+            crypto.load_certificate(crypto.FILETYPE_PEM, self.signers_cert)
```
dandds: I would make a memoized property with the signers cert, since we parse it more than once.

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+            .get_pubkey()
+            .to_cryptography_key()
+        )
+        attrs = self.signed_binary_data["signer_infos"][0]["signed_attrs"]
+        signedData = None
+
+        if attrs is not None and not isinstance(attrs, core.Void):
+            signedData = attrs.dump()
+            signedData = b"\x31" + signedData[1:]
```
dandds: Can you explain what this bit does?

dandds: Also, picky, but probably use our variable name conventions: `signed_data`

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+
+        try:
+            public_key.verify(
+                bytes(self.signed_binary_data["signer_infos"][0]["signature"]),
+                signedData,
+                padding.PKCS1v15(),
+                getattr(hashes, self.hashing_algorithm.upper())(),
+            )
+            return True
+        except Exception:
```
dandds: Not a hill I'll die on (or hold up the PR for) but if you can find a more specific exception to catch, great.

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+def pdf_signature_validations(pdf=None):
+    signatures = []
+    start_byte = 0
+
+    while True:
+        start = start_byte + 1
+        n = pdf.find(b"/ByteRange", start)
+
+        if n == -1:
+            break
```
dandds: I think I'd find a regex iterator to find the indices more readable, like

```
import re
...
[m.start() for m in re.finditer(b"/ByteRange", pdf)]
```

but I'll defer to you. NBD, just preference.

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+        sig = signature.to_dict
+        response["signatures"].append(sig)
+
+        if not sig["is_valid"]:
+            response["result"] = "FAILURE"
+        elif response["result"] is not "FAILURE":
+            response["result"] = "OK"
+
+    if len(signatures) == 0:
+        response["result"] = "FAILURE"
```
dandds: If the result is just a string that's OK or FAILURE, is there a reason to not just use `True` or `False`?

tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+                "hashed_binary_data": "b879a15e19eece534dc63019d3fe539ff4a3efbf8e8f5403a8bdae26a9b713ea",
+                "hashing_algorithm": "sha256",
+                "is_valid": True,
+                "is_valid_cert": True,
+                "is_valid_hash": True,
+                "is_valid_signature": True,
+                "signers_serial": 9_662_248_800_192_484_626,
+            },
+            {
+                "cert_common_name": "VADER.DARTH.9012345678",
```
dandds: The fact that you never had to ask me what the common name was for these rando certs I made is a good sign 😁

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+    def is_cert_valid(self):
+        """
+        Takes the signing certificate and runs it through the CRLCache
+        checker. Returns a boolean.
+        """
+        try:
+            cache = CRLCache(
+                "ssl/server-certs/ca-chain.pem",
+                crl_locations=["ssl/client-certs/client-ca.der.crl"],
+            )
```
dandds: Can we make this configurable? I'd suggest just doing straight-up dependency injection. Let `PDFSignature` take `CRLCache` instance as an init argument. That way it never has to worry about CA chain details.

dandds: And, importantly, the caller in the Flask app can just pass `app.crl_cache` for that arg.

tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+        ],
+    }
+
+
+@pytest.mark.skip(reason="Need fixture file")
+def test_signed_pdf_dod_revoked():
+    signed_pdf_dod_revoked = open(
+        "tests/fixtures/signed-pdf-dod_revoked.pdf", "rb"
+    ).read()
+    result = pdf_signature_validations(pdf=signed_pdf_dod_revoked)
```
dandds: Hm, you could argue that checking revocations makes this an integration test, which might be more than we need. The burden for that should be on `CRLCache`, which is tested elsewhere. If you're injecting `CRLCache` like I suggest above, you could just monkeypatch it to raise the revocation error for this kind of test.

tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+                "signers_serial": None,
+            }
+        ],
+    }
+
+
+def test_signed_dod_pdf_signer_cert_expired():
+    #
+    # TODO: Is this good enough? Do we want an expired DOD certificate? This test is using
+    #       a fake DOD certificate.
```
dandds: The chances of us getting a PDF with a sig from an expired DoD cert are pretty low, unfortunately. As with auth, I think we need to make our own little PKI and test that way.

tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+                "hashing_algorithm": "sha256",
+                "is_valid": False,
+                "is_valid_cert": False,
+                "is_valid_hash": True,
+                "is_valid_signature": True,
+                "signers_serial": -180_673_825_300_246_991_177_196,
+            }
+        ],
+    }
+
```
dandds: I'd like a test with an invalid signature. Let's see if we can scrounge one up.

tests/utils/test_pdf_verification.py
```diff
@@ -0,0 +1,151 @@
...
+
+@pytest.mark.skip(reason="TODO")
+def test_crl_check_unavailable():
```
dandds: If we go the dependency injection route, I don't think the signature logic should be concerned about CRL specifics.

atst/utils/pdf_verification.py
```diff
@@ -0,0 +1,225 @@
...
+from cryptography.hazmat.primitives.asymmetric import padding
+
+
+class PDFSignature:
+    def __init__(self, byte_range_start=None, pdf=None):
+        self.pdf = pdf
+        self.byte_range_start = byte_range_start
+        self._signers_cert = None
+
+        # assert byte_range_start != -1 and self.start != -1 and self.stop != -1
```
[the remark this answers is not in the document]
drummond-dds: Deleted in afeca9ac

atst/utils/pdf_verification.py
drummond-dds: Fixed in 33a3f1a1

atst/utils/pdf_verification.py
drummond-dds: I got this from the reference implementation.

It takes the binary data and then pops off the first digit and replaces it with `b"\x31" (`1`)`.

```python
example = b"\x30\x32"
# => b'02'

example = b"\x31" + example[1:]
# => b'12'

```

drummond-dds: Standardized naming here 9725a4f5

atst/utils/pdf_verification.py
drummond-dds: Added in 336e8ab7

tests/utils/test_pdf_verification.py
drummond-dds: I removed this in 24770967 since we are delegating this to the `crl_checker` that is injected in

atst/utils/pdf_verification.py
drummond-dds: Fixed in 18691072

atst/utils/pdf_verification.py
drummond-dds: I'd like something more specific but I'm unsure what likely relevant errors we would have. I think this is fine since we are only wrapping one function call in the `try/except`.

tests/utils/test_pdf_verification.py
drummond-dds: This has been removed

tests/utils/test_pdf_verification.py
dandds: @drummond-dds I still think we need this test. The simplest thing to do would be to monkeypatch one of the arguments to `public_key.verify` so that it's random data.

drummond-dds: I added in a test case for this 9c2f8072efe7a4d2b3b467eaea49579cd82211e5

drummond-dds: @dandds I reimplemented the previous test case using `pytest-mock` 5572452cc04d4366a6d9ce6accdbb6623c22e0f5

PR #714, opened 2019-03-19
CCPO permission sets

dandds:
This is a follow-on PR to #701 .

This adds the concept of site-wide permission sets, primarily for CCPO. Previously, a `user` could have a relationship to a specific `role` (now `permission_set`). Now, a `user` can have relationships to many `permission_sets` through a join table. When we check if a user can take an action in a portfolio, we check if the permission exists in their `user` permissions (i.e., they're a CCPO with site-wide permissions) or in the permissions attached to their portfolio role. 

**notes**
- If you want to test this manually, you'll need to recreate Sam. You can't just re-run the `seed_sample.py` script, because that will leave Sam alone if he already exists. You'll need to drop him from the users table or recreate your database.
- This PR also restores a bunch of tests that I skipped in the last PR.
- This PR doesn't add any interface or ability to elevate users to CCPO.


atst/domain/authz.py
```diff
@@ -36,10 +39,6 @@ def check_atat_permission(cls, user, permission, message):
     def can_view_audit_log(cls, user):
         return Authorization.has_atat_permission(user, Permissions.VIEW_AUDIT_LOG)
 
-    @classmethod
```
montana-mil: should we still have a way to check if the user is CCPO?

atst/models/user.py
```diff
@@ -1,19 +1,29 @@
...
+class User(
+    Base, mixins.TimestampsMixin, mixins.AuditableMixin, mixins.PermissionsMixin
+):
     __tablename__ = "users"
 
     id = types.Id()
     username = Column(String)
-    atat_role_id = Column(UUID(as_uuid=True), ForeignKey("permission_sets.id"))
 
-    atat_role = relationship("PermissionSet")
```
montana-mil: 🔪 💯

atst/domain/authz.py
dandds: No, everything should be a permission check now, not a role check. CCPO will have most (maybe all) permissions site-wide.

atst/domain/users.py
```diff
@@ -52,18 +55,6 @@ def get_or_create_by_dod_id(cls, dod_id, **kwargs):
 
         return user
 
-    @classmethod
```
drummond-dds: Do we still have a way of revoking a permission?

dandds: @drummond-dds No; we don't know what that functionality looks like yet. Since this method was tied to the old role-based implementation and we don't know requirements for updating site-wide roles, I just removed it.

PR #716, opened 2019-03-19
Portfolio admin users table

montana-mil:
## Description
Read only table of members in a portfolio. This does not include drop downs, archive user button, add user, or save button.

There is still some work to do on the icon for the `Learn more...` link, captured in this bug: https://www.pivotaltracker.com/story/show/164742052

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164446620

## Screenshots
Screen Shot 2019-03-19 at 10 58 13 AM

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -168,13 +168,50 @@
   .member-list {
     .panel {
       @include shadow-panel;
+      padding-bottom: 0;
+    }
+
+    .member-list-header {
+      margin: 2 * $gap 5 * $gap;
+      padding: inherit;
+      overflow: auto;
+
+      .left {
+        float: left;
+        padding-bottom: 0.8rem;
+      }
+
+      .icon-link {
+        float: right;
+        margin-top: 0.8rem;
+      }
+
+      .icon {
+
+      }
+    }
+
+    .subheading {
+      font-size: 1.4rem;
+      color: $color-gray;
     }
 
     table {
       box-shadow: 0 6px 18px 0 rgba(144,164,183,0.3);
+
       thead {
         th:first-child {
           padding-left: 3 * $gap;
+        }
+
+        tr:first-child {
+          padding: 0 2 * $gap 0 5 * $gap;
+        }
+
+        td {
+          font-weight: bold;
+          font-size: 1.4rem;
+          border-top: 0;
         }
       }
 
@@ -186,8 +223,29 @@
         color: $color-gray;
       }
 
-      td {
-        border-bottom: 1px solid $color-gray-lightest;
+      td:first-child {
+        padding: 2 * $gap 2 * $gap 2 * $gap 5 * $gap;
+      }
+
+      tbody {
+        td {
+          border-bottom: 1px solid $color-gray-lightest;
+          font-size: 1.6rem;
+          border-top: 0;
+          padding: 3 * $gap 2 * $gap;
+        }
+
+        .green {
+          color: $color-green;
+        }
+
+        .name {
+          font-weight: bold;
+
+          .you {
+            font-size: 1.2rem;
+          }
+        }
       }
 
       .add-member-link {
--- a/templates/fragments/admin/portfolio_members.html
+++ b/templates/fragments/admin/portfolio_members.html
@@ -35,17 +35,17 @@
       </thead>
 
       <tbody>
-        {% for member in portfolio.members %}
+        {% for member_data in members_data %}
         <tr>
-            <td class='name'>{{ member.user_name }}
-              {% if member.user == user %}
+            <td class='name'>{{ member_data.member.user_name }}
+              {% if member_data.member.user == user %}
                 <span class='you'>(<span class='green'>you</span>)</span>
               {% endif %}
             </td>
-            {% set heading_perms = ['edit_application', 'view_portfolio_funding', 'view_portfolio_reports', 'edit_portfolio_name'] %}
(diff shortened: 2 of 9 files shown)


templates/fragments/admin/portfolio_members.html
```diff
@@ -0,0 +1,64 @@
...
+            <td class='name'>{{ member.user_name }}
+              {% if member.user == user %}
+                <span class='you'>(<a>you</a>)</span>
+              {% endif %}
+            </td>
+            {% set heading_perms = ['edit_application', 'view_portfolio_funding', 'view_portfolio_reports', 'edit_portfolio_name'] %}
+
+            {% for perm in heading_perms %}
+              {% if perm in member.permissions %}
+                <td><a>Edit Access</a></td>
```
dandds: Why does this have an anchor?

templates/fragments/admin/portfolio_members.html
```diff
@@ -0,0 +1,64 @@
...
+
+      <tbody>
+        {% for member in portfolio.members %}
+        <tr>
+            <td class='name'>{{ member.user_name }}
+              {% if member.user == user %}
+                <span class='you'>(<a>you</a>)</span>
+              {% endif %}
+            </td>
+            {% set heading_perms = ['edit_application', 'view_portfolio_funding', 'view_portfolio_reports', 'edit_portfolio_name'] %}
```
dandds: Here, you're checking individual permissions that are in the relevant permission sets. I don't think that's the best way to do this because the individual permissions might change (at least more frequently than the permission sets do). Also, we should never be using raw strings for this sort of check.

You could do this a number of ways, but I'd add a `has_permission_set` property to the `PortfolioRole` model, like:

```
@property
def has_permission_set(self, name):
    return first_or_none([prms.name == name for prms in self.permission_sets])
```

And then you can loop a list of the permission set names in the template like you're already doing (`[PermissionSets.EDIT_PORTFOLIO_ADMIN  ...]`). These are here: https://github.com/dod-ccpo/atst/blob/master/atst/domain/permission_sets.py#L14-L17

styles/components/_portfolio_layout.scss
```diff
@@ -186,8 +223,28 @@
...
 
+      td:first-child {
+        padding: 2 * $gap 2 * $gap 2 * $gap 5 * $gap;
+      }
+
       td {
         border-bottom: 1px solid $color-gray-lightest;
+        font-size: 1.6rem;
+        border-top: 0;
+        padding: 2 * $gap;
```
dandds: I think the wireframe has a little more space in the table in general than what renders right now. Maybe grab a designer to look at it.

dandds: BTW, this isn't necessary for me to approve and merge.

templates/fragments/admin/portfolio_members.html
montana-mil: it looked like it was going to need to be a link in the future, so i just added it here for styling

montana-mil: these will not be links! fixed.

styles/components/_portfolio_layout.scss
montana-mil: I got a bit more space, but the gray header should be a bit thinner. I couldn't get it easily, so leaving for now.

templates/portfolios/applications/index.html
```diff
@@ -45,13 +45,6 @@ <h3 class='icon-link accordian__title' v-on:click="props.toggle">{{ application.
                     <a class='icon-link' href='{{ url_for("portfolios.edit_application", portfolio_id=portfolio.id, application_id=application.id) }}'>
                       <span>{{ "portfolios.applications.app_settings_text" | translate }}</span>
                     </a>
-                    <div class='separator'></div>
-                  {% endif %}
-                  {% if user_can(permissions.VIEW_PORTFOLIO_USERS) %}
-                    <a class='icon-link' href='{{ url_for("portfolios.application_members", portfolio_id=portfolio.id, application_id=application.id) }}'>
```
dandds: Can we _just_ remove the href here for now? I think this mark-up is correct, and we'll just sub the correct url when we have a real app members page.

tests/models/test_portfolio_role.py
```diff
@@ -307,3 +307,21 @@ def test_can_list_all_permissions():
...
+    role_one = PermissionSets.get(PermissionSets.VIEW_PORTFOLIO_FUNDING)
+    role_two = PermissionSets.get(PermissionSets.VIEW_PORTFOLIO_REPORTS)
+    port_role = PortfolioRoleFactory.create(permission_sets=[role_one, role_two])
+
+    assert port_role.has_permission_set(PermissionSets.VIEW_PORTFOLIO_REPORTS)
+
+
+def test_does_not_have_permission_set():
+    role_one = PermissionSets.get(PermissionSets.VIEW_PORTFOLIO_FUNDING)
+    role_two = PermissionSets.get(PermissionSets.VIEW_PORTFOLIO_REPORTS)
```
dandds: You don't have to change this, but for future reference:

```
perm_sets = PermissionSets.get_many([PermissionSets.VIEW_PORTFOLIO_FUNDING, PermissionSets.VIEW_PORTFOLIO_REPORTS])
```

will work.

montana-mil: good to know!

montana-mil: i think it's good to use this pattern now, so i changed it :)

PR #717, opened 2019-03-20
View Primary Point of Contact (non-PoC)

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/164746444)

```cucumber
Given I am NOT the Primary Point of Contact for a portfolio,
When I view the admin page for a portfolio,
    Then I see a section "Primary Point of contact,"
    And it has the fields listed in notes,
    And there is NO “Update Primary PoC” button.
```

### DIFF
--- a/templates/fragments/primary_point_of_contact.html
+++ b/templates/fragments/primary_point_of_contact.html
@@ -1,7 +1,7 @@
 <div class="panel">
   <div class="panel__content">
-    <h2>Primary Point of Contact</h2>
-    <p>The PoC has the ability to edit all aspects of a portfolio.</p>
+    <h2>{{ "fragments.ppoc.title" | translate }}</h2>
+    <p>{{ "fragments.ppoc.subtitle" | translate }}</p>
 
     <p>
       <strong>
@@ -14,10 +14,12 @@
       {{ portfolio.owner.phone_number | usPhone }}
     </p>
 
-    <div class="flex-reverse-row">
-      <a class="usa-button-primary">
-        Update Primary PoC
-      </a>
-    </div>
+    {% if portfolio.owner == current_user %}
+      <div class="flex-reverse-row">
+        <a class="usa-button-primary">
+          {{ "fragments.ppoc.update_btn" | translate }}
+        </a>
+      </div>
+    {% endif %}
   </div>
 </div>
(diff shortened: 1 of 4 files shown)


templates/fragments/primary_point_of_contact.html
```diff
@@ -14,10 +14,12 @@ <h2>Primary Point of Contact</h2>
       {{ portfolio.owner.phone_number | usPhone }}
     </p>
 
-    <div class="flex-reverse-row">
-      <a class="usa-button-primary">
-        Update Primary PoC
-      </a>
-    </div>
+    {% if portfolio.owner == current_user %}
```
dandds: I think we should do a permission check here because this functionality will probably be available to CCPO. There is a separate permission set a PoC gets. The perms check would be something like `user_can(permissions.EDIT_PORTFOLIO_POC)`. Here's a random example from another template: https://github.com/dod-ccpo/atst/blob/master/templates/portfolios/header.html#L59-L65

tests/routes/portfolios/test_portfolios_index.py
```diff
@@ -48,14 +52,31 @@ def test_portfolio_index_without_existing_portfolios(client, user_session):
...
     )
     assert response.status_code == 200
     assert portfolio.name in response.data.decode()
+    assert translate("fragments.ppoc.update_btn").encode("utf8") in response.data
+
+
+def test_portfolio_admin_screen_when_not_ppoc(client, user_session):
+    portfolio = PortfolioFactory.create()
+    user = UserFactory.create()
+    permission_sets = PermissionSets.get_all()
```
dandds: I wouldn't do `get_all` here. That would give them CCPO-like abilities, which would be a problem in conjunction with the other change I recommended. You probably want a user who has all available portfolio perms sets except for the special PoC set.

drummond-dds: Fixed in 2d739110

templates/fragments/primary_point_of_contact.html
drummond-dds: Fixed in 2d739110

drummond-dds:
@dandds all comments are now addressed

PR #720, opened 2019-03-22
Access Decorator

dandds:
This PR moves our access/authorization checks into a route decorator. 

- The decorator takes the name of a permission as its primary argument. It will check if the current user has that permission at the portfolio or site level. 
- The decorator can also take a function as an "exception" keyword arg. If, given a certain set of arguments, the exception evaluates to `True` and does not raise an `UnauthorizedError`, the user passes the access check. This is useful for cases where things are restricted to just the KO, for example.
- The decorator logs all access attempts. Passed access checks generate INFO logs, failed ones are WARNINGS.

**notes**
- This adds a test file that confirms that every route that needs authorization protection has it, per @drummond-dds . It also has a separate access test for every authz-protected route, demonstrating that access works as expected. I removed lots of existing tests in favor of these.
- I have removed any pre-existing authorization checks in the domain methods and route functions in favor of the new model.
- I removed any domain methods, especially in `Portfolios`, that were basically wrappers over specific kinds of access checks. These have all been changed to more generic equivalents. I've also dropped `user` arguments to the domain methods in most places, since they were generally used for access checks.

### DIFF
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -1,10 +1,11 @@
 from functools import wraps
 
-from flask import g
+from flask import g, current_app as app, request
 
 from . import user_can_access
 from atst.domain.portfolios import Portfolios
 from atst.domain.task_orders import TaskOrders
+from atst.domain.exceptions import UnauthorizedError
 
 
 def evaluate_exceptions(user, permission, exceptions, **kwargs):
@@ -15,28 +16,48 @@ def evaluate_exceptions(user, permission, exceptions, **kwargs):
     )
 
 
+def check_access(permission, message, exceptions, *args, **kwargs):
+    access_args = {"message": message}
+
+    if "portfolio_id" in kwargs:
+        access_args["portfolio"] = Portfolios.get(
+            g.current_user, kwargs["portfolio_id"]
+        )
+    elif "task_order_id" in kwargs:
+        task_order = TaskOrders.get(kwargs["task_order_id"])
+        access_args["portfolio"] = task_order.portfolio
+
+    if exceptions and evaluate_exceptions(
+        g.current_user, permission, exceptions, **access_args, **kwargs
+    ):
+        return True
+
+    user_can_access(g.current_user, permission, **access_args)
+
+    return True
+
+
 def user_can_access_decorator(permission, message=None, exceptions=None):
     def decorator(f):
         @wraps(f)
         def decorated_function(*args, **kwargs):
-            access_args = {"message": message}
-
-            if "portfolio_id" in kwargs:
-                access_args["portfolio"] = Portfolios.get(
-                    g.current_user, kwargs["portfolio_id"]
+            try:
+                check_access(permission, message, exceptions, *args, **kwargs)
+                app.logger.info(
+                    "[access] User {} accessed {}".format(
+                        g.current_user.id, g.current_user.dod_id, request.path
+                    )
                 )
-            elif "task_order_id" in kwargs:
-                task_order = TaskOrders.get(kwargs["task_order_id"])
-                access_args["portfolio"] = task_order.portfolio
 
-            if exceptions and evaluate_exceptions(
-                g.current_user, permission, exceptions, **access_args, **kwargs
-            ):
                 return f(*args, **kwargs)
+            except UnauthorizedError as err:
+                app.logger.warning(
+                    "[access] User {} denied access to {}".format(
+                        g.current_user.id, g.current_user.dod_id, request.path
+                    )
+                )
 
-            user_can_access(g.current_user, permission, **access_args)
-
-            return f(*args, **kwargs)
+                raise (err)
 
         return decorated_function
 
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -1,0 +1,33 @@
+from functools import wraps
+
+from flask import g
+
+from . import user_can_access
+from atst.domain.portfolios import Portfolios
+
+
+def user_can_access_decorator(permission, message=None, exceptions=None):
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            access_args = {"message": message}
+
+            if "portfolio_id" in kwargs:
+                access_args["portfolio"] = Portfolios.get(
+                    g.current_user, kwargs["portfolio_id"]
+                )
+
+            if exceptions:
+                evaluated = [
+                    exc(g.current_user, permission, **access_args) for exc in exceptions
+                ]
+                if True in evaluated:
+                    return True
+
+            user_can_access(g.current_user, permission, **access_args)
+
+            return f(*args, **kwargs)
+
+        return decorated_function
+
+    return decorator
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -15,7 +15,8 @@ def check_access(permission, message, exception, *args, **kwargs):
         access_args["portfolio"] = Portfolios.get(
             g.current_user, kwargs["portfolio_id"]
         )
-    elif "task_order_id" in kwargs:
+
+    if "task_order_id" in kwargs:
         task_order = TaskOrders.get(kwargs["task_order_id"])
         access_args["portfolio"] = task_order.portfolio
 
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -8,12 +8,8 @@ from atst.domain.task_orders import TaskOrders
 from atst.domain.exceptions import UnauthorizedError
 
 
-def evaluate_exceptions(user, permission, exceptions, **kwargs):
-    return (
-        True
-        if True in [exc(user, permission, **kwargs) for exc in exceptions]
-        else False
-    )
+def evaluate_exceptions(user, exceptions, **kwargs):
+    return True if True in [exc(user, **kwargs) for exc in exceptions] else False
 
 
 def check_access(permission, message, exceptions, *args, **kwargs):
@@ -28,7 +24,7 @@ def check_access(permission, message, exceptions, *args, **kwargs):
         access_args["portfolio"] = task_order.portfolio
 
     if exceptions and evaluate_exceptions(
-        g.current_user, permission, exceptions, **access_args, **kwargs
+        g.current_user, exceptions, **access_args, **kwargs
     ):
         return True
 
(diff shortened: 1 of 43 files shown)


tests/domain/test_authz.py
```diff
@@ -58,3 +66,122 @@ def test_has_portfolio_permission():
...
+
+
+def test_user_can_access_decorator_exceptions(set_current_user):
+    rando_calrissian = UserFactory.create()
+    portfolio = PortfolioFactory.create()
+
+    because_i_said_so = lambda *args, **kwargs: True
+
+    @user_can_access_decorator(
+        Permissions.EDIT_PORTFOLIO_NAME, exception=because_i_said_so
```
montana-mil: 😱

tests/domain/test_authz.py
```diff
@@ -58,3 +66,122 @@ def test_has_portfolio_permission():
...
+
+    set_current_user(edit_admin)
+    assert _edit_portfolio_name(portfolio_id=portfolio.id)
+
+    set_current_user(view_admin)
+    with pytest.raises(UnauthorizedError):
+        _edit_portfolio_name(portfolio_id=portfolio.id)
+
+
+def test_user_can_access_decorator_exceptions(set_current_user):
```
drummond-dds: Do we have a test case when the exception is raised?

atst/domain/authz/decorator.py
```diff
@@ -0,0 +1,54 @@
...
+
+def user_can_access_decorator(permission, message=None, exception=None):
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            try:
+                check_access(permission, message, exception, *args, **kwargs)
+                app.logger.info(
+                    "[access] User {} accessed {}".format(
+                        g.current_user.id, g.current_user.dod_id, request.path
```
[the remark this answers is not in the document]
dandds: Good catch. I was initially logging the DoD ID but got worried that was PII in the logs. I forgot to update the format args. Logging the method is a good idea, too. Done in 601ca1c655fb221d8dac9a5dd1cb081b730f4a7e.

atst/routes/portfolios/applications.py
```diff
@@ -90,13 +96,20 @@ def update_application(portfolio_id, application_id):
...
+):
+    env_role = EnvironmentRoles.get(user.id, environment_id)
+    if not env_role:
+        raise UnauthorizedError(user, "access environment {}".format(environment_id))
+
+    return True
+
+
 @portfolios_bp.route("/portfolios/<portfolio_id>/environments/<environment_id>/access")
+@user_can(None, exception=wrap_environment_role_lookup, message="access environment")
```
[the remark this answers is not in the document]
dandds: There aren't permissions associated with `EnvironmentRoles`. Access is entirely based on whether a user has an env role or not. We can talk if that doesn't make sense.

tests/domain/test_authz.py
```diff
@@ -58,3 +66,122 @@ def test_has_portfolio_permission():
...
+    view_admin = UserFactory.create()
+
+    portfolio = PortfolioFactory.create(owner=edit_admin)
+    # factory gives view perms by default
+    PortfolioRoleFactory.create(user=view_admin, portfolio=portfolio)
+
+    # check a site-wide permission
+    assert user_can_access(ccpo, Permissions.VIEW_AUDIT_LOG)
+    with pytest.raises(UnauthorizedError):
+        user_can_access(edit_admin, Permissions.VIEW_AUDIT_LOG)
```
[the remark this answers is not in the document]
dandds: Yup, good catch: 6a18109bdd3594a9a32656d67241cdc8708ab91c

atst/routes/portfolios/applications.py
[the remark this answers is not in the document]
drummond-dds: Makes sense!

tests/domain/test_authz.py
dandds: Revised the test here: dc2781ec82350038890ef8ba31c3ff314a57d9fa

atst/routes/portfolios/index.py
```diff
@@ -37,14 +37,9 @@ def serialize_member(member):
...
-@portfolios_bp.route("/portfolios/<portfolio_id>/admin")
-def portfolio_admin(portfolio_id):
-    portfolio = Portfolios.get_for_update_information(g.current_user, portfolio_id)
-    form = PortfolioForm(data={"name": portfolio.name})
+def render_admin_page(portfolio, form):
     pagination_opts = Paginator.get_pagination_opts(http_request)
-    audit_events = AuditLog.get_portfolio_events(
-        g.current_user, portfolio, pagination_opts
-    )
+    audit_events = AuditLog.get_portfolio_events(portfolio, pagination_opts)
```
drummond-dds: We have a `VIEW_AUDIT_LOG` permission but I don't think we check that here?

atst/routes/portfolios/index.py
```diff
@@ -56,36 +51,41 @@ def portfolio_admin(portfolio_id):
...
-    portfolio = Portfolios.get_for_update_information(g.current_user, portfolio_id)
+    portfolio = Portfolios.get_for_update(portfolio_id)
     form = PortfolioForm(http_request.form)
     if form.validate():
         Portfolios.update(portfolio, form.data)
         return redirect(
             url_for("portfolios.portfolio_applications", portfolio_id=portfolio.id)
         )
     else:
-        return render_template("portfolios/edit.html", form=form, portfolio=portfolio)
```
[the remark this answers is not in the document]
dandds: Yeah, my test threw an error and I discovered the template was already gone! 😬

atst/routes/portfolios/index.py
dandds: :+1: added perms wrappers to the admin template here: 7d5e615e6a7db9b68bf4ba2c5ba9afb30779f8fe

atst/domain/authz/decorator.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def check_access(permission, message, exception, *args, **kwargs):
+    access_args = {"message": message}
+
+    if "portfolio_id" in kwargs:
+        access_args["portfolio"] = Portfolios.get(
+            g.current_user, kwargs["portfolio_id"]
+        )
+    elif "task_order_id" in kwargs:
```
drummond-dds: In many instances we have routes like `/portfolios/<portfolio_id>/task_orders/<task_order_id>`.

In those cases we would never get to the `elif` condition. I think this should just be two `if` statements?

drummond-dds: Or maybe the solution is to have the `task_order_id` check before the `portfolio_id` check?

dandds: :+1: 21af2b8ce97d2ff49e9edb2414b852af2b087749

drummond-dds:
👍 Looks good!

atst/domain/authz/decorator.py
```diff
@@ -0,0 +1,55 @@
...
+    if "portfolio_id" in kwargs:
+        access_args["portfolio"] = Portfolios.get(
+            g.current_user, kwargs["portfolio_id"]
+        )
+
+    if "task_order_id" in kwargs:
+        task_order = TaskOrders.get(kwargs["task_order_id"])
+        access_args["portfolio"] = task_order.portfolio
+
+    if exception is not None and exception(g.current_user, **access_args, **kwargs):
```
patricksmithdds: It took me a while to understand that this is "an exception to the rule" rather than a python "exception". I'm not sure if there's a clearer name?

atst/routes/task_orders/new.py
```diff
@@ -249,9 +251,23 @@ def get_started():
...
+        and "portfolio_id" not in kwargs
+    )
+
+
 @task_orders_bp.route("/task_orders/new/<int:screen>")
 @task_orders_bp.route("/task_orders/new/<int:screen>/<task_order_id>")
 @task_orders_bp.route("/portfolios/<portfolio_id>/task_orders/new/<int:screen>")
+@user_can(
+    Permissions.CREATE_TASK_ORDER,
+    exception=is_new_task_order,
```
patricksmithdds: I think this is the only place that both `exception` is specified and the required permission is not `None`. Other than this case, I was thinking you could have the first parameter be either a callable (in which case, call it like `exception` is being called) or a permission (in which case, just perform the permission check). 

If the initial portfolio creation is split out from the first task order screen (🤔 hasn't that been brought up before 😄 ), I think it _could_ be a cleaner interface -- or, maybe I'm just really stuck on `exception` meaning "error".

atst/routes/task_orders/new.py
```diff
@@ -298,6 +314,11 @@ def new(screen, task_order_id=None, portfolio_id=None):
 @task_orders_bp.route(
     "/portfolios/<portfolio_id>/task_orders/new/<int:screen>", methods=["POST"]
 )
+@user_can(
+    Permissions.CREATE_TASK_ORDER,
+    exception=is_new_task_order,
```
patricksmithdds: I think this exception will always be false because `portfolio_id` is in the URL.

atst/routes/__init__.py
```diff
@@ -143,9 +145,10 @@ def logout():
 
 
 @bp.route("/activity-history")
+@user_can(Permissions.VIEW_AUDIT_LOG, message="view activity log")
```
patricksmithdds: 💯 I think things are a lot cleaner with this decorator! 👏 👏

atst/routes/task_orders/new.py
dandds: Yeah, it feels ugly that the first argument is `None` in almost all of those cases. My justification is that there _might_ be a CCPO-level permission that eventually gives them access to things like the KO/COR review screen, so that eventually a lot of those `None`s would go away.

Agreed about the `exception` name.

>  hasn't that been brought up before

😁 we'll keep dreaming.

atst/domain/authz/decorator.py
dandds: After some time on thesaurus.com, I'll try `override`, I think.

tests/test_access.py
```diff
@@ -0,0 +1,789 @@
...
+
+    return _protected_routes
+
+
+sample_config = make_config({"CRL_STORAGE_PROVIDER": "LOCAL"})
+sample_app = make_app(sample_config)
+_PROTECTED_ROUTES = protected_routes(sample_app)
+
+
+class Null:
```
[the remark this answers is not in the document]
dandds: d'oh. The standard library has the goods every time.

atst/routes/task_orders/new.py
dandds: That's true for the third route, but this route function applies to POST `/task_orders/new/1`, too.

PR #721, opened 2019-03-25
User can only update apps or revoke invites related to their portfolios

leigh-mil:
## Description
This fixes a bug were a user could access the Application edit page and the members page for an app that they were not associated with. Replicate the bug by visiting:
`/portfolios/<portfolio-id-associated-with-current-user>/applications/<unassociated-app-id>/edit`
Users could also revoke invites that were unassociated with them

This fixes the issue by adding to the decorator introduced in #720.

## Pivotal
[https://www.pivotaltracker.com/story/show/164836692](https://www.pivotaltracker.com/story/show/164836692)

### DIFF
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -13,22 +13,22 @@ from atst.domain.exceptions import UnauthorizedError
 def check_access(permission, message, exception, *args, **kwargs):
     access_args = {"message": message}
 
-    if "portfolio_id" in kwargs:
-        access_args["portfolio"] = Portfolios.get(
-            g.current_user, kwargs["portfolio_id"]
-        )
-
     if "application_id" in kwargs:
         application = Applications.get(kwargs["application_id"])
         access_args["portfolio"] = application.portfolio
 
-    if "task_order_id" in kwargs:
+    elif "task_order_id" in kwargs:
         task_order = TaskOrders.get(kwargs["task_order_id"])
         access_args["portfolio"] = task_order.portfolio
 
-    if "token" in kwargs:
+    elif "token" in kwargs:
         invite = Invitations._get(kwargs["token"])
         access_args["portfolio"] = invite.portfolio_role.portfolio
+
+    elif "portfolio_id" in kwargs:
+        access_args["portfolio"] = Portfolios.get(
+            g.current_user, kwargs["portfolio_id"]
+        )
 
     if exception is not None and exception(g.current_user, **access_args, **kwargs):
         return True
--- a/atst/domain/authz/decorator.py
+++ b/atst/domain/authz/decorator.py
@@ -5,6 +5,7 @@ from flask import g, current_app as app, request
 from . import user_can_access
 from atst.domain.portfolios import Portfolios
 from atst.domain.task_orders import TaskOrders
+from atst.domain.applications import Applications
 from atst.domain.exceptions import UnauthorizedError
 
 
@@ -15,6 +16,10 @@ def check_access(permission, message, exception, *args, **kwargs):
         access_args["portfolio"] = Portfolios.get(
             g.current_user, kwargs["portfolio_id"]
         )
+
+    if "application_id" in kwargs:
+        application = Applications.get(kwargs["application_id"])
+        access_args["portfolio"] = application.portfolio
 
     if "task_order_id" in kwargs:
         task_order = TaskOrders.get(kwargs["task_order_id"])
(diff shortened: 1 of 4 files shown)


tests/routes/portfolios/test_applications.py
```diff
@@ -157,6 +157,47 @@ def test_user_without_permission_cannot_update_application(client, user_session)
...
+                "description": "More cool stuff happening here!",
+                "environments": [{"name": "dev"}],
+            }
+        ]
+    )
+    other_application = other_portfolio.applications[0]
+    user_session(portfolio.owner)
+
+    # user can't view application edit form
+    response = client.get(
```
drummond-dds: I think we're using `url_for` for generating urls in apps. Will be less brittle if paths change but route names don't.

```python
url_for("portfolios. edit_application", portfolio_id=portfolio_id, application_id=other_application.id)
```

tests/routes/portfolios/test_applications.py
```diff
@@ -157,6 +157,47 @@ def test_user_without_permission_cannot_update_application(client, user_session)
...
+
+    # user can't post update application form
+    response = client.post(
+        url_for(
+            "portfolios.update_application",
+            portfolio_id=portfolio.id,
+            application_id=other_application.id,
+        ),
+        data={"name": "New Name", "description": "A new description."},
+        follow_redirects=True,
```
drummond-dds: Lets remove the `follow_redirects=True`. If we had a weird scenario where it updated and then redirected to an invalid path we would return a `404` for that missing path.

tests/routes/portfolios/test_invitations.py
```diff
@@ -169,6 +169,58 @@ def test_revoke_invitation(client, user_session):
     assert invite.is_revoked
 
 
+def test_user_can_only_revoke_invites_in_their_portfolio(client, user_session):
+    portfolio = PortfolioFactory.create()
+    other_portfolio = PortfolioFactory.create()
+    user = UserFactory.create()
+    ws_role = PortfolioRoleFactory.create(
```
drummond-dds: In this PR lets use `portfolio_role`. We need a separate chore to remove all `ws_role` references elsewhere now that workspaces are ✂️

tests/routes/portfolios/test_applications.py
```diff
@@ -157,6 +157,47 @@ def test_user_without_permission_cannot_update_application(client, user_session)
...
+    )
+    assert response.status_code == 404
```
drummond-dds: Do you think its worthwhile having a test to confirm the `time_updated` doesn't change for the `application`?

drummond-dds:
Looks good just a few comments

atst/domain/authz/decorator.py
```diff
@@ -16,10 +18,18 @@ def check_access(permission, message, exception, *args, **kwargs):
             g.current_user, kwargs["portfolio_id"]
         )
 
+    if "application_id" in kwargs:
+        application = Applications.get(kwargs["application_id"])
+        access_args["portfolio"] = application.portfolio
```
dandds: This is clever. I was picturing a new domain method that would select by portfolio and app ID, but this is simpler. All of these conditionals are setting `access_args["portfolio"]`, so I think at this point it should be genuine branch logic (`if/elif`) in the other of preference with `portfolio_id` last. For a route like `/portfolios/<portfolio_id>/applications/<application_id>/whatever`, we'd run two queries when really we only need one.

tests/routes/portfolios/test_invitations.py
```diff
@@ -169,6 +169,58 @@ def test_revoke_invitation(client, user_session):
...
+    response = client.post(
+        url_for(
+            "portfolios.resend_invitation",
+            portfolio_id=portfolio.id,
+            token=invite.token,
+        )
+    )
+
+    assert response.status_code == 404
+    assert len(queue.get_queue()) == 0
```
dandds: I'm throwing this comment here because the appropriate test file is not on this PR, but can you add a test confirming the correct behavior for portfolios/task orders, too?

tests/routes/portfolios/test_applications.py
```diff
@@ -157,6 +157,52 @@ def test_user_without_permission_cannot_update_application(client, user_session)
...
+            portfolio_id=portfolio.id,
+            application_id=other_application.id,
+        ),
+        data={"name": "New Name", "description": "A new description."},
+    )
+    assert response.status_code == 404
+    assert time_updated == other_application.time_updated
+
+    # user can't view application members
+    response = client.get(
```
drummond-dds: I think we have one more `url_for`

drummond-dds:
👍 Looks good to me

PR #722, opened 2019-03-26
Render Edit Buttons on Portfolio Members Table

montana-mil:
## Description
Add archive, save, and add member buttons. Also added dropdowns that default to the user's permissions.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164446654

## Screenshot
Screen Shot 2019-03-26 at 11 23 23 AM


atst/routes/portfolios/index.py
```diff
@@ -23,27 +24,53 @@ def portfolios():
...
+def render_admin_page(portfolio, form=None):
     pagination_opts = Paginator.get_pagination_opts(http_request)
     audit_events = AuditLog.get_portfolio_events(portfolio, pagination_opts)
-    members_data = [serialize_member(member) for member in portfolio.members]
+    members_data = [serialize_member_form_data(member) for member in portfolio.members]
+
+    portfolio_form = PortfolioForm(data={"name": portfolio.name})
+    permissions_form = MembersPermissionsForm(
+        data={"members_permissions": members_data}
+    )
```
dandds: Quibble, but I think `member_perms_form` or something like that would be clearer. Just so someone reading the code knows it's the list of members for the template.

templates/components/options_input.html
```diff
@@ -12,19 +12,21 @@
...
-          </div>
-
-          {% if field.description %}
-            <span class='usa-input__help'>{{ field.description | safe }}</span>
-          {% endif %}
-
-          <span v-show='showError'>{{ Icon('alert',classes="icon-validation") }}</span>
-          <span v-show='showValid'>{{ Icon('ok',classes="icon-validation") }}</span>
-        </legend>
+        {% if label == True%}
```
dandds: Would it work to just check if `field.label` is `None`? (Note that `None` is lowercase `none` in Jinja templates because reasons.)

templates/task_orders/_new.html
```diff
@@ -45,3 +45,30 @@ <h1 class="task-order-form__heading subheading">
...
+        {% else %}
+          <td>{{ OptionsInput(permissions_form.access, default="View", label=False) }}</td>
+        {% endif %}
+      {% endfor %}
+
+      <td><button type="button" class='{{ archive_button_class }}'>{{ "portfolios.members.archive_button" | translate }}</button>
+      </td>
+
+  </tr>
+  {% endfor %}
```
dandds: Hm, what's this doing in the task order form base template? Is this just a copy error?

templates/components/options_input.html
drummond-dds: Can this just be `{% if label %}`?

templates/fragments/admin/portfolio_members.html
```diff
@@ -1,7 +1,10 @@
 {% from "components/icon.html" import Icon %}
+{% from "components/options_input.html" import OptionsInput %}
 
 <section class="member-list">
   <div class='responsive-table-wrapper panel'>
+    <form method='POST' autocomplete="off" enctype="multipart/form-data">
```
drummond-dds: Is this form missing an `action` and `csrf` token?

templates/components/options_input.html
montana-mil: yes, fixed! ty

atst/routes/portfolios/index.py
montana-mil: ok that makes sense

templates/task_orders/_new.html
montana-mil: lol

templates/components/options_input.html
montana-mil: i ended up just checking `if label` as george suggested. i think this was just vestigial code from earlier issues i was experiencing 😬

atst/routes/portfolios/index.py
montana-mil: good call, that is way clearer

templates/fragments/admin/portfolio_members.html
montana-mil: talked with @dandds and we are gonna handle that in this card: https://www.pivotaltracker.com/story/show/164446780

PR #723, opened 2019-03-26
Add portfolio user

dandds:
PT story: https://www.pivotaltracker.com/story/show/164446525

This adds the two-step modal for adding a new portfolio user. 

**notes**
- The modal will not let the user proceed with invalid data on the first modal screen. We talked about handling it this way during estimation for the card. I'm curious what should happen in the event that the data doesn't validate on the backend. (This shouldn't happen for a regular user, but I still think the route function logic should handle it somehow.)
- The second modal screen is valid by default since it has "view" as the default permission set for every select field.
- There's a bug in FF with the redirect anchor after a new member is successfully added. FF doesn't jump the user to the members table for some reason. It works in Chrome and IE 10.
- The way the step checkmarks have worked in the past is that they get "checked" when the user finished that screen. That's how I've implemented them here.

The add member button:
Screen Shot 2019-03-26 at 4 29 28 PM

Filled in first screen:
Screen Shot 2019-03-26 at 4 30 26 PM

Second screen:
Screen Shot 2019-03-26 at 4 32 25 PM

Added the user:
Screen Shot 2019-03-26 at 4 33 42 PM


atst/utils/flash.py
```diff
@@ -22,10 +22,9 @@
         "category": "success",
     },
     "new_portfolio_member": {
-        "title_template": "Member added successfully",
+        "title_template": "Success!",
         "message_template": """
-          <p>{{ new_member.user_name }} was successfully invited via email to this portfolio. They do not yet have access to any environments.</p>
-          <p><a href="{{ url_for('portfolios.update_member', portfolio_id=portfolio.id, member_id=new_member.user_id) }}">Add environment access.</a></p>
+          <p>You have successfully invited {{ new_member.user_name }} to the portfolio admin.</p>
```
drummond-dds: More a question for Jill but is "admin" needed at the end of this flash message? Might not even be worth looking into if this is just placeholder text.

js/components/text_input.js
```diff
@@ -97,6 +107,8 @@ export default {
 
       if (!this.modified && this.initialErrors && this.initialErrors.length) {
         valid = false
+      } else if (this.optional && value == '') {
```
drummond-dds: We're doing `===` above.

templates/fragments/admin/portfolio_members.html
```diff
@@ -1,65 +1,69 @@
...
+      <div class="action-group">
         {% if user_can(permissions.EDIT_PORTFOLIO_USERS) %}
-          {% include "fragments/admin/members_edit.html" %}
-        {% elif user_can(permissions.VIEW_PORTFOLIO_USERS) %}
-          {% include "fragments/admin/members_view.html" %}
+          <input
+            type='submit'
+            form="member-perms"
+            class='usa-button usa-button-primary'
+            value='{{ "Save" }}' />
```
drummond-dds: I think this is just a python string?

templates/fragments/admin/portfolio_members.html
```diff
@@ -1,65 +1,69 @@
...
+    <form method='POST' id="member-perms" autocomplete="off" enctype="multipart/form-data">
+      <div class='member-list-header'>
+        <div class='left'>
+          <div class='h3'>{{ "portfolios.admin.portfolio_members_title" | translate }}</div>
+          <div class='subheading'>
+            {{ "portfolios.admin.portfolio_members_subheading" | translate }}
+          </div>
         </div>
+        <a class='icon-link'>
+          <span class='icon'>{{ Icon('info') }}</span>
```
drummond-dds: Can we remove the `<span>`?

atst/utils/flash.py
dandds: Yeah, I just copied it straight out of the wireframe. I think we'll revisit content later.

montana-mil:
I noticed two things, but I'm not sure they should hold up this PR:
* On the first screen of the modal, if you press Enter instead of next, the member is added with view only access (skipping screen 2 of the modal)
* If a member has view only access instead of edit access, the portfolio members page has a large white footer where the save button would be. here's a screenshot:
Screen Shot 2019-03-26 at 5 08 23 PM

dandds:
@montana-mil those are good calls, will fix

PR #728, opened 2019-03-28
Edit Portfolio Members Access Levels

montana-mil:
## Description
This PR adds functionality to the dropdowns for the portfolio members table. When a user with `Edit Access` clicks `Save`, the permissions for users are updated. 

After clicking `Save`, the user is brought back to the members table and will see a Success banner. The success banner will always flash, even if no changes were sent, since in a later story, we will disable the `Save` button altogether until changes have been made.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164446780

## Screenshot
Screen Shot 2019-03-29 at 2 33 47 PM

### DIFF
--- a/atst/routes/portfolios/index.py
+++ b/atst/routes/portfolios/index.py
@@ -90,6 +90,20 @@ def portfolio_admin(portfolio_id):
     return render_admin_page(portfolio)
 
 
+def permission_set_has_changed(old_perm_set_names, new_perm_set_names):
+    has_changed = False
+    for perm_name in new_perm_set_names:
+        base = perm_name[4:]
+        if perm_name.split("_")[0] == "edit":
+            if perm_name not in old_perm_set_names:
+                has_changed = True
+        elif perm_name.split("_")[0] == "view":
+            edit_version = "edit" + base
+            if edit_version in old_perm_set_names:
+                has_changed = True
+    return has_changed
+
+
 @portfolios_bp.route("/portfolios/<portfolio_id>/admin", methods=["POST"])
 @user_can(Permissions.EDIT_PORTFOLIO_USERS, message="view portfolio admin page")
 def edit_portfolio_members(portfolio_id):
--- a/atst/routes/portfolios/index.py
+++ b/atst/routes/portfolios/index.py
@@ -5,6 +5,7 @@ from flask import render_template, request as http_request, g, redirect, url_for
 from . import portfolios_bp
 from atst.domain.reports import Reports
 from atst.domain.portfolios import Portfolios
+from atst.domain.portfolio_roles import PortfolioRoles
 from atst.domain.audit_log import AuditLog
 from atst.domain.common import Paginator
 from atst.forms.portfolio import PortfolioForm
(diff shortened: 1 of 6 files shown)


atst/routes/portfolios/index.py
```diff
@@ -86,6 +90,52 @@ def portfolio_admin(portfolio_id):
...
+    for perm_name in new_perm_set_names:
+        base = perm_name[4:]
+        if perm_name.split("_")[0] == "edit":
+            if perm_name not in old_perm_set_names:
+                has_changed = True
+        elif perm_name.split("_")[0] == "view":
+            edit_version = "edit" + base
+            if edit_version in old_perm_set_names:
+                has_changed = True
+    return has_changed
```
dandds: I was going to tell you to change `PortfolioRoles.update` to do this for you, but I did some digging. It looks like `sqlalchemy` is smart enough to not construct an update or delete query if none of the relationships to the permission sets change, so I don't think we should worry about checking whether to update. I would tear this out; if the updates become a performance problem we'll deal with it then.

atst/routes/portfolios/index.py
```diff
@@ -86,6 +90,52 @@ def portfolio_admin(portfolio_id):
...
+        user_id = subform.user_id.data
+        portfolio_role = PortfolioRoles.get(portfolio.id, user_id)
+        old_perm_set = [perm.name for perm in portfolio_role.permission_sets]
+
+        if permission_set_has_changed(old_perm_set, new_perm_set):
+            PortfolioRoles.update(portfolio_role, new_perm_set)
+            have_any_perms_changed = True
+
+    if have_any_perms_changed:
+        flash("update_portfolio_members", portfolio=portfolio)
```
dandds: Since, on a different card, we're not enabling the "save" button in the UI until the user has actually changed a form input, I think for now we can just always display the flash message here.

atst/routes/portfolios/index.py
```diff
@@ -86,6 +90,52 @@ def portfolio_admin(portfolio_id):
...
+
+
+@portfolios_bp.route("/portfolios/<portfolio_id>/admin", methods=["POST"])
+@user_can(Permissions.EDIT_PORTFOLIO_USERS, message="view portfolio admin page")
+def edit_portfolio_members(portfolio_id):
+    portfolio = Portfolios.get_for_update(portfolio_id)
+    member_perms_form = member_forms.MembersPermissionsForm(http_request.form)
+    have_any_perms_changed = False
+
+    for subform in member_perms_form.members_permissions:
```
dandds: I think we do need to validate the form, just in case someone has injected bad inputs. You can probably just validate the top-level `member_perms_form`. (And add a test for this. Let's talk about what should happen if it fails, since it's not something the user should be able to do in the regular UI.)

tests/routes/portfolios/test_admin.py
```diff
@@ -26,3 +27,103 @@ def test_member_table_access(client, user_session):
...
+        user=rando,
+        portfolio=portfolio,
+        permission_sets=[PermissionSets.get(PermissionSets.VIEW_PORTFOLIO_ADMIN)],
+    )
+
+    user = UserFactory.create()
+    PortfolioRoleFactory.create(
+        user=user,
+        portfolio=portfolio,
+        permission_sets=[PermissionSets.get(PermissionSets.EDIT_PORTFOLIO_ADMIN)],
```
dandds:
```suggestion
        permission_sets=PermissionSets.get_many(
            [PermissionSets.EDIT_PORTFOLIO_ADMIN, PermissionSets.VIEW_PORTFOLIO_ADMIN]
        ),
```
This should fix this test. You're getting an error because the initial POST is successful, but then the user isn't authorized to view the admin page when they're redirected.

atst/routes/portfolios/index.py
montana-mil: blast! i knew i was going off the deep end here.

atst/routes/portfolios/index.py
montana-mil: ah, that's a really good point.

atst/routes/portfolios/index.py
montana-mil: we will rerender the admin page if the form does not validate

PR #729, opened 2019-03-29
configure json logging for nginx in k8s cluster

dandds:
This adds a JSON log format for our NGINX web servers in the kubernetes pods. 

The new log format includes most of the information that was available in the default NGINX logs, with some changes:
- it removes the `time_local` field
- it adds a [`request_id`](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id) field
- it adds a `timestamp` field which uses the [ISO 8601 timestamp from NGINX](https://nginx.org/en/docs/http/ngx_http_log_module.html#var_time_iso8601)
- it adds an `msec` field, which corresponds to the NGINX [`$msec`](https://nginx.org/en/docs/http/ngx_http_log_module.html#var_msec) variable (a UNIX epoch timestamp with milliseconds); this is just for additional info if someone wants millisecond data which isn't available on the other timestamp

We're also passing the `request_id` to the UWSGI environment so that it will be available to the Flask app when we adjust those logs on a subsequent PR.

These changes are already applied. If you tail web server logs for any environment, you'll get JSON entries: `kubectl -n atat logs [pod-name] -f --tail=5`

### DIFF
--- a/deploy/kubernetes/atst-nginx-configmap.yml
+++ b/deploy/kubernetes/atst-nginx-configmap.yml
@@ -8,6 +8,7 @@ data:
   nginx-config: |-
     server {
         server_name www.atat.code.mil atat.code.mil;
+        access_log /var/log/nginx/access.log json;
         listen 8442;
         listen [::]:8442 ipv6only=on;
         if ($http_x_forwarded_proto != 'https') {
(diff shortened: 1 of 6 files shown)


deploy/kubernetes/atst-nginx-configmap.yml
```diff
@@ -75,5 +79,21 @@ data:
...
+    log_format json escape=json
+      '{'
+        '"timestamp":"$time_iso8601",'
+        '"msec":"$msec",'
+        '"request_id":"$request_id",'
+        '"remote_addr":"$remote_addr",'
+        '"remote_user":"$remote_user",'
+        '"request":"$request",'
+        '"status":$status,'
+        '"body_bytes_sent":$body_bytes_sent,'
```
drummond-dds: Is this the request body? If so we should be filtering out any secrets/passwords/tokens

drummond-dds:
One small comment on something I'm not sure about - logging secrets by accident

deploy/kubernetes/atst-nginx-configmap.yml
dandds: No, it's actually just the size of the body in bytes: http://nginx.org/en/docs/http/ngx_http_core_module.html#var_body_bytes_sent

It can be useful to see the number, especially if it's much larger than you'd expect.

PR #730, opened 2019-03-29
Remove Portfolio User

drummond-dds:
### Questions

Do we have a standard way of displaying a flash message in different form sections?

### Pivotal Tracker

* https://www.pivotaltracker.com/story/show/164446534

### Acceptance Criteria

```gherkin
Scenario: 
Given I am viewing the portfolio admin page as a user with Edit privileges for Portfolio Admin,
When I click "remove user" next to a user's row on the table,
Then I see a confirmation dialog per the design,
Then if I click 'Archive user" I am returned to the portfolio admin page at an anchor above the table,
And I see a success banner reading "You have successfully removed [User Name] from the portfolio,"
And the user has No Access to the portfolio.
```

### DIFF
--- a/atst/domain/portfolio_roles.py
+++ b/atst/domain/portfolio_roles.py
@@ -122,6 +122,15 @@ class PortfolioRoles(object):
         return PermissionSets.get_many(perms_set_names)
 
     @classmethod
+    def disable(cls, portfolio_role):
+        portfolio_role.status = PortfolioRoleStatus.DISABLED
+
+        db.session.add(portfolio_role)
+        db.session.commit()
+
+        return portfolio_role
+
+    @classmethod
     def update(cls, portfolio_role, set_names):
         new_permission_sets = PortfolioRoles._permission_sets_for_names(set_names)
         portfolio_role.permission_sets = new_permission_sets
--- a/atst/routes/portfolios/index.py
+++ b/atst/routes/portfolios/index.py
@@ -177,18 +177,16 @@ def portfolio_reports(portfolio_id):
 
 
 @portfolios_bp.route(
-    "/portfolios/<portfolio_id>/members/<member_id>/delete", methods=["POST"]
+    "/portfolios/<portfolio_id>/members/<user_id>/delete", methods=["POST"]
 )
 @user_can(Permissions.EDIT_PORTFOLIO_USERS, message="update portfolio members")
-def remove_member(portfolio_id, member_id):
-    if member_id == str(g.current_user.id):
+def remove_member(portfolio_id, user_id):
+    if str(g.current_user.id) == user_id:
         raise UnauthorizedError(
-            user=g.current_user, action="you cant remove yourself from the portfolio"
+            g.current_user, "you cant remove yourself from the portfolio"
         )
 
-    portfolio = Portfolios.get(g.current_user, portfolio_id)
-    portfolio_role = PortfolioRoles.get(portfolio_id=portfolio_id, user_id=member_id)
-
+    portfolio_role = PortfolioRoles.get(portfolio_id=portfolio_id, user_id=user_id)
     PortfolioRoles.disable(portfolio_role=portfolio_role)
 
     flash("portfolio_member_removed", member_name=portfolio_role.user.full_name)
@@ -196,7 +194,7 @@ def remove_member(portfolio_id, member_id):
     return redirect(
         url_for(
             "portfolios.portfolio_admin",
-            portfolio_id=portfolio.id,
+            portfolio_id=portfolio_id,
             _anchor="portfolio-members",
             fragment="portfolio-members",
         )
--- a/atst/routes/portfolios/index.py
+++ b/atst/routes/portfolios/index.py
@@ -2,8 +2,6 @@ from datetime import date, timedelta
 
 from flask import render_template, request as http_request, g, redirect, url_for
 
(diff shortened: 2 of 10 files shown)


templates/fragments/admin/portfolio_members.html
```diff
@@ -49,6 +52,34 @@
...
+          }}
+
+          <div class="action-group">
+            <form method="POST" action="{{ url_for('portfolios.remove_member', portfolio_id=portfolio.id, member_id=member.user_id)  }}">
+              {{ member_perms_form.csrf_token }}
+              <button class="usa-button usa-button-danger">
+                {{ "portfolios.members.archive_button" | translate }}
+              </button>
+            </form>
+            <a v-on:click="closeModal('d')">Cancel</a>
```
dandds: Should `d` be the `modal_id`? (It doesn't actually matter b/c that method doesn't use the name, but I don't know what `d` is here.)

atst/domain/portfolio_roles.py
```diff
@@ -121,6 +121,15 @@ def _permission_sets_for_names(cls, set_names):
...
         return PermissionSets.get_many(perms_set_names)
 
+    @classmethod
+    def disable(cls, portfolio_role):
+        portfolio_role.status = PortfolioRoleStatus.DISABLED
+
+        db.session.add(portfolio_role)
+        db.session.commit()
+
+        return portfolio_role
```
dandds: Setting the role to DISABLED removes them from the members list because of how we derive that list [here](https://github.com/dod-ccpo/atst/blob/master/atst/models/portfolio.py#L46-L53), but it doesn't strip their access. Like we talked about, we need to check whether the role is ACTIVE when checking perms in `Authorization.has_portfolio_permission`.

tests/routes/portfolios/test_portfolios_index.py
```diff
@@ -81,6 +83,56 @@ def test_portfolio_admin_screen_when_not_ppoc(client, user_session):
...
+    response = client.post(
+        url_for(
+            "portfolios.remove_member",
+            portfolio_id=portfolio.id,
+            member_id=portfolio.owner.id,
+        ),
+        follow_redirects=False,
+    )
+
+    assert response.status_code == 500
```
dandds: I'm surprised this is a 500; the app should catch and return 404 for `UnauthorizedError`, I thought.

atst/routes/portfolios/index.py
```diff
@@ -143,3 +145,30 @@ def portfolio_reports(portfolio_id):
...
+
+@portfolios_bp.route(
+    "/portfolios/<portfolio_id>/members/<member_id>/delete", methods=["POST"]
+)
+@user_can(Permissions.EDIT_PORTFOLIO_USERS, message="update portfolio members")
+def remove_member(portfolio_id, member_id):
+    if member_id == str(g.current_user.id):
+        raise UnauthorizedError(
+            user=user, message="you cant remove yourself from the portfolio"
+        )
```
dandds: I'm a little torn. This is definitely one-off logic, but I sort of feel like the route functions shouldn't be responsible for ever raising `UnauthorizedError`. Let's leave it for now, but I think eventually we might need another kwarg that takes a function on the `user_can` decorator to signify an additional check.

tests/routes/portfolios/test_portfolios_index.py
```diff
@@ -81,6 +83,56 @@ def test_portfolio_admin_screen_when_not_ppoc(client, user_session):
...
+    assert (
+        PortfolioRoles.get(portfolio_id=portfolio.id, user_id=portfolio.owner.id).status
+        == PortfolioRoleStatus.ACTIVE
+    )
+
```
dandds: Besides any `Authorization` tests you add, can we get another test in here to confirm that the archived user can't hit a portfolio route?

tests/routes/portfolios/test_portfolios_index.py
drummond-dds: I think it should be `401 Unauthorized`

drummond-dds: Fixed in 6d3c2aa9

templates/fragments/admin/portfolio_members.html
drummond-dds: Fixed in 318b6fe3

atst/domain/portfolio_roles.py
drummond-dds: Added in 0a50b775d987dfdd765c62b850a0e30fd216c8d5

tests/routes/portfolios/test_portfolios_index.py
drummond-dds: I added in 0a50b775d987dfdd765c62b850a0e30fd216c8d5

tests/domain/test_authz.py
```diff
@@ -101,6 +102,14 @@ def test_user_can_access():
             view_admin, Permissions.EDIT_PORTFOLIO_NAME, portfolio=portfolio
         )
 
+    # check when portfolio_role is disabled
+    view_admin_pr = PortfolioRoles.get(portfolio_id=portfolio.id, user_id=view_admin.id)
```
dandds: I would just capture the factory instance on line 79 rather than look it up again.

drummond-dds: Fixed in ee1648a

PR #732, opened 2019-04-01
ATST Json logging

dandds:
This adds a simple JSON logger and some log context filters to the app. JSON logging will be enabled either when the `FLASK_ENV` is `prob` or when `LOG_JSON` is true. (You can demo it for yourself with `LOG_JSON=true ./script/server`.

`RequestContextFilter` injects the request ID and current user ID into the log record if they're available. `JsonFormatter` constructs the log as a JSON. We can iterate on the fields and structure of the JSON; this is just a first pass to build out the infrastructure.

**note**
The `werkzeug` access logs are not lovely when you're running in locally:
```
{"timestamp": "2019-04-01T18:02:00.071555+00:00", "version": 1, "request_id": null, "user_id": null, "severity": "INFO", "tags": null, "message": "127.0.0.1 - - [01/Apr/2019 14:02:00] \"GET /static/img/us_flag_small.png HTTP/1.1\" 304 -"}
```
`werkzeug` has a default message formatter that you can't override. `werkzeug` isn't our production Flask server, though, `uWSGI` is. `uWSGI` logging will be handled separately.

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -29,11 +29,16 @@ from atst.utils.form_cache import FormCache
 from atst.utils.json import CustomJSONEncoder
 from atst.queue import queue
 
+from logging.config import dictConfig
+from atst.utils.logging import JsonFormatter, RequestContextFilter
+
 
 ENV = os.getenv("FLASK_ENV", "dev")
 
 
 def make_app(config):
+    if ENV == "prod" or config.get("LOG_JSON"):
+        apply_json_logger()
 
     parent_dir = Path().parent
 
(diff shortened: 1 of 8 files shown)


atst/app.py
```diff
@@ -29,11 +29,16 @@
 from atst.utils.json import CustomJSONEncoder
 from atst.queue import queue
 
+from logging.config import dictConfig
```
drummond-dds: Where is the `dictConfig` used? I see we declare a variable named that on line 240

dandds: In the apply_json_logger in this file. Kinda hidden. It seems like a weird way to configure a logger, but it's what the docs recommend (vs. letting it build the default logger and then changing it in-place).

PR #733, opened 2019-04-01
Add last login timestamp

leigh-mil:
## Description
Add last login timestamp. Still waiting on final design for timestamp location and styling

## Pivotal
[https://www.pivotaltracker.com/story/show/164910919](https://www.pivotaltracker.com/story/show/164910919)

## Screenshot
Screen Shot 2019-04-03 at 12 48 51 PM


atst/domain/users.py
```diff
@@ -82,6 +83,12 @@ def update(cls, user, user_delta):
 
         return user
 
+    @classmethod
+    def update_last_login(cls, user):
+        setattr(user, "last_login", datetime.now())
```
dandds: Why not just `user.late_login = datetime.now()`?

tests/domain/test_users.py
```diff
@@ -65,3 +66,15 @@ def test_update_user_with_dod_id():
...
     assert "dod_id" in str(excinfo.value)
+
+
+def test_update_user_with_last_login():
+    new_user = UserFactory.create(last_login=datetime.now())
+    Users.update_last_login(new_user)
+    last_login = new_user.last_login
+
+    with pytest.raises(UnauthorizedError):
+        Users.update(new_user, {"last_login": datetime.now()})
```
dandds: Let's talk through this. I don't know if there's a good reason you shouldn't also be able to update the `last_login` with `Users.update` (I also don't know whether you can or not is relevant to this test).

templates/footer.html
```diff
@@ -7,4 +7,9 @@
       <span>{{ "footer.jedi_help_link_text" | translate }}</span>
     </a>
   </div>
+  {% if g.last_login %}
+    <div class="">
+      Last Login: <local-datetime timestamp='{{ g.last_login }}' format='MMM D YYYY H:mm Z'></local-datetime>
```
dandds: Does this STIG say we need to indicate the TZ? If not, I'd say that's a design question.

dandds: One other thought: if we do indicate the TZ, we might want to use a human-friendly string like "EST" instead of the offset.

atst/domain/auth.py
```diff
@@ -50,9 +52,18 @@ def get_current_user():
         return False
 
 
+def get_last_login():
+    last_login = session.get("last_login")
+    if last_login and session.get("user_id"):
+        return last_login
+    else:
+        return False
```
dandds: Very small, but maybe try:

```
return session.get("user_id") and session.get("last_login")
```

instead.

PR #738, opened 2019-04-02
Update Point of Contact

drummond-dds:
### Ticket

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/164447295)

### Acceptance Criteria

```gherkin
	Given I am the Primary Point of Contact for a portfolio,
	When I click the ‘Change Primary Point of Contact’ button,
	Then I see a modal with an info box per the design
	Then if I click Confirm I see a modal with the following:
	A warning with text per the design,
	A selection box reading “select primary point of contact,” populated with all portfolio users,
	But Pending users are grayed out and “(Pending)” is next to their name
	And if I select one and click Confirm,
	Then that user is set to be the Primary Point of Contact,
	And that user is given “Edit” access for all portfolio privileges,
	And I am set to NOT be the PoC,
	Then I am returned to the Portfolio Admin page,
	And I am sent to an anchor at the top of the Primary Point of Contact section,
	And I see a success banner reading “You have successfully added [new PoC name] as Point of Contact. You are no longer the PoC.” (see https://promptworks.invisionapp.com/share/CYQR5UKA5HM#/screens/349427846)
	And I can see the new PoC’s profile information in the PoC box.
```

### DIFF
--- a/atst/domain/portfolio_roles.py
+++ b/atst/domain/portfolio_roles.py
@@ -122,6 +122,28 @@ class PortfolioRoles(object):
         return PermissionSets.get_many(perms_set_names)
 
     @classmethod
+    def make_ppoc(cls, portfolio_role):
+        portfolio = portfolio_role.portfolio
+        original_owner_role = PortfolioRoles.get(
+            portfolio_id=portfolio.id, user_id=portfolio.owner.id
+        )
+        PortfolioRoles.revoke_ppoc_permissions(portfolio_role=original_owner_role)
+        PortfolioRoles.add(
+            user=portfolio_role.user,
+            portfolio_id=portfolio.id,
+            permission_sets=PortfolioRoles.PORTFOLIO_PERMISSION_SETS,
+        )
+
+    @classmethod
+    def revoke_ppoc_permissions(cls, portfolio_role):
+        permission_sets = [
+            permission_set.name
+            for permission_set in portfolio_role.permission_sets
+            if permission_set.name != PermissionSets.PORTFOLIO_POC
+        ]
+        PortfolioRoles.update(portfolio_role=portfolio_role, set_names=permission_sets)
+
+    @classmethod
     def disable(cls, portfolio_role):
         portfolio_role.status = PortfolioRoleStatus.DISABLED
 
--- a/atst/routes/portfolios/index.py
+++ b/atst/routes/portfolios/index.py
@@ -6,12 +6,14 @@ from . import portfolios_bp
 from atst.domain.reports import Reports
 from atst.domain.portfolios import Portfolios
 from atst.domain.portfolio_roles import PortfolioRoles
+from atst.domain.permission_sets import PermissionSets
+from atst.domain.users import Users
 from atst.domain.audit_log import AuditLog
 from atst.domain.common import Paginator
+from atst.domain.exceptions import NotFoundError
 from atst.forms.portfolio import PortfolioForm
 import atst.forms.portfolio_member as member_forms
 from atst.models.permissions import Permissions
-from atst.domain.permission_sets import PermissionSets
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.utils.flash import formatted_flash as flash
 from atst.domain.exceptions import UnauthorizedError
@@ -70,12 +72,19 @@ def render_admin_page(portfolio, form=None):
     member_perms_form = member_forms.MembersPermissionsForm(
         data={"members_permissions": members_data}
     )
+
+    assign_ppoc_form = member_forms.AssignPPOCForm()
+    assign_ppoc_form.user_id.choices += [
+        (user.id, user.full_name) for user in portfolio.users
+    ]
+
     return render_template(
         "portfolios/admin.html",
         form=form,
         portfolio_form=portfolio_form,
         member_perms_form=member_perms_form,
         member_form=member_forms.NewForm(),
+        assign_ppoc_form=assign_ppoc_form,
         portfolio=portfolio,
         audit_events=audit_events,
         user=g.current_user,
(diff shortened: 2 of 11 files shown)


templates/components/multi_step_modal_form.html
```diff
@@ -25,13 +25,16 @@
...
   <multi-step-modal-form inline-template :steps={{ step_count }}>
     <div>
-      <a class='icon-link modal-link' v-on:click="openModal('{{ name }}')">
+      <a class='{{ link_classes }}' v-on:click="openModal('{{ name }}')">
         {{ button_text }}
-        {{ Icon('plus-circle-solid') }}
+
+        {% if button_icon %}
+          {{ Icon('plus-circle-solid') }}
+        {% endif %}
```
dandds: You're passing a string for the icon name in `add_new_portfolio_member.html` and then treating that arg as a boolean here without passing it to the `Icon` component. I also think `button_icon` should be a kwarg with a default, even if it's just empty string.

tests/routes/portfolios/test_portfolios_index.py
```diff
@@ -26,6 +30,112 @@ def test_update_portfolio_name(client, user_session):
...
+    )
+    assert (
+        Permissions.EDIT_PORTFOLIO_POC
+        in PortfolioRoles.get(
+            portfolio_id=portfolio.id, user_id=new_ppoc.id
+        ).permissions
+    )
+
+    #
+    # TODO: assert ppoc change is in audit trail
```
dandds: I know this is just a todo, but I think that should be a separate test in `tests/domain/test_audit_log.py`. Whether it appears in the audit log is a side effect of the change that the route function under test shouldn't know about.

atst/routes/portfolios/index.py
```diff
@@ -86,6 +97,40 @@ def portfolio_admin(portfolio_id):
     return render_admin_page(portfolio)
 
 
+@portfolios_bp.route("/portfolios/<portfolio_id>/update_ppoc", methods=["POST"])
+@user_can(Permissions.EDIT_PORTFOLIO_POC, message="update portfolio ppoc")
+def update_ppoc(portfolio_id):
+    user_id = http_request.form.get("user_id")
```
dandds: What happens when there is no `user_id`? One thing you could do is build the form the way you did in the GET route and just validate it. Then you could just handle it like a regular form validation error and you wouldn't have to raise `NotFoundError` below.

atst/routes/portfolios/index.py
```diff
@@ -66,12 +69,20 @@ def render_admin_page(portfolio, form=None):
     member_perms_form = member_forms.MembersPermissionsForm(
         data={"members_permissions": members_data}
     )
+
+    assign_ppoc_form = member_forms.AssignPPOCForm()
+    assign_ppoc_form.user_id.choices = [("", "- Select -")]
```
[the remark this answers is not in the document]
drummond-dds: Done in 21663caf

atst/routes/portfolios/index.py
```diff
@@ -66,12 +69,20 @@ def render_admin_page(portfolio, form=None):
...
+
+    for user in portfolio.users:
+        assign_ppoc_form.user_id.choices.append((user.id, user.full_name))
```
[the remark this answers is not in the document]
drummond-dds: Fixed in 18db69c7

atst/routes/portfolios/index.py
```diff
@@ -86,6 +97,40 @@ def portfolio_admin(portfolio_id):
...
+    if Users.get(user_id) not in portfolio.users:
+        raise NotFoundError("user not in portfolio")
+
+    #
+    # Reset original PPOCs permissions to default for a portfolio
+    #
+    original_portfolio_role = PortfolioRoles.get(
+        portfolio_id=portfolio.id, user_id=owner.id
+    )
+    PortfolioRoles.reset_default_permission_sets(portfolio_role=original_portfolio_role)
```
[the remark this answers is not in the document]
drummond-dds: I removed the PPOC specific permissions and retained original here 0493e983

atst/routes/portfolios/index.py
```diff
@@ -86,6 +97,40 @@ def portfolio_admin(portfolio_id):
...
+    user = Users.get(user_id)
+    PortfolioRoles.add(
+        user=user,
+        portfolio_id=portfolio.id,
+        permission_sets=[PermissionSets.PORTFOLIO_POC],
+    )
+
+    return redirect(
+        url_for("portfolios.portfolio_applications", portfolio_id=portfolio.id)
+    )
```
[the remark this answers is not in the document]
drummond-dds: Fixed in 09e1a588

atst/routes/portfolios/index.py
drummond-dds: Added a test case 39c512ca

atst/routes/portfolios/index.py
```diff
@@ -86,6 +97,40 @@ def portfolio_admin(portfolio_id):
...
+
+    #
+    # Add PORTFOLIO_POC permissions to new ppoc
+    #
+    user = Users.get(user_id)
+    PortfolioRoles.add(
+        user=user,
+        portfolio_id=portfolio.id,
+        permission_sets=[PermissionSets.PORTFOLIO_POC],
+    )
```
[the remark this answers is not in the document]
drummond-dds: Fixed in 4f69335 & b77c0c33

templates/components/multi_step_modal_form.html
drummond-dds: Fixed in dad123dc

tests/routes/portfolios/test_portfolios_index.py
drummond-dds: Removed comment in cd330eb0

drummond-dds: @dandds do we need to test this or are we confident the audit trail is working from other tests in the system?

drummond-dds:
I can squash the commits for this PR before merge

tests/routes/portfolios/test_portfolios_index.py
dandds: I think the audit trail could use more tests, but I don't think you need to do it on this PR.

atst/domain/portfolio_roles.py
```diff
@@ -121,6 +121,28 @@ def _permission_sets_for_names(cls, set_names):
...
+        portfolio = portfolio_role.portfolio
+        original_owner_role = PortfolioRoles.get(
+            portfolio_id=portfolio.id, user_id=portfolio.owner.id
+        )
+        PortfolioRoles.revoke_ppoc_permissions(portfolio_role=original_owner_role)
+        PortfolioRoles.add(
+            user=portfolio_role.user,
+            portfolio_id=portfolio.id,
+            permission_sets=[PermissionSets.PORTFOLIO_POC],
+        )
```
dandds: This method still needs to do this part: 

> when a user is promoted to PPOC, they get the special PPOC perm sets _and_ all of the edit/view perm sets if they didn't already have them.

Right now, I can make a user who only has view perms PPOC and they still only have view perms after the change.

drummond-dds: Added in 9b6e69ed

drummond-dds:
@dandds disabling pending users will be a follow up ticket https://www.pivotaltracker.com/story/show/165150876

PR #740, opened 2019-04-03
uWSGI JSON logging

dandds:
This adds JSON-formatted logs for uWSGI, our application server. 

**notes**
- It also adds config to turn on JSON logging for the atst, the Flask app. The atst JSON logger was added but not enabled in #732 .
- This relies on the uWSGI alpine logger package I added in #739 .
- There are three kinds of log outputs in uWSGI: general info logs ("I started or stopped"), request logs (like general NGINX/Apache access logs), and our Flask app logs. It's not possible to log the granular uWSGI request log info as JSON because of config limitations. See discussion of the problem [here](https://stackoverflow.com/questions/46534829/how-to-configure-uwsgi-to-encode-logging-as-json-except-app-output) and [here](https://github.com/unbit/uwsgi/issues/1654). I opted to implement a version of what's recommended in [this gist](https://gist.github.com/amgorb/2f8a834f7f1c400c50a8b616f4391117). We create three separate loggers for general info logs, request logs, and application logs. Granular request info is logged as key/value pairs in the message field for request log JSON.

**Testing**

- I made samples of the NGINX and uWSGI config that live in `sample-server`. I thought this would be useful for quickly playing with uWSGI and NGINX config in a way that's closer to our production environment. There's a README in that directory for running the sample. 
- The config has been applied to test.atat.code.mil. You can see the uWSGI logs with:

```
kubectl -n atat-test logs [pod name] -c atst -f
```


sample-server/README
```diff
@@ -0,0 +1,17 @@
...
+- pipenv and honcho: If you have the regular atst dependencies installed, you already have pipenv and honcho
+
+## Run
+
+The sample server runs at localhost:8000, so you cannot have another copy of the app running on that port.
+
+Start both NGINX and uWSGI with honcho:
+
+```
+pipenv run honcho start -f sample-config/Procfile
```
leigh-mil: Do you mean `sample-server/Procfile` here?

leigh-mil:
Looks good! Just one question about the Readme!

sample-server/README
dandds: Sure did. Fixed by amending the commit, thanks.

PR #741, opened 2019-04-03
Render Resend Invite Button Bug

montana-mil:
## Description
The link to resend an invitation for on officer on the manage invitation page was showing up in the case that the user was the Mission Owner and that officer. Now, the link will only show when the user is the MO and _not_ the officer in question (and also the invitation has not yet been accepted).

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164702989

### DIFF
--- a/tests/routes/portfolios/test_task_orders.py
+++ b/tests/routes/portfolios/test_task_orders.py
@@ -289,6 +289,159 @@ class TestTaskOrderInvitations:
         assert response.status_code == 404
         assert time_updated == other_task_order.time_updated
 
+    def test_does_not_render_resend_invite_if_user_is_mo_and_user_is_ko(
+        self, client, user_session
+    ):
+        task_order = TaskOrderFactory.create(
+            portfolio=self.portfolio,
+            creator=self.portfolio.owner,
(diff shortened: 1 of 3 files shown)


tests/routes/portfolios/test_task_orders.py
```diff
@@ -289,6 +289,159 @@ def test_user_can_only_invite_to_task_order_in_their_portfolio(
...
+
+        user_session(self.portfolio.owner)
+        response = client.get(
+            url_for(
+                "portfolios.task_order_invitations",
+                portfolio_id=self.portfolio.id,
+                task_order_id=task_order.id,
+            )
+        )
+        assert "Resend Invitation" in response.data.decode()
```
dandds: You will almost never hear me say there are too many tests, but this might be too many tests. Since we know that the section for each officer is templated and we're checking the user against the officer type every time, I think it's sufficient to just have two tests to check a basic example (like MO is/is not COR). If we had tests that checked our templates in isolation from the routes, I might feel differently.

montana-mil: 👍

PR #742, opened 2019-04-03
Disable save buttons

leigh-mil:
## Description
The Save buttons on forms is disabled until the user makes a change in the form. 
There is a new macro `SaveButton` that is used in conjunction with a new vue component `base-form`, which can be used to wrap any form. If a form requires its own vue component, you will need to add the `FormMixin` to the component to make `SaveButton` work.

## Pivotal
[https://www.pivotaltracker.com/story/show/164929862](https://www.pivotaltracker.com/story/show/164929862)

## Screenshots
Screen Shot 2019-04-03 at 10 05 54 AM
Screen Shot 2019-04-03 at 10 06 01 AM
Screen Shot 2019-04-03 at 10 06 09 AM


js/components/checkbox_input.js
```diff
@@ -10,6 +10,7 @@ export default {
       this.$root.$emit('field-change', {
         value: e.target.checked,
         name: this.name,
+        parent_uid: this.$parent._uid,
```
dandds: Do you think all of these need a `this.$parent && this.$parent._uid` guard?

montana-mil:
this is really cool! i noticed that if you change something, then change it back to the original (so it's in an unchanged state again), the `Save` button stays active. I feel like that's ok, but I wanted to point it out in case it's something you want to discuss.

js/components/date_selector.js
```diff
@@ -128,6 +133,16 @@ export default {
...
 
+  methods: {
+    _emitChange: function(name, value, valid) {
+      this.$root.$emit('field-change', {
+        value,
+        name,
+        parent_uid: this.$parent && this.$parent._uid,
+      })
+    },
+  },
```
dandds: Could this just be moved to a utility function so that we don't have to rewrite it in every input component?

js/mixins/form.js
```diff
@@ -8,7 +8,23 @@ export default {
...
+        if (event['parent_uid'] === this._uid) {
+          this.disabled = false
+        }
       }
     },
   },
+
+  data: function() {
+    return {
+      disabled: this.disableSave,
```
dandds: `disabled` makes me think the whole form is disabled. I think something like `canSave` or `changed` might be better. (I tend to prefer `changed`, because the form mixin, and the forms in general, don't know anything about their save buttons.)

templates/fragments/admin/portfolio_members.html
```diff
@@ -1,71 +1,69 @@
...
-          {% include "fragments/admin/add_new_portfolio_member.html" %}
-        {% endif %}
-      </div>
-    </div>
+      {% endif %}
 
+        <div class="members-table-footer">
+          <div class="action-group">
+            {% if user_can(permissions.EDIT_PORTFOLIO_USERS) %}
+              {{ SaveButton(text='Save') }}
```
dandds: Did you check that this form still submits correctly? The `form` attribute [can be important](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form), and it's only available for `input` elements. I think your Jinja component might need to be more configurable: a submit button could either be a `button` or an `input`, and should accept any other attrs like `form`.

templates/portfolios/task_orders/invitations.html
```diff
@@ -59,7 +60,7 @@ <h4>{{ ("task_orders.invitations." + officer_type + ".edit_title")  | translate}
             {{ Icon("x") }}
             <span>Cancel</span>
           </a>
-          <input type='submit' class='usa-button usa-button-primary' value='Save Changes' />
+          {{ SaveButton(text='Save Changes') }}
```
dandds: I don't know if it matters whether this is a `button` or an `input`, but can you confirm things still work in every place you're swapping elements?

templates/fragments/admin/portfolio_members.html
```diff
@@ -9,51 +10,59 @@
...
+              {% include "fragments/admin/members_view.html" %}
+            {% endif %}
+          </tbody>
 
-    {% endif %}
+        </table>
+        <div class="members-table-footer">
+          <div class="action-group save">
+            {% if user_can(permissions.EDIT_PORTFOLIO_USERS) %}
+              {{ SaveButton(text="Save", element="input", other_attrs="form=member-perms") }}
```
dandds: I wouldn't pass `other_attrs` this way. It will render the attribute values without quotation marks, which is technically incorrect HTML, and then we have to rely on the browser to interpret it correctly. You could pass it as a dictionary of attributes and values and then render it with a dictionary comprehension.

templates/components/save_button.html
```diff
@@ -0,0 +1,8 @@
+{% macro SaveButton(text, element="button", additional_classes="", other_attrs="") -%}
+  {% set class = "usa-button usa-button-primary" + additional_classes %}
+  <{{ element }} type="submit" class="{{ class }}" tabindex="0" v-bind:disabled="!changed" value="{{ text }}" {{ other_attrs }}>
+    {% if element == 'button'%}
+      {{ text }}
+    {% endif %}
+  </{{ element }}>
```
dandds: I think it would be simpler to just have branch logic to render either `<input>` or `<button>`.

js/components/forms/base_form.js
```diff
@@ -0,0 +1,54 @@
...
+    checkboxinput,
+    levelofwarrant,
+  },
+  mixins: [FormMixin],
+  methods: {
+    closeModal: function(name) {
+      this.activeModal = null
+      this.$root.$emit('modalOpen', false)
+      if (this.$root.allyHandler) this.$root.allyHandler.disengage()
+    },
```
dandds: We should chat about the modal stuff. I don't understand why we need it here, and if we do, there's a modal mixin we should be using instead of inlining the methods.

templates/components/save_button.html
```diff
@@ -0,0 +1,15 @@
...
+    {% set form_attr = 'form="{{ form }}"' %}
+  {% else %}
+    {% set form_attr = "" %}
+  {% endif %}
+  {% if element == "button" %}
+    <button type="submit" class="{{ class }}" tabindex="0" v-bind:disabled="!changed" {{ form_attr }} >
+      {{ text }}
+    </button>
+  {% elif element == 'input' %}
+    <input type="submit" class="{{ class }}" tabindex="0" v-bind:disabled="!changed" value="{{ text }}" {{ form_attr }} />
```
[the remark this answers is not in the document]
leigh-mil: I tried that but then there was a empty form attribute which was causing the form to not submit!

PR #743, opened 2019-04-03
Emit plaintext logging for audit trail

montana-mil:
## Description
Every time the app emits a create, update, or delete event, the information is logged to screen. This PR also adds tests to the `auditable` mixin. 

These are the same as the `audit_events` that are displayed in the app and saved to the database. We are just displaying them as `info` level logging now as well.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/164906483

### DIFF
--- a/atst/models/audit_event.py
+++ b/atst/models/audit_event.py
@@ -25,6 +25,19 @@ class AuditEvent(Base, TimestampsMixin):
     display_name = Column(String())
     action = Column(String(), nullable=False)
 
+    @property
+    def log(self):
+        return {
+            "user_id": str(self.id),
+            "portfolio_id": str(self.portfolio_id),
+            "changed_state": self.changed_state,
+            "event_details": self.event_details,
+            "resource_type": self.resource_type,
+            "resource_id": str(self.resource_id),
+            "display_name": self.display_name,
+            "action": self.action,
+        }
+
     def save(self, connection):
         attrs = inspect(self).dict
 
--- a/atst/models/audit_event.py
+++ b/atst/models/audit_event.py
@@ -28,7 +28,6 @@ class AuditEvent(Base, TimestampsMixin):
     @property
     def log(self):
         return {
-            "user_id": str(self.id),
             "portfolio_id": str(self.portfolio_id),
             "changed_state": self.changed_state,
             "event_details": self.event_details,
--- a/atst/models/mixins/auditable.py
+++ b/atst/models/mixins/auditable.py
@@ -1,5 +1,5 @@
 from sqlalchemy import event, inspect
-from flask import g
+from flask import g, current_app as app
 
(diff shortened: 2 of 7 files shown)


tests/domain/test_authz.py
```diff
@@ -193,9 +183,9 @@ def _do_something(*args, **kwargs):
         "atst.domain.authz.decorator.check_access", lambda *a, **k: True
     )
     _do_something()
-    assert len(mock_logger.messages) == 1
-    assert "accessed" in mock_logger.messages[0]
-    assert "GET" in mock_logger.messages[0]
+    assert len(mock_logger.messages) == 2
```
dandds: Little thing, but usually in these situations I'd record the length of the messages list in a variable before the action I expect to log (`_do_something()` in this case) and then assert that the length of the log is equal to that plus one after the action. Similarly, I'd use `-1` as the list indexes below to get the last log message. That was the test is more flexible is a bunch of things get logged before we call `_do_something`.

atst/models/mixins/auditable.py
```diff
@@ -32,6 +32,7 @@ def create_audit_event(connection, resource, action):
         )
 
         audit_event.save(connection)
+        return audit_event
```
dandds: I'm wondering if we should emit the log before we save the event to the database. It seems like the log is actually more important than the database entry (one is for security auditors, the other is just for users to know the history of their portfolio) and I wouldn't want it to not get logged if there are database transaction probkems. Could you try:

```
        app.logger.info(
            "Audit Event {}".format(action.lower()),
            extra={"audit_event": event.log, "tags": ["audit_event", "insert"]},
        )
        audit_event.save(connection)
```

Then we don't have to do it in each audit method.

atst/models/audit_event.py
```diff
@@ -25,6 +25,19 @@ class AuditEvent(Base, TimestampsMixin):
     display_name = Column(String())
     action = Column(String(), nullable=False)
 
+    @property
+    def log(self):
+        return {
+            "user_id": str(self.id),
```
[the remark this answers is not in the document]
montana-mil: oh jeez that was a typo. but if we don't need it, i'll take it out!

atst/models/mixins/auditable.py
montana-mil: that's a really good point. Is there anything we should do to indicate that the `audit_event` might not have succeeded?

montana-mil: it looks like the action on an `after_insert` event is called `create`. I ended up renaming the tests, too.

dandds: > Is there anything we should do to indicate that the audit_event might not have succeeded?

Good q. I tend to think not, because the db audit log table and the app logger don't need to know anything about each other.

PR #744, opened 2019-04-04
record portfolio_role permission_set changes with bulk_replace event

dandds:
This fixes the problem outlined in this PT card: https://www.pivotaltracker.com/story/show/164648971

We were not recording permission set changes for the portfolio role in the audit log because changes to the join table did not trigger the `after_insert` or `after_update` events we use to track all other changes that appear in the audit log. Instead, I've added a separate event listener for [`bulk_replace` ORM events](https://docs.sqlalchemy.org/en/latest/orm/events.html#sqlalchemy.orm.events.AttributeEvents.bulk_replace) on the `permission_set` attribute. I've set the decorator to provide the [`raw` instance state](https://docs.sqlalchemy.org/en/latest/orm/events.html#sqlalchemy.orm.events.AttributeEvents.params.raw) so that we can get the previous permission set and build a correct `change_set` for the audit event.

The drawback to this method is that it doesn't correspond to a database write the way the other event types we use do. After a lot of research, I couldn't find an appropriate event tied directly to a database write that would let us track permission sets changes. It would only be in very unusual circumstances that this hook creates an incorrect audit event, though (the actual perm sets update is rolled back but the audit event insertion succeeds).

### DIFF
--- a/atst/models/mixins/auditable.py
+++ b/atst/models/mixins/auditable.py
@@ -11,14 +11,15 @@ ACTION_DELETE = "delete"
 
 class AuditableMixin(object):
     @staticmethod
-    def create_audit_event(connection, resource, action):
+    def create_audit_event(connection, resource, action, changed_state=None):
         user_id = getattr_path(g, "current_user.id")
         portfolio_id = resource.portfolio_id
         resource_type = resource.resource_type
         display_name = resource.displayname
         event_details = resource.event_details
 
-        changed_state = resource.history if action == ACTION_UPDATE else None
+        if changed_state is None:
+            changed_state = resource.history if action == ACTION_UPDATE else None
 
         audit_event = AuditEvent(
             user_id=user_id,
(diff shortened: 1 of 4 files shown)


leigh-mil:
I noticed two bugs when updating perms in the form: 
1) The first time you update member perms for a portfolio, and audit event is created for each user whether or not you actually updated their perms
2) When you save the form it creates an audit event for the user you changed perms for and every member that is listed beneath them in the table

dandds:
@leigh-mil I think you're seeing this happen because the `seed_sample` script was creating portfolio users without any permissions. When you submit that perms form the first time, it's giving them the default view perms they didn't have yet. I updated the `seed_sample` script, so if you rotate your sample data that shouldn't happen anymore.

>     every member that is listed beneath them in the table
I haven't been able to reproduce this. Can you see if it still happens for you after you update the seed data?

There is one other unrelated bug I noticed with that table. It currently strips the PPoC of their special PPoC perms, because the PPoC perms aren't included in the list of perms that are submitted by that table. That should be taken care of on [this](https://www.pivotaltracker.com/story/show/164446865) PT card.

PR #745, opened 2019-04-04
Delete user update

rachel-dtr:
## Description
Updating language for the newly implemented deleting member from portfolio team

##[Pivotal Card](https://www.pivotaltracker.com/story/show/165103806)


templates/fragments/admin/portfolio_members.html
```diff
@@ -59,12 +59,12 @@
       {% for member in portfolio.members %}
         {% set modal_id = "portfolio_id_{}_user_id_{}".format(portfolio.id, member.user_id) %}
         {% call Modal(name=modal_id, dismissable=False) %}
-          <h1>Are you sure you want to archive this user?</h1>
+          <h1>Are you sure you want to delete this member?</h1>
```
montana-mil: It might be a good idea to move the copy from this file into the `translations.yaml` file. Then you can do this type of thing: https://github.com/dod-ccpo/atst/blob/master/templates/fragments/edit_user_form.html#L37

If that's something you're comfortable with! Or if you wanted to pair on that, I could help tomorrow :)

PR #746, opened 2019-04-04
Application roles

dandds:
This PR does a few things related to application roles and general authorization:

- I added an application roles table to the database. I've only included the bare minimum in the `ApplicationRole` model. Eventually, it should probably share a mixin with `PortfolioRole` model that includes invitation methods, etc.
- I added application permission sets and a couple missing environment permissions. This part might need some iteration as we do app settings development, but should hopefully be enough for a start.
- I added application checks to the `authz` module.
- I updated and reorganized some of the `authz` module code. I noticed that `check_portfolio_permission` was checking both the user's portfolio and atat permissions, but `has_portfolio_permission` wasn't. This would have introduced bugs in the UI, because our `user_can` checks in the template rely on `has_portfolio_permission`. I switched around the responsibilities of those methods so that `has_portfolio_permission` checks the user's portfolio and atat permissions.


tests/domain/test_authz.py
```diff
@@ -144,6 +177,36 @@ def _edit_portfolio_name(*args, **kwargs):
...
+    rando = UserFactory.create()
+
+    portfolio = PortfolioFactory.create(
+        owner=port_admin, applications=[{"name": "Mos Eisley"}]
+    )
+    app = portfolio.applications[0]
+    ApplicationRoleFactory.create(application=app, user=app_user)
+
+    @user_can_access_decorator(Permissions.VIEW_APPLICATION)
+    def _stroll_into_mos_eisley(*args, **kwargs):
```
leigh-mil: 😆 Great function name!

PR #747, opened 2019-04-08
Disable edit ppoc

leigh-mil:
## Description
When viewing the Portfolio Members table, permissions for the Point of Contact are read-only, for all users.

## Pivotal
[https://www.pivotaltracker.com/story/show/164446865](https://www.pivotaltracker.com/story/show/164446865)

Also solves issue in this bug: [https://www.pivotaltracker.com/story/show/165184807](https://www.pivotaltracker.com/story/show/165184807)

## Screenshot
Screen Shot 2019-04-08 at 3 56 54 PM

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -295,6 +295,37 @@
       background: $color-red;
     }
 
+    select {
+      padding-left: 1.2rem
+    }
+
+    .members-table-ppoc {
+      select::-ms-expand {
+        display: none;
+      }
+
+      select {
+        -webkit-appearance: none;
+        -moz-appearance: none;
+        appearance: none;
+        display: block;
+        width: 100%;
+        float: right;
+        margin: 5px 0px;
+        padding: 0px 24px;
+        background-image: none;
+        -ms-word-break: normal;
+        word-break: normal;
+        padding-right: 3rem;
+        padding-left: 1.2rem;
+      }
+
+      select:hover {
+        box-shadow: none;
+        color: $color-base;
+      }
+    }
+
     .members-table-footer {
       float: right;
       padding: 3 * $gap 0;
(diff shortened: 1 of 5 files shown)


templates/fragments/admin/members_edit.html
```diff
@@ -2,8 +2,9 @@
 
 {% for subform in member_perms_form.members_permissions %}
   {% set modal_id = "portfolio_id_{}_user_id_{}".format(portfolio.id, subform.user_id.data) %}
+  {% set ppoc = subform.member.data == portfolio.owner.full_name %}
```
dandds: I would check the IDs here. There might be two John Smiths associated with a portfolio eventually.

tests/routes/portfolios/test_admin.py
```diff
@@ -130,3 +131,32 @@ def test_rerender_admin_page_if_member_perms_form_does_not_validate(
     )
     assert response.status_code == 200
     assert "Portfolio Administration" in response.data.decode()
+
+
+def test_cannot_update_portfolio_owner_perms(client, user_session):
```
dandds: We should stick with "ppoc" instead of "owner" for test names, at least. Eventually we should move all the language from one to the other.

styles/components/_portfolio_layout.scss
```diff
@@ -295,6 +295,32 @@
...
+        width: 100%;
+        float: right;
+        margin: 5px 0px;
+        padding: 0px 24px;
+        background-image: none;
+        -ms-word-break: normal;
+        word-break: normal;
+        padding-right: 3rem;
+        padding-left: 1.2rem;
+      }
```
dandds: The border hover state is a little weird/unexpected for the PPoC perms, I think. Not a big deal and I'd approve this with that unchanged.

PR #748, opened 2019-04-08
Display app users

drummond-dds:
## Ticket information

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/165059837)

```cucumber
Scenario: 
Given I am logged into AT-AT as a user with "view only" access to an application (set at the portfolio level),
When I click on Team Settings,
Then I see the Team Settings table per the design,
And the count of users is up at the top as "Members,"
And each user is listed that has been added to the app,
And the "Environment Access" contains the number of environments to which that user has access of any kind,
And their Environment & Team Mgmt & delete access permissions are displayed,
And Delete is a yes/no rather than the "view/edit" in the design,
And I can click the carat/plus sign to expand the list of environments per the design,
But there are no dropdowns to change permissions,
And there are no buttons to edit or add or save anything.
```

### Table open 

Screen Shot 2019-04-17 at 09 38 47

### Table closed

Screen Shot 2019-04-17 at 09 38 35

### DIFF
--- a/atst/routes/portfolios/applications.py
+++ b/atst/routes/portfolios/applications.py
@@ -9,6 +9,7 @@ from flask import (
 
 from . import portfolios_bp
 from atst.domain.environment_roles import EnvironmentRoles
+from atst.domain.environments import Environments
 from atst.domain.exceptions import UnauthorizedError
 from atst.domain.applications import Applications
 from atst.domain.portfolios import Portfolios
@@ -16,6 +17,8 @@ from atst.forms.application import NewApplicationForm, ApplicationForm
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.models.permissions import Permissions
 from atst.utils.flash import formatted_flash as flash
+from atst.domain.permission_sets import PermissionSets
+from atst.utils.localization import translate
 
 
 @portfolios_bp.route("/portfolios/<portfolio_id>/applications")
@@ -138,4 +141,45 @@ def delete_application(portfolio_id, application_id):
 
     return redirect(
         url_for("portfolios.portfolio_applications", portfolio_id=portfolio_id)
+    )
+
+
+def permission_str(member, edit_perm_set):
+    if member.has_permission_set(edit_perm_set):
+        return translate("portfolios.members.permissions.edit_access")
+    else:
+        return translate("portfolios.members.permissions.view_only")
+
+
+@portfolios_bp.route("/portfolios/<portfolio_id>/applications/<application_id>/team")
+@user_can(Permissions.VIEW_APPLICATION, message="view portfolio applications")
+def application_team(portfolio_id, application_id):
+    application = Applications.get(application_id)
+    portfolio = Portfolios.get(g.current_user, portfolio_id)
+
+    environment_users = {}
+    for member in application.members:
+        user_id = member.user.id
+        environment_users[user_id] = {
+            "permissions": {
+                "delete_access": permission_str(
+                    member, PermissionSets.DELETE_APPLICATION_ENVIRONMENTS
+                ),
+                "environment_management": permission_str(
+                    member, PermissionSets.EDIT_APPLICATION_ENVIRONMENTS
+                ),
+                "team_management": permission_str(
+                    member, PermissionSets.EDIT_APPLICATION_TEAM
+                ),
+            },
+            "environments": Environments.for_user(
+                user=member.user, application=application
+            ),
+        }
+
+    return render_template(
+        "portfolios/applications/team.html",
+        application=application,
+        portfolio=portfolio,
+        environment_users=environment_users,
     )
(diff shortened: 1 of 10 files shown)


atst/routes/portfolios/applications.py
```diff
@@ -140,3 +143,44 @@ def delete_application(portfolio_id, application_id):
...
+    if member.has_permission_set(edit_perm_set):
+        return translate("portfolios.members.permissions.edit_access")
+    else:
+        return translate("portfolios.members.permissions.view_only")
+
+
+@portfolios_bp.route("/portfolios/<portfolio_id>/applications/<application_id>/team")
+@user_can(Permissions.VIEW_APPLICATION, message="view portfolio applications")
+def application_team(portfolio_id, application_id):
+    application = Applications.get(application_id)
```
dandds: Can you rebase and update this to pass the `portfolio_id` kwarg for scoping?

atst/routes/portfolios/applications.py
```diff
@@ -140,3 +143,44 @@ def delete_application(portfolio_id, application_id):
...
+    portfolio = Portfolios.get(g.current_user, portfolio_id)
```
dandds: I don't think you need to get the portfolio here or pass it to the `render_template` function below.

templates/portfolios/applications/team.html
```diff
@@ -0,0 +1,104 @@
...
+    ) }}
+
+  {% else %}
+    <div class='subheading'>
+      {{ 'portfolios.applications.team_settings.subheading' | translate }}
+    </div>
+
+    <section class="member-list" id="portfolio-members">
+      <div class='responsive-table-wrapper panel'>
+        {% if g.matchesPath("portfolio-members") %}
```
dandds: You might need to update some of the IDs in this template.

styles/components/_accordion_table.scss
```diff
@@ -110,3 +110,28 @@
     }
   }
 }
+
+#portfolio-members {
```
dandds: I think you might want to change this ID to `application-members` or, better, make a more generic class since there will probably be similar lists/tables. Something in here is also messing with rendering in the portfolio members table:

Screen Shot 2019-04-17 at 11 15 55 AM

templates/portfolios/applications/team.html
```diff
@@ -0,0 +1,104 @@
...
+                <span>
+                  {{ "portfolios.applications.team_settings.section.table.delete_access" | translate }}
+                </span>
+                <span>
+                  {{ "portfolios.applications.team_settings.section.table.environment_management" | translate }}
+                </span>
+                <span>
+                  {{ "portfolios.applications.team_settings.section.table.team_management" | translate }}
+                </span>
+              </span>
```
dandds: Ah, now this one actually seems like a table since we have real column data. 🙃 

The problem is that when you expand and show the list of environments... that's not longer table data. You could make it a nested list inside of the `td` for the toggle link, but I doubt you could resolve the CSS easily to show the table row and the list as separate elements (since the list is a grandchild of the row). I dunno, maybe play with it for 15 minutes or so and see if it's possible.

tests/routes/portfolios/test_applications.py
```diff
@@ -329,3 +329,38 @@ def test_delete_application(client, user_session):
     # app and envs are soft deleted
     assert len(port.applications) == 0
     assert len(application.environments) == 0
+
+
+def test_application_team_with_permissions(client, user_session):
```
dandds: I think simple access test like this should live in `tests/test_access` and follow the test naming convention there: `test_portfolio_application_team_access`, in this case.

atst/routes/portfolios/applications.py
drummond-dds: Added in e814f890

tests/routes/portfolios/test_applications.py
drummond-dds: Added in 52669a02

atst/routes/portfolios/applications.py
drummond-dds: Removed in 20149268

templates/portfolios/applications/team.html
drummond-dds: Renamed in e1cca580

styles/components/_accordion_table.scss
drummond-dds: I changed the id scoping. I think I will tackle the CSS thing when I make the table more generic to allow multiple drop downs. I think there is going to be a fair amount of restructuring and it makes sense to do that in one go.

templates/portfolios/applications/team.html
drummond-dds: I think this will be tackled in the "Multiple drop downs per table row" problem I need to solve for my other table view.

PR #750, opened 2019-04-09
Emit event from root when modal closes

leigh-mil:
## Description
Closing a modal was freezing the site. Updating the event when a modal closes to emit from the root component fixed the issue. 

## Pivotal
[https://www.pivotaltracker.com/story/show/165214427](https://www.pivotaltracker.com/story/show/165214427)

### DIFF
--- a/js/mixins/modal.js
+++ b/js/mixins/modal.js
@@ -4,7 +4,7 @@ export default {
   methods: {
     closeModal: function(name) {
       this.activeModal = null
-      this.$emit('modalOpen', false)
+      this.$root.$emit('modalOpen', { isOpen: false, name: name })
       if (this.allyHandler) this.allyHandler.disengage()
     },
 


js/mixins/modal.js
```diff
@@ -4,7 +4,7 @@ export default {
   methods: {
     closeModal: function(name) {
       this.activeModal = null
-      this.$emit('modalOpen', false)
+      this.$root.$emit('modalOpen', { isOpen: false })
```
dandds: Should we emit the name in this object, too?

leigh-mil: I did that because in index.js, on the close event the name of the `activeModal` is not being updated. Should we update it there?
[https://github.com/dod-ccpo/atst/blob/5236c94901d8371d60658eb6f42314080353cd7d/js/index.js#L76-L86](https://github.com/dod-ccpo/atst/blob/5236c94901d8371d60658eb6f42314080353cd7d/js/index.js#L76-L86)

dcorrigan: I don't think we have to update it there. The point of emitting events is that any other part of the code can listen for them, and the shape of the event we emit should be the same for open and close. Whether or not that info gets used to up to the listener. We're also not using the name argument in `closeModal` at all right now, and we probably should be.

PR #754, opened 2019-04-10
Delete applications

dandds:
PT story: https://www.pivotaltracker.com/n/projects/2160940/stories/165067652

This adds the functionality for "deleting" an application. That deletion cascades to environments and all the associated user roles at the app and env level. When an environment is deleted, that information is pushed to the CSP via the `csp.cloud` module.

**notes**
- This PR restores the USWDS default alerts. The designs are leaning on those default styles, so I changed our Jinja component for alerts to use them too.
- I added soft deletes for the following tables: `applications`, `application_roles`, `environments`, `environment_roles`. The `deleted` property is available via a mixin. A model with a relationship to a deletable resource should define `primaryjoin` so that it filters out deleted resources. Domain methods that `get` resources should filter out deleted resources. Deleted resources should only be available through direct queries, and we shouldn't be using those anywhere.

Screen Shot 2019-04-10 at 5 11 08 PM
Screen Shot 2019-04-10 at 5 11 28 PM
Screen Shot 2019-04-10 at 5 11 39 PM


drummond-dds:
We discussed handling CSP deletion errors and made up a follow up chore for when the CSP API behavior is known https://www.pivotaltracker.com/story/show/165315237

deploy/kubernetes/test/atst-envvars-configmap.yml
```diff
@@ -12,3 +12,4 @@ data:
   RQ_QUEUES: atat-test
   CRL_STORAGE_PROVIDER: CLOUDFILES
   LOG_JSON: "true"
+  DEBUG: "false"
```
drummond-dds: What is this changing?

templates/portfolios/applications/edit.html
```diff
@@ -12,7 +14,38 @@
...
+        <div class="form-row">
+          <div class="form-col form-col--half">
+            {{ TextInput(form.name) }}
+          </div>
+          <div class="form-col form-col--half">
+            {% if user_can(permissions.DELETE_APPLICATION) %}
+            <div class="usa-input">
+              <label for="delete-application">
+                <div class="usa-input__title">
+                  &nbsp;
```
drummond-dds: Whats the empty label for?

dandds: the empty label is just there to push the button to the same horizontal height as the input field. we could do it with CSS, but we'd have to hard-code the input field's height to match it.

deploy/kubernetes/test/atst-envvars-configmap.yml
dandds: That turns off the Flask auto-reload. I meant to make a separate PR for it, but it's applied to the cluster already.

PR #761, opened 2019-04-15
Environment lists

drummond-dds:
Paired with @dandds to extract out some styles for the accordion table from @montana-mil's PR and make them generic so they can be reused across the app.

### DIFF
--- a/atst/routes/portfolios/applications.py
+++ b/atst/routes/portfolios/applications.py
@@ -57,7 +57,7 @@ def create_application(portfolio_id):
         )
 
 
-def environments_obj(application):
+def get_environments_obj_for_app(application):
     environments_obj = {}
 
     for env in application.environments:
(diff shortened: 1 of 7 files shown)


atst/routes/portfolios/applications.py
```diff
@@ -57,23 +57,28 @@ def create_application(portfolio_id):
         )
 
 
+def environments_obj(application):
```
dandds: This is a little Ruby-ish. I think a more Pythonic name would be something like `environment_data_for_app`.

montana-mil: what makes it ruby/what makes this name pythonic?

dandds: Ruby methods tend to have names that are just nouns, like `repo.pull_requests`. Python doesn't put quite as much emphasis on OO and, importantly, we're not in a class context here. Python functions tend to have names that describe what they do, so even `get_environment_obj_for_app` might be better.

montana-mil: oh i see, i didnt even realize it was a function. the new name definitely helps!

PR #762, opened 2019-04-15
Updating terms and capitalization

rachel-dtr:
## Description

This is the continuation of the effort to update capitalization across the site to be consistent, as well as updating some terms to be correct/less redundant.

### DIFF
--- a/atst/forms/portfolio_member.py
+++ b/atst/forms/portfolio_member.py
@@ -13,28 +13,28 @@ class PermissionsForm(BaseForm):
     member = StringField()
     user_id = HiddenField()
     perms_app_mgmt = SelectField(
-        None,
+        "App management",
         choices=[
             (PermissionSets.VIEW_PORTFOLIO_APPLICATION_MANAGEMENT, "View only"),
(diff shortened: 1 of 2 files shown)


dandds:
Ah, actually this needs a rebase @rachel-dtr . There's a merge conflict in the translations file. I or someone else can help tomorrow if you need it.

montana-mil:
i fixed the conflict :)

atst/forms/portfolio_member.py
```diff
@@ -13,28 +13,28 @@ class PermissionsForm(BaseForm):
     member = StringField()
     user_id = HiddenField()
     perms_app_mgmt = SelectField(
-        None,
+        "App management",
```
drummond-dds: Depending on how aggressively we want to use `translations.yml` we can also use them in forms.

See https://github.com/dod-ccpo/atst/blob/master/atst/forms/ccpo_review.py#L13

montana-mil: that's a good call

rachel-dtr: ok I think I did that...

PR #771, opened 2019-04-17
PPoC in members table

leigh-mil:
## Description
Add the PPoC label under the PPoC and display them first in the members table. Added some minor design tweaks:
 - disabled fields for PPoC perms should be grey
- moved you and PPoC labels under the member's name
Also fixed issue in the view only version of the members table that had some text hard coded in the template which was causing the perms to not render correctly. 

## Pivotal
[https://www.pivotaltracker.com/story/show/165227381](https://www.pivotaltracker.com/story/show/165227381)

## Screenshots
Screen Shot 2019-04-17 at 11 36 58 AM
Screen Shot 2019-04-17 at 11 37 13 AM

### DIFF
--- a/templates/fragments/admin/members_edit.html
+++ b/templates/fragments/admin/members_edit.html
@@ -3,17 +3,20 @@
 {% for subform in member_perms_form.members_permissions %}
   {% set modal_id = "portfolio_id_{}_user_id_{}".format(portfolio.id, subform.user_id.data) %}
   {% set ppoc = subform.user_id.data == portfolio.owner.id %}
+  {% set archive_button_class = 'button-danger-outline' %}
 
   <tr {% if ppoc %}class="members-table-ppoc"{% endif %}>
     <td class='name'>{{ subform.member.data }}
-      {% if subform.user_id.data == user.id %}
-        <span class='you'>(<span class='green'>you</span>)</span>
-        {% set archive_button_class = 'usa-button-disabled' %}
-      {% elif ppoc %}
-        {% set archive_button_class = 'usa-button-disabled' %}
-      {% else %}
-        {% set archive_button_class = 'button-danger-outline' %}
-      {% endif %}
+      <div>
+        {% if ppoc %}
+          {% set archive_button_class = 'usa-button-disabled' %}
+          <span class='you'>PPoC</span>
+        {% endif %}
+        {% if subform.user_id.data == user.id %}
+          {% set archive_button_class = 'usa-button-disabled' %}
+          <span class='you'>(<span class='green'>you</span>)</span>
+        {% endif %}
+      </div>
     </td>
 
     <td>{{ OptionsInput(subform.perms_app_mgmt, label=False, disabled=ppoc) }}</td>
--- a/templates/fragments/admin/members_view.html
+++ b/templates/fragments/admin/members_view.html
@@ -1,18 +1,24 @@
 {% for subform in member_perms_form.members_permissions %}
+  {% set ppoc = subform.user_id.data == portfolio.owner.id %}
+  {% set heading_perms = [subform.perms_app_mgmt, subform.perms_funding, subform.perms_reporting, subform.perms_portfolio_mgmt] %}
 
   <tr>
     <td class='name'>{{ subform.member.data }}
-      {% if subform.member.data == user.full_name %}
-        <span class='you'>(<span class='green'>you</span>)</span>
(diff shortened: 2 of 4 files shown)


templates/fragments/admin/members_edit.html
```diff
@@ -1,15 +1,25 @@
 {% from "components/confirmation_button.html" import ConfirmationButton %}
 
-{% for subform in member_perms_form.members_permissions %}
-  {% set modal_id = "portfolio_id_{}_user_id_{}".format(portfolio.id, subform.user_id.data) %}
-  {% set ppoc = subform.user_id.data == portfolio.owner.id %}
+{% set ppoc_subform = [] %}
+
+{% for member in member_perms_form.members_permissions %}
+  {% if member.user_id.data == portfolio.owner.id %}
+    {% if ppoc_subform.append(member) %}{% endif %}
```
dandds: I have a few questions here:
- Does Jinja not allow regular statement like `{% ppoc_subform.append(member) %}` ?
- Does `ppoc_subform` need to be a list? There's only one PPoC.
- This might make the other points moot, but this seems like enough logic that it should live in the route function, i.e., we should be doing this ordering when we construct the `member_perms_form` object. It is technically a rendering concern, but practically I think the logic would live more comfortably in regular Python code.

templates/fragments/admin/members_edit.html
```diff
@@ -1,15 +1,25 @@
...
+  <tr {% if is_ppoc %}class="members-table-ppoc"{% endif %}>
     <td class='name'>{{ subform.member.data }}
-      {% if subform.user_id.data == user.id %}
-        <span class='you'>(<span class='green'>you</span>)</span>
+      {% if is_ppoc %}
         {% set archive_button_class = 'usa-button-disabled' %}
-      {% elif ppoc %}
+        <br>
+        <span class='you'>PPoC{% if subform.user_id.data == user.id %} (<span class='green'>you</span>){% endif %}</span>
+        </span>
```
dandds: I think this is a stray closing tag.

templates/fragments/admin/members_edit.html
```diff
@@ -1,15 +1,25 @@
...
+      {% elif subform.user_id.data == user.id %}
+        <br>
+        <span class='you'>(<span class='green'>you</span>)</span>
```
dandds: I'm finding this a little hard to follow. I think that it would be easier to read if "PPoC" and "you" were in self-contained spans. You could do something like this:

```
  {% set archive_button_class = 'button-danger-outline' %}
  <tr {% if is_ppoc %}class="members-table-ppoc"{% endif %}>
    <td class='name'>{{ subform.member.data }}
      <p style="margin: 0">
        {% if is_ppoc %}
          {% set archive_button_class = 'usa-button-disabled' %}
          <span class='you'>PPoC</span>
        {% endif %}
        {% if subform.user_id.data == user.id %}
          {% set archive_button_class = 'usa-button-disabled' %}
          <span class='you'>(<span class='green'>you</span>)</span>
        {% endif %}
      </p>
    </td>
```
Don't use inline styles like I did here and probably come up with a more generic class name than "you", but you get the idea.

templates/fragments/admin/members_view.html
```diff
@@ -1,20 +1,41 @@
-{% for subform in member_perms_form.members_permissions %}
+{% set ppoc_subform = [] %}
+
+{% for member in member_perms_form.members_permissions %}
+  {% if member.user_id.data == portfolio.owner.id %}
+    {% if ppoc_subform.append(member) %}{% endif %}
+  {% endif %}
+{% endfor %}
```
dandds: Same questions/comments as above. Another advantage to sorting the form's list of members in the route function is that you wouldn't have to repeat this.

templates/fragments/admin/members_view.html
```diff
@@ -1,20 +1,41 @@
...
+{% macro DisplayMember(subform, is_ppoc=False) %}
   <tr>
     <td class='name'>{{ subform.member.data }}
-      {% if subform.member.data == user.full_name %}
+      {% if is_ppoc %}
+        <br>
+        <span class='you'>PPoC{% if subform.user_id.data == user.id %} (<span class='green'>you</span>){% endif %}</span>
+        </span>
+      {% elif subform.user_id.data == user.id %}
+        <br>
```
dandds: Same comments as above.

PR #777, opened 2019-04-18
Form for editing app env settings

leigh-mil:
## Description
Creates the form for editing the app's environment members' roles via the App Settings page. As per the new design, you can only update the team for one environment at a time. 
This also updated some code left over from the way env roles were previously assigned, including refactoring a bunch of tests!

## Pivotal
[https://www.pivotaltracker.com/story/show/165156617](https://www.pivotaltracker.com/story/show/165156617)


atst/routes/applications/settings.py
```diff
@@ -23,17 +25,39 @@ def get_environments_obj_for_app(application):
...
+
+
 @applications_bp.route("/applications/<application_id>/settings")
 @user_can(Permissions.VIEW_APPLICATION, message="view application edit form")
 def settings(application_id):
     application = Applications.get(application_id)
     form = ApplicationForm(name=application.name, description=application.description)
+    env_data = serialize_env_member_form_data(application)
+    env_forms = {}
+    for data in env_data:
```
dandds: Nitpicky, but something more specific than `data` would be good.

atst/routes/applications/settings.py
```diff
@@ -53,11 +77,43 @@ def update(application_id):
...
+@user_can(Permissions.ASSIGN_ENVIRONMENT_MEMBER, message="update application")
+def update_env_roles(application_id):
+    application = Applications.get(application_id)
+    env_roles_form = EnvironmentRolesForm(http_request.form)
+
+    if env_roles_form.validate():
+        env_data = env_roles_form.data
+        Environments.update_env_roles_by_environment(
+            environment_id=env_data["env_id"], team_roles=env_data["team_roles"]
+        )
```
dandds: There should probably be a success flash message here, I think.

atst/routes/applications/settings.py
```diff
@@ -53,11 +77,43 @@ def update(application_id):
...
+    else:
+        return render_template(
+            "portfolios/applications/edit.html",
+            application=application,
+            form=ApplicationForm(
+                name=application.name, description=application.description
+            ),
+            environments_obj=get_environments_obj_for_app(application=application),
+            env_forms=env_roles_form,
         )
```
dandds: I find spots like this challenging, because if the user is playing by the rules (submitting data using the regular web page, instead of sending random data via curl or something) they should never have form validation errors. Can you at least add a TODO for this branch? There are a few spots like this and we think we need a pattern to handle them.

atst/routes/applications/settings.py
```diff
@@ -53,11 +77,43 @@ def update(application_id):
             )
         )
     else:
+        env_data = serialize_env_member_form_data(application)
+        env_forms = {}
+        for data in env_data:
+            env_forms[data["env_id"]] = EnvironmentRolesForm(data=data)
+
         return render_template(
             "portfolios/applications/edit.html",
```
dandds: Can we update this template name to `settings.html`? It's getting confusing.

atst/routes/applications/settings.py
```diff
@@ -53,11 +77,43 @@ def update(application_id):
...
             application=application,
             form=form,
             environments_obj=get_environments_obj_for_app(application=application),
+            env_forms=env_forms,
+        )
```
dandds: I wonder if we could use a helper function like the [`render_admin_page`](https://github.com/dod-ccpo/atst/blob/master/atst/routes/portfolios/admin.py#L63) we have for portfolio routes. This template gets rendered three times in this file.

tests/domain/test_environments.py
```diff
@@ -1,9 +1,11 @@
 import pytest
+import random
```
dandds: You're not using random in this test file.

tests/domain/test_environments.py
```diff
@@ -1,9 +1,11 @@
...
 
 from atst.domain.environments import Environments
 from atst.domain.environment_roles import EnvironmentRoles
 from atst.domain.portfolio_roles import PortfolioRoles
```
dandds: Not using `PortfolioRoles` any more, either.

atst/routes/applications/settings.py
```diff
@@ -53,11 +77,43 @@ def update(application_id):
...
+
+
+@applications_bp.route(
+    "/applications/<application_id>/update_env_roles", methods=["POST"]
```
dandds: This route function only ever updates one environment at a time, right? If that's the case, I think that this should be an environment-specific route: `/environments/<environment_id>/roles`. You could still put the environment ID as a field in the form, but you would use it to build the URL for the individual form `action` in the HTML instead of having it be a hidden input.

tests/routes/applications/test_settings.py
```diff
@@ -120,6 +166,89 @@ def test_user_without_permission_cannot_update_application(client, user_session)
...
+        follow_redirects=True,
+    )
+
+    assert response.status_code == 404
+    assert env_role_1.role == CSPRole.BASIC_ACCESS.value
+    assert (
+        EnvironmentRoles.get(env_role_2.user.id, environment.id).role
+        == CSPRole.BASIC_ACCESS.value
+    )
+    assert not EnvironmentRoles.get(app_role.user.id, environment.id)
```
dandds: My preference is to have basic access tests live in `tests/test_access.py`. Those tests mock out pretty much everything the route function is supposed to do and _only_ care about whether a particular user can hit the route successfully. Your first test here is fine, but maybe change the test name; it's testing that the route does what it's supposed to. Your second should probably be removed in favor of something simpler over in that `test_access.py` file.

atst/domain/environments.py
```diff
@@ -62,39 +63,51 @@ def get(cls, environment_id):
...
+
+        for member in team_roles:
+            new_role = member["role"]
+            user = Users.get(member["user_id"])
+            Environments.update_env_role(
+                environment=environment, user=user, new_role=new_role
+            )
+
+    @classmethod
+    def update_env_roles_by_member(cls, member, env_roles):
```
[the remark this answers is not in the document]
dandds: I'm not telling you to delete it, btw; just asking.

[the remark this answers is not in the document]
leigh-mil: It is not used in the app, but @montana-mil pointed out that it will be used when we edit the app members via the team page, so I left it in and updated the name to make its intended behavior more clear!

atst/routes/applications/settings.py
leigh-mil: Yes! I was intending to refactor that in my next PR for this story!

atst/domain/environments.py
```diff
@@ -62,39 +63,51 @@ def get(cls, environment_id):
         return env
 
     @classmethod
-    def update_environment_roles(cls, portfolio_role, ids_and_roles):
+    def update_env_role(cls, environment, user, new_role):
```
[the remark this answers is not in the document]
leigh-mil: I not sure what the reason is to have these methods here! I left them in this file because that is where they were previously living

atst/routes/applications/settings.py
leigh-mil: I was planning on adding that in when I do the front end part of this card!

atst/forms/app_settings.py
```diff
@@ -0,0 +1,15 @@
+from wtforms.fields import StringField, HiddenField, RadioField, FieldList, FormField
+
+from .forms import BaseForm
+from .data import ENV_ROLES
+
+
+class EnvMemberRoleForm(BaseForm):
```
dandds: I think that nested forms shouldn't inherit from `BaseForm` because they'll do redundant things (like flash the validation error message or cache themselves separately).

PR #778, opened 2019-04-19
Removing KO from language

rachel-dtr:
## Description

Getting rid of any mentions of "KO" in AT-AT as it was recently disclosed that not all organizations in the DoD use that abbreviation for contracting officer.

### DIFF
--- a/translations.yaml
+++ b/translations.yaml
@@ -91,7 +91,7 @@ forms:
     to_number: Task order number
     loa: Line of accounting (LOA) &#35;1
     custom_clauses_label: Task order custom clauses (optional)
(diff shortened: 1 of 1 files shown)


montana-mil:
are we still using the getting started doc? i found another one here: https://github.com/dod-ccpo/atst/blob/master/templates/help/docs/getting-started.html#L14

montana-mil:
found these jawns too
Screen Shot 2019-04-19 at 11 45 18 AM

rachel-dtr:
@montana-mil looks like those are no longer used (requests & fin verification)

PR #779, opened 2019-04-19
Disable Next Button in Modal

montana-mil:
## Description
When the `add new member` multi-step modal opens, the `Next` button needs to be disabled until all of the required fields in that first step are complete. There was a bug in that the modal was tracking validity for all fields in all forms on the page instead of just the fields on the modal form.

This caused the `Next` button to be active if, for instance, a user changed a permission and then open the add member modal. In order to fix this, we need to check that the field in question is actually part of the modal. I needed to emit the field's parent uid when it mounted in order to verify that it belongs to the modal, so there is also a refactor of the js event emitter file.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/165246981

### DIFF
--- a/js/components/forms/multi_step_modal_form.js
+++ b/js/components/forms/multi_step_modal_form.js
@@ -27,7 +27,6 @@ export default {
       step: 0,
       fields: {},
       invalid: true,
-      parent_uid: '',
     }
   },
 
@@ -54,7 +53,7 @@ export default {
     handleValidChange: function(event) {
       const { name, valid, parent_uid } = event
       // check that this field is in the modal and not on some other form
-      if (parent_uid === this.parent_uid) {
+      if (parent_uid === this._uid) {
         this.fields[name] = valid
         this._checkIsValid()
       }
@@ -67,7 +66,6 @@ export default {
     handleFieldMount: function(event) {
       const { name, optional, parent_uid } = event
       this.fields[name] = optional
-      this.parent_uid = parent_uid
     },
     handleModalOpen: function(_bool) {
       this.step = 0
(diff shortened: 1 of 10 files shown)


js/components/forms/multi_step_modal_form.js
```diff
@@ -51,18 +52,22 @@ export default {
       }
     },
     handleValidChange: function(event) {
-      const { name, valid } = event
-      this.fields[name] = valid
-      this._checkIsValid()
+      const { name, valid, parent_uid } = event
+      // check that this field is in the modal and not on some other form
+      if (parent_uid === this.parent_uid) {
```
dandds: I think I'm missing something here. The `multiStepModalForm` _is_ the parent to the fields, right? Could we just compare `this._uid === parent_uid`? If that's the case, I don't think `parent_uid` needs to be a `data` on this component attribute, either.

montana-mil: i didn't know that `this._uid` was a thing! cool, that works

js/components/text_input.js
```diff
@@ -27,6 +27,7 @@ export default {
     paragraph: String,
     noMaxWidth: String,
     optional: Boolean,
+    parent_uid: String,
```
dandds: I don't think we need this `parent_uid` here.

js/components/forms/multi_step_modal_form.js
```diff
@@ -51,17 +51,20 @@ export default {
...
+      }
     },
     _checkIsValid: function() {
       const valid = !Object.values(this.fields).some(field => field === false)
       this.invalid = !valid
       return valid
     },
     handleFieldMount: function(event) {
-      const { name, optional } = event
+      const { name, optional, parent_uid } = event
```
dandds: I wouldn't unpack `parent_uid` if we're not using it here.

PR #780, opened 2019-04-19
Toggle multiple

drummond-dds:
I modified the toggler to allow multiple dropdowns on a page. The HTML structure is also more generic which will allow more flexibility in future designs.

### One toggle

Screen Shot 2019-04-19 at 14 27 42

### Multiple toggles

Screen Shot 2019-04-19 at 14 27 35
Screen Shot 2019-04-19 at 14 26 15

### DIFF
--- a/templates/components/toggle_list.html
+++ b/templates/components/toggle_list.html
@@ -9,12 +9,6 @@
   </span>
 {% endmacro %}
 
-{% macro ToggleList(item_name, item_type, length) %}
-  <div>
-    {{ caller() }}
-  </div>
-{% endmacro %}
-
 {% macro ToggleSection(section_name) %}
   <div v-show="selectedSection === '{{ section_name }}'">
     {{ caller() }}
--- a/templates/components/toggle_list.html
+++ b/templates/components/toggle_list.html
@@ -1,45 +1,22 @@
-{% from "components/icon.html" import Icon %}
-
-{% macro ToggleItem(action, item_type, length, icon) %}
-  <span
-    v-on:click="props.toggle"
-    class="icon-link icon-link--large accordion-table__item__toggler">
-    {{ action }} {{ item_type }} ({{ length }}){{ Icon(icon) }}
+{% macro ToggleButton(action, open_html, close_html, section_name) %}
+  <span v-on:click="toggleSection('{{ section_name }}')">
+    <div v-if="selectedSection === '{{ section_name }}'">
+      {{ close_html }}
+    </div>
+    <div v-else>
+      {{ open_html }}
+    </div>
   </span>
 {% endmacro %}
 
 {% macro ToggleList(item_name, item_type, length) %}
-  <li is="toggler" class="accordion-table__item">
-    <template slot-scope="props">
-      <div class="accordion-table__item-content">
-        <span v-on:click="props.toggle">
-          {{ item_name }}
-        </span>
+  <div>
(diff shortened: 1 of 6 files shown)


templates/components/toggle_list.html
```diff
@@ -1,45 +1,22 @@
...
+    <div v-if="selectedSection === '{{ section_name }}'">
+      {{ close_html }}
+    </div>
+    <div v-else>
+      {{ open_html }}
+    </div>
   </span>
 {% endmacro %}
 
 {% macro ToggleList(item_name, item_type, length) %}
```
dandds: Not sure if there's a point to `ToggleList` not if its only job is to wrap content in a `div`. (Is that necessary for Vue reasons?)

templates/fragments/applications/environments.html
```diff
@@ -22,16 +22,68 @@
...
+        <toggler inline-template>
+          <li class="accordion-table__item">
+            <div class="accordion-table__item-content">
+              <span>
+                {{ name }}
+              </span>
+              <span class="icon-link">
+                {% macro EditEnvironmentButton() %}
+                  {{ Icon('edit') }}
+                {% endmacro %}
```
dandds: Rather than macros that don't take arguments, I think it would be better to use the Jinja block set syntax to just set a variable: https://github.com/dod-ccpo/atst/blob/master/templates/fragments/admin/change_ppoc.html#L8

templates/components/toggle_list.html
drummond-dds: Removed in a8d92940

templates/fragments/applications/environments.html
drummond-dds: Changed in 29655e45ddd18717acfed2b53e968088323b42a5

dandds: It looks like this one didn't change, though?

templates/fragments/applications/environments.html
```diff
@@ -22,16 +22,72 @@
...
+                  ToggleButton(
+                    open_html=EditEnvironmentButton(),
+                    close_html=EditEnvironmentButton(),
+                    section_name="edit"
+                  )
+                }}
+              </span>
+              <span class="icon-link icon-link--large accordion-table__item__toggler">
+                {% set open_environments_button %}
+                  {{ "common.show" | translate }} Environments ({{ members_list | length }})
```
dandds: Are these supposed to have the up/down arrows? Did those get removed in the designs?

templates/fragments/applications/environments.html
```diff
@@ -22,16 +22,72 @@
...
+                  <div class="right">{{ member.role }}</div>
+                </li>
+                {% endfor %}
+              </ul>
+            {% endcall %}
+
+            {% call ToggleSection(section_name="edit") %}
+              <ul>
+                <li class="accordion-table__item__expanded">
+                  <div>
```
dandds: Do we need this extra div?

drummond-dds: This can stay for now. Its being totally replaced in my follow up ticket.

templates/fragments/applications/environments.html
drummond-dds: Added back in a35a6320

drummond-dds:
@dandds fixed application toggle in b5795a30

PR #781, opened 2019-04-22
Route rearrange

dandds:
This rearranges and updates our routes in a few ways:

- Wherever possible, routes are unnested (`/applications/:id` instead of `/portfolios/:id/applications/:id`).
- Route function names should not redundantly include the name of the resource they correspond to (`portfolios.create`, not `portfolios.create_portfolio`).
- Routes are roughly organized by major functionality or by screen. A lot of functionality lives on the portfolio admin screen now, so those routes are in `routes/portfolios/admin.py`, for example.
- Tests have also been updated and rearranged as much as possible to conform to the same organization as the route functions. (I might have missed a few here or there.)

If there are additional changes or reorganization you'd like to see, we should talk about doing them on a separate PR; this one is already huge. Please do look for omissions or mistakes in what's supposed to be here, though.

**notes**
- We should consider storing the portfolio on `g`. (We could also then log the portfolio ID for every relevant log action, among other things.)  I think we should save that work for a subsequent PR, though.
- I wanted to rearrange the template files but this PR is already a major change to the repo and that would take even longer. I think we can live with the template file organization as tech debt for a little longer.


dandds:
@leigh-mil Thanks! Yes, Ieft a TODO for those. I want to handle streamlining the routes for new TOs separately. We need `/portfolios/<port_id>/task_orders/new/1` since there's no TO ID for that TO yet, but we should think through the other cases.

PR #782, opened 2019-04-22
Only show active members for PPOC Update Dropdown

montana-mil:
## Description
The dropdown for changing the PPOC was populating with all members, even if pending or disabled. This PR changes the dropdown to only populate with active portfolio members.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/165451605

## Screenshots
Screen Shot 2019-04-22 at 11 00 44 AM

### DIFF
--- a/atst/routes/portfolios/admin.py
+++ b/atst/routes/portfolios/admin.py
@@ -70,9 +70,11 @@ def render_admin_page(portfolio, form=None):
     )
 
     assign_ppoc_form = member_forms.AssignPPOCForm()
-    assign_ppoc_form.user_id.choices += [
-        (user.id, user.full_name) for user in portfolio.users if user != portfolio.owner
-    ]
+    for pf_role in portfolio.roles:
+        if pf_role.user != portfolio.owner and pf_role.is_active:
+            assign_ppoc_form.user_id.choices += [
+                (pf_role.user.id, pf_role.user.full_name)
+            ]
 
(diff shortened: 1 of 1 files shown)


atst/routes/portfolios/admin.py
```diff
@@ -70,9 +70,10 @@ def render_admin_page(portfolio, form=None):
     )
 
     assign_ppoc_form = member_forms.AssignPPOCForm()
-    assign_ppoc_form.user_id.choices += [
-        (user.id, user.full_name) for user in portfolio.users if user != portfolio.owner
-    ]
+    for user in portfolio.users:
+        pf_role = PortfolioRoles.get(portfolio.id, user.id)
+        if user != portfolio.owner and pf_role.is_active:
+            assign_ppoc_form.user_id.choices += [(user.id, user.full_name)]
```
dandds: I think it might be cleaner to come at it from the other direction now that we care about the portfolio role here. The `portfolio` has a direct association to `portfolio_role`, so you can do something like:

```
for pf_role in portfolio.roles
    if pf_role.is_active and pf_role.user != portfolio.owner:
        assign_ppoc_form.user_id.choices += [(pf_role.user.id, pf_role.user.full_name)]
```
(I didn't run that code! Something like that should work.)

montana-mil: true true i like this angle much better

PR #785, opened 2019-04-23
Check user is in app before adding env role

montana-mil:
## Description
When adding a user to an environment, we need to make sure that the user is already in the environment's application. We check this in the route. If that check fails, a `NotFoundError` exception is raised and we log a message with the is of the user that is in question.

We default the application role permission sets to `VIEW_APPLICATION`.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/165517253

### DIFF
--- a/atst/domain/application_roles.py
+++ b/atst/domain/application_roles.py
@@ -9,7 +9,9 @@ class ApplicationRoles(object):
         return PermissionSets.get_many(set_names)
 
     @classmethod
-    def create(cls, user, application, permission_set_names):
+    def create(
+        cls, user, application, permission_set_names=[PermissionSets.VIEW_APPLICATION]
+    ):
         application_role = ApplicationRole(user=user, application_id=application.id)
 
         application_role.permission_sets = ApplicationRoles._permission_sets_for_names(
--- a/atst/domain/application_roles.py
+++ b/atst/domain/application_roles.py
@@ -6,12 +6,11 @@ from atst.domain.permission_sets import PermissionSets
 class ApplicationRoles(object):
     @classmethod
     def _permission_sets_for_names(cls, set_names):
+        set_names = set(set_names).union({PermissionSets.VIEW_APPLICATION})
         return PermissionSets.get_many(set_names)
 
     @classmethod
-    def create(
-        cls, user, application, permission_set_names=[PermissionSets.VIEW_APPLICATION]
-    ):
+    def create(cls, user, application, permission_set_names):
         application_role = ApplicationRole(user=user, application_id=application.id)
 
         application_role.permission_sets = ApplicationRoles._permission_sets_for_names(
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -70,24 +70,23 @@ class Environments(object):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if environment.application.has_member(user.id):
-            if new_role is None:
-                updated = EnvironmentRoles.delete(user.id, environment.id)
-            else:
-                env_role = EnvironmentRoles.get(user.id, environment.id)
-                if env_role and env_role.role != new_role:
-                    env_role.role = new_role
-                    updated = True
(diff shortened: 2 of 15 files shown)


atst/domain/application_roles.py
```diff
@@ -9,7 +9,9 @@ def _permission_sets_for_names(cls, set_names):
         return PermissionSets.get_many(set_names)
 
     @classmethod
-    def create(cls, user, application, permission_set_names):
+    def create(
+        cls, user, application, permission_set_names=[PermissionSets.VIEW_APPLICATION]
```
dandds: We shouldn't use mutable kwargs in Python: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments. If you want, you can steal what I did on my branch: https://github.com/dod-ccpo/atst/blob/application-invitations/atst/domain/application_roles.py (I haven't PR'd it yet.)

atst/domain/environments.py
```diff
@@ -32,6 +33,7 @@ def create_many(cls, application, names):
 
     @classmethod
     def add_member(cls, environment, user, role):
+        ApplicationRoles.create(user=user, application=environment.application)
```
dandds: I worry that this is a weird side effect for this method. I think that the domains shouldn't know or be responsible for much above their tier in the hierarchy, and the hierarchy goes portfolio -> application -> environment. Since we're not using this method in production code at the moment, I think you can safely leave out the addition.

montana-mil: ok, so we will need to check for this in the routes whenever we change something about env users, correct?

montana-mil: i think i wasn't following before -- this makes sense now. It would be a bad idea to add the member to an application if we are just adding them to an environment. They should already be in the app if we are adding to an env.

atst/routes/applications/settings.py
```diff
@@ -42,6 +44,13 @@ def serialize_env_member_form_data(application):
     return environments_list
 
 
+def check_users_are_in_application(user_ids, application):
+    for user_id in user_ids:
+        if not application.has_member(user_id):
+            raise NotFoundError("application user {}".format(user_id))
+    return True
```
dandds: The `has_member` method calls `users`, which makes a new `set` of users from the roles and then makes a new `set` of IDs from those users every time we call it, which is a lot of work. I would do something like:

```
existing_ids = [role.user_id for role in application.roles]
for user_id in user_ids:
    if not user_id in existing_ids:
       raise etc.
```

atst/routes/applications/settings.py
```diff
@@ -100,6 +109,19 @@ def update_env_roles(environment_id):
...
+
+        try:
+            user_ids = [user["user_id"] for user in env_roles_form.data["team_roles"]]
+            check_users_are_in_application(user_ids, application)
+        except NotFoundError as err:
+            app.logger.warning(
+                "Security violation for user {}, {} {}".format(
+                    g.current_user.id, request.method, request.path
+                ),
+                extra={"tags": ["update", "failure"], "security_warning": True},
```
dandds: I don't know that we can assume it's a security violation, just that it's potentially one. If Amanda and Erica talk about how they don't need Brandon on the app team anymore, and Amanda deletes his app role entirely a minute before Erica tries to set him to "no access" for every environment, Erica will generate this warning. I'd go with something that just describes what happened, like, "user <:id> requested environment role change for unauthorized user <:other_id> in application <:app_id>".

atst/routes/applications/settings.py
montana-mil: should i still keep the `has_member()` function, in that case?

atst/routes/applications/settings.py
montana-mil: i feel like having the security warning in the `extras` still makes sense though. As something to be able to filter for.

atst/routes/applications/settings.py
montana-mil: i removed it since we aren't using it anywhere else in the app

PR #790, opened 2019-04-29
Add DoD Disclaimer

drummond-dds:
# Add DoD Disclaimer

## Ticket

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/164911276)

> "The application must display the Standard Mandatory DoD Notice and Consent Banner before granting access to the application."
>
> SV-83971r2_rule

## Implementation notes

This does not store any cookies and will always display when visiting the `/` logged out page. I'm not sure if there are any app entry points other than the `Sign in with CAC` button on the home page that I need to consider. I purposely avoided storing any data in a cookie so we don't need to worry about invalidating it based on logout or time since last set.

## Screenshots

Screen Shot 2019-04-29 at 15 46 03

### DIFF
--- a/js/components/fullpagenotice.js
+++ b/js/components/fullpagenotice.js
@@ -1,0 +1,33 @@
+import checkboxinput from './checkbox_input'
+import FormMixin from '../mixins/form'
+
+export default {
+  name: 'fullpagenotice',
+
+  mixins: [FormMixin],
+
+  components: {
+    checkboxinput,
+  },
+
+  data: function() {
+    return {
+      agree: false,
+      visible: true,
+    }
+  },
+
+  mounted: function() {
+    this.$root.$on('field-change', this.handleValidChange)
+  },
+
+  methods: {
+    handleValidChange: function(event) {
+      this.agree = event.value
+    },
+
+    agreeToTermsClick: function() {
+      this.visible = false
+    },
+  },
+}
--- a/js/components/fullpagenotice.js
+++ b/js/components/fullpagenotice.js
@@ -1,33 +1,15 @@
-import checkboxinput from './checkbox_input'
-import FormMixin from '../mixins/form'
-
 export default {
   name: 'fullpagenotice',
 
-  mixins: [FormMixin],
-
-  components: {
-    checkboxinput,
-  },
-
   data: function() {
     return {
-      agree: false,
-      visible: true,
+      visible: false,
     }
   },
 
-  mounted: function() {
-    this.$root.$on('field-change', this.handleValidChange)
-  },
-
   methods: {
-    handleValidChange: function(event) {
-      this.agree = event.value
-    },
-
-    agreeToTermsClick: function() {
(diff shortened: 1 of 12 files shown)


templates/fragments/dod_notice.html
```diff
@@ -0,0 +1,43 @@
+<fullpagenotice inline-template key="notice">
+  <div class="full-page-notice" v-show="visible">
+    <div class='usa-alert usa-alert-info'>
+      <div class='usa-alert-body'>
+        <h3>You are accessing a U.S. Government (USG) Information System (IS) that is provided for USG-authorized use only.</h3>
```
dandds: Do we want all this text in the translations file? It's boilerplate from the USG and shouldn't change, but...?

drummond-dds: I'm thinking that since this is a standard government notice it won't change. I think theres also the argument that having it in the translations file might encourage people to make modifications when they shouldn't. I vote to keep it here but I can move it if we want consistency.

js/components/fullpagenotice.js
```diff
@@ -0,0 +1,33 @@
...
+
+  data: function() {
+    return {
+      agree: false,
+      visible: true,
+    }
+  },
+
+  mounted: function() {
+    this.$root.$on('field-change', this.handleValidChange)
```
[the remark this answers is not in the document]
drummond-dds: This is going to be removed since the checkbox is not required

styles/sections/_login.scss
```diff
@@ -19,6 +19,10 @@
         max-width: 25rem;
         margin: ($gap * 2) auto;
       }
+
+      .login-banner__button {
+        margin-right: 0;
+      }
```
dandds: Little thing, but the hover state for this button is weird for me on FF.

templates/fragments/dod_login.html
```diff
@@ -0,0 +1,33 @@
+<dodlogin inline-template key="notice">
+  <div>
+    <div class="usa-button usa-button-primary usa-button-big login-banner__button" v-on:click="displayNotice()">
+      {{ "login.login_button" | translate }}
+    </div>
+
+    <div class="dod-login-notice" v-if="visible" v-cloak>
+      <div class='usa-alert usa-alert-info'>
+        <div class='usa-alert-body'>
+          <h3>You are accessing a U.S. Government (USG) Information System (IS) that is provided for USG-authorized use only.</h3>
```
dandds: Try using the `usa-alert` header styles here so that the header aligns with the icon. (You can find them in the `alert` component).

drummond-dds: Fixed in b05de8632f4546b8d11d2928926f85f991093502

styles/sections/_login.scss
drummond-dds: Fixed in 950d9adb79d09d6241e8d4cdb8d034d7d93f7f87

PR #794, opened 2019-05-02
Fix padding issue on IE

drummond-dds:
Having the nested span applies padding twice on IE

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/165067811)

## Before
Screen Shot 2019-05-02 at 10 08 49

## After
Screen Shot 2019-05-02 at 10 09 00


dandds:
This will be taken care of by #797

PR #797, opened 2019-05-02
Moar style updates

dandds:
This fixes a few issues Upma identified:

- We need to be consistent with the FA icon set we're using.
- The portfolio breadcrumbs were misaligned.
- The new application page needed to be updated and fixed.
- The application environments table on the app settings page was missing some elements.

I've done those and also implemented our grid system for the app settings and team tables.


styles/components/_accordion_table.scss
```diff
@@ -43,9 +38,9 @@
 
   .accordion-table__item-content {
     padding-top: $gap * 2;
-    padding-right: $gap;
```
drummond-dds: This can just be `paddding: $gap * 2;` now

dandds: Good call. Amended the last commit to fix this.

templates/portfolios/applications/team.html
```diff
@@ -43,22 +43,22 @@
...
+              {{ "portfolios.applications.team_settings.section.table.delete_access" | translate }}
+            </div>
+            <div class="col col--grow">
+              {{ "portfolios.applications.team_settings.section.table.environment_management" | translate }}
+            </div>
+            <div class="col col--grow">
+              {{ "portfolios.applications.team_settings.section.table.team_management" | translate }}
+            </div>
+            <div class="col col--grow">
+              &nbsp;
```
montana-mil: what's this?

dandds: Non-breaking space. Browsers will strip out basic whitespace when they parse the HTML (like `<span>    a bunch of whitespace in     and around   </span>` gets reduced to `<span> a bunch of whitespace in and around </span>`). Browsers don't do that with non-breaking spaces (they count as text content). Without it, that `div` wouldn't have any content and the browser wouldn't render the columns correctly.

dandds: ^ I even had to update that comment to use `pre` tags so GH would render the extra space.

PR #799, opened 2019-05-02
Edit env form view part 2

leigh-mil:
## Description
This PR adds in the functionality to edit a users environment role the drop down/toggle radio fields to select the env role for app members. 

## Pivotal
[https://www.pivotaltracker.com/story/show/165517273](https://www.pivotaltracker.com/story/show/165517273)
Also fixes this bug: [https://www.pivotaltracker.com/story/show/165745260](https://www.pivotaltracker.com/story/show/165745260)

## Invision
[https://promptworks.invisionapp.com/share/KYRMXSXQZC9#/screens/359137879](https://promptworks.invisionapp.com/share/KYRMXSXQZC9#/screens/359137879)

## Screenshots
Open form:
Screen Shot 2019-05-02 at 4 38 47 PM
Edit user role:
Screen Shot 2019-05-02 at 4 31 36 PM
Alert on successful save:
Screen Shot 2019-05-02 at 4 31 47 PM
User is listed under their new role:
Screen Shot 2019-05-02 at 4 32 22 PM

### DIFF
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -73,7 +73,7 @@ class Environments(object):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if new_role == None:
+        if new_role is None:
             updated = EnvironmentRoles.delete(user.id, environment.id)
         else:
             env_role = EnvironmentRoles.get(user.id, environment.id)
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -126,7 +126,7 @@ class Environments(object):
         for env_role in env_roles:
             members_list.append(
                 {
-                    "user_id": env_role.user_id,
+                    "user_id": str(env_role.user_id),
                     "user_name": env_role.user.full_name,
                     "role": role,
                 }
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -73,7 +73,7 @@ class Environments(object):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if new_role is None:
+        if new_role == "no_access":
             updated = EnvironmentRoles.delete(user.id, environment.id)
         else:
             env_role = EnvironmentRoles.get(user.id, environment.id)
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -98,7 +98,7 @@ class Environments(object):
         environment = Environments.get(environment_id)
 
         for member in team_roles:
-            new_role = member["role"]
+            new_role = member["role_name"]
             user = Users.get(member["user_id"])
             Environments.update_env_role(
                 environment=environment, user=user, new_role=new_role
@@ -115,24 +115,12 @@ class Environments(object):
 
     @classmethod
     def get_members_by_role(cls, env, role):
-        env_roles = (
+        return (
             db.session.query(EnvironmentRole)
             .filter(EnvironmentRole.environment_id == env.id)
             .filter(EnvironmentRole.role == role)
             .all()
         )
-
-        members_list = []
-        for env_role in env_roles:
-            members_list.append(
-                {
-                    "user_id": str(env_role.user_id),
-                    "user_name": env_role.user.full_name,
-                    "role": role,
-                }
-            )
-
-        return members_list
 
     @classmethod
     def revoke_access(cls, environment, target_user):
--- a/atst/domain/environments.py
+++ b/atst/domain/environments.py
@@ -73,7 +73,7 @@ class Environments(object):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if new_role == "no_access":
+        if new_role == None:
             updated = EnvironmentRoles.delete(user.id, environment.id)
         else:
             env_role = EnvironmentRoles.get(user.id, environment.id)
@@ -112,6 +112,27 @@ class Environments(object):
             Environments.update_env_role(
                 environment=environment, user=member, new_role=new_role
             )
+
+    @classmethod
+    def get_members_by_role(cls, env, role):
+        env_roles = (
+            db.session.query(EnvironmentRole)
+            .filter(EnvironmentRole.environment_id == env.id)
+            .filter(EnvironmentRole.role == role)
+            .all()
+        )
+
+        members_list = []
+        for env_role in env_roles:
+            members_list.append(
+                {
+                    "user_id": env_role.user_id,
+                    "user_name": env_role.user.full_name,
+                    "role": role,
+                }
+            )
+
+        return members_list
 
     @classmethod
     def revoke_access(cls, environment, target_user):
--- a/atst/forms/app_settings.py
+++ b/atst/forms/app_settings.py
@@ -8,14 +8,13 @@ from .data import ENV_ROLES
 class MemberForm(FlaskForm):
     user_id = HiddenField()
     user_name = StringField()
-    role = RadioField(choices=ENV_ROLES, default="no_access")
+    role_name = RadioField(choices=ENV_ROLES, default="no_access")
 
     @property
     def data(self):
         _data = super().data
-        for field in _data:
-            if field == "role" and _data[field] == "no_access":
-                _data[field] = None
+        if "role_name" in _data and _data["role_name"] == "no_access":
+            _data["role_name"] = None
         return _data
 
 
--- a/atst/forms/app_settings.py
+++ b/atst/forms/app_settings.py
@@ -1,21 +1,33 @@
 from flask_wtf import FlaskForm
-from wtforms.fields import FieldList, FormField, HiddenField, RadioField
+from wtforms.fields import FieldList, FormField, HiddenField, RadioField, StringField
 
 from .forms import BaseForm
 from .data import ENV_ROLES
 
 
-class EnvMemberRoleForm(FlaskForm):
+class MemberForm(FlaskForm):
     user_id = HiddenField()
+    user_name = StringField()
     role = RadioField(choices=ENV_ROLES, default="no_access")
 
     @property
     def data(self):
         _data = super().data
-        _data.pop("csrf_token", None)
+        for field in _data:
+            if field == "role" and _data[field] == "no_access":
+                _data[field] = None
         return _data
 
 
-class EnvironmentRolesForm(BaseForm):
-    team_roles = FieldList(FormField(EnvMemberRoleForm))
+class RoleForm(FlaskForm):
+    role = HiddenField()
+    members = FieldList(FormField(MemberForm))
+
+
+class EnvironmentRolesForm(FlaskForm):
+    team_roles = FieldList(FormField(RoleForm))
     env_id = HiddenField()
+
+
+class AppEnvRolesForm(BaseForm):
+    envs = FieldList(FormField(EnvironmentRolesForm))
--- a/atst/forms/app_settings.py
+++ b/atst/forms/app_settings.py
@@ -1,14 +1,19 @@
 from flask_wtf import FlaskForm
-from wtforms.fields import StringField, HiddenField, RadioField, FieldList, FormField
+from wtforms.fields import FieldList, FormField, HiddenField, RadioField
 
 from .forms import BaseForm
 from .data import ENV_ROLES
 
 
 class EnvMemberRoleForm(FlaskForm):
-    name = StringField()
     user_id = HiddenField()
-    role = RadioField(choices=ENV_ROLES, coerce=BaseForm.remove_empty_string)
+    role = RadioField(choices=ENV_ROLES, default="no_access")
+
+    @property
+    def data(self):
+        _data = super().data
+        _data.pop("csrf_token", None)
+        return _data
 
 
 class EnvironmentRolesForm(BaseForm):
--- a/atst/routes/applications/settings.py
+++ b/atst/routes/applications/settings.py
@@ -31,7 +31,7 @@ def sort_env_users_by_role(env):
     users_list = []
     no_access_users = env.application.users - env.users
     no_access_list = [
-        {"user_id": user.id, "user_name": user.full_name, "role": "no_access"}
+        {"user_id": str(user.id), "user_name": user.full_name, "role": "no_access"}
         for user in no_access_users
     ]
     users_list.append({"role": "no_access", "members": no_access_list})
@@ -82,6 +82,8 @@ def settings(application_id):
         form=form,
         environments_obj=environments_obj,
         members_form=members_form,
+        active_toggler=http_request.args.get("active_toggler"),
+        active_toggler_section=http_request.args.get("active_toggler_section"),
     )
 
 
@@ -104,6 +106,8 @@ def update_environment(environment_id):
                 application_id=application.id,
                 fragment="application-environments",
                 _anchor="application-environments",
+                active_toggler=environment.id,
+                active_toggler_section="edit",
             )
         )
     else:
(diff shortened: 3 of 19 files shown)


atst/routes/applications/settings.py
```diff
@@ -47,11 +47,15 @@ def sort_env_users_by_role(env):
...
 def data_for_env_members_form(environment):
     data = {"env_id": environment.id, "team_roles": []}
-    for user in environment.users:
+    for user in environment.application.users:
         env_role = EnvironmentRoles.get(user.id, environment.id)
-        data["team_roles"].append(
-            {"name": user.full_name, "user_id": user.id, "role": env_role.displayname}
-        )
+
+        if env_role:
```
montana-mil: i don't think it matters but I think you can do
```role = env_role if env_role else "no_access"```

File: atst/domain/environments.py [hunk 2]
```diff
@@ -73,7 +73,7 @@ def get(cls, environment_id):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if new_role is None:
+        if new_role == "no_access":
```
dandds: I don't think the domain methods should know about the `no_access` string. That should be sanitized on the form data before it gets to here (i.e., it really should be `None` at this point). The idea of the domain methods is that they implement our business logic as abstractly as possible and shouldn't be too dependent on implementation details from the frontend or form data.

atst/forms/app_settings.py
```diff
@@ -1,16 +1,43 @@
...
     user_id = HiddenField()
-    role = RadioField(choices=ENV_ROLES, coerce=BaseForm.remove_empty_string)
+    user_name = StringField()
+    role = RadioField(choices=ENV_ROLES, default="no_access")
 
+    @property
+    def data(self):
+        _data = super().data
+        _data.pop("csrf_token", None)
+        return _data
```
dandds: If we need to do this more than once it should probably be implemented on a mixin or a separate function.

atst/routes/applications/settings.py
```diff
@@ -20,37 +19,44 @@ def get_environments_obj_for_app(application):
...
-    return users_dict
+    users_list.append(no_access)
+
+    for role in CSPRole:
+        users_list.append(
+            {
+                "role": role.value,
+                "members": Environments.get_members_by_role(env, role.value),
+            }
+        )
```
dandds: I think looping to make the list of "no access" users and then looping separately to build the lists of members with roles could be cleaner. `env.application.users` returns a `set`, and you can subtract sets from other sets in Python (i.e., `{1,2,3} - {1,2}` gives `{3}`). You could keep track of `env.application.users` in a variable and then just make sets from the members for each role to subtract from it in your loop on line 41. Whatever is left at the end is the list of no-access users.

atst/forms/app_settings.py
```diff
@@ -1,16 +1,43 @@
...
+
+class EnvironmentRolesForm(FlaskForm):
+    team_roles = FieldList(FormField(RoleForm))
     env_id = HiddenField()
+
+    @property
+    def data(self):
+        _data = super().data
+        _data.pop("csrf_token", None)
+        return _data
```
dandds: Should this live in a mixin or form utility function?

atst/domain/environments.py
dandds: I don't think the domain methods should know form or frontend details like the "no access" string. This should be sanitized in the form data before it's passed to this method.

atst/forms/app_settings.py
leigh-mil: actually, none of them require this to validate properly! :-D

templates/fragments/applications/edit_environments.html
```diff
@@ -84,28 +102,92 @@ <h4>{{ role }}</h4>
...
-            {% call ToggleSection(section_name="members") %}
-                <div class='app-team-settings-link'>Need to add someone new to the team? <a href='{{ url_for("applications.team", application_id=application.id) }}'>Jump to Team Settings</a></div>
-                {% for role, members in members_by_role.items() %}
-                  {{ RolePanel(users=members, role=role) }}
-                {% endfor %}
+            {% call ToggleSection(section_name="members", classes="environment-roles") %}
+              {% for env_form in members_form.envs %}
+                {% if env_form.env_id.data == env['id'] %}
+                  <div class='app-team-settings-link'>
+                    Need to add someone new to the team? <a href='{{ url_for("applications.team", application_id=application.id) }}'>Jump to Team Settings</a>
```
drummond-dds: We're going to have to go through this PR and extract as much text as we can into translations.yaml

templates/fragments/applications/edit_environments.html
```diff
@@ -84,28 +102,92 @@ <h4>{{ role }}</h4>
...
+                  </div>
+                  <form action="{{ url_for('applications.update_env_roles', environment_id=env['id']) }}" method="post">
+                    {{ members_form.csrf_token }}
+                    {{ env_form.env_id() }}
```
drummond-dds: I forget what this line is doing.

templates/fragments/applications/edit_environments.html
```diff
@@ -84,28 +102,92 @@ <h4>{{ role }}</h4>
...
+                                {{ Icon('edit', classes="icon--medium") }}
+                              </span>
+                              <div v-show="selectedSection === member.user_id" class='environment-role__user-field'>
+                                <div class="usa-input">
+                                  <fieldset data-ally-disabled="true" class="usa-input__choices" v-on:change="onInput">
+                                    <ul v-for='(role, roleinputindex) in roles' v-bind:id="'envs-{{ loop.index0 }}-team_roles-' + roleindex + '-members-' + memberindex + '-role'">
+                                      <li>
+                                        <input
+                                          v-bind:checked="member.role === role.role"
+                                          v-bind:name="'envs-{{ loop.index0 }}-team_roles-' + roleindex + '-members-' + memberindex + '-role'"
```
drummond-dds: Lets think of a cleaner/dryer way of formulating the ID/names for form inputs. Maybe we can have a function in the vue component that accepts arguments and formulates it?

js/components/forms/edit_environment_role.js
```diff
@@ -1,63 +1,100 @@
...
+      this.removeUser(userId)
+      this.addUser(userInfo, newRole)
+      this.toggleSection()
+    },
+  },
+
+  mounted: function() {
+    this.$parent.$on('role-change', data => {
+      this.updateRoles(data.userId, data.value)
+    })
```
dandds: It seems curious to me that the component both emits and watches for its own `role-change` events. Is there a `onInput` can't just do the update? Alternatively: can't this just be a method call instead of an event?

templates/fragments/applications/edit_environments.html
```diff
@@ -84,28 +102,92 @@ <h4>{{ role }}</h4>
...
+                    {{ members_form.csrf_token }}
+                    {{ env_form.env_id() }}
+                    <edit-environment-role inline-template v-bind:initial-roles='{{ env_form.team_roles.data | tojson }}'>
+                      <div>
+                        <div v-for='(role, roleindex) in roles' class='environment-role'>
+                          <h4 v-if='checkNoAccess(role)'>Unassigned (No Access)</h4>
+                          <h4 v-else v-html='role.role'></h4>
+                          <ul class='environment-role__users'>
+                            <div v-if="role.members && !role.members.length" class='environment-role__no-user'>Currently no members are in this role</div>
+                            <li v-for='(member, memberindex) in role.members' class="environment-role__user" v-bind:class="{'unassigned': checkNoAccess(member)}">
```
dandds: Seems confusing that you can `checkNoAccess` on both a role and a member within a role?

templates/fragments/applications/edit_environments.html
```diff
@@ -84,28 +102,92 @@ <h4>{{ role }}</h4>
...
+                                </div>
+                              </div>
+                              <input
+                                v-bind:id="'envs-{{ loop.index0 }}-team_roles-' + roleindex + '-members-' + memberindex + '-user_id'"
+                                v-bind:name="'envs-{{ loop.index0 }}-team_roles-' + roleindex + '-members-' + memberindex + '-user_id'"
+                                type="hidden"
+                                v-bind:value='member.user_id'>
+                            </li>
+                          </ul>
+                        </div>
```
dandds: One thing worth considering is that if we were using more traditional component hierarchies, we would probably have a separate Vue component for each item within a loop here:
```
<edit-environment-role>
  <role>
    <member>
      <radioRoles /> 
    </member>
  </role>
</edit-environment-role>
```
You would pass the relevant info from `edit-environment-role` to each child `role` like:

```
<role v-for="'(role, roleindex) in roles'" v-bind:role="role" v-bind:index="index">
[more vue goes here]
</role>
```

Then, one of the deeper child components would be responsible for emitting your `role-change` event and the parent would watch for it. The advantage to something like this is that each component has a smaller set of responsibilities. We can talk about this IRL if you want. I'm not saying we have to do it, but given the complexity here I think we should really consider it.

js/components/forms/edit_environment_role.js
```diff
@@ -1,63 +1,100 @@
...
+        this.selectedSection = null
+      } else {
+        this.selectedSection = sectionName
+      }
+    },
+
+    onInput: function(e) {
+      this.changed = true
+      this.$parent.$emit('role-change', {
+        event: e,
```
drummond-dds: I think this was in for debugging?

js/components/forms/edit_environment_role.js
```diff
@@ -1,63 +1,100 @@
...
   },
 
-  mounted: function() {
-    this.$root.$on('revoke-' + this.applicationId, this.revoke)
-  },
-
   methods: {
-    change: function(e) {
-      this.new_role = e.target.value
+    sanitizeValues: function(roles) {
```
drummond-dds: This is only run once on `this.initialRoles`. I'm wondering if this belongs in the `methods` section or if it should live within the `data: function()`? I think it could belong in either place.

styles/sections/_application_edit.scss
```diff
@@ -50,10 +53,35 @@
...
+        left: -0.1rem;
+        border: solid 1px $color-gray-light;
+        width: 20rem;
+        z-index: 3;
+
+        .usa-input {
+          margin: 0;
+
+          li {
+            background-color: white;
```
drummond-dds: We can use `$color-white`

templates/components/toggle_list.html
```diff
@@ -9,8 +9,8 @@
   </span>
 {% endmacro %}
 
-{% macro ToggleSection(section_name) %}
-  <div v-show="selectedSection === '{{ section_name }}'">
+{% macro ToggleSection(section_name, classes) %}
```
drummond-dds: Should we default `classes=""`?

js/components/forms/edit_environment_role.js
leigh-mil: woops! you are correct!

templates/fragments/applications/edit_environments.html
leigh-mil: It is because both a role and a member are objects with a key 'role', but if we do the refactor to make the data structure simpler this will change!

templates/fragments/applications/edit_environments.html
montana-mil: it looks like it's probably a `HiddenField`

atst/domain/environments.py
```diff
@@ -73,7 +73,7 @@ def get(cls, environment_id):
     def update_env_role(cls, environment, user, new_role):
         updated = False
 
-        if new_role is None:
+        if new_role == None:
```
montana-mil: why'd we switch this? https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html

templates/fragments/applications/edit_environments.html
drummond-dds: Fixed in 0cf30e9

js/components/forms/edit_environment_role.js
drummond-dds: This has now been removed

styles/sections/_application_edit.scss
drummond-dds: Fixed in 43b8e01

atst/domain/environments.py
drummond-dds: Fixed in 43b8e01

atst/forms/app_settings.py
```diff
@@ -1,16 +1,33 @@
...
-    role = RadioField(choices=ENV_ROLES, coerce=BaseForm.remove_empty_string)
+    user_name = StringField()
+    role = RadioField(choices=ENV_ROLES, default="no_access")
 
+    @property
+    def data(self):
+        _data = super().data
+        for field in _data:
+            if field == "role" and _data[field] == "no_access":
+                _data[field] = None
```
dandds: I don't think you need a `for` loop here. Just:
```
if "role" in _data and _data["role"] == "no_access":
    [do a thing]
```

templates/fragments/applications/edit_environments.html
leigh-mil: As per our IRL convo, we are going to add a chore for that refactor to be done at a later time.

atst/routes/applications/settings.py
```diff
@@ -67,15 +71,19 @@ def check_users_are_in_application(user_ids, application):
...
 
     return render_template(
         "portfolios/applications/settings.html",
         application=application,
         form=form,
-        environments_obj=get_environments_obj_for_app(application=application),
+        environments_obj=environments_obj,
+        members_form=members_form,
+        active_toggler=http_request.args.get("active_toggler"),
+        active_toggler_section=http_request.args.get("active_toggler_section"),
```
dandds: I'm not going to ask you to change this, but do we know for a fact that the page is supposed to reload with that section open?

atst/routes/applications/settings.py
```diff
@@ -159,11 +178,23 @@ def update_env_roles(environment_id):
...
+                fragment="application-environments",
+                _anchor="application-environments",
+                active_toggler=environment.id,
+                active_toggler_section="members",
+            )
         )
-        return redirect(url_for("applications.settings", application_id=application.id))
     else:
         # TODO: Create a better pattern to handle when a form doesn't validate
         # if a user is submitting the data via the web page then they
```
dandds: While we're here, can you update this to return 400 for an invalid submission?

dandds: (and remove the TODO 😁 )

js/components/forms/edit_environment_role.js
```diff
@@ -1,63 +1,98 @@
...
+      roles.forEach(role => {
+        role.members.forEach(member => {
+          if (member.role === null) {
+            member.role = 'no_access'
```
dandds: If you can do it with a minimum of effort, I'd like to distinguish the outer role group from an individual member's role anywhere those concepts apply. Something like `role_category` for the group vs `role` for the member.

js/components/forms/edit_environment_role.js
```diff
@@ -1,63 +1,98 @@
...
+
+    getUserInfo: function(userId) {
+      for (let role of this.roles) {
+        for (let member of role.members) {
+          if (member.user_id === userId) {
+            return member
+          }
+        }
+      }
+    },
```
dandds: We have `ramda` as a dependency for this kind of thing. You could try using it to clean this up a little. like this person illustrates: https://gist.github.com/cezarneaga/e7377357d62a2b2909685c1fb94125bb

atst/routes/applications/settings.py
leigh-mil: I reviewed it with UX, and they said it made sense to have it opened after the form submits!

js/components/forms/edit_environment_role.js
leigh-mil: Tried using `ramda`, but was unable to get it to work!

PR #800, opened 2019-05-03
Replace sass-convert with prettier

drummond-dds:
Using `sass-convert` was great but it had some flaws:

1. We couldn't run it with the `check` option and return a `1` exit status when a format was required.
2. It was just an extra tool we were introducing to the project.

This PR moves to using `prettier` which addresses those to issues.


drummond-dds:
I'd love for it to also sort styles but there isn't support for that.

PR #801, opened 2019-05-03
Application Team Table Permissions

montana-mil:
## Description
This PR sets up permissions editing for users with team management permissions. It creates a form to handle permission changes and environment role changes. A success banner posts when changes are saved.

The Save button is disabled until a change has been made. Because we have nested vue components within the `base-form` component, I needed to add a `watch` option for the `OptionsInput` which emits a Boolean to all of its parents when that component changes. Before this addition, the change was only being tracked by the immediate parent, so a grandparent component (in this case, `base-form`) never knew about changes to grandchildren components.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/165060777

## Screenshots
Screen Shot 2019-05-02 at 3 26 59 PM
Screen Shot 2019-05-02 at 3 27 07 PM

### DIFF
--- a/atst/forms/team.py
+++ b/atst/forms/team.py
@@ -12,27 +12,25 @@ class PermissionsForm(FlaskForm):
     perms_env_mgmt = SelectField(
         translate("portfolios.applications.members.new.manage_envs"),
         choices=[
-            (None, "View only"),
+            ("", "View only"),
             (PermissionSets.EDIT_APPLICATION_ENVIRONMENTS, "Edit access"),
         ],
-        filters=[BaseForm.remove_empty_string],
     )
     perms_team_mgmt = SelectField(
         translate("portfolios.applications.members.new.manage_team"),
         choices=[
-            (None, "View only"),
+            ("", "View only"),
             (PermissionSets.EDIT_APPLICATION_TEAM, "Edit access"),
         ],
-        filters=[BaseForm.remove_empty_string],
     )
     perms_del_env = SelectField(
-        choices=[(None, "No"), (PermissionSets.DELETE_APPLICATION_ENVIRONMENTS, "Yes")],
-        filters=[BaseForm.remove_empty_string],
+        choices=[("", "No"), (PermissionSets.DELETE_APPLICATION_ENVIRONMENTS, "Yes")]
     )
 
     @property
     def data(self):
         _data = super().data
+        _data.pop("csrf_token", None)
         permission_sets = []
         for field in _data:
             if _data[field] is not None:
@@ -46,6 +44,13 @@ class MemberForm(FlaskForm):
     user_name = StringField()
     environment_roles = FieldList(FormField(EnvironmentForm))
     permission_sets = FormField(PermissionsForm)
+
+    @property
+    def data(self):
+        _data = super().data
+        _data.pop("csrf_token", None)
+
+        return _data
 
 
 class TeamForm(BaseForm):
--- a/atst/forms/team.py
+++ b/atst/forms/team.py
@@ -1,5 +1,6 @@
 from flask_wtf import FlaskForm
 from wtforms.fields import FormField, FieldList, HiddenField, StringField
+from wtforms.validators import Required
 
 from .application_member import EnvironmentForm
 from .forms import BaseForm
@@ -43,7 +44,7 @@ class PermissionsForm(FlaskForm):
 
 
 class MemberForm(FlaskForm):
-    user_id = HiddenField()
+    user_id = HiddenField(validators=[Required()])
     user_name = StringField()
     environment_roles = FieldList(FormField(EnvironmentForm))
     permission_sets = FormField(PermissionsForm)
--- a/atst/forms/team.py
+++ b/atst/forms/team.py
@@ -1,0 +1,51 @@
+from flask_wtf import FlaskForm
+from wtforms.fields import FormField, FieldList, HiddenField
+
+from .application_member import EnvironmentForm
+from .forms import BaseForm
+from atst.forms.fields import SelectField
+from atst.domain.permission_sets import PermissionSets
+from atst.utils.localization import translate
+
+
+class PermissionsForm(FlaskForm):
+    perms_env_mgmt = SelectField(
+        translate("portfolios.applications.members.new.manage_envs"),
+        choices=[
+            (None, "View only"),
+            (PermissionSets.EDIT_APPLICATION_ENVIRONMENTS, "Edit access"),
+        ],
+        filters=[BaseForm.remove_empty_string],
+    )
+    perms_team_mgmt = SelectField(
+        translate("portfolios.applications.members.new.manage_team"),
+        choices=[
+            (None, "View only"),
+            (PermissionSets.EDIT_APPLICATION_TEAM, "Edit access"),
+        ],
+        filters=[BaseForm.remove_empty_string],
+    )
+    perms_del_env = SelectField(
+        choices=[(None, "No"), (PermissionSets.DELETE_APPLICATION_ENVIRONMENTS, "Yes")],
+        filters=[BaseForm.remove_empty_string],
+    )
+
+    @property
+    def data(self):
+        _data = super().data
+        permission_sets = []
+        for field in _data:
+            if _data[field] is not None:
+                permission_sets.append(_data[field])
+
+        return permission_sets
+
+
+class MemberForm(FlaskForm):
+    user_id = HiddenField()
+    environment_roles = FieldList(FormField(EnvironmentForm))
+    permission_sets = FormField(PermissionsForm)
+
+
+class TeamForm(BaseForm):
+    members = FieldList(FormField(MemberForm))
--- a/atst/forms/team.py
+++ b/atst/forms/team.py
@@ -9,22 +9,25 @@ from atst.utils.localization import translate
 
 
 class PermissionsForm(FlaskForm):
-    perms_env_mgmt = SelectField(
-        translate("portfolios.applications.members.new.manage_envs"),
-        choices=[
-            ("", "View only"),
-            (PermissionSets.EDIT_APPLICATION_ENVIRONMENTS, "Edit access"),
-        ],
-    )
     perms_team_mgmt = SelectField(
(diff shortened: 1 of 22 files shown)


atst/routes/applications/team.py
```diff
@@ -2,61 +2,102 @@
...
+        )
+
+    team_data = []
+    for member in application.members:
+        user_id = str(member.user.id)
+        user_name = member.user.full_name
+        permission_sets = {
+            "perms_env_mgmt": PermissionSets.EDIT_APPLICATION_ENVIRONMENTS
+            if member.has_permission_set(PermissionSets.EDIT_APPLICATION_ENVIRONMENTS)
+            else "",
```
dandds: I might break this logic into it's own little helper func, like:
```
def get_form_permission_set_value(member, perm):
    if member.has_permission_set(perm):
        return perm
    else:
        return ""
```
Something like that, just for readability.

atst/forms/team.py
```diff
@@ -0,0 +1,57 @@
...
+        ],
+    )
+    perms_del_env = SelectField(
+        choices=[("", "No"), (PermissionSets.DELETE_APPLICATION_ENVIRONMENTS, "Yes")]
+    )
+
+    @property
+    def data(self):
+        _data = super().data
+        _data.pop("csrf_token", None)
```
dandds: I wonder why this hasn't been a problem on other nested forms. 🤔

atst/routes/applications/team.py
```diff
@@ -2,61 +2,102 @@
...
-        }
-
     env_roles = [
         {"environment_id": e.id, "environment_name": e.name}
         for e in application.environments
     ]
-    member_form = NewMemberForm(data={"environment_roles": env_roles})
+    new_member_form = NewMemberForm(data={"environment_roles": env_roles})
+
+    if not application.members:
```
dandds: Are there errors if you don't return early in this case? It seems like if there are no members, nothing should happen in the monster loop on line 36, and we'll just give the template a `team_form` that doesn't have any data.

atst/routes/applications/team.py
```diff
@@ -2,61 +2,102 @@
...
+
+
+@applications_bp.route("/application/<application_id>/team", methods=["POST"])
+@user_can(Permissions.EDIT_APPLICATION_MEMBER, message="update application member")
+def update_team(application_id):
+    application = Applications.get(application_id)
+    form = TeamForm(http_request.form)
+
+    if form.validate():
+        # TODO check that all users coming through are app members
```
dandds: Wouldn't that just be a check for `None` on the result of line 89?

atst/routes/applications/team.py
```diff
@@ -2,61 +2,102 @@
...
+        for member in form.members:
+            app_role = ApplicationRoles.get(member.data["user_id"], application.id)
+            new_perms = [perm for perm in member.data["permission_sets"] if perm != ""]
```
dandds: I would do this filtering in the `data` method of the form instead (i.e. just don't add empty strings to the `permission_sets` list).

styles/components/_accordion_table.scss
```diff
@@ -44,8 +44,20 @@
   .accordion-table__item-content {
     padding-top: $gap * 2;
     padding-right: $gap;
-    padding-left: $gap * 2;
     padding-bottom: $gap * 3;
+
+    .first {
+      padding: 1rem;
+    }
```
dandds: I think this will get overwritten when you rebase, but you should check out positional selectors like `first-child`: https://developer.mozilla.org/en-US/docs/Web/CSS/:first-child

templates/portfolios/applications/team.html
```diff
@@ -24,104 +25,69 @@
...
-            <div class="responsive-table-wrapper__title">
-              <div class="h3">
-                {{ "portfolios.applications.team_settings.section.title" | translate({ "application_name": application.name }) }}
+      <base-form inline-template>
+        <div class='responsive-table-wrapper panel'>
+          {% if g.matchesPath("application-members") %}
+            {% include "fragments/flash.html" %}
+          {% endif %}
+          <form method='POST' id="team" action='{{ url_for("applications.update_team", application_id=application.id) }}' autocomplete="off" enctype="multipart/form-data">
+          {{ team_form.csrf_token }}
```
dandds: Not the biggest deal, but I might move the `form` tags into the "edit" version of the list. We don't need them in the markup for the view-only.

templates/fragments/applications/edit_team.html
```diff
@@ -0,0 +1,51 @@
+{% from "components/options_input.html" import OptionsInput %}
+
+{% for member_form in team_form.members %}
+  {% set environment_roles_form = member_form.environment_roles %}
+  {% set permissions_form = member_form.permission_sets %}
+
+    {% macro PermissionField(field) %}
+      {% set display_value = dict(field.choices).get(field.data) %}
+      <span class="user-permission{% if field.data %} green{% endif %}">{{ display_value }}</span>
+    {% endmacro %}
```
dandds: I don't think you're using this macro in this template.

atst/forms/team.py
montana-mil: It has been for some -- @leigh-mil helped me with this

templates/fragments/applications/edit_team.html
```diff
@@ -0,0 +1,47 @@
+{% from "components/options_input.html" import OptionsInput %}
+
+{% for member_form in team_form.members %}
+  {% set environment_roles_form = member_form.environment_roles %}
+  {% set permissions_form = member_form.permission_sets %}
+
+  <form method='POST' id="team" action='{{ url_for("applications.update_team", application_id=application.id) }}' autocomplete="off" enctype="multipart/form-data">
```
dandds: Because the form tags are inside the `for` loop, you're getting a form for every user and only the first one is getting updated properly. Just move this to line 2, and the closing tag outside `endfor`.

atst/models/application.py
```diff
@@ -28,6 +28,7 @@ class Application(
         back_populates="application",
         primaryjoin="and_(Environment.application_id==Application.id, Environment.deleted==False)",
     )
+    # TODO: filter condition on this relationship?
```
drummond-dds: Is this a follow up ticket on PT?

montana-mil: just made a chore here: https://www.pivotaltracker.com/story/show/165867842

atst/routes/applications/team.py
```diff
@@ -2,64 +2,122 @@
...
+
+@applications_bp.route("/application/<application_id>/team", methods=["POST"])
+@user_can(Permissions.EDIT_APPLICATION_MEMBER, message="update application member")
+def update_team(application_id):
+    application = Applications.get(application_id)
+    form = TeamForm(http_request.form)
+
+    if form.validate():
+        for member in form.members:
+            app_role = ApplicationRoles.get(member.data["user_id"], application.id)
```
drummond-dds: How to handle case where a team member is deleted before the form is submitted has been discussed. This will cause some updates to save where as once the 404 is raised the others won't save. Using transactional commit rollbacks has been discussed and a [PT ticket](https://www.pivotaltracker.com/story/show/165868121) has been created for follow up

templates/fragments/applications/edit_team.html
```diff
@@ -0,0 +1,48 @@
...
+
+            {% set close_html %}
+              {{ "portfolios.applications.team_settings.environments" | translate }} ({{ environment_roles_form | length }}) {{ Icon('caret_up') }}
+            {% endset %}
+
+            {{
+              ToggleButton(
+                open_html=open_html,
+                close_html=close_html,
+                section_name="environments"
```
drummond-dds: Since this will be in a `for` loop I think you need the `section_name` to be unique. Otherwise it will always open the same toggler on click

templates/fragments/applications/edit_team.html
```diff
@@ -0,0 +1,48 @@
...
+              )
+            }}
+          </div>
+        </div>
+        {% call ToggleSection(section_name="environments") %}
```
drummond-dds: See above comment about `section_name` uniqueness in the `for` loop

templates/fragments/applications/read_only_team.html
```diff
@@ -0,0 +1,45 @@
...
+
+          {% set close_html %}
+            {{ "portfolios.applications.team_settings.environments" | translate }} ({{ member.environment_roles | length }}) {{ Icon('caret_up') }}
+          {% endset %}
+
+          {{
+            ToggleButton(
+              open_html=open_html,
+              close_html=close_html,
+              section_name="environments"
```
drummond-dds: Should be unique here and in `ToggleSection`

atst/routes/applications/team.py
montana-mil: thanks!

tests/domain/test_application_roles.py
```diff
@@ -33,3 +33,30 @@ def test_enabled_application_role():
     ApplicationRoles.enable(app_role)
 
     assert app_role.status == ApplicationRoleStatus.ACTIVE
+
+
+def test_get():
```
drummond-dds: Can we throw in an exception that it raises `NotFoundError` for an invalid ID?

templates/fragments/applications/edit_team.html
montana-mil: It looks like because there is a new `toggler` component for each iteration of the `for` loop, the scope of the `section_name` is separate for each. It seems counterintuitive, but this works as is!

js/mixins/form.js
```diff
@@ -16,8 +16,7 @@ export default {
       const { name, valid, parent_uid } = event
       if (typeof this[name] !== undefined) {
         this.fields[name] = valid
-
-        if (parent_uid === this._uid) {
+        if (event['parent_uid'] === this._uid || event['watch']) {
```
leigh-mil: What will happen if multiple components on the page have `watch` set to `true`?

montana-mil: this should still work in that case, since each event emits a `watch`. this is a bit of a trick that just works for now.

PR #795, opened 2019-05-02
Edit environment save fail

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/165067811)

Clearing out the name for the "Edit environment name" form and then clicking submit would show incorrect and confusing errors. It would show "Error occurred" and also a "Success" banner. This PR shows the expected error and also opens the correct form dropdown where the errors are displayed for the missing environment name.

## After
Screen Shot 2019-05-02 at 13 19 39

## Before
Screen Shot 2019-05-01 at 1 30 18 PM_big

PR #803, opened 2019-05-03
Edit environment save fail refactor

drummond-dds:
This is a refactor of https://github.com/dod-ccpo/atst/pull/795

We are depending on the front end to display and report errors. Don't need to have backend logic to handle it for now.

### DIFF
--- a/js/components/toggler.js
+++ b/js/components/toggler.js
@@ -6,13 +6,17 @@ export default {
 
   mixins: [FormMixin],
 
+  props: {
+    initialSelectedSection: String,
+  },
+
   components: {
     textinput,
   },
 
   data: function() {
     return {
-      selectedSection: null,
+      selectedSection: this.initialSelectedSection,
     }
   },
 
--- a/js/components/toggler.js
+++ b/js/components/toggler.js
@@ -6,10 +6,6 @@ export default {
 
   mixins: [FormMixin],
 
-  props: {
-    initialSelectedSection: String,
-  },
-
   components: {
     textinput,
   },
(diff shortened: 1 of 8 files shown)


js/components/toggler.js
```diff
@@ -6,23 +6,60 @@ export default {
...
       } else {
         this.selectedSection = sectionName
       }
     },
+
+    handleSubmit: function(event) {
+      if (this.invalid) {
+        event.preventDefault()
+      }
+    },
```
dandds: This seems like too much for the toggler at this point. Would it be feasible to try two things:

- Make a small Vue form component for the "edit environment" form. It would sit below the toggler in the Vue component tree and be responsible for this stuff.
- Pull out the validation logic into a mixin that the above component and the multi step modal component can share? (It looks like the logic is the same.)

templates/fragments/applications/edit_environments.html
```diff
@@ -97,10 +98,9 @@ <h4>{{ role }}</h4>
...
               <ul>
                 <li class="accordion-table__item__expanded">
-                  {% set edit_form = env['edit_form'] %}
-                  <form action="{{ url_for('applications.update_environment', environment_id=env['id']) }}" method="post">
+                  <form action="{{ url_for('applications.update_environment', environment_id=env['id']) }}" method="post" v-on:submit="handleSubmit">
                     {{ edit_form.csrf_token }}
-                    {{ TextInput(edit_form.name) }}
+                    {{ TextInput(edit_form.name, validation='requiredField') }}
                     {{
                       SaveButton(
```
dandds: I notice that I can erase what's in the environment name field so that the input is empty and it doesn't disable the submit button. (You can't submit it, but the button still appears active.) I think the button should be disabled, too, if the form is invalid.

drummond-dds: Fixed in https://github.com/dod-ccpo/atst/pull/803/commits/0e3217d3ac7ece73cc06d7c62c220a49e8435b5e

js/components/toggler.js
drummond-dds: Moved in 334f3d8

drummond-dds:
@dandds ready for a re review

PR #809, opened 2019-05-06
Resource finder

dandds:
PT bug: https://www.pivotaltracker.com/story/show/165776075

The bug involved an application user not being able to see the environments table on the user page. The problem was more general: the `user_can` function provided to the Jinja templates only checked portfolio-level permissions. I updated the function to check application, portfolio, and sitewide permissions so that it's analogous to our route access decorator.

I made the decision here to load the largest resources--application, portfolios, and task orders--once and place the results on `g`. Now the context processors and access functions don't have to do redundant lookups. 

We can evaluate whether we want the route functions to rely on the copies of the resources now available on `g`. That should come in a future PR and after some discussion, though. (Mostly I think it's bad timing to make a PR that touches every route function.)

**NOTES**
We had this [chore](https://www.pivotaltracker.com/story/show/165426643) to investigate whether we were running redundant queries in these contexts. SQLAlchemy says it implements an [identity map within the session](https://docs.sqlalchemy.org/en/13/orm/session_basics.html#is-the-session-a-cache) which should do some basic caching. The context processor and the access decorator both query for the portfolio, so I checked on master to see if we were issuing redundant queries: we are. We might be doing some operation that invalidates the session's map in the interval between the two lookups, but either way, there are two lookups. I checked in both SQLAlchemy's own output and by enabling verbose logging for postgres. This PR reduces the portfolio lookups. I'll add more notes to the chore.

### DIFF
--- a/atst/utils/context_processors.py
+++ b/atst/utils/context_processors.py
@@ -1,6 +1,6 @@
 from operator import attrgetter
 
-from flask import request as http_request, g
+from flask import g
 from sqlalchemy.orm.exc import NoResultFound
 
 from atst.database import db
@@ -10,7 +10,7 @@ from atst.models.permissions import Permissions
 from atst.domain.portfolios.scopes import ScopedPortfolio
 
 
-def get_portfolio_from_context(view_args):
+def get_resources_from_context(view_args):
     query = None
 
     if "portfolio_id" in view_args:
@@ -20,14 +20,14 @@ def get_portfolio_from_context(view_args):
 
     elif "application_id" in view_args:
         query = (
-            db.session.query(Portfolio)
+            db.session.query(Portfolio, Application)
             .join(Application, Application.portfolio_id == Portfolio.id)
             .filter(Application.id == view_args["application_id"])
         )
 
     elif "environment_id" in view_args:
         query = (
-            db.session.query(Portfolio)
+            db.session.query(Portfolio, Application)
             .join(Application, Application.portfolio_id == Portfolio.id)
             .join(Environment, Environment.application_id == Application.id)
             .filter(Environment.id == view_args["environment_id"])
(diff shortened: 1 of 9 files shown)


atst/utils/context_processors.py
```diff
@@ -1,67 +1,105 @@
...
+from atst.models import (
+    Application,
+    Environment,
+    Portfolio,
+    PortfolioInvitation,
+    PortfolioRole,
+    TaskOrder,
+)
 from atst.models.permissions import Permissions
 from atst.domain.portfolios.scopes import ScopedPortfolio
```
montana-mil: if we want to keep these in order, this line should be moved up

atst/utils/context_processors.py
```diff
@@ -1,67 +1,105 @@
...
+            g.current_user, g.application, permission
+        )
+    elif g.portfolio:
+        return Authorization.has_portfolio_permission(
+            g.current_user, g.portfolio, permission
+        )
+    else:
+        return Authorization.has_atat_permission(g.current_user, permission)
+
+    return False
```
montana-mil: this won't ever be called, will it? since the loop ends in an `else`?

atst/utils/context_processors.py
```diff
@@ -1,67 +1,105 @@
...
 
-    def user_can(permission):
-        if portfolio:
-            return Authorization.has_portfolio_permission(
-                g.current_user, portfolio, permission
-            )
-        return False
 
-    if not portfolio is None:
+def user_can_view(permission):
```
montana-mil: i like this name so much better!

atst/utils/context_processors.py
```diff
@@ -1,67 +1,103 @@
...
+    elif g.portfolio:
+        return Authorization.has_portfolio_permission(
+            g.current_user, g.portfolio, permission
+        )
+    else:
+        return Authorization.has_atat_permission(g.current_user, permission)
+
+
+def portfolio():
+    if not g.portfolio is None:
```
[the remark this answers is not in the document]
dandds: I don't know; I looked around PEP 8 (the Python styleguide) a while and couldn't find any specific recommendation. There is an example of your form in the third bullet item [here](https://www.python.org/dev/peps/pep-0008/#programming-recommendations), which seems like a tacit endorsement.

PR #810, opened 2019-05-07
Add new application members to an empty application.

dandds:
The link button to add a new application member to an app that did not
have any ("empty state")  was not wired up. This:

- Moves the two steps of the new member modal into macros so that they
  can be reused with different button/link types
- Wires the "invite a new team member" button on the application team
  page
- Makes a minor CSS tweak so that icons nested under `.empty-state` in
  the modal aren't enormous

**Note**
The "add a new application member" workflow is currently not easily testable for DDS users, since they don't get access to our sample applications unless they log in as one of the test users.

### DIFF
--- a/templates/fragments/applications/add_new_application_member.html
+++ b/templates/fragments/applications/add_new_application_member.html
@@ -1,136 +1,15 @@
-{% from "components/icon.html" import Icon %}
-{% from "components/text_input.html" import TextInput %}
-{% from "components/checkbox_input.html" import CheckboxInput %}
 {% from "components/multi_step_modal_form.html" import MultiStepModalForm %}
 
-{% set step_one %}
-  <div class="modal__form--header">
-    <h1>Invite new member</h1>
-  </div>
-  <div class='form-row'>
-    <div class='form-col form-col--half'>
-      {{ TextInput(new_member_form.user_data.first_name, validation='requiredField') }}
-    </div>
-    <div class='form-col form-col--half'>
-      {{ TextInput(new_member_form.user_data.last_name, validation='requiredField') }}
-    </div>
-  </div>
-  <div class='form-row'>
-    <div class='form-col form-col--half'>
-      {{ TextInput(new_member_form.user_data.email, validation='email') }}
-    </div>
-    <div class='form-col form-col--half'>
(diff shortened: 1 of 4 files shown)


templates/fragments/applications/add_new_application_member.html
```diff
@@ -1,136 +1,15 @@
...
-    </div>
-  </div>
-{% endset %}
+{% import "fragments/applications/new_member_modal_content.html" as member_steps %}
+
 {{ MultiStepModalForm(
     'add-app-mem',
     member_form,
     url_for("applications.create_member", application_id=application.id),
-    [step_one, step_two],
```
drummond-dds: I know this is copy/paste from the old implementation but I think its cleaner to include the variable name in the keyword arguments. Can we change them to

```
{{ MultiStepModalForm(
  name='app-app-mem',
  form=member_form,
  action=url_for("applications.create_member", application_id=application.id),
  ....
```

drummond-dds:
Looks good. I know its copy/paste but can you add in keyword names into the macro call in this PR?

templates/fragments/applications/add_new_application_member.html
montana-mil: +1 for this

templates/fragments/applications/new_member_modal_content.html
```diff
@@ -0,0 +1,128 @@
+{% from "components/icon.html" import Icon %}
+{% from "components/text_input.html" import TextInput %}
+{% from "components/checkbox_input.html" import CheckboxInput %}
+{% from "components/multi_step_modal_form.html" import MultiStepModalForm %}
```
montana-mil: i don't think this one is used anymore here

montana-mil:
I poked around on the modals and things seem to be working fine :)

templates/fragments/applications/add_new_application_member.html
dandds: 👍 fixed on the rebase

templates/fragments/applications/new_member_modal_content.html
dandds: 👍 fixed on the rebase

PR #811, opened 2019-05-08
App team form submit

dandds:
This fixes an IE 10 bug with https://www.pivotaltracker.com/n/projects/2160940/stories/165060777.

I was wrong; the `form` attribute on `input` elements does not work on IE 10. Submit inputs or buttons need to be inside the form. That was the specific issue in this case.

In order to do that, I needed to detangle the `MultiStepModalForm` button from the modal itself. Left as-is, the modal form would be nested inside the application team form, which also won't work. I made some extensive changes to the template files so that we can put the multi-step modals elsewhere on the page, separate from the buttons or links that open them.

Once confusing thing I found in our modal usage is that _every_ Vue component tracks `activeModal`. This means that some components just update their local copy of `activeModal`, and the root component doesn't know about it. I added an extra event listener to ensure that they all update `activeModal` correctly. In a future refactor, _only_ the root component should track `activeModal`. I don't feel comfortable making that change at the moment, though.


templates/components/modal.html
```diff
@@ -1,7 +1,7 @@
 {% from "components/icon.html" import Icon %}
 
 {% macro Modal(name, dismissable=False, classes="") -%}
-  <div v-show="activeModal === '{{name}}'" v-cloak>
+  <div v-show="this.$root.activeModal === '{{name}}'" v-cloak>
```
leigh-mil: This change makes it so the root component emits `activeModal` and not the component it is called from, right?

dandds: Almost: any component can emit a modal event, but we only check what the root component thinks the `activeModal` is. Only the root is responsible for tracking that; otherwise, it's possible different components could think `activeModal` is different for them.

PR #812, opened 2019-05-10
Updating content across the site

rachel-dtr:
## Description

Continuing to clean up some content across the site.

### DIFF
--- a/atst/forms/portfolio_member.py
+++ b/atst/forms/portfolio_member.py
@@ -14,29 +14,29 @@ class PermissionsForm(BaseForm):
     perms_app_mgmt = SelectField(
         translate("forms.new_member.app_mgmt"),
         choices=[
-            (PermissionSets.VIEW_PORTFOLIO_APPLICATION_MANAGEMENT, "View only"),
-            (PermissionSets.EDIT_PORTFOLIO_APPLICATION_MANAGEMENT, "Edit access"),
+            (PermissionSets.VIEW_PORTFOLIO_APPLICATION_MANAGEMENT, "View"),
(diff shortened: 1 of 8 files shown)


atst/utils/flash.py
```diff
@@ -162,7 +160,7 @@
         "title_template": translate("flash.success"),
         "message_template": """
             {{ "flash.application.deleted" | translate({"application_name": application_name}) }}
-            <a href="#">Undo</a>.
+            <a href="#">translate("common.undo")</a>
```
drummond-dds: I think we want to call this as 

```
<a href="#">{{ "common.undo" | translate }}</a>
```

atst/forms/portfolio_member.py
```diff
@@ -14,29 +14,29 @@ class PermissionsForm(BaseForm):
     perms_app_mgmt = SelectField(
         translate("forms.new_member.app_mgmt"),
         choices=[
-            (PermissionSets.VIEW_PORTFOLIO_APPLICATION_MANAGEMENT, "View only"),
-            (PermissionSets.EDIT_PORTFOLIO_APPLICATION_MANAGEMENT, "Edit access"),
+            (PermissionSets.VIEW_PORTFOLIO_APPLICATION_MANAGEMENT, "View"),
```
drummond-dds: These should reference the `translations.yaml`. I can help you get this sorted on Monday if you need help finding the correct translation keys.

drummond-dds:
Specs are failing because it looks like we have hard coded strings in the test files. We can get an engineer to pair with you on this to get them passing again.

montana-mil:
Ready for another look!

PR #814, opened 2019-05-10
Add in missing "delete environment" translations

drummond-dds:


translations.yaml
```diff
@@ -45,6 +45,8 @@ common:
 components:
   modal:
     close: Close
+    destructive_message: You will no longer be able to access this environment
```
dandds: Should environments-specific language be in `components.modal`? You could just make the resource noun a variable and pass it in.

drummond-dds: Fixed in 1f0b48703ce99fe01b02d0819d3698805b83ea7b

drummond-dds:
@dandds ready for re review

PR #816, opened 2019-05-13
Notification System

rthoward:
PR for https://www.pivotaltracker.com/story/show/165518816

Adds a new `NotificationSender` which builds up on our email system to send notifications to some set of recipients which are stored in the database. For the time being this set of recipients is determined by this rule: `"send emails to all recipients and log all notifications somewhere"`.

Currently some tests which assert the state of the queue are failing, because they trigger an error which causes the notification system to put an email task on the queue. The easiest sane way to fix that is probably to disable the notification system in test, but I'd like some input on that because I never prefer sprinkling `if env == 'DEV':`s around the codebase (even in "infrastructure" code like this).

- [ ] Fix queue tests
- [ ] Solicit feedback on email subject and body


atst/utils/notification_sender.py
```diff
@@ -0,0 +1,25 @@
...
+        recipients = self._get_recipients(type_)
+        self.logger.info(
+            "Sending a notification to these recipients: {}\n\n{}".format(recipients, body))
+        self.queue.send_mail(recipients, self.EMAIL_SUBJECT, body)
+
+    def _get_recipients(self, type_):
+        return [
+            recipient.email
+            for recipient in db.session.query(NotificationRecipient).all()
+        ]
```
dandds: Would a query that just loads the emails as a list work? I don't think we need to model object for anything. (It would also be nice to put the query lookup on the enqueued job too, but maybe not worth the effort on this first cut.)

atst/routes/errors.py
```diff
@@ -20,6 +20,7 @@ def log_error(e):
 
 def handle_error(e, message="Not Found", code=404):
     log_error(e)
+    current_app.notification_sender.send(message)
```
dandds: I think we should have some kind of finer-grained logic for determining when to send notifications here. (For example, probably not for every 404.) Not sure off the top of my head what that logic looks like, though. It should probably be configurable, like a whitelist of HTTP error codes or ranges.

atst/utils/notification_sender.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __init__(self, queue: ATSTQueue, logger: Logger):
+        self.queue = queue
+        self.logger = logger
+
+    def send(self, body, type_=None):
+        recipients = self._get_recipients(type_)
+        self.logger.info(
+            "Sending a notification to these recipients: {}\n\n{}".format(recipients, body))
+        self.queue.send_mail(recipients, self.EMAIL_SUBJECT, body)
```
dandds: > Sending a notification to these recipients

Should this happen closer to when the email is actually sent, i.e., on the enqueued job itself? I am wondering whether this should be its own separate job-type, similar to `ATSTQueue.send_mail` with the logging added. In that case, you'd use the `rq` workers logger. I'm not sold on this, just thinking aloud.

atst/utils/notification_sender.py
rthoward: Very good call! I'll do that

PR #817, opened 2019-05-14
Delete an application member

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/165062523)

```gherkin
Scenario: 
Given I am logged in as a user with Edit Team privs on an application and looking at the Team settings page
When I click Delete Member,
Then I see a modal per the design,
And the modal has a warning banner reading "Warning! This action is permanent. {Username} will no longer be able to access this application,"
And there is text reading "Please type the word 'DELETE' to confirm:",
And there is a text box,
And there is a Delete User button that is grayed out until/unless the text box contains 'delete' (case insensitive),
And there is a cancel button,
Then I click Delete User,
Then I am taken back to the page I was on, to the table I was at, with a success banner - "You have successfully deleted {username} from {Application-name}"
And the user is set to No Access for all environments in the application.
```

### DIFF
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -3,6 +3,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from atst.database import db
 from . import BaseDomainClass
 from atst.domain.application_roles import ApplicationRoles
+from atst.domain.environment_roles import EnvironmentRoles
 from atst.domain.environments import Environments
 from atst.domain.exceptions import NotFoundError
 from atst.domain.users import Users
@@ -100,3 +101,15 @@ class Applications(BaseDomainClass):
                 Environments.add_member(environment, user, env_role_data.get("role"))
 
         return application_role
+
+    @classmethod
+    def remove_member(cls, application, user):
+        application_role = ApplicationRoles.get(
+            user_id=user.id, application_id=application.id
+        )
+
+        db.session.delete(application_role)
+        db.session.commit()
+
+        for env in application.environments:
+            EnvironmentRoles.delete(user_id=user.id, environment_id=env.id)
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -103,13 +103,13 @@ class Applications(BaseDomainClass):
         return application_role
 
     @classmethod
-    def remove_member(cls, application, user):
+    def remove_member(cls, application, user_id):
         application_role = ApplicationRoles.get(
-            user_id=user.id, application_id=application.id
+            user_id=user_id, application_id=application.id
         )
 
         db.session.delete(application_role)
         db.session.commit()
 
         for env in application.environments:
-            EnvironmentRoles.delete(user_id=user.id, environment_id=env.id)
+            EnvironmentRoles.delete(user_id=user_id, environment_id=env.id)
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -108,7 +108,8 @@ class Applications(BaseDomainClass):
             user_id=user_id, application_id=application.id
         )
 
-        db.session.delete(application_role)
+        application_role.status = ApplicationRoleStatus.DISABLED
+        db.session.add(application_role)
         db.session.commit()
 
         for env in application.environments:
--- a/atst/routes/applications/team.py
+++ b/atst/routes/applications/team.py
@@ -158,3 +158,31 @@ def create_member(application_id):
             _anchor="application-members",
         )
     )
+
+
+@applications_bp.route(
+    "/applications/<application_id>/members/<user_id>/delete", methods=["POST"]
+)
+# TODO: Is this correct??
+@user_can(Permissions.EDIT_APPLICATION_MEMBER, message="remove application member")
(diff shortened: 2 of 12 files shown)


atst/domain/applications.py
```diff
@@ -100,3 +101,15 @@ def create_member(
...
 
         return application_role
+
+    @classmethod
+    def remove_member(cls, application, user):
+        application_role = ApplicationRoles.get(
+            user_id=user.id, application_id=application.id
+        )
+
+        db.session.delete(application_role)
```
dandds: I'm not sure about the answer to this question, but we should talk through whether this should actually delete the role or set it to disabled: https://github.com/dod-ccpo/atst/blob/master/atst/models/application_role.py#L15

atst/routes/applications/team.py
```diff
@@ -158,3 +158,31 @@ def create_member(application_id):
             _anchor="application-members",
         )
     )
+
+
+@applications_bp.route(
+    "/applications/<application_id>/members/<user_id>/delete", methods=["POST"]
+)
+# TODO: Is this correct??
+@user_can(Permissions.EDIT_APPLICATION_MEMBER, message="remove application member")
```
dandds: You mean, is EDIT_APPLICATION_MEMBER the correct permission? Yes. We could add DELETE_APPLICATION_MEMBER in the same permission set for granularity, which might be a good idea.

atst/routes/applications/team.py
```diff
@@ -158,3 +158,31 @@ def create_member(application_id):
...
+def remove_member(application_id, user_id):
+    application_role = ApplicationRoles.get(
+        application_id=application_id, user_id=user_id
+    )
+
+    Applications.remove_member(application=g.application, user=application_role.user)
```
dandds: `remove_member` also does `ApplicationRoles.get`. I don't think we should do it in both places.

tests/domain/test_applications.py
```diff
@@ -155,3 +158,25 @@ def test_for_user():
...
+    )
+
+    Applications.remove_member(application=application, user=member_role.user)
+
+    with pytest.raises(NotFoundError):
+        ApplicationRoles.get(user_id=user.id, application_id=application.id)
+
+    #
+    # TODO: Why does above raise NotFoundError and this returns None
+    #
```
dandds: I don't know the original rationale for `EnvironmentRoles.get` returning one or `None`. If you can change it to `raise` without breaking anything, please do. (This brings up the whole "don't commit until you're really done making changes" question again, but we can punt on that a while longer.)

templates/portfolios/applications/team.html
```diff
@@ -109,6 +112,35 @@
...
+          {% set delete_modal_id = "delete-user-{}".format(member_form.id) %}
+          {% call Modal(name=delete_modal_id, dismissable=True) %}
+            <h1>
+              {{ "portfolios.applications.remove_member.header" | translate }}
+            </h1>
+
+            {{
+              Alert(
+                title=("portfolios.applications.remove_member.alert.title" | translate),
+                message=("portfolios.applications.remove_member.alert.message" | translate({"user_name": member_form.user_name.data})),
```
dandds: The modal still has application-specific language in it.

translations.yaml
```diff
@@ -408,6 +408,12 @@ portfolios:
     app_settings_text: App settings
     create_button_text: Create
     csp_console_text: CSP console
+    remove_member:
+      alert:
+        message: '{user_name} will no longer be able to access this application'
+        title: Warning! This action is permanent.
```
dandds: This is in at least three places now. Can we consolidate?

atst/routes/applications/team.py
dandds: I guess you could argue that the domain method should be self-contained. At a minimum, could you confirm that SQLAlchemy's basic ID cache is working here and we don't issue two equivalent SELECT statements?

tests/domain/test_applications.py
drummond-dds: I just did a quick spike and I think we should defer this to a future follow up PR

atst/routes/applications/team.py
drummond-dds: Added in f1299dc4b0e22af8f07714dbb56c9f9009a652f8

translations.yaml
drummond-dds: Fixed in b884c8a762eb2ad92476345513aafd35babcbdcb

templates/portfolios/applications/team.html
drummond-dds: I think this is in the requirements

```cucumber
And the modal has a warning banner reading "Warning! This action is permanent. {Username} will no longer be able to access this application,"
```

drummond-dds: Screen Shot 2019-05-14 at 15 13 56

atst/domain/applications.py
drummond-dds: Fixed in 9927b22783e2deaac19e239fe068cd5e6e3d22e5

PR #819, opened 2019-05-14
Catch None values for IsNumber validator

montana-mil:
## Description
If a user checked the box to invite an officer on the `Oversight` page of the Task Order form without entering a DOD ID, a server error would occur. The validator, `IsNumber` was only catching `ValueError`s, and the `NoneType` was throwing a `TypeError`.

With this PR, `IsNumber` catches both errors and handles them in the same way -- namely, a validation error is thrown and the user is asked to fill in the DOD ID.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/166005243

## Screenshots
Screen Shot 2019-05-14 at 4 49 00 PM

### DIFF
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -32,7 +32,7 @@ def IsNumber(message=translate("forms.validators.is_number_message")):
     def _is_number(form, field):
         try:
             int(field.data)
-        except ValueError:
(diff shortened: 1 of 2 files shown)


montana-mil:
@dandds i updated an existing test! `test_IsNumber_rejects_anything_else`

PR #820, opened 2019-05-14
Set Delete Access only if User has Env Mgmt Permission

montana-mil:
## Description
When a user does not have environment management permission, it doesn't make sense to be able to set the delete environment access for that user. This PR hardcodes "No" under the `Delete Access` section on the team page for any user who has `view_only` access for `perms_env_mgmt`.

I also added a new CSS class called `dropdown-row-text` to vertically center text in rows that also have a dropdown.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/165893607

## Screenshots
Screen Shot 2019-05-14 at 5 53 10 PM
Read Only version:
Screen Shot 2019-05-14 at 5 52 54 PM

### DIFF
--- a/js/components/tables/team_member.js
+++ b/js/components/tables/team_member.js
@@ -15,7 +15,7 @@ export default {
 
   data: function() {
     return {
-      edit_access: this.initialValue.includes('edit') ? true : false,
+      edit_access: this.initialValue.includes('edit'),
       open: false,
     }
   },
--- a/js/components/tables/team_member.js
+++ b/js/components/tables/team_member.js
@@ -2,7 +2,7 @@ import EnvironmentRole from '../environment_role'
 import optionsinput from '../options_input'
 
 export default {
-  name: 'team',
+  name: 'team-member',
 
   props: {
     initialValue: String,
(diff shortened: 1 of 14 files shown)


js/components/tables/team_member.js
```diff
@@ -0,0 +1,31 @@
...
+  },
+
+  components: {
+    EnvironmentRole,
+    optionsinput,
+  },
+
+  data: function() {
+    return {
+      edit_access: this.initialValue.includes('edit') ? true : false,
```
dandds: Do you need the ternary? Does `edit_access: this.initialValue.includes('edit')` work?

js/components/tables/team_member.js
```diff
@@ -0,0 +1,31 @@
...
+      open: false,
+    }
+  },
+
+  methods: {
+    flip: function() {
```
dandds: Can we get a more specific name than `flip`? Like `hasEnvEditAccess` or something?

montana-mil:
Because of the new design pivot, we are tabling this functionality for now.

PR #821, opened 2019-05-15
We dont need this in vue template

drummond-dds:
`this` was breaking vue templates in Microsoft Edge

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/165935329)
* [Breaking commit](https://github.com/dod-ccpo/atst/commit/457b1f9523c73466d767fcd366c859b62518729f#diff-91300a21de63234ef6d7d70fed0f90dcL4)

### DIFF
--- a/templates/components/modal.html
+++ b/templates/components/modal.html
@@ -1,7 +1,7 @@
 {% from "components/icon.html" import Icon %}
 
 {% macro Modal(name, dismissable=False, classes="") -%}
-  <div v-show="this.$root.activeModal === '{{name}}'" v-cloak>
+  <div v-show="$root.activeModal === '{{name}}'" v-cloak>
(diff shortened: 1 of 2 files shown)


templates/components/modal.html
```diff
@@ -1,7 +1,7 @@
 {% from "components/icon.html" import Icon %}
 
 {% macro Modal(name, dismissable=False, classes="") -%}
-  <div v-show="this.$root.activeModal === '{{name}}'" v-cloak>
+  <div v-show="$root.activeModal === '{{name}}'" v-cloak>
```
dandds: There's another instance of this we should get rid of in `templates/components/multi_step_modal_form.html`.

drummond-dds: I amended the commit to include this

PR #822, opened 2019-05-15
App members edit

dandds:
This covers two PT stories: https://www.pivotaltracker.com/story/show/165059841 and https://www.pivotaltracker.com/story/show/165156615

This covers displaying more of the edit functionality for the application team on the members table. The following was implemented:
- displaying the user's environment role in the environments toggle
- making the user's environment role editable

**note**
It wasn't specified in the designs, but I made it so that changing the environment role updates the displayed role in the environments list. It seemed unintuitive to not do that.

Screen Shot 2019-05-15 at 10 08 42 AM


js/components/environment_role.js
```diff
@@ -0,0 +1,40 @@
+import optionsinput from './options_input'
+import { emitEvent } from '../lib/emitters'
+
+export default {
+  name: 'environment-role',
+
+  mixins: [],
```
leigh-mil: You may want to clean this up and remove `mixins` and `computed` since they are not being used

styles/components/_portfolio_layout.scss
```diff
@@ -399,10 +393,12 @@
         max-width: 95%;
       }
 
+      .application-list-item__environment__name {
+      }
```
leigh-mil: this class selector isn't being styled, is it here for a reason?

js/components/environment_role.js
dandds: fixed up the commit 👍

styles/components/_portfolio_layout.scss
dandds: fixed up a commit. honeslty dunno how that got there 🤷‍♀

montana-mil: lol it might have been from me somehow. i feel like i wrote this at some point

montana-mil:
If a user doesn't have delete access, the CSS is a bit off on the toggle open:
Screen Shot 2019-05-15 at 4 14 32 PM

templates/fragments/applications/edit_team.html
```diff
@@ -26,30 +30,67 @@
...
+                    </button>
+                    {{ environment_form.environment_id() }}
+                  </div>
+                </div>
+              </environment-role>
             </li>
           {% endfor %}
+        </ul>
+        <div class="accordion-table__item__action-group">
+          <a class="icon-link">
```
montana-mil: does this need a permissions check for `EDIT_APPLICATION_MEMBER`?

styles/components/_accordion_table.scss
```diff
@@ -91,6 +91,30 @@
...
+        .icon-link {
+          padding-top: 0.5rem;
+          float: none;
+        }
+
+        > *:first-child {
+          padding-left: 0;
+        }
+
+        > *:last-child {
```
montana-mil: This is causing the funky CSS when a user does not have `DELETE_APPLICATION_MEMBER` permissions but can see the `add_to_environment` link.

templates/fragments/applications/edit_team.html
montana-mil: or maybe `ASSIGN_ENVIRONMENT_MEMBER`

styles/components/_accordion_table.scss
dandds: Talked about this IRL: this isn't a situation that can happen with the current permission structure. If we change the permission structure down the road, we'll get ugly UI in places (can't/won't future-proof for this).

templates/fragments/applications/edit_team.html
dandds: fixed here: f0502871060fb55279936f0996600d5fb6e1b961

translations.yaml
```diff
@@ -20,6 +20,7 @@ base_public:
 common:
   back: Back
   cancel: Cancel
+  close: Close
```
montana-mil: looks like we can get rid of this one now: https://github.com/dod-ccpo/atst/pull/822/files#diff-598f198d6c48aadf73010c7c45257b5bR50

PR #825, opened 2019-05-15
Updating alert language

rachel-dtr:
## Description

Updating some more language per the content audit.

### DIFF
--- a/atst/utils/flash.py
+++ b/atst/utils/flash.py
@@ -41,9 +41,7 @@ MESSAGES = {
     },
     "task_order_draft": {
         "title_template": translate("task_orders.form.draft_alert_title"),
-        "message_template": """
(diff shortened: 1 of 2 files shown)


atst/utils/flash.py
```diff
@@ -41,9 +41,7 @@
     },
     "task_order_draft": {
         "title_template": translate("task_orders.form.draft_alert_title"),
-        "message_template": """
-        <p>Please complete your task order before submitting it for approval.</p>
-        """,
+        "message_template": translate("task_orders.form.draft_alert_message"),
```
leigh-mil: Did this change how this appears at all on the site?
I'm asking because the original text included html paragraph tags, which may have affected the presentation.

leigh-mil: as per our irl convo -- removing the tags did not change how the alert was displayed

PR #826, opened 2019-05-15
Raise error if CRL download fails.

dandds:
PT card: https://www.pivotaltracker.com/story/show/166039596

The download method for Libcloud objects returns a boolean, which means
that our CRL download could fail silently. The RackspaceCRLProvider
would not raise an error until it tried to open the full path for the
downloaded resource. This checks the return status of the download call
and raises an error if the download failed. For reference:

https://libcloud.readthedocs.io/en/latest/storage/api.html#libcloud.storage.base.StorageDriver.download_object


drummond-dds:
👍 looks good

PR #827, opened 2019-05-15
Create Notification System

richard-dds:
PR for https://www.pivotaltracker.com/story/show/165518816

Adds a new NotificationSender which builds up on our email system to send notifications to some set of recipients which are stored in the database. For the time being this set of recipients is determined by this rule: "send emails to all recipients and log all notifications somewhere".


atst/queue.py
```diff
@@ -41,5 +44,14 @@ def send_mail(self, recipients, subject, body):
     def _send_mail(self, recipients, subject, body):
         app.mailer.send(recipients, subject, body)
 
+    @classmethod
+    def _send_notification_mail(self, recipients, subject, body):
+        app.logger.info(
+            "Sending a notification to these recipients: {}\n\n{}".format(
```
drummond-dds: I think we need to log the subject to meet the AC?

PR #834, opened 2019-05-20
Prevent multiple active sessions

richard-dds:
https://www.pivotaltracker.com/story/show/165518844

Only allow one active session for a given user. Achieves this by saving the user's last session id in the database and manually deleting that previous session when the user logs in again.

This feature is disabled in dev for convenience, but can be enabled by setting`LIMIT_CONCURRENT_SESSIONS = true` in the configuration.


alembic/versions/ab1167fc8260_add_user_last_session_id.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = 'ab1167fc8260'
+down_revision = '404bb5bb3a0e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('users', sa.Column('last_session_id', postgresql.UUID(as_uuid=True), server_default=sa.text('uuid_generate_v4()'), nullable=True))
```
dandds: Do we need a server default on this column? It's always external data, right?

atst/utils/session_limiter.py
```diff
@@ -0,0 +1,19 @@
+from atst.domain.users import Users
+
+
+class SessionLimiter(object):
+    def __init__(self, config, session, redis):
+        self.limit_logins = config["LIMIT_CONCURRENT_SESSIONS"]
```
dandds: Would it be better to just pass this value in by itself, rather than the whole config?

rthoward: I'm not sure. In this case, it does feel a little unnecessary to pass the entire config, but I could imagine that with a more complex object with more config, having a bunch of config params to pass in the constructor might be too cumbersome.

atst/app.py
```diff
@@ -249,10 +254,18 @@ def make_mailer(app):
...
+
+
 def make_notification_sender(app):
     app.notification_sender = NotificationSender(queue)
 
 
+def make_session_limiter(app, session, config):
+    app.session_limiter = SessionLimiter(config, session, app.redis)
+
+
```
dandds: Looks like this function made it in twice.

rthoward: Whoops! Must have been a bad rebase.

richard-dds: Whoops! Must have been a bad rebase

atst/utils/session_limiter.py
richard-dds: I'm not sure. In this case, it does feel a little unnecessary to pass the entire config, but I could imagine that with a more complex object with more config, having a bunch of config params to pass in the constructor might be too cumbersome.

PR #838, opened 2019-05-21
Filter deleted environment roles from top-level domain method.

dandds:
PT story: https://www.pivotaltracker.com/story/show/166158006

### DIFF
--- a/tests/domain/test_environment_roles.py
+++ b/tests/domain/test_environment_roles.py
@@ -14,3 +14,13 @@ def test_get_for_application_and_user():
     assert len(roles) == 1
     assert roles[0].environment == env1
     assert roles[0].user == user
+
+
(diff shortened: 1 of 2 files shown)


tests/domain/test_environment_roles.py
```diff
@@ -14,3 +14,14 @@ def test_get_for_application_and_user():
     assert len(roles) == 1
     assert roles[0].environment == env1
     assert roles[0].user == user
+
+
+def test_get_for_application_and_user_does_not_return_deleted():
+    user = UserFactory.create()
+    application = ApplicationFactory.create()
+    env1 = EnvironmentFactory.create(application=application)
+    EnvironmentFactory.create(application=application)
```
montana-mil: do we need this line?

dandds: amended!

PR #843, opened 2019-05-23
Use application_role.id to reference users in team page forms.

dandds:
More work on this PT story: https://www.pivotaltracker.com/story/show/160432365

Very small PR to change our user/role ID handling for the members form on the application team page. Membership in a resource should be decoupled from the users table.

### DIFF
--- a/atst/domain/application_roles.py
+++ b/atst/domain/application_roles.py
@@ -48,6 +48,18 @@ class ApplicationRoles(object):
         return app_role
 
     @classmethod
+    def get_by_id(cls, id_):
+        try:
+            return (
+                db.session.query(ApplicationRole)
+                .filter(ApplicationRole.id == id_)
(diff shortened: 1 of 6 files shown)


atst/domain/application_roles.py
```diff
@@ -47,6 +47,17 @@ def get(cls, user_id, application_id):
 
         return app_role
 
+    @classmethod
+    def get_by_id(cls, id_):
+        try:
+            return (
+                db.session.query(ApplicationRole)
+                .filter(ApplicationRole.id == id_)
```
montana-mil: Should we check the status here?

dandds: 👍 Rebased. I think we need to do a more general sweep for this on a separate PR; the `get` method in the same module doesn't check, either, for example.

PR #844, opened 2019-05-24
Create encrypted hidden field

leigh-mil:
## Description
Create a new WTForms field that encrypts the value in a hidden field. 

### To do
- Fix places where the front end was effected by the changes (ie. places where we relied on using the value in the hidden fields)

## Pivotal
https://www.pivotaltracker.com/story/show/166097221


leigh-mil:
Tabling PR for now because of design pivot.

PR #846, opened 2019-05-28
Just clean out tables for seed sample reset

drummond-dds:
Seed sample remove script was breaking constantly. Just clean out the tables and stop maintaining broken script.

### DIFF
--- a/script/remove_sample_data.py
+++ b/script/remove_sample_data.py
@@ -5,109 +5,27 @@ import sys
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 sys.path.append(parent_dir)
 
-import ctypes
 import sqlalchemy
-from sqlalchemy import event as sqlalchemy_event
 
 from atst.database import db
 from atst.app import make_config, make_app
 
-from atst.models.audit_event import AuditEvent
-from atst.models.environment import Environment
-from atst.models.environment_role import EnvironmentRole
-from atst.models.application import Application
-from atst.models.permission_set import PermissionSet
-from atst.models.user import User
-from atst.models.portfolio_role import PortfolioRole
-from atst.models.portfolio import Portfolio
-from atst.models.mixins import AuditableMixin
 
-from atst.domain.environments import Environments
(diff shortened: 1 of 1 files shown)


script/remove_sample_data.py
```diff
@@ -5,109 +5,24 @@
...
-    "4567890123",
-    "5678901234",
-    "6789012345",
-    "2342342342",  # Andy
-    "3453453453",  # Sally
-    "4564564564",  # Betty
-    "6786786786",
-]
+    for t in meta.sorted_tables:
+        conn.execute("ALTER TABLE {} DISABLE trigger ALL;".format(t))
```
rthoward: `execute()` has query string interpolation built-in, so you don't need to use format (though it's not dangerous to do so here since `t` doesn't come from user input): ```conn.execute("ALTER TABLE ? DISABLE trigger ALL;", t)```

rthoward: Though I realize now that the query string is passed directly to our dbapi implementation, which may or may not use `?` for its interpolation sigil. If you want it to be db-agnostic, `sqlalchemy.text()`[0] is the way to go (or `format()`).

[0]: https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.text

leigh-mil:
Ignore my previous comments! I forgot to run seed_roles before I ran seed_sample!

script/remove_sample_data.py
```diff
@@ -5,109 +5,24 @@
...
-    "3456789012",
-    "4567890123",
-    "5678901234",
-    "6789012345",
-    "2342342342",  # Andy
-    "3453453453",  # Sally
-    "4564564564",  # Betty
-    "6786786786",
-]
+    for t in meta.sorted_tables:
```
dandds: I'd prefer if we didn't have to add the seed data again after running this. Could you add a list of exception tables to skip? I think the only exception is `permission_sets` at the moment.

drummond-dds: Fixed in update

PR #852, opened 2019-05-30
/home no longer redirects and has add portfolio button

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/166330641)

### Questions

Does the "Add new portfolio" endpoint exist? I think we will need to add that in with a follow up PR?

### Acceptance Criteria 

> When a user logs into AT-AT, instead of being taken to a portfolio, they 
> should be taken to a landing page.
>
> For now that page has a placeholder for future help text, and a large button 
> reading "Add New Portfolio."

### DIFF
--- a/atst/routes/__init__.py
+++ b/atst/routes/__init__.py
@@ -58,34 +58,7 @@ def helpdocs(doc=None):
 
 @bp.route("/home")
 def home():
-    user = g.current_user
-    num_portfolios = len([role for role in user.portfolio_roles if role.is_active])
-
-    if num_portfolios == 0:
-        return redirect(url_for("portfolios.portfolios"))
-    elif num_portfolios == 1:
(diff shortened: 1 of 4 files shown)


drummond-dds:
I see that the "Add new portfolio" workflow is my next ticket. I can make the url change in that PR.

atst/routes/__init__.py
```diff
@@ -58,34 +58,7 @@ def helpdocs(doc=None):
...
-            )
-    else:
-        portfolios = Portfolios.for_user(g.current_user)
-        first_portfolio = sorted(portfolios, key=lambda portfolio: portfolio.name)[0]
-        return redirect(
-            url_for(
-                "applications.portfolio_applications", portfolio_id=first_portfolio.id
-            )
-        )
+    return render_template("home.html")
```
dandds: lol

PR #854, opened 2019-05-30
Use application_role_id on environment_roles.

dandds:
In the future, an `application_invitation` will not refer to a `user` until someone accepts the invitation; they'll only reference an `application_role`. When a user is invited to an application, the inviter can specify the environments the invitee should have access to.  For this to be possible, an `environment_role` should reference an `application_role`, because no `user` entity will be known at that time.

In addition to updating all the models and domain methods necessary for this change, this commit deletes unused code and tests that were dependent on `environment_roles` having a `user_id` foreign key.

### DIFF
--- a/alembic/versions/d2390c547dca_environment_roles_has_relation_to_.py
+++ b/alembic/versions/d2390c547dca_environment_roles_has_relation_to_.py
@@ -1,0 +1,54 @@
+"""environment_roles has relation to application_roles
+
+Revision ID: d2390c547dca
+Revises: ab1167fc8260
+Create Date: 2019-05-29 12:34:45.928219
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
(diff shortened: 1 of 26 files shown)


alembic/versions/d2390c547dca_environment_roles_has_relation_to_.py
```diff
@@ -0,0 +1,54 @@
+"""environment_roles has relation to application_roles
+
+Revision ID: d2390c547dca
+Revises: c5deba1826be
```
leigh-mil: I think you need to rebase off of master and change the migration that it revises to `ab1167fc8260`

leigh-mil:
I was unable to remove a user from a portfolio, but I think this may also be solved by rebasing!
I checked all of the app and env member forms and they appear to be working.

dandds:
@leigh-mil thanks, amended the commit! Removing a portfolio member worked for me after the rebase.

PR #855, opened 2019-05-30
Update TO table

leigh-mil:
## Description
This the first PR for this card. It updates the TO table to only have the columns required in the new TO uploader/form and removes unnecessary code related to the table update. 

I skipped a bunch of tests instead of deleting them because I wanted to leave them in as a reminder to implement similar tests after I change the TO form.

## Pivotal


tests/models/test_task_order.py
```diff
@@ -9,79 +9,35 @@
 
 
 class TestTaskOrderStatus:
+    @pytest.mark.skip(reason="See if still needed after implementing CLINs")
```
dandds: We will definitely need a TO status, but it will be different from what we have. We can let these hang out til the right implementation becomes clear.

PR #856, opened 2019-05-31
Add CLIN table

leigh-mil:
## Description
Create table for CLINs. 

## Pivotal


atst/models/clin.py
```diff
@@ -0,0 +1,38 @@
...
+class CLIN(Base, mixins.TimestampsMixin):
+    __tablename__ = "clins"
+
+    id = types.Id()
+
+    task_order_id = Column(ForeignKey("task_orders.id"), nullable=False)
+    task_order = relationship("TaskOrder")
+
+    number = Column(String, nullable=False)
+    loa = Column(ARRAY(String), server_default="{}", nullable=False)
```
dandds: I think plural might be clearer.

atst/models/clin.py
```diff
@@ -0,0 +1,38 @@
...
+from sqlalchemy.orm import relationship
+
+from atst.models import Base, mixins, types
+
+
+class JEDICLINType(Enum):
+    JEDI_CLIN_1 = "jedi clin 0001"
+    JEDI_CLIN_2 = "jedi clin 0002"
+    JEDI_CLIN_3 = "jedi clin 0003"
+    JEDI_CLIN_4 = "jedi clin 0004"
```
dandds: Should these values just be the same as the descriptions in the `jedi_clin_description` property?

atst/models/clin.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class CLIN(Base, mixins.TimestampsMixin):
```
dandds: Should this get the auditable mixin? (Genuine question; I don't know the answer.)

atst/models/clin.py
leigh-mil: We decided to not do that because if the text changes it will mess up the database. I think I will just get rid of it because I think the descriptions will end up in the form drop downs selector.

atst/models/clin.py
leigh-mil: I think the task order should have the AuditableMixin, but I'm not sure if CLIN should have it too.

PR #857, opened 2019-05-31
New Portfolio Workflow

drummond-dds:
* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/166330650)

---

As an AT-AT user, I can click the "New Portfolio" button, which takes me to the New Portfolio page.

Fields:

- [x] Portfolio Name
- [x] Portfolio Description
- [x] DoD Component
- [x] Demographic Fields (Survey questions from current TO builder)
- [x] Submit button, activated when all fields are entered

When submitted

- [x] new portfolio is created (with no task orders)
- [x] user is taken to landing page for that new portfolio. 
- [x] That user is also set as the PPoC for the new portfolio.


atst/domain/portfolios/portfolios.py
```diff
@@ -16,9 +16,31 @@ class PortfolioError(Exception):
 
 class Portfolios(object):
     @classmethod
-    def create(cls, user, name, defense_component=None):
+    def create(
+        cls,
+        user,
+        name,
+        description,
+        defense_component=None,
```
drummond-dds: This is very verbose. I wonder if I should have a list of white listed fields and then get/set them after passing in the formdata?

dandds: Since we're using a form object to validate the data, I think you can just pass a dictionary.

atst/domain/portfolios/portfolios.py
```diff
@@ -16,9 +16,31 @@ class PortfolioError(Exception):
...
+            name=name,
+            defense_component=defense_component,
+            description=description,
+            app_migration=app_migration,
+            complexity=complexity,
+            complexity_other=complexity_other,
+            dev_team=dev_team,
+            dev_team_other=dev_team_other,
+            native_apps=native_apps,
+            team_experience=team_experience,
```
dandds: I would just pass the info dictionary to the top-level method (`Portfolios.create(user, portfolio_info)`) and then splat it here: `PortfoliosQuery.create(**portfolio_info)`.

drummond-dds: Fixed

atst/domain/portfolios/portfolios.py
drummond-dds: Fixed

drummond-dds:
I also updated the seed sample script

atst/routes/portfolios/index.py
```diff
@@ -19,6 +20,40 @@ def portfolios():
...
+            "complexity": form.data["complexity"],
+            "complexity_other": form.data["complexity_other"],
+            "defense_component": form.data["defense_component"],
+            "description": form.data["description"],
+            "dev_team": form.data["dev_team"],
+            "dev_team_other": form.data["dev_team_other"],
+            "name": form.data["name"],
+            "native_apps": form.data["native_apps"],
+            "team_experience": form.data["team_experience"],
+        }
```
dandds: I don't think you have to do this. You can just

```
portfolios = Portfolios.create(user=g.current_user, portfolio_attrs=form.data)
```
As long as the form only has fields defined that are permissible `Portfolio` attributes, that will work. (The form does the whitelisting for us.)

PR #858, opened 2019-05-31
Sticky CTA Header

richard-dds:
# Description
Create a reusable floating, sticking header component which contains a subheader and call-to-action buttons.

# Pivotal 
https://www.pivotaltracker.com/story/show/166330656

# Screenshot
Screen Recording 2019-05-31 at 04 16 PM

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -6,6 +6,7 @@ import classes from '../styles/atat.scss'
 import Vue from 'vue/dist/vue'
 import VTooltip from 'v-tooltip'
 import VueSticky from 'vue-sticky'
+import stickybits from 'stickybits'
 
 import dodlogin from './components/dodlogin'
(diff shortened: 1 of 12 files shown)


leigh-mil:
Not sure why it is doing this, but the style of the div with `class='sticky-cta'` is defaulting to `top: 2px`, which creates a little gap between the top and the header.
I don't see it in your code anywhere, could it be coming from `VueSticky`?  
Screen Shot 2019-06-03 at 3 23 36 PM

richard-dds:
> Looks great! Works well on FF.
> 
> Is there anything we can do about the very small gap between the top of the sticky and browser window? (Looks like it's ajust a few pixels on FF, but I see it in your gif too.)

@dandds We definitely can. It's set at 10 pixels currently, but I can bump that down. Do we want any space?

PR #860, opened 2019-06-03
Basic Task Order Review Page

montana-mil:
## Description
Front end work for the Task Order Review page. This includes mostly mocked data from the back end. The "More"/"Less" text in the description is reactive.

This does not include the sticky header or top navigation. The CLIN tables are not quite right, but this work was beginning to block others' work. The submit modal is also not implemented. 

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/166330688
This is the first of at least two PRs for this story.

## Screenshots
Screen Shot 2019-06-04 at 10 58 12 AM


styles/sections/_task_order.scss
```diff
@@ -67,6 +67,43 @@
...
+    margin-top: $gap * 3;
+    width: 33.77%;
+    max-height: 45rem;
+    background-color: $color-gray-lightest;
+  }
+
+  .semi-collapsed {
+    overflow: hidden;
+    height: $gap * 5.5;
+  }
```
dandds: I think the semi-collapsed text is going to appear on every portfolio page, so maybe we should move this some place generic.

templates/portfolios/task_orders/review.html
```diff
@@ -1,113 +1,104 @@
...
-                </li>
-              </ul>
-              <div class="task-order__loa-add-item">
-                <button v-on:click="addLOA" class="icon-link" tabindex="0" type="button">{{ Icon('plus') }} Add another LOA</button>
-              </div>
-            </div>
-
-            {{ TextInput(form.custom_clauses, paragraph=True) }}
-          </div>
+    <div class="col to-sidebar">
```
leigh-mil: Would it make sense to put the sidebar for the Funds into its own macro? This component needs to be used in a few places and that will make future updates to this box apply everywhere!

PR #861, opened 2019-06-03
New TO form

leigh-mil:
## Description
First PR implementing the new TO form, only adds functionality to add a TO number and save.

## Pivotal


atst/routes/task_orders/new.py
```diff
@@ -1,353 +1,56 @@
...
-    if workflow.task_order:
-        template_args["task_order"] = workflow.task_order
-        if http_request.args.get("ko_edit"):
-            template_args["ko_edit"] = True
-            template_args["next"] = url_for(
-                "task_orders.ko_review", task_order_id=task_order_id
-            )
-            url_args["next"] = template_args["next"]
+        task_order = TaskOrders.get(task_order_id)
+        form = TaskOrderForm(number=task_order.number)
```
dandds: Nitpicky, but this will make two instances of the `TaskOrderForm` (the first on line 15). You could make a little helper func that takes `task_order_id` as an arg and returns the correct `form`, or just make a very explicit `if` structure, like I suggest on line 44.

atst/routes/task_orders/new.py
```diff
@@ -1,353 +1,56 @@
...
+def update(portfolio_id, task_order_id=None):
+    form_data = http_request.form
+    form = TaskOrderForm(form_data)
+
+    if form.validate():
+        task_order = (
+            TaskOrders.update(task_order_id, **form.data)
+            if task_order_id
+            else TaskOrders.create(g.current_user, portfolio_id, **form.data)
+        )
```
dandds: I think this is a little hard to read. Can we be explicit:

```
task_order = None
if task_order_id:
    TaskOrders.update(task_order_id, **form.data)
[etc.]
```

tests/routes/task_orders/test_new.py
```diff
@@ -186,87 +91,15 @@ def test_task_order_form_shows_errors(client, user_session, task_order):
 
 
 @pytest.mark.skip(reason="Update after implementing new TO form")
```
dandds: Are these coming later?

leigh-mil: Yes, I was going to add these back in when I add in the functionality where the `Review task order` button is only active when the form is complete.

atst/forms/task_order.py
```diff
@@ -1,131 +1,26 @@
...
-        translate("forms.task_order.csp_estimate_label"),
-        description=translate("forms.task_order.csp_estimate_description"),
-        validators=[
-            FileAllowed(
-                ["pdf", "png"], translate("forms.task_order.file_format_not_allowed")
-            )
-        ],
-        render_kw={"accept": ".pdf,.png,application/pdf,image/png"},
-    )
     clin_01 = DecimalField(
```
montana-mil: seems like we are changing clins to the form `clin_001`, etc. Should we do that here, too?

atst/forms/task_order.py
```diff
@@ -151,135 +46,7 @@ class UnclassifiedFundingForm(FundingForm):
     )
 
 
-class OversightForm(BaseForm):
```
montana-mil: 👋

atst/utils/flash.py
```diff
@@ -2,25 +2,16 @@
 from atst.utils.localization import translate
 
 MESSAGES = {
-    "application_member_removed": {
```
montana-mil: 🎉

templates/navigation/global_sidenav.html
```diff
@@ -29,12 +29,6 @@
             <li><span class="sidenav__text">You have no portfolios yet</span></li>
           {% endif %}
         </ul>
-        <div class="sidenav__divider--small"></div>
```
montana-mil: 🥒

tests/routes/task_orders/test_new.py
```diff
@@ -1,20 +1,28 @@
...
-    TaskOrderFactory,
     PortfolioFactory,
     PortfolioRoleFactory,
+    TaskOrderFactory,
+    UserFactory,
 )
 
 
+@pytest.fixture
+def task_order():
```
montana-mil: is this better than just using a taskorder factory object?

atst/forms/task_order.py
leigh-mil: Good idea! I'll do that in my next PR

tests/routes/task_orders/test_new.py
leigh-mil: The fixture was originally in the tests, just buried deep within the code. When I deleted out the unnecessary tests I moved it to the top.

PR #863, opened 2019-06-04
Add CLIN fields to Task Order Form

leigh-mil:
## Description
First PR of several for the new/edit TO form. This adds in all of the functionality except the LOA field and did not implement any design (both will be coming in follow up PRs). 

## Pivotal
https://www.pivotaltracker.com/story/show/166330968

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -37,13 +37,12 @@ class TaskOrders(BaseDomainClass):
 
         for clin in task_order.clins:
             db.session.delete(clin)
-            db.session.commit()
 
         if number != task_order.number:
             task_order.number = number
             db.session.add(task_order)
-            db.session.commit()
 
+        db.session.commit()
         TaskOrders.create_clins(task_order_id, clins)
 
         return task_order
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -19,9 +19,9 @@ class TaskOrders(BaseDomainClass):
     UNCLASSIFIED_FUNDING = []
 
     @classmethod
-    def create(cls, creator, portfolio_id, number, clins):
+    def create(cls, creator, portfolio_id, number, clins, pdf):
         task_order = TaskOrder(
-            portfolio_id=portfolio_id, creator=creator, number=number
+            portfolio_id=portfolio_id, creator=creator, number=number, pdf=pdf
         )
         db.session.add(task_order)
         db.session.commit()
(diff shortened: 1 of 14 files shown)


atst/domain/task_orders.py
```diff
@@ -18,27 +19,50 @@ class TaskOrders(BaseDomainClass):
...
         task_order = TaskOrders.get(task_order_id)
-        for key, value in kwargs.items():
-            setattr(task_order, key, value)
+        task_order.pdf = pdf
 
-        db.session.add(task_order)
-        db.session.commit()
+        for clin in task_order.clins:
+            db.session.delete(clin)
+            db.session.commit()
```
drummond-dds: I'm not sure if we should `db.session.commit()` here or wait until everything has been added to the session? It could live on line 46 or even on line 48?

js/components/forms/to_form.js
```diff
@@ -0,0 +1,41 @@
...
+    textinput,
+    uploadinput,
+  },
+
+  props: {
+    initialClinCount: Number,
+  },
+
+  data: function() {
+    const clins = this.initialClinCount == 0 ? [''] : []
```
drummond-dds: I'm finding this assignment a little confusing. If we have clins we set an empty array?

leigh-mil: The `clins` array determines the number of blank CLIN fields that appear on the page. So if a TO form does not have any CLINs already, we will want 1 blank one on the page by default.  If a TO has CLINs, then we do not show any blank/new ones but will add them if they click the 'add clin' button

templates/task_orders/edit.html
```diff
@@ -1,36 +1,256 @@
...
+              <clin-fields v-bind:initial-clin-index='clinIndex' inline-template>
+                <div>
+                  <div class="usa-input">
+                    <fieldset data-ally-disabled="true" class="usa-input__choices">
+                        <legend>
+                          <div class="usa-input__title">
+                          </div>
+                        </legend>
+                        <select :id="'clins-' + clinIndex + '-jedi_clin_type'" :name="'clins-' + clinIndex + '-jedi_clin_type'">
+                          <option value="JEDI_CLIN_1">CLIN 01 : Unclassified</option>
```
drummond-dds: Can we use `translations.yml` here? I think its `forms.task_order.clin_01_label` etc

js/components/forms/to_form.js
dandds: I think we talked about maybe just recording this as an integer. Would that work?

PR #865, opened 2019-06-04
New application invitation flow

dandds:
PT: https://www.pivotaltracker.com/story/show/160432365

Much like #859 . This makes it so that when you invite an application member, a `user` entry from the database is not associated with the new `application_role` until the invitation recipient accepts the invite. This fixes the issue where you could enter a DOD ID that already exists in the database and get that user's information, instead of the information you entered on the invite.

I also deleted some now-unused code.


atst/routes/applications/team.py
```diff
@@ -125,6 +125,17 @@ def update_team(application_id):
         return (render_team_page(application), 400)
 
 
+def send_application_invitation(invitee_email, inviter_name, token):
+    body = render_template(
+        "emails/application/invitation.txt", owner=inviter_name, token=token
+    )
+    queue.send_mail(
+        [invitee_email],
+        "{} has invited you to a JEDI cloud application".format(inviter_name),
```
drummond-dds: Should this be in the `translations.yaml`?

atst/domain/applications.py
```diff
@@ -76,33 +82,50 @@ def delete(cls, application):
...
         for env_role_data in environment_roles_data:
-            role = env_role_data.get("role")
-            if role:
-                environment = Environments.get(env_role_data.get("environment_id"))
-                Environments.add_member(
-                    environment, application_role, env_role_data.get("role")
+            env_role_name = env_role_data.get("role")
+            environment_id = env_role_data.get("environment_id")
+            if env_role_name is not None:
+                # pylint: disable=cell-var-from-loop
```
drummond-dds: Just curious what this does

atst/routes/applications/team.py
```diff
@@ -135,19 +146,19 @@ def create_member(application_id):
...
+                inviter=g.current_user,
+                user_data=form.user_data.data,
+                permission_sets_names=form.permission_sets.data,
                 environment_roles_data=form.environment_roles.data,
             )
 
-            invite_service = InvitationService(
-                g.current_user, member, form.user_data.data.get("email")
+            send_application_invitation(
+                invite.email, g.current_user.full_name, invite.token
```
drummond-dds: Maybe we can have the keyword arguments here so we don't confuse the email/name for inviter/invitee

drummond-dds:
@dandds looks good, just some small things

atst/models/application_role.py
```diff
@@ -51,12 +51,17 @@ class ApplicationRole(
...
+        if self.invitations:
+            return self.invitations[-1]
+
     @property
     def user_name(self):
         if self.user:
             return self.user.full_name
-        else:
-            return None
+        elif self.latest_invitation:
```
montana-mil: 👍

montana-mil:
So if you add a user with an already-in-use DOD ID, they show up as the name you gave them when you added them, right? Are DOD IDs no longer unique? If they are unique, what does it mean for the user that was there first?

atst/domain/applications.py
dandds: I don't entirely understand why, but pylint objects to using the temporary variable `environment_id` in a closure there. That comments disables the pylint error.

dandds:
@montana-mil Yes, that's how it works now. DOD IDs are still unique on the `users` table. We store user information on the invitation and that's what's displayed until someone accepts the invitation. It doesn't affect users who already exist because the invitation user info is stored separately now.

atst/routes/applications/team.py
dandds: :+1: b6e51351c7561d250466ce12ed684f0cc1bd46b7

atst/routes/applications/team.py
dandds: :+1: b6e51351c7561d250466ce12ed684f0cc1bd46b7

PR #866, opened 2019-06-04
Update node packages

drummond-dds:
Running `yarn audit` shows some potentially vulnerable packages. This PR fixes all the HIGH severity warnings specifically.

## Before

Screen Shot 2019-06-04 at 13 57 00

## After
Screen Shot 2019-06-04 at 13 57 56


dandds:
@drummond-dds This looks good but I want a little more time to poke around and make sure we don't get any breaking JS changes.

drummond-dds:
Sounds good. I saved the commit message as the command I ran so you can replicate the upgrade. A quick run through on my end and things seemed to be working but we should do a more thorough run through.

PR #869, opened 2019-06-05
Filter out deleted application environment roles

dandds:
PT story: https://www.pivotaltracker.com/story/show/166469506

An `ApplicationRole` should filter deleted, dependent `environment_roles`. This was causing errors when an environment was deleted because we iterate the team's `ApplicationRole`s and dependent `EnvironmentRole`s to build the data for that page. This caused errors when an app role referred to an env role that was soft-deleted.

### DIFF
--- a/atst/models/application_role.py
+++ b/atst/models/application_role.py
@@ -51,6 +51,11 @@ class ApplicationRole(
         "PermissionSet", secondary=application_roles_permission_sets
     )
 
+    environment_roles = relationship(
+        "EnvironmentRole",
+        primaryjoin="and_(EnvironmentRole.application_role_id==ApplicationRole.id, EnvironmentRole.deleted==False)",
+    )
+
     @property
     def user_name(self):
         if self.user:
--- a/atst/models/environment_role.py
+++ b/atst/models/environment_role.py
@@ -29,7 +29,7 @@ class EnvironmentRole(
     application_role_id = Column(
         UUID(as_uuid=True), ForeignKey("application_roles.id"), nullable=False
     )
-    application_role = relationship("ApplicationRole", backref="environment_roles")
+    application_role = relationship("ApplicationRole")
 
     def __repr__(self):
         return "<EnvironmentRole(role='{}', user='{}', environment='{}', id='{}')>".format(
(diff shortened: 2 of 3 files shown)


atst/models/application_role.py
```diff
@@ -51,6 +51,11 @@ class ApplicationRole(
         "PermissionSet", secondary=application_roles_permission_sets
     )
 
+    environment_roles = relationship(
+        "EnvironmentRole",
+        primaryjoin="and_(EnvironmentRole.application_role_id==ApplicationRole.id, EnvironmentRole.deleted==False)",
```
drummond-dds: I don't think we need to pass in a string here

```diff
diff --git a/atst/models/application_role.py b/atst/models/application_role.py
index 9bcb0f6e..db08d639 100644
--- a/atst/models/application_role.py
+++ b/atst/models/application_role.py
@@ -1,11 +1,12 @@
 from enum import Enum
-from sqlalchemy import Index, ForeignKey, Column, Enum as SQLAEnum, Table
+from sqlalchemy import and_, Index, ForeignKey, Column, Enum as SQLAEnum, Table
 from sqlalchemy.dialects.postgresql import UUID
 from sqlalchemy.orm import relationship
 from sqlalchemy.event import listen
 
 from atst.utils import first_or_none
 from atst.models import Base, mixins
+from atst.models.environment_role import EnvironmentRole
 from atst.models.mixins.auditable import record_permission_sets_updates
 from .types import Id
 
@@ -53,7 +54,9 @@ class ApplicationRole(
 
     environment_roles = relationship(
         "EnvironmentRole",
-        primaryjoin="and_(EnvironmentRole.application_role_id==ApplicationRole.id, EnvironmentRole.deleted==False)",
+        primaryjoin=and_(
+            EnvironmentRole.application_role_id == id, EnvironmentRole.deleted == False
+        ),
     )
 
     @property
```

montana-mil: weird. it looks like we are doing strings everywhere we have this `primaryjoin` jawn

dandds: @drummond-dds I think I just copied the first instance straight from the docs and kept doing it. I'm loathe to change one instance and I don't think we should change all of them on this PR. If you want to make another PR to change it everywhere, that would be great.

drummond-dds: I will make this global change in a follow up PR

PR #871, opened 2019-06-05
safer-eval upgrade

drummond-dds:
Security issue with `safer-eval`. 

Updating `parcel-builder` since `safer-eval` is only a dependency of `serialize-to-js` which only a dependency of `parcel-builder`.

Running `yarn upgrade safer-eval` or `yarn upgrade serialize-to-js` changes nothing however `yarn upgrade parcel-builder` does.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,40 +2,51 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


drummond-dds:
Yup looks like the warning has gone now. I opened this PR because I thought that other one might get held up. Closing this one out.

PR #874, opened 2019-06-06
Task order uploader

dandds:
PT story: https://www.pivotaltracker.com/story/show/166332118

This doesn't complete the PT story to the designs; that will come later. The PR does two main things:

- Rewrite our uploader component (Vue and Jinja) to be closer to the new designs.
- Implement the task order upload field on the TO form.

**notes**
- I've used our existing "invalid" state instead of the new one in the designs. There is a separate story for implementing the new validation states: https://www.pivotaltracker.com/story/show/166503998
- I'm not including the other text around the upload field; we'll add that when we finalize the form and have a better idea what the shape of the HTML is in that template.
- I made some adjustments to the routes file and TO edit template so that the form route and general layout are correct.

**Screenshots**
Empty field:
Screen Shot 2019-06-06 at 2 00 23 PM

Attached PDF:
Screen Shot 2019-06-06 at 2 00 35 PM

Validation error:
Screen Shot 2019-06-06 at 2 01 01 PM


templates/task_orders/edit.html
```diff
@@ -1,36 +1,36 @@
...
-          </div>
-        </form>
-      </base-form>
-    </div>
+    <base-form inline-template>
+      {% if task_order_id %}
+        {% set action = url_for("task_orders.update", portfolio_id=portfolio.id, task_order_id=task_order_id) %}
+      {% else %}
+        {% set action = url_for("task_orders.update", portfolio_id=portfolio.id) %}
+      {% endif %}
```
leigh-mil: good catch on not switching the post url for creating a new TO vs updating an existing one!

PR #875, opened 2019-06-06
Add Sticky Header to Task Order Review Page

montana-mil:
## Description
Adds the header to the top of the TO review page -- the navigation and the sticky header. The navigation now indicates the page is a part of `Funding`.

## Screenshots
Screen Shot 2019-06-07 at 9 59 33 AM

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -373,7 +373,8 @@
 }
 
 .portfolio-funding {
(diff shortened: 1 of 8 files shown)


dandds:
@montana-mil 

> there is an issue with the More/Less link showing over the sticky header.
This is fixed on #876, so don't worry about it. The sticky header just needs a z-index set.

PR #876, opened 2019-06-06
Home page

dandds:
PT story: https://www.pivotaltracker.com/story/show/166330633

This adds the rest of the style and placeholder content for the portfolio landing page. 

Screenshot_2019-06-06 JEDI Cloud

### DIFF
--- a/styles/components/_global_layout.scss
+++ b/styles/components/_global_layout.scss
@@ -23,7 +23,6 @@ body {
 
   .global-panel-container {
     margin: $gap;
-    max-width: $site-max-width;
     flex-grow: 1;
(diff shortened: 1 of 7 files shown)


leigh-mil:
I have a design question -- Some of the descriptive texts for the icons are longer than others (see Admin), so it causes the height of the grey box to get larger and everything jumps up on the page. Would we want to set the height of the box so it doesn't do that?

dandds:
@leigh-mil 

> Would we want to set the height of the box so it doesn't do that?

That's a good point. I'm going to merge this, but we should ask for design guidance and circle back to it.

PR #877, opened 2019-06-07
Portfolio archiving

drummond-dds:
**Specs currently failing due to CI/Docker** 

---

* [Pivotal Tracker](https://www.pivotaltracker.com/story/show/166338359)

```gherkin
Scenario: 
Given I am viewing the Portfolio Admin page as the PPoC or a CCPO user,
Then I see a Deactivate Portfolio button that is only active if there are ZERO applications in the portfolio,
And the button is in its own panel below the "primary point of contact" panel
And I see explanatory content about deactivation,
When I click "Deactivate Portfolio" I am prompted with a modal where I am forced to type "deactivate" to confirm (like the delete modals for environments),
Then the portfolio is soft-deleted
And the user is sent to their home page and the user sees a success banner saying "This portfolio is now deactivated."
```


templates/home.html
```diff
@@ -6,6 +6,7 @@
 {% block content %}
 
 <main class="home">
+  {% include "fragments/flash.html" %}
```
dandds: Can we try putting this below the sticky? I get a success alert like below:

Screen Shot 2019-06-10 at 1 00 19 PM

PR #880, opened 2019-06-10
TO Review Submit Modal

montana-mil:
## Description
When reviewing a Task Order, the user must hit `Submit` and check a box to "sign" the task order. They are then redirected to the funding page for that Task Order.

## Pivotal Tracker
https://www.pivotaltracker.com/story/show/166330884

## Screenshots
Screen Shot 2019-06-11 at 11 06 35 AM
Screen Shot 2019-06-11 at 11 06 39 AM
Screen Shot 2019-06-11 at 11 07 01 AM

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -1,3 +1,4 @@
+import datetime
 from flask import current_app as app
 
 from atst.database import db
@@ -44,6 +45,16 @@ class TaskOrders(BaseDomainClass):
 
         db.session.commit()
         TaskOrders.create_clins(task_order_id, clins)
+
+        return task_order
+
+    @classmethod
+    def sign(cls, task_order, signer_dod_id):
+        task_order.signer_dod_id = signer_dod_id
+        task_order.signed_at = datetime.datetime.now()
+
+        db.session.add(task_order)
+        db.session.commit()
 
         return task_order
 
--- a/atst/routes/task_orders/index.py
+++ b/atst/routes/task_orders/index.py
@@ -1,5 +1,3 @@
-from datetime import date
-
 from flask import g, render_template, url_for, redirect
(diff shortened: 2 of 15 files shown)


atst/domain/task_orders.py
```diff
@@ -47,6 +47,17 @@ def update(cls, task_order_id, number, clins, pdf):
 
         return task_order
 
+    @classmethod
+    def sign(cls, task_order_id, signer_dod_id, signed_at):
+        task_order = TaskOrders.get(task_order_id)
+        task_order.signer_dod_id = signer_dod_id
+        task_order.signed_at = signed_at
```
dandds: I think that this method should be responsible for generating `signed_at`. It seems like the caller shouldn't have to know that detail.

dandds: I would even maybe pass in the user, so that the method signature is:

```
def sign(cls, task_order, user):
```

That way only this method has to know the details of what's in a signature.

atst/routes/task_orders/index.py
```diff
@@ -25,7 +29,32 @@ def view_task_order(task_order_id):
...
+    )
+
+
+@task_orders_bp.route("/task_orders/<task_order_id>/submit", methods=["POST"])
+@user_can(Permissions.CREATE_TASK_ORDER, "submit task order")
+def submit_task_order(task_order_id):
+
+    task_order = TaskOrders.get(task_order_id)
+    TaskOrders.sign(
+        task_order_id=task_order.id,
```
dandds: I would just pass in the TO, rather than make `sign` do another lookup to get it.

atst/domain/task_orders.py
montana-mil: for sure!

PR #882, opened 2019-06-10
Add LOAs to CLINs

leigh-mil:
## Description
Adds functionality to TO form to add multiple lines of accounting. This PR does not cover design or copy at all, the next PR will be for the designs.

## Pivotal
https://www.pivotaltracker.com/story/show/166330968

### DIFF
--- a/js/components/clin_fields.js
+++ b/js/components/clin_fields.js
@@ -1,4 +1,5 @@
 import DateSelector from './date_selector'
+import optionsinput from './options_input'
 import textinput from './text_input'
 
 export default {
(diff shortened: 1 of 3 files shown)


templates/task_orders/edit.html
```diff
@@ -7,15 +7,67 @@
...
+    v-bind:initial-clin-index='{{ index }}'
+    v-bind:initial-loa-count="{{ fields.loas.data | length }}"
+    inline-template>
+    <div>
+      <hr>
+      {{ OptionsInput(fields.jedi_clin_type) }}
+      {{ TextInput(fields.number) }}
+
+      {% for loa in fields.loas %}
+        {{ TextInput(loa) }}
```
dandds: This renders a weird field name, I'm afraid. You probably need to override it with nothing:
Screen Shot 2019-06-10 at 7 15 05 PM

js/components/clin_fields.js
```diff
@@ -1,19 +1,42 @@
...
+    return {
+      clinIndex: this.initialClinIndex,
+      indexOffset: this.initialLoaCount,
+      loas: loas,
+    }
+  },
+
+  methods: {
+    addLoa: function(event) {
+      this.loas = this.loas + 1
```
dandds: You can do `+=` here, I think.

templates/task_orders/edit.html
leigh-mil: I have already done so in my next PR! Do you want me to update it here too?

PR #884, opened 2019-06-10
Funding page Task Order statuses

richard-dds:
# Pivotal
https://www.pivotaltracker.com/story/show/166468879

# Description

Implement new designs for the task order funding page. Includes:

- new CLIN-based task order statuses (draft, unsigned, upcoming, active, expired)
- new sorting method for task orders (by status, then by time_created)
- new colors for status labels

# Screenshot
Screen Shot 2019-06-10 at 1 37 33 PM

# Notes

If you're wondering why I took kind of a strange approach in seed_sample.py, it's because I ran in to some weird issues while trying to create TaskOrders with CLINs, and was only able to get it to work by creating the CLINs separately from the TaskOrders. If I created CLINs directly on the task order, like so,
```
TaskOrderFactory.create(..., clins=[CLINFactory.create(...)])
```
the resulting TO would sometimes include a CLIN, and sometimes wouldn't. Not sure why.

### DIFF
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -15,7 +15,7 @@ class Status(Enum):
     ACTIVE = "Active"
     UPCOMING = "Upcoming"
     EXPIRED = "Expired"
-    UNSIGNED = "Unsigned"
+    UNSIGNED = "Not signed"
 
 
 SORT_ORDERING = {
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from datetime import date, datetime
+from datetime import date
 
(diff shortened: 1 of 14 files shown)


atst/models/task_order.py
```diff
@@ -62,23 +71,36 @@ def is_active(self):
...
+    @property
+    def is_signed(self):
+        return self.signed_at is not None
+
     @property
     def status(self):
-        # TODO: fix task order -- implement correctly using CLINs
-        # Faked for display purposes
-        return Status.ACTIVE
+        today = date.today()
```
drummond-dds: The docs say `date.today()` returns the current local date. 

I'm unsure what the local time zone would be since users could be spread across every geographic time zone. I guess for MVP we are ok with just using server local time?

PR #889, opened 2019-06-11
Funding page CTAs

richard-dds:
# Pivotal
A second PR for this story: https://www.pivotaltracker.com/n/projects/2160940/stories/166330906

# Description
Update the CTAs (buttons) for task orders on the funding page. Will look incomplete, since we've decided to omit buttons for pages that don't yet exist.

I also updated the `begins x | ends y` text to match the designs.

# Screenshot
Screen Shot 2019-06-11 at 2 52 09 PM

### DIFF
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -85,6 +85,14 @@ class TaskOrder(Base, mixins.TimestampsMixin):
         return self.status == Status.UNSIGNED
 
     @property
+    def has_begun(self):
+        return Clock.today() >= self.start_date
+
+    @property
+    def has_ended(self):
+        return Clock.today() >= self.end_date
+
+    @property
     def is_completed(self):
         return all([self.pdf, self.number, len(self.clins)])
 
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -65,12 +65,24 @@ class TaskOrder(Base, mixins.TimestampsMixin):
             raise TypeError("Could not set attachment with invalid type")
 
     @property
+    def is_draft(self):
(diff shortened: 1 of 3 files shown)


atst/models/task_order.py
```diff
@@ -64,14 +64,34 @@ def _set_attachment(self, new_attachment, attribute):
...
     def is_expired(self):
         return self.status == Status.EXPIRED
 
+    @property
+    def is_unsigned(self):
+        return self.status == Status.UNSIGNED
+
+    @property
+    def has_begun(self):
+        return Clock.today() >= self.start_date
```
montana-mil: just curious - is Clock better than datetime?

richard-dds: I added `Clock` yesterday as a little utility class for doing `now()` and `today()`. The only real benefit is uses a UTC timezone, which yesterday we determined was the right timezone to use for TOs.

templates/portfolios/task_orders/index.html
```diff
@@ -18,24 +18,49 @@
...
 {% endmacro %}
 
 {% macro TaskOrderActions(task_order) %}
   <div class="task-order-card__buttons">
-    {% if task_order.is_pending %}
-      {{ ViewLink(task_order, text="Edit") }}
-    {% elif task_order.is_active %}
-      {{ ViewLink(task_order, text="Modify") }}
+    {% if task_order.is_draft %}
+      {{ TaskOrderAction(task_order, text="Edit") }}
```
dandds: I think this needs `route='edit'`. This should go to the update screen for the TO.

PR #894, opened 2019-06-12
Style to form

leigh-mil:
## Description
Initial styling updates to the TO Form. There is still some fine tuning that needs to be done, like the LOA fields.

## Pivotal

## Screenshots
Screen Shot 2019-06-12 at 2 15 37 PM
Screen Shot 2019-06-12 at 2 15 48 PM
Screen Shot 2019-06-12 at 2 15 56 PM

### DIFF
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -56,8 +56,10 @@ def update(portfolio_id=None, task_order_id=None):
         else:
             task_order = TaskOrders.create(g.current_user, portfolio_id, **form.data)
 
-        if http_request.args.get('review'):
-            return redirect(url_for("task_orders.review_task_order", task_order_id=task_order.id))
+        if http_request.args.get("review"):
+            return redirect(
+                url_for("task_orders.review_task_order", task_order_id=task_order.id)
+            )
 
         flash("task_order_draft")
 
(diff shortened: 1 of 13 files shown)


templates/components/totals_box.html
```diff
@@ -1,14 +1,19 @@
-{% macro TotalsBox(task_order) -%}
+{% macro TotalsBox(task_order=None, obligated_funds=0, contract_amount=0) -%}
```
montana-mil: why this change? seems like extra lines of code for the same thing, no?

atst/routes/task_orders/new.py
```diff
@@ -8,11 +8,12 @@
 from atst.utils.flash import formatted_flash as flash
 
 
-def render_task_orders_edit(portfolio_id, task_order_id=None, form=None):
+def render_task_orders_edit(portfolio_id=None, task_order_id=None, form=None):
```
dandds: Do we need `portfolio_id` as an arg at all, now?

atst/routes/task_orders/new.py
```diff
@@ -28,18 +29,16 @@ def render_task_orders_edit(portfolio_id, task_order_id=None, form=None):
 
 
 @task_orders_bp.route("/portfolios/<portfolio_id>/task_orders/new")
-@task_orders_bp.route("/portfolios/<portfolio_id>/task_orders/<task_order_id>/edit")
+@task_orders_bp.route("/task_orders/<task_order_id>/edit")
 @user_can(Permissions.CREATE_TASK_ORDER, message="view new task order form")
-def edit(portfolio_id, task_order_id=None):
+def edit(portfolio_id=None, task_order_id=None):
```
dandds: We can just drop the `portfolio_id` kwarg from the route functions.

PR #895, opened 2019-06-12
More realistic seeds

drummond-dds:
* CLIN start/end dates are no longer the same
* Attachments have PDF extension


tests/factories.py
```diff
@@ -261,7 +261,7 @@ class AttachmentFactory(Base):
     class Meta:
         model = Attachment
 
-    filename = factory.Faker("domain_word")
+    filename = factory.Faker("file_name", extension="pdf")
```
dandds: This sounds like it's going overboard, but I think that for each TO we should upload a real sample PDF. That way if someone tries the download link they'll get a file instead of an error. We can use the same fixture file for all the TOs.

PR #912, opened 2019-06-13
Fix validation quirks in Portfolio form

richard-dds:
- Blank fields should be considered invalid, even if they've been changed
- When the form is submitted with errors, fields that were previously valid should still be considered valid
- Show "you've got errors" flash message at the top of the form

This PR includes some changes to our shared Vue components (text_input and options_input). I did a quick peek on some of the other forms, and they look ok to me.

### DIFF
--- a/js/mixins/form.js
+++ b/js/mixins/form.js
@@ -25,8 +25,8 @@ export default {
     },
 
     handleFieldMount: function(event) {
-      const { name, optional } = event
-      this.fields[name] = optional
+      const { name, optional, valid } = event
+      this.fields[name] = optional || valid
     },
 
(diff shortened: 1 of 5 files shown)


js/mixins/form.js
```diff
@@ -25,8 +25,8 @@ export default {
     },
 
     handleFieldMount: function(event) {
-      const { name, optional } = event
-      this.fields[name] = optional
+      const { name, optional, valid } = event
+      this.fields[name] = optional || valid
```
dandds: I thought this might present problems for ensuring optional fields still have to be validated, but it seems like it works.

PR #927, opened 2019-06-17
Remove unused Clock.now

drummond-dds:


dandds:
Will revisit if time allows.

PR #929, opened 2019-06-17
Fix duped alerts

leigh-mil:
## Description
Removes duplicate flash messages for  logout and errors on TO form. 

## Pivotal
https://www.pivotaltracker.com/story/show/166683697

### DIFF
--- a/atst/domain/auth.py
+++ b/atst/domain/auth.py
@@ -1,7 +1,6 @@
 from flask import g, redirect, url_for, session, request
 
 from atst.domain.users import Users
-from atst.utils.flash import formatted_flash
 
 
 UNPROTECTED_ROUTES = [
@@ -61,7 +60,6 @@ def logout():
     if session.get("user_id"):  # pragma: no branch
         del session["user_id"]
         del session["last_login"]
-        formatted_flash("logged_out")
 
 
 def _unprotected_route(request):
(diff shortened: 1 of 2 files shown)


atst/domain/auth.py
```diff
@@ -61,7 +60,6 @@ def logout():
     if session.get("user_id"):  # pragma: no branch
         del session["user_id"]
         del session["last_login"]
-        formatted_flash("logged_out")
```
drummond-dds: I think I had issues with the logout message not displaying when a session timed out. This was added to fix [this Pivotal Tracker ticket](https://www.pivotaltracker.com/story/show/166472937). Can you double check that this doesn't break the previous bug?

drummond-dds: I think I had issues with the logout message not displaying when a session timed out. This was added to fix [this Pivotal Tracker ticket](https://www.pivotaltracker.com/story/show/166472937). Can you double check that this doesn't break the previous bug?

drummond-dds: You can change the `PERMANENT_SESSION_LIFETIME` to something like `5` seconds to test this out in https://github.com/dod-ccpo/atst/blob/master/config/base.ini#L14

leigh-mil: @drummond-dds I just tested it out and it did not cause a regression on the previous bug!

PR #932, opened 2019-06-17
Allow saving TO as draft

drummond-dds:
### Why

As a user filling out the TO upload form,
I want to save even if I don't have all the fields in place,
So that I can go back to my KO or system of record to get more information later

### Acceptance Criteria

```gherkin
Scenario: 
Given I am filling out the TO Upload page,
When I click "Save,"
Then the page saves and allows fields to be empty, but not otherwise invalid
```


alembic/versions/3d3c71b03e98_nullable_fields_for_clins.py
```diff
@@ -0,0 +1,65 @@
...
+               existing_type=sa.DATE(),
+               nullable=True)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('clins', 'start_date',
+               existing_type=sa.DATE(),
+               nullable=False)
```
leigh-mil: If we have to downgrade this after some records have been created with `Null` in the columns, what will happen?

atst/models/task_order.py
```diff
@@ -86,7 +86,7 @@ def is_unsigned(self):
 
     @property
     def has_begun(self):
-        return Clock.today() >= self.start_date
+        return self.start_date and Clock.today() >= self.start_date
 
     @property
     def has_ended(self):
```
leigh-mil: I think you will need to check to see if `self.end_date` exists, like on line 89.

alembic/versions/3d3c71b03e98_nullable_fields_for_clins.py
dandds: I think the best thing to do in these situations is to not have a downgrade function. If there's not a sensible default that can be applied first, we should remove the downgrade function and leave a comment in its place noting that the migration is irreversible. (Or have the downgrade function raise an error to that effect.)

atst/models/task_order.py
drummond-dds: Added!

alembic/versions/3d3c71b03e98_nullable_fields_for_clins.py
drummond-dds: I removed the downgrade migration

atst/routes/task_orders/new.py
```diff
@@ -57,13 +57,14 @@ def update(portfolio_id=None, task_order_id=None):
...
 
         flash("task_order_draft")
 
-        return redirect(url_for("task_orders.edit", task_order_id=task_order.id))
-    else:
-        return render_task_orders_edit(portfolio_id, task_order_id, form), 400
+        if task_order.is_completed:
+            return redirect(url_for("task_orders.edit", task_order_id=task_order.id))
+
+    return render_task_orders_edit(portfolio_id, task_order_id, form), 400
```
dandds: I think the branch logic in here is getting a little confusing to read, but it works for now.

templates/task_orders/edit.html
```diff
@@ -121,6 +121,9 @@
               type="submit"
               formaction="{{ review_action }}"
               tabindex="0"
+              {% if task_order and task_order.is_draft  %}
+              disabled="disabled"
```
dandds: It worked for me the way I'd want on a quick pass, but I wonder if there are edge cases where setting it explicitly here competes with the Vue binding on 127.

PR #938, opened 2019-06-17
Clear checkbox on TO signature modal when it is closed

leigh-mil:
## Description
Updates the TO signature modal so that when you cancel out of the modal, the checkbox is reset.

## Pivotal
https://www.pivotaltracker.com/story/show/166636970

### DIFF
--- a/atst/routes/task_orders/index.py
+++ b/atst/routes/task_orders/index.py
@@ -4,7 +4,7 @@ from . import task_orders_bp
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.domain.portfolios import Portfolios
 from atst.domain.task_orders import TaskOrders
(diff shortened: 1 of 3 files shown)


atst/routes/task_orders/index.py
```diff
@@ -4,7 +4,7 @@
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.domain.portfolios import Portfolios
 from atst.domain.task_orders import TaskOrders
-from atst.forms.task_order import TaskOrderForm, SignatureForm
```
montana-mil: thanks for taking this out!

montana-mil:
oh weird. i wonder if different browsers are handling this in different ways. im going to try it on IE11 and edge

PR #939, opened 2019-06-17
Fix home page icons

leigh-mil:
## Description
Add more padding around home page icons

## Pivotal
https://www.pivotaltracker.com/story/show/166685583

## Screenshot
Screen Shot 2019-06-17 at 4 45 18 PM

### DIFF
--- a/styles/sections/_home.scss
+++ b/styles/sections/_home.scss
@@ -28,8 +28,16 @@
   .links {
     justify-content: flex-start;
 
+    .icon-link {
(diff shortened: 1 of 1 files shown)


dandds:
@leigh-mil can you check if the same change is need in the portfolio header?

leigh-mil:
> @leigh-mil can you check if the same change is need in the portfolio header?

The spacing in the header looks to be the same as in the designs, so I'm going to go ahead and merge!

PR #940, opened 2019-06-18
Fix CLIN type display

dandds:
PT: https://www.pivotaltracker.com/story/show/166765021

Fixes a bug where the CLIN type was not displaying correctly for saved TOs.

The enum for the CLIN type needs to be coerced in the form data so that the `jedi_clin_field` data is the value of the enum option. Previously the `jedi_clin_field` was being populated with a stringified representation of the enum, like `JEDICLINType.JEDI_CLIN_1`.

### DIFF
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -18,8 +18,18 @@ from atst.forms.validators import FileLength
 from atst.utils.localization import translate
 
 
+def coerce_enum(enum_inst):
+    if getattr(enum_inst, "value", None):
+        return enum_inst.value
+    else:
+        return enum_inst
+
+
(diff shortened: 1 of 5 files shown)


atst/forms/task_order.py
```diff
@@ -18,8 +18,18 @@
 from atst.utils.localization import translate
 
 
+def coerce_enum(enum_inst):
```
drummond-dds: Do you think this should belong somewhere in utils?

dandds: I think we should move it if we wind up needing it somewhere else. Since it's only used in one place right now, I like it living close to home.

PR #956, opened 2019-06-20
Delete some unused code

drummond-dds:


js/index.js
```diff
@@ -2,7 +2,6 @@ import 'svg-innerhtml'
 import 'babel-polyfill'
 import ally from 'ally.js'
 
-import classes from '../styles/atat.scss'
```
dandds: We chatted, but turns out we need this! Can you add it back, and maybe add a comment about why it's there? 🙏

dandds:
We'll revisit this if we have time.

PR #959, opened 2019-07-10
Fix tests that were broken by a fixture CRL expiring.

dandds:
Adjust the broken tests to use our dynamic fixtures for PKI files. Some
tests still rely on these fixtures, but this is a minimal patch to get
the test suite passing again. Eventually all tests should use the pytest
fixtures.


tests/domain/authnid/test_crl.py
```diff
@@ -62,16 +62,27 @@ def test_can_build_crl_list_with_missing_crls():
...
+    ca_key,
+    ca_file,
+    crl_file,
+    rsa_key,
+    make_x509,
+    make_crl,
+    serialize_pki_object_to_disk,
+):
+    good_cert = make_x509(rsa_key(), signer_key=ca_key, cn="luke")
+    bad_cert = make_x509(rsa_key(), signer_key=ca_key, cn="darth")
```
leigh-mil: 😆

PR #964, opened 2019-07-12
Push Docker Images to CSP registries

dandds:
This adds CircleCI config to push docker images from ATAT to the CSP registries.

These changes are more thoroughly described in the commits, but it also:
- Fixes a mistake in the Dockerfile
- Moves the Dockerfile to the repo root
- Removes the submodule and hard-copies the files

## Confirming this Works
The CircleCI config contains filters to ensure we only push images when new commits are merged into master. I amended my last commit to add this filter. There is a scrap version of my branch that has the last commit before it was amended [here](https://github.com/dod-ccpo/atst/tree/push-container-scrap). The commit starts with `884d`. The successful CircleCI job for that commit can be found [here](https://circleci.com/workflow-run/0888aeda-52a7-49c7-ae3f-1d34dccd657b). You can see the resulting image in both the CSP registries in the screenshots below:

Azure:
Screen Shot 2019-07-14 at 4 19 06 PM

AWS (redacted the registry ID):
Screen Shot 2019-07-14 at 4 18 28 PM

If you're curious, I can demo in real-time.


dandds:
> Remove submodule clone part of the config (line 28 on circleci/config.yml)

@mk-dds Done!

PR #966, opened 2019-07-12
Vue testing spike

tomdds:
## Description
This PR includes a proof of concept to render jinja templates to files that can be imported and tested in jest. The python script (`scripts/render_vue_component.py`) will currently render out the checkbox input and upload input templates, including an error treatment for upload input. Currently these solutions and helpers are specifically for fairly simple components, and may need revising with more complex forms that provide multiple props to multiple vue components.

 ## Pivotal Link
https://www.pivotaltracker.com/story/show/167259934


script/render_vue_component.py
```diff
@@ -0,0 +1,50 @@
...
+from wtforms.widgets import CheckboxInput, FileInput
+from wtforms.fields import StringField, FileField
+from wtforms.validators import InputRequired
+from wtforms import Form
+
+from atst.filters import iconSvg
+
+env = Environment(loader=FileSystemLoader("templates"))
+
+env.filters["iconSvg"] = iconSvg
```
dandds: If we were working in the pytest context here, we could use the `app` fixture, which would have a preloaded Jinja env as `app.jinja_env` (with all of our filters applied already), which would mean less setup in this script. We'd also have access to all the other test fixtures. 

You could move this file right into `tests/` and then it will inherit the test context, but (importantly) the test runner will ignore it for a normal test run. You can invoke explicitly with `pytest tests/render_vue_component.py`, so there could be a bash script in the `script/` dir that just does that.

If we do this, all the pieces of functionality we repeat would become pytest fixtures:
```
@pytest.fixture
def env(app):
    return app.jinja_env
```
If you load up a macro that gets used repeatedly, that can also be a fixture:
```
@pytest.fixture
def upload_input_macro(env):
    upload_template = env.get_template("components/upload_input.html")
    up_macro = getattr(upload_template.module, "UploadInput")
```
The every step that writes out a fragment becomes a pytest test:
```
def test_make_upload_input_template(upload_input_macro)
    [do the work here]
```
I know these aren't tests in themselves, but using pytest would take away some of the template setup (especially as we try to render component hierarchies) and give us a scaffold for organizing the code.

PR #970, opened 2019-07-15
Force lodash resolution to higher patch version

tomdds:
## Description
This forces our dependencies to use a version of lodash that includes a fix for this vulnerability: https://snyk.io/vuln/SNYK-JS-LODASH-73638

## Pivotal
https://www.pivotaltracker.com/story/show/167295195

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,9 @@
     "prettier": "^1.16.1",
     "vue-template-compiler": "2.5.15"
   },
+  "resolutions": {
+    "parcel-bundler/**/lodash": "^4.17.13"
+  },
   "browserslist": [
     "last 3 version",
     "> 5%",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -39,6 +39,9 @@
     "prettier": "^1.16.1",
     "vue-template-compiler": "2.5.15"
   },
+  "resolutions": {
+    "parcel-bundler/**/lodash": "^4.17.13"
+  },
```
dandds: I didn't know you could do this, nice.

tomdds: I learned this as part of the investigation once I realized we weren't using it directly: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-how-to-use-it

PR #973, opened 2019-07-16
Split TO Form into 5 Steps

leigh-mil:
## Description
Moves the TO form into multiple steps. Each step is now its own template with its own route and each page with form inputs posts to its own route. The old task order edit and update routes and template have been deleted. 

Some tests have been skipped instead of deleted because they will need to be reimplemented later.
There is also an issue when a page is valid but no changes have been made to the form, so the 'Next' button is inactive. This issue will be fixed in a follow up PR.

## Pivotal


js/components/forms/base_form.js
```diff
@@ -22,9 +25,18 @@ export default {
...
     uploadinput,
   },
   mixins: [FormMixin],
+  directives: {
+    sticky: {
+      inserted: el => {
+        stickybits(el)
+      },
+    },
+  },
```
dandds: Remind me what this does? Do we really want it in the base form component?

tests/routes/task_orders/test_new.py
```diff
@@ -59,55 +94,116 @@ def test_task_orders_update(client, user_session, portfolio):
...
+def test_task_orders_confirm_signature(client, user_session, task_order):
+    user_session(task_order.creator)
+    response = client.get(
+        url_for("task_orders.confirm_signature", task_order_id=task_order.id)
+    )
+    assert response.status_code == 200
+
+
+def test_task_orders_new_flow():
+    pass
```
dandds: Do you need this? I'd either delete or add a skip with an explanatory note about what it's supposed to do.

atst/routes/task_orders/new.py
```diff
@@ -57,18 +44,95 @@ def update(portfolio_id=None, task_order_id=None):
...
+@task_orders_bp.route("/task_orders/<task_order_id>/step_2")
+@user_can(Permissions.CREATE_TASK_ORDER, message="view new task order form")
+def add_number(task_order_id):
+    return render_task_orders_edit(
+        "task_orders/step_2.html", task_order_id=task_order_id
+    )
+
+
+@task_orders_bp.route("/task_orders/<task_order_id>/step_2", methods=["POST"])
+@user_can(Permissions.CREATE_TASK_ORDER, message="view new task order form")
```
dandds: Small thing, but can you update the `message` kwargs on the `user_can` decorators for all these functions? They don't have to be very detailed, but they should make a distinction between viewing a TO form page and updating a TO.

tests/routes/task_orders/test_new.py
leigh-mil: Woops! Forgot to delete that out when I was cleaning up

js/components/forms/base_form.js
leigh-mil: It's needed there because the Sticky header needs it to be sticky and the sticky header has to be nested in the BaseForm in order for the submit buttons to work properly!
Would moving that bit into a mix in make more sense?

mk-dds:
Minor bug: I clicked "Previous" after completing step 2, and "Next" is disabled. If I click out of the box, the Next button becomes available again. 

Screen Shot 2019-07-18 at 9 39 13 AM

atst/routes/task_orders/new.py
```diff
@@ -57,18 +44,95 @@ def update(portfolio_id=None, task_order_id=None):
...
+    next_page = "task_orders.add_number"
+    current_template = "task_orders/step_1.html"
+
+    return update_task_order(
+        form_data,
+        next_page,
+        current_template,
+        portfolio_id=portfolio_id,
+        task_order_id=task_order_id,
+    )
```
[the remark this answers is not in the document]
leigh-mil: To recap our irl convo -- In another PR, I will lock down the pages for step 4 & 5 so a TO can't be reviewed or submitted when it is incomplete.

mk-dds:
Actually I think your PR #976 will fix the bug I pointed out! Feel free to ignore for now if you're planning to return to it later. ^

PR #976, opened 2019-07-17
Activate next buttons on TO form

leigh-mil:
## Description
Fixes an issue when a page is valid but no changes have been made to the form the 'Next' button is disabled. 
This was solved by adding the computed property `canSave` to the form mixin, which is used to set whether or not the Next button is disabled. Also updated the checkbox input vue component to emit an event on mount and whether or not it is valid on field-change, the same as has been implemented on other input vue components.

## Pivotal

### DIFF
--- a/js/mixins/form.js
+++ b/js/mixins/form.js
@@ -1,6 +1,10 @@
 export default {
   props: {
     initialSelectedSection: String,
+    hasChanges: {
+      type: Boolean,
+      default: false,
+    },
   },
 
   mounted: function() {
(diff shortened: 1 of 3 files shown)


js/mixins/form.js
```diff
@@ -44,6 +44,18 @@ export default {
     },
   },
 
+  computed: {
+    canSave: function() {
+      if (!this.invalid) {
+        return true
+      } else if (this.changed && !this.invalid) {
```
mk-dds: Could we use `formValid` on 49 and 51 instead to avoid the double negative?

PR #979, opened 2019-07-18
User can't access TO form review or signature pages with incomplete TO

leigh-mil:
## Description
Adds a check to a user cannot access the TO form review or signature pages unless the TO is complete. This will prevent the data from getting into a weird state and causing issues. 

## Pivotal

### DIFF
--- a/atst/domain/task_orders.py
+++ b/atst/domain/task_orders.py
@@ -1,5 +1,4 @@
 import datetime
-from flask import current_app as app
 
 from atst.database import db
 from atst.models.clin import CLIN
@@ -10,10 +9,6 @@ from . import BaseDomainClass
 class TaskOrders(BaseDomainClass):
     model = TaskOrder
     resource_name = "task_order"
-
-    SECTIONS = {"app_info": ["portfolio_name"], "funding": [], "oversight": []}
-
-    UNCLASSIFIED_FUNDING = []
 
     @classmethod
     def create(cls, creator, portfolio_id, number, clins, pdf):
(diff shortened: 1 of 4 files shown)


atst/domain/task_orders.py
```diff
@@ -70,43 +65,6 @@ def create_clins(cls, task_order_id, clin_list):
...
-
-        return True
-
-    @classmethod
-    def mission_owner_sections(cls):
-        section_list = TaskOrders.SECTIONS
-        if not app.config.get("CLASSIFIED"):
-            section_list["funding"] = TaskOrders.UNCLASSIFIED_FUNDING
-        return section_list
-
```
dandds: Good catch.

atst/routes/task_orders/new.py
```diff
@@ -128,6 +129,11 @@ def submit_form_step_three_add_clins(task_order_id):
 @task_orders_bp.route("/task_orders/<task_order_id>/form/step_4")
 @user_can(Permissions.CREATE_TASK_ORDER, message="view task order form")
 def form_step_four_review(task_order_id):
+    task_order = TaskOrders.get(task_order_id)
+
+    if task_order.is_completed == False:
```
dandds: I wonder if it matters that the `is_completed` property only checks that there _are_ CLINs. What if the CLINs themselves are incomplete? (I don't know the answer to this; probably worth a little discussion.)

leigh-mil: I think I am going to solve this issue in a separate PR by splitting each page into its own form, that way we can validate through wtforms that no fields are empty.

PR #981, opened 2019-07-18
Form styling updates

leigh-mil:
## Description
Made a few styling changes that apply to the to form base file and added a macro for each step's heading.

## Pivotal
https://www.pivotaltracker.com/story/show/167224257

### DIFF
--- a/templates/task_orders/form_header.html
+++ b/templates/task_orders/form_header.html
@@ -13,9 +13,7 @@
       </div>
     </div>
     <div class="form-row">
-      <div>
-        {{ description }}
-      </div>
+      {{ description }}
     </div>
   </div>
 {% endmacro %}
--- a/templates/task_orders/form_header.html
+++ b/templates/task_orders/form_header.html
@@ -1,0 +1,21 @@
+{% macro TOFormStepHeader(title, description, to_number=None) %}
+  <div class="column">
+    <div class="form-row">
+      <div class="form-col form-col--two-thirds">
+        <div class="h2">
+          {{ title }}
+        </div>
+      </div>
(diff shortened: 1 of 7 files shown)


templates/task_orders/form_header.html
```diff
@@ -0,0 +1,21 @@
...
+        </div>
+      </div>
+      <div class="form-col form-col--third">
+        {% if to_number %}
+          <strong>Task Order Number:</strong> {{ to_number }}
+        {% endif %}
+      </div>
+    </div>
+    <div class="form-row">
+      <div>
```
mk-dds: Are we able to remove the double `<div>` here? I don't see the second container doing anything

templates/task_orders/step_5.html
```diff
@@ -2,21 +2,19 @@
...
 {% set action = url_for("task_orders.submit_task_order", task_order_id=task_order_id) %}
 {% set next_button_text = "Next: Confirm & Submit" %}
 {% set previous_button_link = url_for("task_orders.form_step_four_review", task_order_id=task_order_id) %}
 {% set step = "5" %}
 
 {% block to_builder_form_field %}
-  <div>
-    <h1>Signature confirmation: <em>Task Order #{{task_order.number}}</em></h1>
-  </div>
+  {{ TOFormStepHeader("Signature confirmation", "Lorem ipsem", task_order_number) }}
```
mk-dds: I realize this was not a change you made - but is there a reason this one is not using the translation file?

leigh-mil: No reason! But I am working on step 5 now, so I'll move everything into translations in that PR.

PR #982, opened 2019-07-19
TO form review styling

leigh-mil:
## Description
Updates the styling on the review page. 
There are a few design questions I have about how to list multiple CLINS and the text for CLINS 2-4, but right now I kept the display as it was the previous version of the review page.

## Pivotal
https://www.pivotaltracker.com/story/show/167233673

## Screenshot
Screen Shot 2019-07-19 at 12 00 45 PM

### DIFF
--- a/templates/fragments/task_order_review.html
+++ b/templates/fragments/task_order_review.html
@@ -4,82 +4,79 @@
 
 
 <div class="task-order">
-  {{ SemiCollapsibleText() }}
-
-  <hr>
-
-  <div class="h1">
-    {{ "task_orders.review.review_your_task_order" | translate }}
-  </div>
-  <p>
-    {{ "task_orders.review.check_paragraph" | translate }}
-  </p>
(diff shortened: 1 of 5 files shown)


templates/fragments/task_order_review.html
```diff
@@ -3,83 +3,78 @@
...
+      <div class="h2">
+        {{ "task_orders.review.review_your_task_order" | translate }}
       </div>
-      <div>{{task_order.number}}</div>
-
-      <hr>
 
       <div class="h3">
-        {{ "task_orders.review.funding_summary" | translate }}
+        Approved Task Order
```
mk-dds: Is there a reason this one is not translated but the other 2 subheaders on the page are?

tomdds:
@leigh-mil looks good except when I'm running it, I'm seeing almost all the values in the table be cut off 
Screen Shot 2019-07-24 at 10 36 00 AM

PR #988, opened 2019-07-30
Save draft TO on cancel

richard-dds:
https://www.pivotaltracker.com/story/show/167224230

### DIFF
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -53,7 +53,11 @@ class CLINForm(FlaskForm):
 
     def validate(self, *args, **kwargs):
         valid = super().validate(*args, **kwargs)
-        if self.start_date.data > self.end_date.data:
+        if (
+            self.start_date.data
+            and self.end_date.data
+            and self.start_date.data > self.end_date.data
+        ):
             self.start_date.errors.append(
                 translate("forms.task_order.start_date_error")
             )
--- a/atst/routes/task_orders/new.py
+++ b/atst/routes/task_orders/new.py
@@ -21,7 +21,10 @@ def render_task_orders_edit(template, portfolio_id=None, task_order_id=None, for
     else:
         render_args["form"] = form or TaskOrderForm()
 
-    render_args["cancel_url"] = url_for(
+    render_args["cancel_save_url"] = url_for(
+        "task_orders.portfolio_funding", portfolio_id=portfolio_id, save=True
+    )
+    render_args["cancel_discard_url"] = url_for(
         "task_orders.portfolio_funding", portfolio_id=portfolio_id
     )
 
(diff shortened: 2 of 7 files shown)


templates/task_orders/builder_base.html
```diff
@@ -26,13 +27,21 @@
           {% endif %}
 
           <a
-            href="{{ cancel_url }}"
+            v-on:click="openModal('cancel')"
```
leigh-mil: Is `cancel_url` used at all anymore? If not we should remove it from the render args in `render_task_orders_edit()`

atst/routes/task_orders/new.py
```diff
@@ -86,6 +89,33 @@ def submit_form_step_one_add_pdf(portfolio_id=None, task_order_id=None):
     )
 
 
+@task_orders_bp.route("/task_orders/<task_order_id>/form/cancel", methods=["POST"])
+@user_can(Permissions.CREATE_TASK_ORDER, message="cancel task order form")
+def cancel_edit(task_order_id):
+    save = http_request.args.get("save", False)
```
leigh-mil: There needs to be a second route here that uses `portfolio_id` for when you generate a new TO. Right now when I try to create a new TO I'm getting an error because it can't build this route with out a `task_order_id`

richard-dds: @leigh-mil I think I addressed this. A unit test and some manual testing seems to confirm.

templates/task_orders/builder_base.html
```diff
@@ -26,13 +27,23 @@
...
           </a>
         </span>
       {% endcall %}
 
+      {% call Modal(name='cancel', dismissable=True) %}
+        <div class="task-order__modal-cancel">
+          <h1>Do you want to save this draft?</h1>
+          <div class="task-order__modal-cancel_buttons">
+            <button formaction="{{ cancel_discard_url }}" class="usa-button usa-button-primary" type="submit">No, delete it</button>
+            <button formaction="{{ cancel_save_url }}" class="usa-button usa-button-primary" type="submit">Yes, save for later</button>
```
dandds: When I click this with validation errors, I get the flash message at the top of the TO list page. We probably need to pop that message off the flash stack before it appears.

richard-dds:
@leigh-mil One of my tests was failing because `CLINForm.validate()` was being called with a start_date and end_date of None. Could you check my fix (https://github.com/dod-ccpo/atst/pull/988/commits/1cd35445eab0cc7c60bd036dd2c92da264141dc0) and tell me if that looks ok?

atst/forms/task_order.py
```diff
@@ -53,7 +53,11 @@ class CLINForm(FlaskForm):
 
     def validate(self, *args, **kwargs):
         valid = super().validate(*args, **kwargs)
-        if self.start_date.data > self.end_date.data:
+        if (
+            self.start_date.data
+            and self.end_date.data
+            and self.start_date.data > self.end_date.data
+        ):
```
leigh-mil: @richard-dds Your fix looks great! thanks!

PR #999, opened 2019-08-05
Page for CCPO users list

leigh-mil:
## Description
Creates a page where a CCPO user can view a list of all of the CCPO users.  Adds new permissions to the `ATAT_PERMISSION_SETS` that correspond with the ability to view/edit/add/remove CCPO users.

When testing locally, you will need to reset your database so Sam (the CCPO user) gets the new CCPO permissions.

## Pivotal
https://www.pivotaltracker.com/story/show/167498650

## Screenshot
Screen Shot 2019-08-05 at 4 19 22 PM

### DIFF
--- a/atst/domain/permission_sets.py
+++ b/atst/domain/permission_sets.py
@@ -17,7 +17,7 @@ class PermissionSets(object):
     EDIT_PORTFOLIO_ADMIN = "edit_portfolio_admin"
     PORTFOLIO_POC = "portfolio_poc"
     VIEW_AUDIT_LOG = "view_audit_log"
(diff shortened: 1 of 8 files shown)


atst/domain/permission_sets.py
```diff
@@ -56,7 +57,17 @@ def get_many(cls, perms_set_names):
         "display_name": "View Audit Log",
         "description": "",
         "permissions": [Permissions.VIEW_AUDIT_LOG],
-    }
+    },
+    {
+        "name": PermissionSets.EDIT_CCPO_USERS,
```
dandds: I might call this MANAGE_CCPO_USERS, since it encompasses view and edit perms.

PR #1011, opened 2019-08-08
Remove CCPO user

leigh-mil:
## Description
Adds in functionality to revoke a user's CCPO permissions. Recording changing users CCPO perms in the activity log will be in an upcoming PR. 

## Pivotal
https://www.pivotaltracker.com/story/show/167716449

## Screenshots
Screen Shot 2019-08-08 at 2 48 56 PM
Screen Shot 2019-08-08 at 2 49 05 PM

### DIFF
--- a/atst/routes/ccpo.py
+++ b/atst/routes/ccpo.py
@@ -26,9 +26,7 @@ def activity_history():
 @user_can(Permissions.VIEW_CCPO_USER, message="view ccpo users")
 def users():
     users = Users.get_ccpo_users()
-    data = {}
-    for user in users:
-        data[user] = CCPOUserForm(obj=user)
+    data = [(user, CCPOUserForm(obj=user)) for user in users]
     return render_template("ccpo/users.html", data=data)
 
 
--- a/atst/routes/ccpo.py
+++ b/atst/routes/ccpo.py
@@ -56,3 +56,12 @@ def confirm_new_user():
     Users.give_ccpo_perms(user)
     flash("ccpo_user_added", user_name=user.full_name)
     return redirect(url_for("ccpo.users"))
+
+
+@bp.route("/ccpo-users/remove-access/<user_id>", methods=["POST"])
+@user_can(Permissions.DELETE_CCPO_USER, message="remove ccpo user")
+def remove_access(user_id):
+    user = Users.get(user_id)
+    # update user to remove perms
+    # flash alert to confirm removing ccpo perms
+    return redirect(url_for("ccpo.users"))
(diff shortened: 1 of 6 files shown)


atst/routes/ccpo.py
```diff
@@ -26,7 +26,10 @@ def activity_history():
 @user_can(Permissions.VIEW_CCPO_USER, message="view ccpo users")
 def users():
     users = Users.get_ccpo_users()
-    return render_template("ccpo/users.html", users=users)
+    data = {}
+    for user in users:
+        data[user] = CCPOUserForm(obj=user)
```
dandds: Do you need to use a dictionary here? It's generally best to avoid using mutable objects like `user` as dictionary keys. Could you do a nested list? `data = [[user, CCPOUserForm(obj=user)] for user in users]`

dandds: Or a tuple might be more Pythonic: `data = [(user, CCPOUserForm(obj=user)) for user in users]`

atst/routes/ccpo.py
```diff
@@ -56,3 +59,12 @@ def confirm_new_user():
     Users.give_ccpo_perms(user)
     flash("ccpo_user_added", user_name=user.full_name)
     return redirect(url_for("ccpo.users"))
+
+
+@bp.route("/ccpo-users/remove-access/<user_id>", methods=["POST"])
+@user_can(Permissions.DELETE_CCPO_USER, message="remove ccpo user")
+def remove_access(user_id):
+    user = Users.get(user_id)
+    Users.revoke_ccpo_perms(user)
```
dandds: If you call this method on someone who isn't a CCPO, nothing happens, right?

atst/routes/ccpo.py
```diff
@@ -26,7 +26,10 @@ def activity_history():
 @user_can(Permissions.VIEW_CCPO_USER, message="view ccpo users")
 def users():
     users = Users.get_ccpo_users()
-    return render_template("ccpo/users.html", users=users)
+    data = {}
+    for user in users:
+        data[user] = CCPOUserForm(obj=user)
+    return render_template("ccpo/users.html", data=data)
```
dandds: Can we call it something more specific than `data`?

atst/routes/ccpo.py
leigh-mil: Correct! The function replaces the users `permission_sets` with an empty array, so it doesn't actually matter if the user currently has ccpo perms or not.

dandds: :+1: Someone removing CCPO perms from a user who doesn't have them is a pretty benign  edge case, so I don't think we need to handle it yet.

PR #1012, opened 2019-08-08
Apply config changes for CSP file uploads.

dandds:
This applies configuration changes for the Flask app and adds changes to the Dockerfile so that the build can make a CSP-specific JS bundle. It adds `write_dotenv` script that creates the appropriate `.env` file for the `parcel` bundler depending on how the `CSP` environment variable is set.

- Configure K8s environment variables for Flask CSP usage
- Supply default CSP config setting to Flask app
- Declare the CSP arg in the Dockerfile
- Supply extra Docker build args to CD
- Fix top-level reference to boto3 in file_upload module
- Add back missing sample NGINX config for docker-compose build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,6 +129,7 @@ workflows:
           requires:
             - app_setup
       - azure-acr/build_and_push_image:
+          extra-build-args: "--build-arg CSP=azure"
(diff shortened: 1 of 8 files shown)


dandds:
@tomdds Could you glance at this again? Had to fix a conflict.

PR #1013, opened 2019-08-08
Update TO form cancel button behavior

richard-dds:
Think clarified that the intended behavior of the "No, delete it" option on the cancel button modal is to actually delete the current draft TO.

### DIFF
--- a/alembic/versions/fda6bd7e1b65_clin_delete_cascade.py
+++ b/alembic/versions/fda6bd7e1b65_clin_delete_cascade.py
@@ -1,0 +1,26 @@
+"""clin delete cascade
+
+Revision ID: fda6bd7e1b65
+Revises: e0c6eb21771f
+Create Date: 2019-08-07 16:37:02.451277
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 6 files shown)


alembic/versions/fda6bd7e1b65_clin_delete_cascade.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = 'fda6bd7e1b65'
+down_revision = 'e0c6eb21771f'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_constraint("clins_task_order_id_fkey", "clins")
+    op.create_foreign_key("clins_task_order_id_fkey", "clins", "task_orders", ["task_order_id"], ["id"], ondelete="cascade")
```
richard-dds: I'm curious about how people feel about adding this `ON DELETE CASCADE` to this foreign key in the database when the ORM will handle the cascading deletes with or without it. I feel like it may be good idea to have it to enforce that rule in the database, even if it is redundant.

tomdds: I think it's a good idea to have the parity between the ORM and DB if possible.

PR #1016, opened 2019-08-09
Scripts for finding accidental secrets in the repo.

dandds:
This adds the following:
- A detect-secrets dependency and a related script
  (`script/detect_secrets`) to find and alert developers to secrets
  added to the code. By default, the script will search staged and new,
  unstaged files. It can optionally search only staged files.
- A whitelist, `.secrets.baseline`, that tracks instances of secrets or
  false positives already in the repo.
- Modifies `script/test` to detect secrets as part of the test suite.
- Updates to the README regarding the use of detect-secrets.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,187 @@
+{
+  "exclude": {
+    "files": null,
+    "lines": null
+  },
+  "generated_at": "2019-08-13T09:59:21Z",
+  "plugins_used": [
+    {
+      "base64_limit": 4.5,
+      "name": "Base64HighEntropyString"
+    },
+    {
+      "name": "BasicAuthDetector"
+    },
+    {
+      "hex_limit": 3,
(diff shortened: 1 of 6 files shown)


leigh-mil:
lgtm!

.secrets.baseline
```diff
@@ -0,0 +1,199 @@
...
+    },
+    {
+      "name": "PrivateKeyDetector"
+    },
+    {
+      "name": "SlackDetector"
+    },
+    {
+      "name": "StripeDetector"
+    }
```
tomdds: Are any of these extraneous defaults that we should remove (e.g. `StripeDetector` `SlackDetector`)

dandds: I was going to cast a wide net, but the vendor-specific plugins don't make a difference in the baseline output. I removed them and amended the commit.

PR #1018, opened 2019-08-09
Allow creation of new users in dev mode

tomdds:
## Description

Makes 2 changes to support creating new users in dev mode:
* Adds a new route, `/dev-new-user` which takes `first_name`, `last_name`, and `dod_id` query parameters (all required), creates the new user and logs you in as that user.
* Adds the ability to pass a `dod_id` query parameter to `/login-dev` to login as a user if they exist in the db

## Pivotal
https://www.pivotaltracker.com/story/show/167785247

### DIFF
--- a/atst/routes/dev.py
+++ b/atst/routes/dev.py
@@ -11,11 +11,13 @@ from flask import (
 import pendulum
 
 from . import redirect_after_login_url, current_user_setup
+from atst.domain.exceptions import AlreadyExistsError, NotFoundError
 from atst.domain.users import Users
 from atst.domain.permission_sets import PermissionSets
 from atst.forms.data import SERVICE_BRANCHES
 from atst.queue import queue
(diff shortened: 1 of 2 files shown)


atst/routes/dev.py
```diff
@@ -110,31 +112,75 @@ def random_service_branch():
...
+
+    new_user = {
+        "first_name": first_name,
+        "last_name": last_name,
+        "email": "{}@example.com".format(first_name),
+        "service_branch": random_service_branch(),
+        "phone_number": "1234567890",
+        "citizenship": "United States",
+        "designation": "Military",
+        "date_latest_training": pendulum.date(2018, 1, 1),
```
dandds: I think we should limit this to `first_name` and `last_name`. Users with an incomplete profile are directed to complete their profile first, and that's the flow we want to test.

PR #1019, opened 2019-08-13
Add validations and error states for TO upload form

richard-dds:
Adds validations for both hidden fields of the TO file upload form (filename and object_name). Also adds an error state to the form when there's an error uploading the file to the CSP.

I threw in a small fix to `scripts/seed_sample.py`.

Screen Shot 2019-08-13 at 1 40 12 PM

### DIFF
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -74,7 +74,9 @@ class AttachmentForm(BaseForm):
     object_name = HiddenField(
         id="attachment_object_name",
         validators=[
-            Length(max=40, message=translate("forms.attachment.object_name.length_error"))
+            Length(
+                max=40, message=translate("forms.attachment.object_name.length_error")
+            )
         ],
     )
     accept = ".pdf,application/pdf"
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -71,7 +71,12 @@ class AttachmentForm(BaseForm):
             Length(max=100, message="Filename may be no longer than 100 characters.")
         ],
     )
-    object_name = HiddenField(id="attachment_object_name", validators=[Length(max=40)])
+    object_name = HiddenField(
+        id="attachment_object_name",
+        validators=[
+            Length(max=40, message="Object name may be no longer than 40 characters.")
+        ],
+    )
     accept = ".pdf,application/pdf"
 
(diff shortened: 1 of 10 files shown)


richard-dds:
@tomdds Looks like they're passing!

atst/forms/task_order.py
```diff
@@ -65,10 +65,23 @@ def validate(self, *args, **kwargs):
 
 
 class AttachmentForm(BaseForm):
-    filename = HiddenField(id="attachment_filename")
-    object_name = HiddenField(id="attachment_object_name")
+    filename = HiddenField(
+        id="attachment_filename",
+        validators=[
+            Length(max=100, message="Filename may be no longer than 100 characters.")
```
dandds: Can we put both messages in the translations file, too?

script/seed_sample.py
```diff
@@ -167,20 +168,19 @@ def add_task_orders_to_portfolio(portfolio):
     yesterday = today - timedelta(days=1)
     five_days = timedelta(days=5)
 
-    with open("tests/fixtures/sample.pdf", "rb") as fp:
-        pdf = FileStorage(fp, content_type="application/pdf")
+    pdf = {"filename": "sample_task_order.pdf", "object_name": str(uuid4())}
```
dandds: Since we only make the UUID once here, I was curious why we don't get an error for the unique constraint on `attachments.object_name` in the database when we attach the PDF to multiple TOs. It looks like it's because we do `Attachment.get_or_create` in the pdf setter method in the `TaskOrder` model. Why are we doing `get_or_create` there instead of just `create`? I imagine it's to handle the case where someone is just viewing a previously submitted form and submits with the same file info, but then how do we prevent object_name collisions like we have here in the sample script? (I'm just wondering out loud.)

richard-dds: @dandds That's a good call. I'm not sure why we started using `get_or_create` here. It's just a historical thing that I didn't think to change. I'll make this uuid unique, and I'll peruse the codebase and see if making it a `.create()` call would be an issue.

dandds: Doesn't need to happen on this PR, but I wonder if we need a more complex db constraint so that every task order has to have a unique object_name on its associated attachment.

PR #1023, opened 2019-08-14
Env roles bug fix

leigh-mil:
## Description
There was an issue when a new member was added to an application -- environment roles were being created with the role `No Access`, but an env role should be created if the user does not have access to the env. This was happening because `No Access` was not being properly filtered and converted to `None` by the form.
I fixed the issue by moving the data property into the EnvironmentForm that is used in the NewForm (creates a new app member), so now the form properly handles the value `No Access`.

## Pivotal
https://www.pivotaltracker.com/story/show/165793573

### DIFF
--- a/atst/forms/team.py
+++ b/atst/forms/team.py
@@ -3,7 +3,7 @@ from wtforms.fields import FormField, FieldList, HiddenField, RadioField, String
 from wtforms.validators import Required
 
 from .application_member import EnvironmentForm as BaseEnvironmentForm
-from .data import ENV_ROLES, ENV_ROLE_NO_ACCESS as NO_ACCESS
+from .data import ENV_ROLES
 from .forms import BaseForm
 from atst.forms.fields import SelectField
 from atst.domain.permission_sets import PermissionSets
@@ -17,13 +17,6 @@ class EnvironmentForm(BaseEnvironmentForm):
         default=None,
         filters=[lambda x: None if x == "None" else x],
     )
-
-    @property
-    def data(self):
-        _data = super().data
-        if "role" in _data and _data["role"] == NO_ACCESS:
-            _data["role"] = None
-        return _data
 
 
 class PermissionsForm(FlaskForm):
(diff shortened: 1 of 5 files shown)


tests/models/test_environments.py
```diff
@@ -42,3 +42,31 @@ def test_audit_event_for_environment_deletion(session):
...
+    )
+
+    EnvironmentRoleFactory.create(
+        application_role=application_role_dev,
+        environment=dev_environment,
+        role=CSPRole.BASIC_ACCESS.value,
+    )
+
+    assert developer in dev_environment.users
+    assert user not in dev_environment.users
```
dandds: I don't think I understand this test. The problem was that we were creating `EnvironmentRole` instances where `role` was set to `No Access`, but it looks likes this test just asserts that someone isn't in an environment's user list if we don't make a role for them.

tests/models/test_environments.py
```diff
@@ -42,3 +42,31 @@ def test_audit_event_for_environment_deletion(session):
...
+
+def test_users_does_not_include_no_access():
+    owner = UserFactory.create()
+    developer = UserFactory.create()
+    user = UserFactory.create()
+
+    portfolio = PortfolioFactory.create(owner=owner)
+    application = Applications.create(
+        portfolio, "my test application", "It's mine.", ["dev", "staging", "prod"]
+    )
```
dandds: Avoid using domain methods in tests, if you can. It makes the tests fragile.

atst/forms/team.py
```diff
@@ -18,13 +18,6 @@ class EnvironmentForm(BaseEnvironmentForm):
         filters=[lambda x: None if x == "None" else x],
     )
 
-    @property
-    def data(self):
-        _data = super().data
-        if "role" in _data and _data["role"] == NO_ACCESS:
-            _data["role"] = None
-        return _data
-
```
dandds: Are we using this version of `EnvironmentForm` anywhere now? Can we delete it?

leigh-mil: We use that form as part of the MemberForm (line 59)

tests/models/test_environments.py
leigh-mil: Woops, I meant to delete this! It is one of the tests I wrote when I was trying to figure out where things were going wrong.

PR #1024, opened 2019-08-15
Application members should not see deleted applications.

dandds:
This updates the `Portfolios.for_user` method to screen out deleted
ApplicationRole entities. For extra assurance, we also mark application
roles as disabled when they are deleted.

PT story: https://www.pivotaltracker.com/story/show/167299614


leigh-mil:
Looks good to me!

PR #1025, opened 2019-08-15
Center all modals

leigh-mil:
## Description
Fixes a bug where modals were appearing at the top of the screen.

## Pivotal
https://www.pivotaltracker.com/story/show/167854946

## Screenshot
Screen Shot 2019-08-15 at 11 03 43 AM
Screen Shot 2019-08-15 at 11 03 50 AM


rthoward:
I haven't checked all of the modals on the site since I don't know where to find them all, but from what I can see this looks great!

PR #1029, opened 2019-08-16
Remove validation icons on app settings and portfolio admin pages

leigh-mil:
## Description
Validation icons were appearing on the app settings and portfolio admin pages even if changes were not made to the name. This pr fixes that by adding in a check when the vue components mounts, so that if the form is inset into a page, it will not display the valid icon on mount.

## Pivotal
https://www.pivotaltracker.com/story/show/167929495

## Screenshots
Screen Shot 2019-08-16 at 9 25 00 AM
Screen Shot 2019-08-16 at 9 25 06 AM

### DIFF
--- a/js/components/text_input.js
+++ b/js/components/text_input.js
@@ -31,7 +31,7 @@ export default {
       type: Boolean,
       default: false,
     },
-    optional: Boolean,
+    insetForm: Boolean,
   },
 
   data: function() {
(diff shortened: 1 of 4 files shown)


js/components/text_input.js
```diff
@@ -31,7 +31,7 @@ export default {
       type: Boolean,
       default: false,
     },
-    optional: Boolean,
```
dandds: Are you sure we're not using the `optional` option anywhere? A quick grep of the templates folder shows it being passed in a few uses of the `TextInput` macro.

leigh-mil: `optional` was listed as a prop twice (line 29), so I deleted the dupe!

dandds: ahh, good catch

PR #1030, opened 2019-08-16
File upload fixes

richard-dds:
Fixes some of the bugs in the TO upload form:

## Large file uploads result in "Request Entity Too Large" from nginx even though the uploads are happening in the client

This was caused by the fact that the upload form still contains an `<input type="file">`. Though we're only using it for display purposes, the value of that field (the file) was still being sent to the server which caused nginx to choke. By disabling this form when the file is attached, the value won't be sent to the server. Then we enable it again when the file is removed so that the user can interact with the input again.

## File uploads larger than 64MB don't work on Azure

I implemented a client-side check for filesize which displays an error message and doesn't upload the file. The copy for this message hasn't been confirmed yet. 64MB seems like a large enough size that it should be acceptable for either CSP.

### DIFF
--- a/translations.yaml
+++ b/translations.yaml
@@ -156,6 +156,7 @@ forms:
       length_error: Filename may be no longer than 100 characters.
   task_order:
     upload_error: There was an error uploading your file. Please try again. If you encounter repeated problems uploading this file, please contact CCPO.
+    size_error: The file you have selected is too large. Please choose a file no larger than 64MB.
     app_migration:
       both: 'Yes, migrating from both an on-premise data center <strong>and</strong> another cloud provider'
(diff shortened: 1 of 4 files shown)


translations.yaml
```diff
@@ -156,6 +156,7 @@ forms:
       length_error: Filename may be no longer than 100 characters.
   task_order:
     upload_error: There was an error uploading your file. Please try again. If you encounter repeated problems uploading this file, please contact CCPO.
+    size_error: The file you have selected is too large. Please choose a file no larger than 64MB.
```
dandds: Can we update the "Your file may not exceed 1MB" text on the file input field, too?

PR #1032, opened 2019-08-16
Do not create AuditEvent if updating the User.last_login

leigh-mil:
## Description
Fixes a bug where every time a user logs in, an AuditEvent is created because the user's last_login is updated. 

## Pivotal
https://www.pivotaltracker.com/story/show/167766390

### DIFF
--- a/atst/models/mixins/auditable.py
+++ b/atst/models/mixins/auditable.py
@@ -54,7 +54,8 @@ class AuditableMixin(object):
 
     @staticmethod
     def audit_update(mapper, connection, target):
-        if AuditableMixin.get_changes(target):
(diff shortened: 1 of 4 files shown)


atst/models/mixins/auditable.py
```diff
@@ -54,7 +54,8 @@ def audit_delete(mapper, connection, target):
 
     @staticmethod
     def audit_update(mapper, connection, target):
-        if AuditableMixin.get_changes(target):
+        changes = AuditableMixin.get_changes(target)
+        if changes and not "last_login" in changes:
```
dandds: I would reimplement the necessary logic directly in the `User` model. I don't think that custom logic for a single model should live in the mixin.

PR #1034, opened 2019-08-19
Center modals in ie11

leigh-mil:
## Description
Fix bug in IE11 where modals were not centered horizontally on the screen.

## Pivotal
https://www.pivotaltracker.com/story/show/167854946

## Screenshot
Screen Shot 2019-08-19 at 11 20 57 AM


dandds:
Merging this since you're away @leigh-mil

PR #1040, opened 2019-08-27
Max line width

leigh-mil:
## Description
Set max line width to 80 characters for `<p>` elements and audit log items. 
Please let me know if I missed anything that needs a max-width applied!

## Pivotal
https://www.pivotaltracker.com/story/show/168031763

## Screenshots
Screen Shot 2019-08-27 at 5 19 20 PM
Screen Shot 2019-08-27 at 5 19 43 PM
Screen Shot 2019-08-27 at 5 19 51 PM

### DIFF
--- a/styles/core/_util.scss
+++ b/styles/core/_util.scss
@@ -76,3 +76,7 @@
 .user-permission {
   font-weight: $font-normal;
 }
+
+p {
+  max-width: 80ch;
+}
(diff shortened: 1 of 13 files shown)


leigh-mil:
@dandds to follow up on your questions -- I did test in IE11 using browser stack and it looks good!
I checked in with the designers yesterday, and although it does look a little weird in some places they would rather adhere to the USWDS than expand the line width. The one exception for this is the audit log, which becomes more difficult to parse with a shorter line length, so the max-width will be removed there.

PR #1050, opened 2019-08-29
Update CloudProviderInterface for provision job consumption

tomdds:
## Description

Initial pass at building the general interface for CSP provisioning

This gist sketches out the example jobs that consume the interface: https://gist.github.com/tomdds/cd379ba95ce81d88aee3f50c7db04b04

## Pivotal
https://www.pivotaltracker.com/story/show/168148066

### DIFF
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -40,16 +40,15 @@ class CloudProviderInterface:
         """
         raise NotImplementedError()
 
-    def create_environment_baseline_roles(self, auth, csp_environment_id):
-        """Provision the baseline set of roles that align with the available
-        environment roles in the ATAT system.
+    def create_environment_baseline(self, auth, csp_environment_id):
+        """Provision the necessary baseline entities (such as roles) in the given environment
 
         Arguments:
             auth -- Object representing authorization for the CSP
             csp_environment_id -- ID of the CSP Environment to provision roles against.
 
         Returns:
-            dict: Returns dict of role name => csp role IDs.
+            dict: Returns dict that associates the resource identities with their ATAT representations.
         """
         raise NotImplementedError()
 
@@ -118,7 +117,7 @@ class MockCloudProvider(CloudProviderInterface):
     def create_atat_admin_user(self, auth, csp_environment_id):
         return {"id": uuid4().hex, "credentials": {}}
 
-    def create_environment_baseline_roles(self, auth, csp_environment_id):
+    def create_environment_baseline(self, auth, csp_environment_id):
         return {
             CSPRole.BASIC_ACCESS: uuid4().hex,
             CSPRole.NETWORK_ADMIN: uuid4().hex,
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -4,12 +4,13 @@ from atst.models.environment_role import CSPRole
 
 
 class CloudProviderInterface:
-    def create_environment(self, auth_credentials, user):
+    def create_environment(self, auth_credentials, user, environment):
         """Create a new environment in the CSP.
 
         Arguments:
             auth_credentials -- Object containing CSP account credentials
             user -- ATAT user authorizing the environment creation
+            environment -- ATAT Environment model
 
         Returns:
             string: ID of created environment
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -4,22 +4,11 @@ from atst.models.environment_role import CSPRole
 
 
 class CloudProviderInterface:
-    def get_auth(self, auth_credentials):
-        """Validate credentials and create auth object
-
-        Arguments:
-            auth_credentials -- Object containing credentials
-
-        Returns:
-            object: An object to be passed into subsequent calls
-        """
-        raise NotImplementedError()
-
-    def create_environment(self, auth, user):
+    def create_environment(self, auth_credentials, user):
         """Create a new environment in the CSP.
 
         Arguments:
-            auth -- Object representing authorization for the CSP
+            auth_credentials -- Object containing CSP account credentials
             user -- ATAT user authorizing the environment creation
 
         Returns:
@@ -27,12 +16,12 @@ class CloudProviderInterface:
         """
         raise NotImplementedError()
 
-    def create_atat_admin_user(self, auth, csp_environment_id):
+    def create_atat_admin_user(self, auth_credentials, csp_environment_id):
         """Creates a new, programmatic user in the CSP. Grants this user full permissions to administer
         the CSP. Returns a dictionary containing user details, including user's API credentials.
 
         Arguments:
-            auth -- Object representing authorization for the CSP
+            auth_credentials -- Object containing CSP account credentials
             csp_environment_id -- ID of the CSP Environment the admin user should be created in
 
         Returns:
@@ -40,11 +29,11 @@ class CloudProviderInterface:
         """
         raise NotImplementedError()
 
-    def create_environment_baseline(self, auth, csp_environment_id):
+    def create_environment_baseline(self, auth_credentials, csp_environment_id):
         """Provision the necessary baseline entities (such as roles) in the given environment
 
         Arguments:
-            auth -- Object representing authorization for the CSP
+            auth_credentials -- Object containing CSP account credentials
             csp_environment_id -- ID of the CSP Environment to provision roles against.
 
         Returns:
@@ -52,11 +41,11 @@ class CloudProviderInterface:
         """
         raise NotImplementedError()
 
-    def create_or_update_user(self, auth, user_info, csp_role_id):
+    def create_or_update_user(self, auth_credentials, user_info, csp_role_id):
         """Creates a user or updates an existing user's role.
 
         Arguments:
-            auth -- Object representing authorization for the CSP
+            auth_credentials -- Object containing CSP account credentials
             user_info -- object containing user data, if it has a csp_user_id
                          it will try to update a user with that id
             csp_role_id -- The id of the role the user should be given in the CSP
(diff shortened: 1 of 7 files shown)


atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+    def get_auth(self, auth_credentials):
+        """Validate credentials and create auth object
+
+        Arguments:
+            auth_credentials -- Object containing credentials
+
+        Returns:
+            object: An object to be passed into subsequent calls
+        """
+        raise NotImplementedError()
```
dandds: I'm wondering what the right pattern is for the AWS consumer. After creating an environment, it needs to:
1. Create a regular boto IAM client.
2. Add an inline policy that allows it to assume a role in the new account.
3. The previous item returns credentials that can be used to make a new IAM client with access to the new account.
4. Then it needs to use the new client to generate a brand new IAM user for itself in the new account.

I guess number 1 is [`atat_root_auth`](https://gist.github.com/tomdds/cd379ba95ce81d88aee3f50c7db04b04#file-comparison-py-L23) in your gist, and then 2-4 are implementation details hidden under the hood of [`create_atat_admin_user`](https://gist.github.com/tomdds/cd379ba95ce81d88aee3f50c7db04b04#file-comparison-py-L32-L34)?

dandds: Is the `auth` object just an instance of a boto/azure-SDK client? For number 2 above, boto needs an "sts" client, whereas most of the other work requires IAM clients. Is the interface expecting that a "starter" client gets passed in to the function, and then it knows how to make alternative clients as-needed? If the implementations need to make their own clients on the fly anyway, is their value to having the consumer pass one version of the client around? Could we just pass the credentials as the first argument to every call instead? (I'm not advocating for these alternatives, per se, just poking at the implementation.)

dandds: In fact, making a generic auth/client object is difficult, because both boto and the Azure SDK require that you build different clients for different kinds of functionality (user management, policy management, file storage, etc.).

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+    def create_environment_baseline_roles(self, auth, csp_environment_id):
+        """Provision the baseline set of roles that align with the available
+        environment roles in the ATAT system.
+
+        Arguments:
+            auth -- Object representing authorization for the CSP
+            csp_environment_id -- ID of the CSP Environment to provision roles against.
+
+        Returns:
+            dict: Returns dict of role name => csp role IDs.
```
dandds: We might want this to be more generic, like `create_environment_baseline`. ATAT will probably be responsible for more than just baseline roles.

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+
+        Arguments:
+            auth_credentials -- Object containing credentials
+
+        Returns:
+            object: An object to be passed into subsequent calls
+        """
+        raise NotImplementedError()
+
+    def create_environment(self, auth, user):
```
dandds: I think this should have a minimal amount of info from the ATAT environment: we could use the portfolio-application-environment names together to build a unique name for the CSP environment, for example. Maybe we should just pass the environment object and let that be an implementation detail we work out later?

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+            string: ID of created environment
         """
         raise NotImplementedError()
 
-    def delete_application(self, cloud_id):  # pragma: no cover
-        """Delete an application in the cloud with the provided cloud_id. Returns
-        True for success or raises an error.
+    def create_atat_admin_user(self, auth, csp_environment_id):
+        """Creates a new, programmatic user in the CSP. Grants this user full permissions to administer
+        the CSP. Returns a dictionary containing user details, including user's API credentials.
```
dandds: Can we spec out the dictionary in the docstring? Or do you think that should come later?

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
-    def create_user(self, user):  # pragma: no cover
-        """Create an account in the CSP for specified user. Returns the ID of
-        the created user.
+    def create_or_update_user(self, auth, user_info, csp_role_id):
+        """Creates a user or updates an existing user's role.
+
+        Arguments:
+            auth -- Object representing authorization for the CSP
+            user_info -- object containing user data, if it has a csp_user_id
+                         it will try to update a user with that id
```
dandds: `user_info` will probably be an instance of the `EnvironmentRole` model; is it worth being explicit about that?

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+            object: An object to be passed into subsequent calls
+        """
+        raise NotImplementedError()
+
+    def create_environment(self, auth, user):
+        """Create a new environment in the CSP.
+
+        Arguments:
+            auth -- Object representing authorization for the CSP
+            user -- ATAT user authorizing the environment creation
```
dandds: Is `user` an instance of our `User` model? Should we be explicit about that?

dandds: I know we're in dynamic-language-land, I'm just wondering if we should distinguish this `user` from the `user_info`  in `create_or_update_user`.

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+        This will likely be a static property elsewhere once a CSP is chosen.
         """
         raise NotImplementedError()
 
-    def calculator_url(self):  # pragma: no cover
-        """Returns a URL for the CSP's estimate calculator."""
+    def get_environment_login_url(self, environment):
+        """Returns the login url for a given environment
+        This may move to be a computed property on the Environment domain object
+        """
```
dandds: In the vein of the previous methods where we pass `csp_user_id`, we should probably expect the environment's `cloud_id` here.

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+        """Creates a user or updates an existing user's role.
+
+        Arguments:
+            auth -- Object representing authorization for the CSP
+            user_info -- object containing user data, if it has a csp_user_id
+                         it will try to update a user with that id
+            csp_role_id -- The id of the role the user should be given in the CSP
+
+        Raises:
+            NotImplementedError: [description]
```
dandds: Should this have a return type specified?

atst/domain/csp/cloud.py
```diff
@@ -1,84 +1,144 @@
...
+
+        Arguments:
+            auth -- Object representing authorization for the CSP
+            csp_user_id -- CSP internal user identifier
 
-        This _does not_ return a token, but is intended to perform any necessary
-        setup to allow a token to be generated in the future (for example,
-        add the user to the cloud entity in some fashion).
+        Returns:
+            None
```
dandds: I don't know what's Pythonic in this situation. `True` to indicate success, or else it raises and error?

tests/domain/test_environments.py
```diff
@@ -15,13 +15,6 @@
 )
 
 
-def test_create_environments():
-    application = ApplicationFactory.create()
-    environments = Environments.create_many(application, ["Staging", "Production"])
-    for env in environments:
-        assert env.cloud_id is not None
```
dandds: NBD, but this might be the only test that directly invokes that `create_many` class method.

atst/domain/csp/cloud.py
tomdds: I think the structure of the details might depend on the CSP in question? Unless this is enough:
```python
{
    "user_id": string,
    "credentials": dict, # contents depends on csp
}
```

atst/domain/csp/cloud.py
tomdds: 👍

atst/domain/csp/cloud.py
tomdds: Yeah, I was a little more lax here, because I suspect this might just get stored as a property on Environment rather than be part of this interface. Thoughts on that?

atst/domain/csp/cloud.py
tomdds: I included the user field because you at least need an email address for the root user of the environment. We could scale it back to `root_user_email`?

atst/domain/csp/cloud.py
tomdds: I was hoping there'd be some synthesis of the credentials that could be used as an auth object, but it seems like both CSP SDKs have enough variance in their implementations for interacting with different services that it probably does make sense to just pass credential objects around.

atst/domain/csp/cloud.py
dandds: You're probably right; that might be over-anticipating things.

atst/domain/csp/cloud.py
dandds: I think it's mostly a design question. For the most part, we've tried to treat the models as data-bags that can calculate easily derived properties. Any CSP-specific information (like the URL format) should live in the CSP modules, so the model would have to import the CSP interface. Which is maybe just fine?

tomdds: I was thinking there'd be a login_url property on Environment, either directly or in the json column, so you'd get it from the model when you wanted to display it to users. I mostly think that and the calculator URL should get moved since neither are actually CSP interactions, they're more metadata about the CSP/environment.

atst/domain/csp/cloud.py
tomdds: What do you think besides `user_id` should come back from this?

tests/domain/test_environments.py
tomdds: Should I mark this as skip rather than deleting it entirely, since this will eventually do a create many, check their computed status is pending, and that a job was issued for each?

richard-dds:
@tomdds @dandds Would it make sense for us to add type annotations to some of the arguments and return types for this interface? The docstrings are good, but annotations may cut down on some of the ambiguity.

PR #1051, opened 2019-08-29
Remove LOAs

graham-dds:
`Yeet` LOAs out of the app.

Addresses: https://www.pivotaltracker.com/story/show/168195826


alembic/versions/4a3122ffe898_remove_loas.py
```diff
@@ -0,0 +1,32 @@
...
+revision = '4a3122ffe898' # pragma: allowlist secret
+down_revision = 'fda6bd7e1b65' # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint('clins_task_order_id_fkey', 'clins', type_='foreignkey')
+    op.create_foreign_key(None, 'clins', 'task_orders', ['task_order_id'], ['id'])
```
dandds: The foreign key constraint needs a name in order for the downgrade to work.

graham-dds: Dropped in `clins_task_order_id_fkey` in place of both `None`s. Upgrades and downgrades seemed to work without any errors locally

dandds:
:+1:

PR #1053, opened 2019-08-30
Stickybits!

leigh-mil:
## Description
The side nav and top bars are now fixed and the sticky CTA rests below! Besides lots of styling, I updated the way that the sticky directive is called, so now we can pass in different attributes and options from [stickybits](https://dollarshaveclub.github.io/stickybits/).

Also fixes some bugs related to the sticky CTA: the sticky CTA did not extend the full width of the screen, and there was a bug in IE11 were it would disappear on the 3rd step of the TO form.

## Pivotal
https://www.pivotaltracker.com/story/show/168030766
https://www.pivotaltracker.com/story/show/168176725
https://www.pivotaltracker.com/story/show/167929953

### DIFF
--- a/js/components/forms/base_form.js
+++ b/js/components/forms/base_form.js
@@ -34,8 +34,12 @@ export default {
   mixins: [FormMixin],
   directives: {
     sticky: {
-      inserted: el => {
-        stickybits(el)
+      inserted: (el, binding) => {
+        var customAttributes
+        if (binding.expression) {
+          customAttributes = JSON.parse(binding.expression)
+        }
+        stickybits(el, customAttributes)
       },
     },
   },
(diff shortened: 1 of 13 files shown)


js/components/forms/base_form.js
```diff
@@ -34,8 +34,12 @@ export default {
   mixins: [FormMixin],
   directives: {
     sticky: {
-      inserted: el => {
-        stickybits(el)
+      inserted: (el, binding) => {
+        var customAttributes
+        if (binding.expression) {
+          customAttributes = JSON.parse(binding.expression)
```
dandds: It's a little gross to have to parse JSON here, but I guess there's no other way to get the config from the template.

PR #1054, opened 2019-09-03
Refactor CLIN summary page

graham-dds:
New Layout             |  Old Layout
:-------------------------:|:-------------------------:
image  |  image

[Mockups](https://thinkcompany.invisionapp.com/share/Y4SWVUKC7K6#/screens/380530986)

This PR rearranges and slightly changes some style elements on the the CLIN Summary Page:
 - changes copy in a few places
 - groups CLINs into one table, instead of one table per CLIN
 - orders CLINs according to [acceptance criteria](https://www.pivotaltracker.com/story/show/168200268)

This PR does **not** address some of the business logic present in the mockups:
 - the CLIN Value column in the CLIN table is a placeholder value
 - I did not investigate the totals in the obligated funds and task order funds sections

### DIFF
--- a/atst/models/clin.py
+++ b/atst/models/clin.py
@@ -36,6 +36,10 @@ class CLIN(Base, mixins.TimestampsMixin):
         ]
 
     @property
+    def type(self):
+        return "Base" if self.number[0] == "0" else "Option"
+    
+    @property
     def is_completed(self):
         return all(
             [
--- a/atst/models/clin.py
+++ b/atst/models/clin.py
@@ -38,7 +38,7 @@ class CLIN(Base, mixins.TimestampsMixin):
     @property
     def type(self):
         return "Base" if self.number[0] == "0" else "Option"
-    
+
     @property
     def is_completed(self):
         return all(
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -47,6 +47,10 @@ class TaskOrder(Base, mixins.TimestampsMixin):
         "CLIN", back_populates="task_order", cascade="all, delete-orphan"
     )
 
+    @property
+    def sorted_clins(self):
+        return sorted(self.clins, key=lambda clin: (clin.number[1:], clin.number[0]))
+
     @hybrid_property
     def pdf(self):
         return self._pdf
--- a/templates/fragments/task_order_review.html
+++ b/templates/fragments/task_order_review.html
@@ -49,7 +49,7 @@
             </thead>
 
             <tbody>
-              {% for clin in task_order.clins %}
+              {% for clin in task_order.sorted_clins %}
               <tr>
                 <td>{{ clin.number }}</td>
                 <td>{{ clin.type }}</td>
--- a/templates/fragments/task_order_review.html
+++ b/templates/fragments/task_order_review.html
@@ -7,7 +7,10 @@
   <div class="form-row">
     <div class="form-col">
       <div class="h2">
-        {{ "task_orders.review.review_your_task_order" | translate }}
+        {{ "task_orders.review.review_your_funding" | translate }}
+      </div>
+      <div>
+        {{ TotalsBox(task_order=task_order) }}
       </div>
 
       <div class="h3">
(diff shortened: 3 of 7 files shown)


atst/models/task_order.py
```diff
@@ -47,6 +47,15 @@ class TaskOrder(Base, mixins.TimestampsMixin):
         "CLIN", back_populates="task_order", cascade="all, delete-orphan"
     )
 
+    @property
+    def sorted_clins(self):
+        return sorted(
+            self.clins,
+            key=lambda clin: (clin.number[-1], clin.number[0:-1])
+            if clin.number
+            else tuple(),
```
graham-dds: When I was trying to use `@hybrid_property` for clin sorting, I noticed that I had to have this conditional to make some tests pass? Not sure if it's actually needed anymore, but I thought I'd keep it in just in case.

templates/fragments/task_order_review.html
```diff
@@ -32,40 +35,36 @@
...
-                  {% if clin.is_obligated() %}
-                    {{ "common.yes" | translate }}
-                  {% else %}
-                    {{ "common.no" | translate }}
-                  {% endif %}
+                 {{ clin.start_date | formattedDate }} -  {{ clin.end_date | formattedDate }}
                 </td>
-                <td>{{ clin.start_date | formattedDate }}</td>
-                <td>{{ clin.end_date | formattedDate }}</td>
+                {# TODO: Swap in total CLIN amount #}
```
graham-dds: This will be addressed in https://www.pivotaltracker.com/story/show/168195824

translations.yaml
```diff
@@ -396,8 +398,8 @@ task_orders:
   sign:
     digital_signature_description: I acknowledge that the uploaded task order contains the required KO signature.
 JEDICLINType:
-  JEDI_CLIN_1: 'Base CLIN 0001: Unclassified IaaS and PaaS'
-  JEDI_CLIN_2: 'CLIN 2: Classified Cloud Services - 0002'
+  JEDI_CLIN_1: 'Unclassified IaaS and PaaS (IDIQ CLIN 0001)'
+  JEDI_CLIN_2: 'Classified Cloud Services  (IDIQ CLIN 0002)'
   JEDI_CLIN_3: 'CLIN 3:'
```
graham-dds: Any idea what CLIN 3 or 4 should say?

templates/fragments/task_order_review.html
```diff
@@ -32,40 +35,36 @@
           {{ "task_orders.review.funding_summary" | translate }}
         </div>
 
-        {% for clin in task_order.clins %}
-          <div>
-            {{ "{}".format(clin.jedi_clin_type) | translate}}
-          </div>
+
```
dandds: Not sure where this is in the partial, but it looks like the designs have `hr` around the "Task Order Number" section. Some of the spacing looks a little different too, but I'd wait for Think to review the build before we address that.

atst/models/clin.py
```diff
@@ -35,6 +35,10 @@ def is_obligated(self):
             JEDICLINType.JEDI_CLIN_3,
         ]
 
+    @property
+    def _type(self):
```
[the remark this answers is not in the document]
graham-dds: Will this conflict at all with https://docs.python.org/3/library/functions.html#type?

tests/domain/test_task_orders.py
```diff
@@ -133,6 +133,41 @@ def test_update_adds_clins():
...
+        },
+    ]
+    task_order = TaskOrders.create(
+        creator=task_order.creator,
+        portfolio_id=task_order.portfolio_id,
+        number="0000000000",
+        clins=clins,
+        pdf={"filename": "sample.pdf", "object_name": "1234567"},
+    )
+    assert [clin.number for clin in task_order.sorted_clins] == ["0001", "1011", "1002"]
```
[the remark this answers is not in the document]
graham-dds: From what I could tell from the acceptance criteria, numbers are sorted first by the last digit, and then by the first three digits. The example given by Ben was:
```
0001
1001
2001
0002
1002
```

So given 1011, 1002, 0001, 2002, I think we'd want 0001, 1011, 1002, 2002

templates/fragments/task_order_review.html
graham-dds: 👍

atst/models/task_order.py
graham-dds: > Do we allow users to save draft CLINs without a number (but with other info)?

I don't think so? I think I'll remove the conditional here.

tests/domain/test_task_orders.py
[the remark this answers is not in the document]
graham-dds: Also, refactored that test. It's _much_ better with the factory. I didn't notice that I was actually using the create method on the model and not the factory 🤦‍♂️

atst/models/clin.py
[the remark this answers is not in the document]
dandds: It won't conflict because `type` is a top-level function, and this property is always called as a method. I would either drop the underscore or rename it. The leading `_` indicates to a reader that it's supposed to be private ([PEP8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)).

PR #1055, opened 2019-09-03
TO Form Step 3/CLIN page styling

leigh-mil:
## Description
Update styling on the CLIN page as per new designs. 
Since the obligated and total funds are covered in [this story](https://www.pivotaltracker.com/story/show/168195824), I did not add in the new field for total funds. I also left out the tooltips for now because we do not have copy for them.

## Pivotal
https://www.pivotaltracker.com/story/show/168200145

## Invision
https://thinkcompany.invisionapp.com/share/Y4SWVUKC7K6#/screens/380530989

## Screenshots
Screen Shot 2019-09-03 at 1 47 00 PM
Screen Shot 2019-09-03 at 1 47 07 PM
Screen Shot 2019-09-03 at 1 47 21 PM


js/components/clin_fields.js
```diff
@@ -113,8 +122,20 @@ export default {
         } else if (event.name.includes(END_DATE)) {
           if (!!event.value) this.endDate = new Date(event.value)
           this.validatePop()
+        } else if (event.name.includes(NUMBER)) {
+          if (!!event.value) this.clinNumber = event.value
```
dandds: In FF, if I delete the entire CLIN number with backspace, it leaves the first digit I entered displayed in the header.
Screen Shot 2019-09-03 at 4 02 07 PM

dandds: And it leaves the entire CLIN there if I select and delete all my input at once.

styles/sections/_task_order.scss
```diff
@@ -243,3 +278,12 @@
...
 }
+
+.clin-card {
+  width: 70rem;
+
+  &__title.h4 {
+    margin-top: 0;
+    padding-bottom: $gap * 2.5;
+  }
+}
```
dandds: The max-width in the design seems slightly narrower (maybe not worth pursuing). The form inputs are all constrained to the same max-width, though, and are definitely narrower in the designs.

templates/components/date_picker.html
```diff
@@ -26,6 +28,10 @@
           {{ label }}
         </div>
 
+        {% if pop_end_alert %}
+          {{ Alert(message="task_orders.form.pop_end_alert" | translate) }}
+        {% endif %}
```
dandds: Since `DatePicker` is supposed to be a generic component, I think it would be better to inject this HTML fragment using the `caller` pattern. You can find an example in the `CheckboxInput` component: templates/components/checkbox_input.html.

PR #1056, opened 2019-09-04
Mock implementation of CSP interface

richard-dds:


atst/domain/csp/cloud.py
```diff
@@ -115,28 +119,84 @@ def get_environment_login_url(self, environment) -> str:
...
+
+    # TODO: All of these constants
+    AUTH_EXCEPTION = GeneralCSPException("Authentication failure.")
+    NETWORK_EXCEPTION = GeneralCSPException("Network failure.")
+
+    NETWORK_FAILURE_PCT = 7
+    ENV_CREATE_FAILURE_PCT = 12
+    ATAT_ADMIN_CREATE_FAILURE_PCT = 12
+
+    def __init__(self, with_delay=True, with_failure=True):
```
dandds: After having talked, I wonder if we should default these kwargs to `False` so that we can easily test happy path when unit testing. We'd only rely on the chaos monkey on the deployed sites; for testing failures, I think we'd mock and stub instead, right?

richard-dds: @dandds I pulled in the change that you and I made while working on the env provisioning jobs. Now `config/test.ini` declares `CSP=mock-test`, which is the happy-path, deterministic, non-delaying implementation. Does that work?

PR #1058, opened 2019-09-04
Remove CLIN

leigh-mil:
## Description
Adds functionality to remove a CLIN card from the TO form. When the 'X' is clicked, a modal pops up to confirm that the user wants to remove that CLIN.
Also fixes some styling bugs that I noticed: fixed rendering TOs with CLINs pre-filled (some of the fields were marked as optional), and fixed the hr elements that disappeared.

## Pivotal
https://www.pivotaltracker.com/story/show/168200227

## Screenshots
Screen Shot 2019-09-04 at 4 51 18 PM
Screen Shot 2019-09-04 at 4 51 23 PM

### DIFF
--- a/templates/task_orders/step_3.html
+++ b/templates/task_orders/step_3.html
@@ -16,8 +16,6 @@
   <clin-fields
     {% if fields %}
       v-bind:initial-clin-index='{{ index }}'
-      v-bind:initial-clin-type="'{{ fields.jedi_clin_type.data }}'"
-      v-bind:initial-amount='{{ fields.obligated_amount.data or 0 }}'
       v-bind:initial-start-date="'{{ fields.start_date.data | string }}'"
       v-bind:initial-end-date="'{{ fields.end_date.data | string }}'"
       v-bind:initial-clin-number="'{{ fields.number.data | string }}'"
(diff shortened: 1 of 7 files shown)


templates/task_orders/step_3.html
```diff
@@ -318,33 +325,64 @@
...
         </div>
       </div>
+
+      <div v-show="$root.activeModal === removeModalId" v-cloak>
+        <div :id='"modal--" + removeModalId' class='modal modal--dismissable'>
+          <div class='modal__container'>
+            <div class='modal__dialog' role='dialog' aria-modal='true'>
+              <div class='modal__body'>
+                <div class="task-order__modal-cancel">
+                  <h1 v-html='"{{ 'task_orders.form.clin_remove_text' | translate }}" + clinTitle + "?"'></h1>
```
dandds: Is there a reason to use the `v-html` directive here and not just make it regular Jinja templating?

leigh-mil: It is because I need to interpolate the `clinTitle` into the text!

dandds: Aaah gotcha, yup. I didn't read closely enough.

PR #1060, opened 2019-09-05
App settings redesign

leigh-mil:
## Description
Lo-fi update of the App settings page based on the new designs.

## Pivotal
https://www.pivotaltracker.com/story/show/168300846

## Screenshots
Screen Shot 2019-09-06 at 3 42 12 PM
Screen Shot 2019-09-06 at 3 42 20 PM
Screen Shot 2019-09-06 at 3 42 23 PM


Pipfile
```diff
@@ -26,6 +26,7 @@ azure-storage = "*"
 azure-storage-common = "*"
 boto3 = "*"
 celery = "*"
+inflect = "*"
```
dandds: I'm not sure how I feel about pulling in this dependency. We're only using `pluralize` in one place, which you could do by checking if there's more than one environment. I'm also not sure that `numberToWords` adds enough value to be worth it; we could probably get away with just using integers in MVP. The bar for including production dependencies is higher in this app because of the nature of the project; let's talk about it.

atst/filters.py
```diff
@@ -67,6 +68,14 @@ def renderAuditEvent(event):
         return render_template("audit_log/events/default.html", event=event)
 
 
+def numberToWords(integer):
+    return inflect.engine().number_to_words(integer)
```
dandds: If we're going to use `pluralize`, you'd probably want to make one instance of `engine` that's shared by the whole app. That way you're not allocating an extra object every time you do one of these operations.

templates/fragments/applications/add_new_environment.html
```diff
@@ -28,15 +28,9 @@
...
-        <a class='icon-link' v-on:click="toggle">
-          {{ "portfolios.applications.add_environment" | translate }}
-          {{ Icon('plus') }}
-        </a>
-      </div>
-    </div>
+    <a class='icon-link' v-on:click="toggle">
+      {{ Icon('plus') }}
+      {{ "portfolios.applications.add_environment" | translate }}
+    </a>
```
dandds: The "create environment" CTA overlaps "cancel" when I open this. I know this is low-fi, but all the basic functions should still be be accessible.
Screen Shot 2019-09-09 at 1 51 44 PM

atst/routes/applications/settings.py
```diff
@@ -264,3 +338,110 @@ def delete_environment(environment_id):
             fragment="application-environments",
         )
     )
+
+
+@applications_bp.route("/application/<application_id>/team", methods=["POST"])
+@user_can(Permissions.EDIT_APPLICATION_MEMBER, message="update application member")
+def update_team(application_id):
```
dandds: Is this being used in the template at the moment, or is it just a transplant from the teams file? I'm tempted to say delete; when this functionality is re-implemented, the route function probably won't look like this anymore.

templates/fragments/applications/add_new_environment.html
leigh-mil: Good catch!

PR #1061, opened 2019-09-06
Record job failures with application context.

dandds:
AT-AT needs to be able to track which user tasks failed and why. To
accomplish this we:

- Enabled Celery's results backend, which logs task results to a data
  store; a Postgres table, in our case.
  (https://docs.celeryproject.org/en/latest/userguide/tasks.html#result-backends)
- Created tables to track the relationships between the relevant models
  (Environment, EnvironmentRole) and their task failures.
- Added an `on_failure` hook that tasks can use. The hook will add
  records to the job failure tables.

Now a resource like an `Environment` has access to it task failures
through the corresponding failure table.

Notes:
- It might prove useful to use a real foreign key to the Celery results
  table eventually. I did not do it here because it requires that we
  explicitly store the Celery results table schema as a migration and
  add a model for it. In the current implementation, AT-AT can be
  agnostic about where the results live.
- We store the task results indefinitely, so it is important to specify
  tasks for which we do not care about the results (like `send_mail`)
  via the `ignore_result` kwarg.


atst/jobs.py
```diff
@@ -1,14 +1,32 @@
...
 
 from atst.queue import celery
+from atst.database import db
+from atst.models import EnvironmentJobFailure, EnvironmentRoleJobFailure
 
 
-@celery.task()
+class RecordFailure(celery.Task):
+    def on_failure(self, exc, task_id, args, kwargs, einfo):
+        if "environment_id" in kwargs:
```
richard-dds: I wonder if it would be clearer to make separate `RecordFailure` classes for each failure type (`RecordEnvironmentFailure` or `RecordEnvironmentRoleFailure`). That way the failure type would be clearly defined on the task itself. It would also save us from the possible but unlikely issue that someone defined a task with both `environment_id` and `environment_role_id`, or whatever resources we may need to record failures on in the future.

PR #1069, opened 2019-09-11
CLIN obligated and total funding

graham-dds:
Mostly addresses: https://www.pivotaltracker.com/story/show/168195824
And covers the summary table on: https://www.pivotaltracker.com/story/show/168200268

#### CLIN Funding Valid State
Screen Shot 2019-09-11 at 3 01 28 PM

#### CLIN Funding Invalid States
Screen Shot 2019-09-11 at 3 02 12 PM

#### Step 4 Summary Table
image

#### Task Order Card on Portfolio Page
image


js/components/clin_dollar_amount.js
```diff
@@ -0,0 +1,180 @@
...
+import { formatDollars } from '../lib/dollars'
+import { emitEvent } from '../lib/emitters'
+
+export default {
+  name: 'clindollaramount',
+
+  components: {
+    MaskedInput,
+  },
+
```
dandds: Can you try creating a mixin that this component and `textinput` could share? You can see some examples in the `js/mixins` directory.

https://vuejs.org/v2/guide/mixins.html

atst/forms/task_order.py
```diff
@@ -47,9 +48,25 @@ class CLINForm(FlaskForm):
         format="%m/%d/%Y",
         validators=[Optional()],
     )
+    total_amount = DecimalField(
+        label=translate("task_orders.form.total_funds_label"),
+        validators=[
+            NumberRange(
+                0,
+                1000000000,
```
dandds: We should at least move this number out to a constant var, and maybe even put it in the application config.

templates/components/clin_dollar_amount.html
```diff
@@ -0,0 +1,70 @@
...
+        <div v-bind:class="['usa-input usa-input--validation--dollars', { 'usa-input--error': showError, 'usa-input--success': showValid}]">
+          {% if field %}
+          <label for='{{ field.name }}'> 
+          {% else %}
+          <label :for='name'>
+          {% endif  %}
+
+            {% if type=="obligated" %} 
+            <div class="usa-input__title">{{ 'task_orders.form.obligated_funds_label' | translate }}</div>
+            {% else %}
```
leigh-mil: could you use `field.label` here instead of this if/else statement?

atst/forms/task_order.py
```diff
@@ -89,6 +106,16 @@ def validate(self, *args, **kwargs):
...
 
+        if (
+            isinstance(self.total_amount.data, Number)
+            and isinstance(self.obligated_amount.data, Number)
+            and self.total_amount.data < self.obligated_amount.data
+        ):
+            self.obligated_amount.errors.append(
+                translate("forms.task_order.clin_funding_errors.obligated_amount_error")
+            )
+            valid = False
```
dandds: I should have called this out on the PR that added the other validations, but I don't think doing this in the top-level `validate` method is the right approach. We should be using custom validators:

https://wtforms.readthedocs.io/en/stable/validators.html#custom-validators

A custom validator gets the entire form object and the individual field as arguments, so you can still check values in other fields. Then you can just raise `ValidationError` with the correct error text. I like this better because it seems to be the way the library intends for you to do this kind of validation.

alembic/versions/f03333c42bdb_add_total_amount_to_clins.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = '30ea1cb20807' # pragma: allowlist secret
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('clins', sa.Column('total_amount', sa.Numeric(scale=2), nullable=True))
+    op.execute("UPDATE clins SET total_amount = 0")
+    op.alter_column('clins', 'total_amount', nullable=False)
```
dandds: Looks like you have this set as nullable in the model. Everything but the TO foreign key is nullable on `clins` at the moment. Is there a strong reason to make `total_amount` non-nullable?

js/components/clin_fields.js
```diff
@@ -101,14 +121,37 @@ export default {
...
   },
 
   methods: {
+    clinChangeEvent: function() {
+      emitEvent('clin-change', this, {
+        id: this._uid,
+        obligatedAmount: this.initialObligated,
+        totalAmount: this.initialTotal,
+      })
+    },
```
dandds: Can we pair and talk through the event chain in this component? I'm having trouble wrapping my head around everything that's happening here.

js/components/clin_fields.js
```diff
@@ -153,9 +196,30 @@ export default {
...
+
+    validateFunding: function() {
+      if (this.totalAmount && this.obligatedAmount) {
+        this.fundingValid = this.checkFundingValid()
+      }
+
+      emitEvent('field-change', this, {
+        name: OBLIGATED_AMOUNT,
+        valid: this.checkFundingValid(),
```
dandds: Do we want to emit this event if we don't have `this.totalAmount && this.obligatedAmount`? And should we just call `this.checkFundingValid()` once inside `validateFunding`?

js/components/clin_fields.js
```diff
@@ -186,6 +250,20 @@ export default {
...
+          return '<1%'
+        } else if (percentage > 99 && percentage < 100) {
+          return '>99%'
+        } else {
+          return `${percentage.toFixed(0)}%`
+        }
+      } else {
+        return '0%'
+      }
+    },
```
dandds: We can probably get away with returning string values here for now. If the designers want more control over the styling and content (like a badge) we might need to do something more complicated.

templates/components/clin_dollar_amount.html
graham-dds: Talked about this on Slack -- since field could be `None`, I'll keep this if/ else statement in

PR #1074, opened 2019-09-16
Implement simple locking system for environments

richard-dds:
A simple locking system for environments to prevent duplicate cloud resources from being created.


atst/jobs.py
```diff
@@ -44,32 +45,60 @@ def send_notification_mail(recipients, subject, body):
...
+        )
+
+    claimed = db.session.query(Model).filter_by(id=resource.id).one()
+
+    try:
+        yield claimed
+    finally:
+        db.session.query(Model).filter(Model.id == resource.id).filter(
+            Model.claimed_at != None
+        ).update({"claimed_at": sql.null()}, synchronize_session=False)
```
dandds: ❤️ I love this

I know this is in draft so you're probably still feeling things out, but should we move this into a utils module under `models`?

atst/jobs.py
```diff
@@ -44,32 +45,60 @@ def send_notification_mail(recipients, subject, body):
...
+
+
+def do_create_environment(csp: CloudProviderInterface, environment_id=None):
     environment = Environments.get(environment_id)
 
-    if environment.cloud_id is not None:
-        # TODO: Return value for this?
-        return
+    with claim_for_update(environment) as environment:
```
dandds: So this will just fail with `ClaimFailedException`, right?

dandds: Do we want to also update the query in `Environments.get_environments_pending_creation` to ignore claimed environments?

dandds: (And all of the `do_` functions are going to use `claim_for_update`, right?

dandds:
Sorry, just dropping in comments as I think of them. Slightly unrelated to this PR, but the `text("'null'")` portions of the dispatch queries didn't work for me when I fired up the beat jobs. I can see in the database that there are plenty of environments with `cloud_id` but nothing else set, but those queries return empty lists. Switching that SQL snippet to Python `None` worked, but breaks the tests. I haven't poked at the failing tests very much.

I wonder if we want to adjust the sample script to create environments that are mostly (or maybe entirely provisioned). When you kick off the beat worker the first time against a database with our sample data set, it immediately goes into turbo mode.

richard-dds:
@dandds Thanks for the feedback! 

Since I was surprised that I even needed to use `text("'null'")` in the first place to query for an unset `jsonb` type, and then even more surprised to find that I needed `sql.null()` to set a column to a null value, and now you're saying that it's not working for you, I really wonder what's going on with these `None`s and `null`s.

richard-dds:
@dandds  Another thing I thought of: if we want to make these claims expire, just in case the job fails partway through and the `finally` block is never reached, I could easily make the `claimed_at` column something more like `claimed_until`, and instead of checking for `null`ness we would check to see if `claimed_at < now()`. Just an idea, because as it stands it's very possible for something to stay claimed indefinitely if there's an unexpected failure. Do you think that's something we should worry about?

atst/jobs.py
richard-dds: Yeah, the idea is that it will just fail instead of waiting, since it could be a few minutes. That way we could also handle the exception in the task and kick off a retry with an appropriate timeout.

Having the queries ignore claimed environments is also a great idea. Otherwise we'll still end up with unnecessary jobs on the queue that will just catch a `ClaimFailedException` and retry.

And yes, I expect that all of those functions will use `claim_for_update`.

atst/jobs.py
richard-dds: For sure. Originally I thought it could be a model mixin, but for now I do think I prefer it as a plain function. I'll move it

tests/test_jobs.py
```diff
@@ -73,7 +76,7 @@ def csp():
 def test_create_environment_job(session, csp):
     user = UserFactory.create()
```
dandds: Nit-pick, but can we remove the extra user from these tests if we don't need it?

atst/models/utils.py
```diff
@@ -0,0 +1,49 @@
...
+                or_(Model.claimed_until == None, Model.claimed_until < func.now()),
+            )
+        )
+        .update({"claimed_until": claim_until}, synchronize_session="fetch")
+    )
+    if rows_updated < 1:
+        raise ClaimFailedException(resource)
+
+    # Fetch the claimed resource
+    claimed = db.session.query(Model).filter_by(id=resource.id).one()
```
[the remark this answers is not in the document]
richard-dds: I feel like that should exist, but I'm having trouble figuring out how to do it. There are a few ways in sqlalchemy to tell it to add a `RETURNING ...` statement to your query, but they seem to be for the core api only. I'm surprised that I can't find an easy way for me to tell it to return all columns and give them to me in a model instance.

PR #1075, opened 2019-09-16
Mock data for testing

leigh-mil:
## Description
Mocking some functionality and labels for Think's testing:
1. Adds pending tag for pending app invites
2. adds resend and revoke invite links for pending app invites
3. adds processing label to an environment named 'QA'
4. adds a link in the add app member form for users to get info on how to find a DoD ID

### DIFF
--- a/templates/portfolios/applications/settings.html
+++ b/templates/portfolios/applications/settings.html
@@ -128,9 +128,7 @@
             {% for member in members %}
               <tr>
                 <td>
-                  <p>
-                    {{ member.user_name }}
-                  </p>
+                  {{ member.user_name }}<br>
                   {% if member.role_status == 'pending' %}
                     <span class='label label--purple'>INVITE PENDING</span>
                   {% endif %}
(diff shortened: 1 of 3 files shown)


templates/portfolios/applications/settings.html
```diff
@@ -121,12 +121,21 @@
...
             </tr>
           </thead>
           <tbody>
             {% for member in members %}
               <tr>
-                <td>{{ member.user_name }}</td>
+                <td>
+                  <p>
+                    {{ member.user_name }}
+                  </p>
```
dandds: Nit-picky, but I wouldn't do `p` followed by `span`. `p` is block, `span` is inline. I don't think you need the `p`, so maybe just remove it.

templates/portfolios/applications/settings.html
```diff
@@ -137,6 +146,16 @@
...
                   {% endfor %}
                 </td>
+                <td>
+                  {% if member.role_status == 'pending' %}
+                    <p>
+                      <a href="#">Resend Invite</a>
+                    </p>
+                    <p>
+                      <a href="#">Revoke Invite</a>
+                    </p>
```
dandds: I'd probably just use a `br` instead of two `p`s.

PR #1077, opened 2019-09-17
First pass at Cloud Provision Exceptions

tomdds:
## Description

Set of exceptions for the CSP interface that tries to account for the different types of issues that can occur during CSP provisioning.

## Pivotal
https://www.pivotaltracker.com/story/show/168148134

### DIFF
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -78,21 +78,6 @@ class UnknownServerException(GeneralCSPException):
         return "A server error occured: {}".format(self.server_error)
 
 
-class EnvironmentExistsException(GeneralCSPException):
-    """If the environment you're attempting to provision either already exists
-    or is in the process of being created, throw this exception
-    """
-
-    def __init__(self, env_identifier):
-        self.env_identifier = env_identifier
-
-    @property
-    def message(self):
-        return "The environment {} already exists or is already being created".format(
-            self.env_identifier
-        )
-
-
 class EnvironmentCreationException(GeneralCSPException):
     """If there was an error in creating the environment
     """
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -93,7 +93,7 @@ class EnvironmentCreationException(GeneralCSPException):
         )
 
 
-class UserProvisioningError(GeneralCSPException):
+class UserProvisioningException(GeneralCSPException):
     """Failed to provision a user
     """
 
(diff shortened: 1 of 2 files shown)


atst/domain/csp/cloud.py
```diff
@@ -11,6 +11,138 @@ class GeneralCSPException(Exception):
...
+        return "An operation for this entity is already in progress: {}".format(
+            self.operation_desc
+        )
+
+
+class AuthorizationException(GeneralCSPException):
+    """Throw this for instances when there is a problem with the auth credentials:
+    * Missing credentials
+    * Incorrect credentials
+    * Credentials not authorized for current action? (should this be it's own error?)
```
dandds: Yeah, I think we should have `AuthenticationException` (bad credentials) and `AuthorizationException` (missing permissions), assuming both CSPs communicate that difference.

atst/domain/csp/cloud.py
```diff
@@ -11,6 +11,138 @@ class GeneralCSPException(Exception):
...
+        self.connection_error = connection_error
+
+    @property
+    def message(self):
+        return "Could not connect to cloud provider: {}".format(self.connection_error)
+
+
+class UnknownServerException(GeneralCSPException):
+    """An error occured on the CSP side (5xx) and we don't know why
+    """
```
dandds: woof

atst/domain/csp/cloud.py
```diff
@@ -11,6 +11,138 @@ class GeneralCSPException(Exception):
...
+
+    @property
+    def message(self):
+        return "A server error occured: {}".format(self.server_error)
+
+
+class EnvironmentExistsException(GeneralCSPException):
+    """If the environment you're attempting to provision either already exists
+    or is in the process of being created, throw this exception
+    """
```
dandds: Is this distinct from `OperationInProgressException` in all cases?

atst/domain/csp/cloud.py
tomdds: Yeah, I think they both have the distinction, I'll split them.

atst/domain/csp/cloud.py
tomdds: Yeah, they both have a few error types for 5xx errors that all summarize to "Something went wrong on our end" without any details. I wanted to log these explicitly since there'll likely be a direct support line to escalate these errors to.

atst/domain/csp/cloud.py
tomdds: 🤔 This might warrant more discussion with the introduction of the locking behavior for environment provisioning. Concurrent modification exceptions are listed for a lot of endpoints in both APIs, so they can occur with user upsert and other entity behaviors (e.g. downstream deletes). Maybe `EnvironmentExistsException` only fires for completed environments that somehow get re-requested?

tomdds:
@dandds I think updating the mock implementation is fine for this PR, just want to get an iteration or 2 on the errors themselves before committing to them in the mock.

atst/domain/csp/cloud.py
dandds: > Maybe EnvironmentExistsException only fires for completed environments that somehow get re-requested?

Assuming that's knowable from what the CSP tells us, that makes sense to me.

tomdds: I think I may drop this one then. I was operating under an assumption that inputs to the create environment (account/subscription) request were expected to be unique and there could be unexpected collisions based on how we constructed them, but that doesn't appear to be the case. Anything like this, though, we could specify in EnvironmentCreationException instead.

PR #1084, opened 2019-09-20
Revoke app invite

leigh-mil:
## Description
Hook up the functionality to revoke an invite to the existing link. 
This is implemented so when an invite is revoked, the associated app role is disabled, and all related env roles are deleted. This PR also updates the seed sample script so the app members created now have invites so you can revoke them!

## Pivotal
https://www.pivotaltracker.com/story/show/168459532

## Screenshot
Screen Shot 2019-09-20 at 12 01 57 PM


atst/domain/applications.py
```diff
@@ -117,13 +116,4 @@ def invite(
...
-        application_role.deleted = True
-
-        for env in application_role.application.environments:
-            EnvironmentRoles.delete(
-                application_role_id=application_role.id, environment_id=env.id
-            )
-
-        db.session.add(application_role)
-        db.session.commit()
+        ApplicationRoles.disable(application_role)
```
dandds: Is there a good reason to leave this class method if it's just calling another without modification?

atst/domain/application_roles.py
```diff
@@ -70,3 +71,24 @@ def update_permission_sets(cls, application_role, new_perm_sets_names):
...
+
+    @classmethod
+    def disable(cls, application_role):
+        cls._update_status(application_role, ApplicationRoleStatus.DISABLED)
+        application_role.deleted = True
+
+        for env in application_role.application.environments:
+            EnvironmentRoles.delete(
+                application_role_id=application_role.id, environment_id=env.id
+            )
```
dandds: This might be outside the scope of your PR, but `EnvironmentRoles.delete` is deleting the roles from the db instead of marking them deleting, which is what `disable` is doing for the parent role here. It seems like they should do the same thing.

atst/routes/applications/settings.py
```diff
@@ -379,3 +380,30 @@ def update_member(application_id, application_role_id):
             _anchor="application-members",
         )
     )
+
+
+@applications_bp.route(
+    "/applications/<application_id>/members/<application_role_id>/revoke_invite",
+    methods=["POST"],
+)
+@user_can(Permissions.DELETE_APPLICATION_MEMBER, message="revoke appliction invitation")
```
dandds: Small typo

atst/routes/applications/settings.py
```diff
@@ -379,3 +380,30 @@ def update_member(application_id, application_role_id):
...
+    app_role = ApplicationRoles.get_by_id(application_role_id)
+    invite = app_role.latest_invitation
+
+    if invite.is_pending:
+        ApplicationInvitations.revoke(invite.token)
+        flash(
+            "application_invite_revoked",
+            user_name=app_role.user_name,
+            application_name=g.application.name,
+        )
```
dandds: I think you need another side to this if-block, even if it just adds a generic error message to `flash`.

templates/portfolios/applications/settings.html
```diff
@@ -138,6 +138,22 @@ <h1>{{ Icon('avatar') }} {{ member.user_name }}</h1>
...
+          {% if user_can(permissions.DELETE_APPLICATION_MEMBER) and member.role_status == 'pending' %}
+            {% set revoke_invite_modal = "revoke_invite_{}".format(member.role_id) %}
+            {% call Modal(name=revoke_invite_modal, dismissable=True) %}
+              <div class="task-order__modal-cancel">
+                <form method="post" action="{{ url_for('applications.revoke_invite', application_id=application.id, application_role_id=member.role_id) }}">
+                  {{ member.form.csrf_token }}
+                  <h1>Do you want to revoke the invite for {{ member.user_name }}</h1>
+                  <div class="task-order__modal-cancel_buttons">
+                    <button class="usa-button usa-button-primary" type="submit">Yes, revoke it</button>
+                    <button type='button' v-on:click='closeModal("{{revoke_invite_modal}}")' class="usa-button usa-button-primary">No, do not revoke it</button>
```
dandds: Translations file 😬 ?

atst/domain/application_roles.py
```diff
@@ -70,3 +71,24 @@ def update_permission_sets(cls, application_role, new_perm_sets_names):
...
+    @classmethod
+    def _update_status(cls, application_role, new_status):
+        application_role.status = new_status
+        db.session.add(application_role)
+        db.session.commit()
+
+        return application_role
+
+    @classmethod
+    def disable(cls, application_role):
```
dandds: Can you add a test for this method?

atst/domain/application_roles.py
leigh-mil: I thought that a while back we decided to actually delete the environment roles, but I can switch it to a soft delete instead.

dandds: You might be right; why don't we leave as-is for now and we can revisit the question once we have more overall clarity around the way we handle environment members.

tests/domain/test_application_roles.py
```diff
@@ -66,3 +66,21 @@ def test_get_by_id():
...
+    application = ApplicationFactory.create()
+    user = UserFactory.create()
+    member_role = ApplicationRoleFactory.create(application=application, user=user)
+    environment = EnvironmentFactory.create(application=application)
+    environment_role = EnvironmentRoleFactory.create(
+        application_role=member_role, environment=environment
+    )
+    assert member_role == ApplicationRoles.get(
+        user_id=user.id, application_id=application.id
+    )
```
dandds: Why are you re-fetching the member role here? If you need an updated copy from the dp, you can do `session.refresh(member_role)` with the pytest `session` fixture.

tests/domain/test_application_roles.py
```diff
@@ -66,3 +66,21 @@ def test_get_by_id():
...
+
+
+def test_disable():
+    application = ApplicationFactory.create()
+    user = UserFactory.create()
+    member_role = ApplicationRoleFactory.create(application=application, user=user)
+    environment = EnvironmentFactory.create(application=application)
+    environment_role = EnvironmentRoleFactory.create(
+        application_role=member_role, environment=environment
+    )
```
dandds: Should we assert that this is deleted?

tests/domain/test_application_roles.py
dandds: Nevermind, I misread this. I still don't understand why we're asserting this here, though.

PR #1088, opened 2019-09-23
Make application creation multistep

graham-dds:
This PR is one that will address https://www.pivotaltracker.com/story/show/168412308

This PR simply splits the current flow of `https://<csp>.atat.code.mil/portfolios/<profile>/applications/new` into a multistep flow. 

In following PRs, the ability to add members will be added to this flow. Additionally, copy and design will be changed to be in line with Think's designs.

The confirmation modal was also removed with this PR, as this wasn't present in Think's new designs.

### DIFF
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -1,4 +1,5 @@
 from . import BaseDomainClass
+from flask import g
 from atst.database import db
 from atst.domain.application_roles import ApplicationRoles
 from atst.domain.environment_roles import EnvironmentRoles
@@ -26,7 +27,7 @@ class Applications(BaseDomainClass):
         db.session.add(application)
 
         if environment_names:
-        Environments.create_many(user, application, environment_names)
+            Environments.create_many(user, application, environment_names)
 
         db.session.commit()
         return application
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -49,7 +49,10 @@ class Applications(BaseDomainClass):
             application.name = new_data["name"]
         if "description" in new_data:
             application.description = new_data["description"]
-
+        if "environment_names" in new_data:
+            Environments.create_many(
+                g.current_user, application, new_data["environment_names"]
+            )
         db.session.add(application)
         db.session.commit()
 
--- a/atst/domain/applications.py
+++ b/atst/domain/applications.py
@@ -19,12 +19,13 @@ class Applications(BaseDomainClass):
     resource_name = "application"
 
     @classmethod
-    def create(cls, user, portfolio, name, description, environment_names):
+    def create(cls, user, portfolio, name, description, environment_names=None):
         application = Application(
             portfolio=portfolio, name=name, description=description
         )
         db.session.add(application)
 
+        if environment_names:
         Environments.create_many(user, application, environment_names)
 
         db.session.commit()
--- a/atst/utils/flash.py
+++ b/atst/utils/flash.py
@@ -7,6 +7,13 @@ MESSAGES = {
         "message_template": "Portfolio '{{portfolio_name}}' has been deleted",
         "category": "success",
     },
+    "application_created": {
+        "title_template": translate("flash.success"),
+        "message_template": """
+            {{ "flash.application.created" | translate({"application_name": application_name}) }}
+        """,
+        "category": "success",
+    },
     "application_deleted": {
         "title_template": translate("flash.success"),
         "message_template": """
--- a/js/components/forms/new_application/name_and_description.js
+++ b/js/components/forms/new_application/name_and_description.js
@@ -1,0 +1,19 @@
+import FormMixin from '../../../mixins/form'
+import textinput from '../../text_input'
+import * as R from 'ramda'
+
+export default {
+  name: 'application-name-and-description',
+
+  mixins: [FormMixin],
+
+  components: {
+    textinput,
+  },
+
+  data: function() {
+    return {
+      errors: [],
+    }
+  },
+}
--- a/templates/portfolios/applications/new/step_1.html
+++ b/templates/portfolios/applications/new/step_1.html
@@ -1,0 +1,51 @@
+{% extends "portfolios/applications/base.html" %}
+
+{% from "components/alert.html" import Alert %}
+{% from "components/text_input.html" import TextInput %}
+{% from 'components/save_button.html' import SaveButton %}
+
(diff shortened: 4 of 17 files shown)


js/components/forms/new_application/name_and_description.js
```diff
@@ -0,0 +1,19 @@
+import FormMixin from '../../../mixins/form'
+import textinput from '../../text_input'
+import * as R from 'ramda'
+
+export default {
```
graham-dds: This is essentially a stub at this point, but more logic may be added if/ when we put more logic into the different steps of this form, such as the ability to cancel a submission, or a confirmation modal before proceeding.

atst/utils/flash.py
```diff
@@ -7,6 +7,25 @@
         "message_template": "Portfolio '{{portfolio_name}}' has been deleted",
         "category": "success",
     },
+    "application_environment_members_updated": {
+        "title_template": "Application environment members updated",
+        "message_template": "Application environment members have been updated",
+        "category": "success",
+    },
```
dandds: Looks like this isn't used, but maybe you want it for your second PR? (I'd move the text to the translations file, though.)

atst/utils/flash.py
```diff
@@ -7,6 +7,25 @@
...
+        """,
+        "category": "success",
+    },
+    "application_draft": {
+        "title_template": translate("flash.success"),
+        "message_template": """
+            {{ "flash.application.draft" | translate({"application_name": application_name}) }}
+        """,
+        "category": "success",
+    },
```
dandds: This isn't use either. Is this a leftover from when we were thinking we might have draft states?

templates/portfolios/applications/new/step_1.html
```diff
@@ -0,0 +1,57 @@
...
+              {{ TextInput(form.name, optional=False) }}
+            </div>
+          </div>
+          <div class="form-row">
+            <div class="form-col form-col--two-thirds">
+              {{ TextInput(form.description, paragraph=True, optional=False) }}
+            </div>
+          </div>
+
+          <div> {# this extra div prevents this bug: https://www.pivotaltracker.com/story/show/160768940 #}
```
dandds: The PT card is gone, so a lot of good that comment does, but I'm 99% sure it refers to an issue with the environments list and you don't need to reproduce it here.

tests/routes/applications/test_new.py
```diff
@@ -1,21 +1,92 @@
...
+    application = ApplicationFactory.create()
+    user_session(application.portfolio.owner)
+    response = client.get(
+        url_for(
+            "applications.view_new_application_step_1",
+            portfolio_id=application.portfolio.id,
+            application_id=application.id,
+        )
+    )
+    assert response.status_code == 200
```
dandds: Do you want to check that application info is on the page?

tests/routes/applications/test_new.py
```diff
@@ -1,21 +1,92 @@
...
+        data={"name": "Test Application", "description": "This is only a test"},
+    )
+    assert response.status_code == 302
+    assert len(portfolio.applications) == 1
+    assert portfolio.applications[0].name == "Test Application"
+    assert portfolio.applications[0].description == "This is only a test"
+
+
+def test_post_name_and_description_for_update(client, user_session):
+    application = ApplicationFactory.create()
```
dandds: Nit-picky, but I like to hard-code the initial attributes (name and descrip) in cases like these so it's obvious to the reader that they changed by the time we make assertions.

tests/routes/applications/test_new.py
```diff
@@ -1,21 +1,92 @@
...
+        url_for(
+            "applications.update_new_application_step_1",
+            portfolio_id=application.portfolio.id,
+            application_id=application.id,
+        ),
+        data={"name": "Test Application", "description": "This is only a test"},
+    )
+    assert response.status_code == 302
+
+    updated_application = Applications.get(application.id)
```
dandds: You can also use the pytest `session` fixture and do `sesson.refresh(application)`, I think.

atst/domain/applications.py
```diff
@@ -48,6 +50,11 @@ def update(cls, application, new_data):
             application.name = new_data["name"]
         if "description" in new_data:
             application.description = new_data["description"]
+        # TODO: only allow creating environments once
```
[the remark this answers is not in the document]
graham-dds: > You can add new environments any time.
Ahh okay. But is it true that we should be careful about not allowing users to edit the names of environments at any time? I think that's what I meant to say with this TODO.

> Is this to make sure users don't somehow re-use your form page for application creation? I would handle that in the route function, if that's the case.

This sounds like the right move.

atst/utils/flash.py
graham-dds: > Is this a leftover from when we were thinking we might have draft states? 
Yep -- I'll remove it 🗑

tests/routes/applications/test_new.py
graham-dds: I'll give this a try  👍

tests/routes/applications/test_new.py
graham-dds: Yeah I was thinking that, but wasn't sure how to test it. Will I need to use BeautifulSoup for this? I see other tests using `response.data.decode()`, but I'm not totally sure what that's doing.

atst/domain/applications.py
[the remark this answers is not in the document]
dandds: > But is it true that we should be careful about not allowing users to edit the names of environments at any time? I think that's what I meant to say with this TODO.

I don't think this method would ever change environments; it would just add new ones every time. In this specific context, I think that's fine.

> I'm still not 100% sure about how the environments in this application object are related (or not) to the environments that are provisioned on the CSP.

We make a corresponding environment in the CSP for every environment object in the application. (We have a diagram [here](https://drive.google.com/drive/folders/1kLMXLuBWde-7eSnzOkvWbm7NsgE8RAOo)). We could whiteboard a little tomorrow if you want.

tests/routes/applications/test_new.py
dandds: The `data` property on the response just returns the response body (ultimately it just calls [`werkzeug.wrapper.BaseResponse.get_data()`](https://werkzeug.palletsprojects.com/en/0.15.x/wrappers/#werkzeug.wrappers.BaseResponse.get_data). It returns the body as a byte string, so we call `decode` on it to get a regular Python string. You can do things like:

```
assert application.name in response.data.decode()
```

PR #1091, opened 2019-09-24
AWS Implementation of CloudProviderInterface

richard-dds:
https://www.pivotaltracker.com/story/show/168148337

Implement the `AWSCloudProvider`.

Includes some tests which use a mock version of `boto3`. I tried using [moto](https://github.com/spulec/motourl), but some endpoints weren't implemented and I couldn't figure out how to add them myself.


atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+    def create_environment(
+        self, auth_credentials: Dict, user: User, environment: Environment
+    ):
+
+        org_client = self._get_client("organizations")
+
+        # Create an account. Requires organizations:CreateAccount permission
+        account_request = org_client.create_account(
+            Email=user.email,
+            # TODO: Is a uuid fine here, or do we want something human-readable?
```
dandds: Given that we're going to use a federated organization model, I think the answer to this is yes. Some combo of the portfolio/application/environment name would make sense, but we'd need to talk about the parameters and whether they necessitate constraints on those fields in AT-AT. (AWS naming rules are [here](https://docs.aws.amazon.com/organizations/latest/userguide/orgs_reference_limits.html#name-limits) for reference.)

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+        # A waiter is a boto3 helper which can be configured to poll a given status
+        # endpoint until it succeeds or fails. boto3 has many built in waiters, but none
+        # for the organizations service so we're building our own here.
+        waiter_config = {
+            "version": 2,
+            "waiters": {
+                "AccountCreated": {
+                    "operation": "DescribeCreateAccountStatus",
+                    "delay": 20,
+                    "maxAttempts": 10,
```
dandds: Does it throw an exception after hitting this max or just return the failure state? Also, should we move this out to a constant?

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+            # Poll until the CreateAccount request either succeeds or fails.
+            account_waiter.wait(
+                CreateAccountRequestId=account_request["CreateAccountStatus"]["Id"]
+            )
+        except WaiterError:
+            # TODO: Possible failure reasons:
+            # 'ACCOUNT_LIMIT_EXCEEDED'|'EMAIL_ALREADY_EXISTS'|'INVALID_ADDRESS'|'INVALID_EMAIL'|'CONCURRENT_ACCOUNT_MODIFICATION'|'INTERNAL_FAILURE'
+            raise EnvironmentCreationException(
+                environment.id, "Failed to create account."
+            )
```
dandds: We might need to clarify this in the docstrings, but I think the intention with the error classes was that the second argument should be serialized info from the original exception, like `[Exception type]: [reason the SDK gives why the thing failed]`.

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+
+        # We need to re-fetch this since the Waiter throws away the success response for some reason.
```
dandds: Still can't believe this, it's like if a deli told you your sandwich order was done but then they threw the sandwich in the trash

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
         self._delay(1, 5)
         if credentials != self._auth_credentials:
             raise self.AUTHENTICATION_EXCEPTION
+
+
+class AWSCloudProvider(CloudProviderInterface):
+    BASELINE_POLICIES = [
```
dandds: Could you leave a comment explaining that this is a temporary standin?

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+
+    def create_atat_admin_user(
+        self, auth_credentials: Dict, csp_environment_id: str
+    ) -> Dict:
+        """
+        Create an IAM user within a given account.
+        """
+
+        # Create a policy which allows user to assume a role within the account.
+        # Another async call.
```
dandds: I still favor putting this in a background task, but is this actually async? We don't have a waiter for it, do we?

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+        # credentials that can be used to make subsequent API calls
+        credentials = assumed_role_object["Credentials"]
+
+        # Use the temporary credentials that AssumeRole returns to make a new connection to IAM
+        iam_client = self.boto3.client(
+            "iam",
+            aws_access_key_id=credentials["AccessKeyId"],
+            aws_secret_access_key=credentials["SecretAccessKey"],
+            aws_session_token=credentials["SessionToken"],
+        )
```
dandds: It's a little weird to see a client being initialized outside `_get_client` because we use that everywhere else, but I guess we'd have to normalize the credentials dictionary we get back from AWS to use it here?

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+
+        # Create the user with a PermissionBoundary
+        permission_boundary_arn = "arn:aws:iam::aws:policy/AlexaForBusinessDeviceSetup"
```
dandds: Two things:
- Can we move this to a constant?
- It occurred to me that this might be a custom policy we have to create in the new environment first. If have to do even more work here, and we're going with a federated model, it makes me want to re-assess whether we need an AT-AT IAM user or we should just do role-assumption every time. I can see arguments for both.

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+        access_key = iam_client.create_access_key(UserName=self.root_account_username)[
+            "AccessKey"
+        ]
+        credentials = {
+            "key": access_key["AccessKeyId"],
+            "secret_key": access_key["SecretAccessKey"],
+        }
+
+        # TODO: Create real policies in account.
+
```
dandds: If we're keeping the IAM user approach, should we delete the inline policy we added on 580, just by way of cleanup?

atst/domain/csp/cloud.py
```diff
@@ -442,3 +445,266 @@ def _authorize(self, credentials):
...
+                ]
+                }
+            ]
+        }
+        """
+        )
+        policy_dict["Statement"][0]["Resource"][0] = policy_dict["Statement"][0][
+            "Resource"
+        ][0].format(account_id, self.root_account_policy_name)
+        return json.dumps(policy_dict)
```
dandds: Not a high priority, but maybe this could just be a Jinja template?

tests/mock_boto3.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+@pytest.fixture(scope="function")
+def mock_boto3(request):
+    marks = request.node.get_closest_marker("mock_boto3")
+    if marks:
+        mock_config = marks.args[0] if len(marks.args) else {}
+    else:
+        mock_config = {}
+    return MockBoto3(mock_config)
```
dandds: Neat, I hadn't seen this pytest pattern implemented before.

atst/domain/csp/cloud.py
richard-dds: I believe that either a timeout or an outright error will both result in a `WaiterError`, which we're handling.

atst/domain/csp/cloud.py
richard-dds: @dandds From the `WaiterError` I have access to this info: ```{'name': 'AccountCreated', 'reason': 'Waiter encountered a terminal failure state'}```. What do you think the second arg should be in that case?

atst/domain/csp/cloud.py
richard-dds: 🍔 🗑

atst/domain/csp/cloud.py
richard-dds: I normalized the credentials everywhere and switched to `_get_client` here

atst/domain/csp/cloud.py
richard-dds: I agree with what you're saying here, but given that we're planning on changing our user approach anyway, maybe we can put this off?

atst/domain/csp/cloud.py
dandds: Let's punt on really fine-grained error handling for now and circle back.

PR #1094, opened 2019-09-25
updates to script/test and JS coverage

dandds:
This makes a few changes to our tests:

1. Measure JS test coverage for entire js directory. This gives us more accurate numbers for our JS test coverage.
2. Render Vue component templates as part of test script. This ensures that any changes made to the Jinja components are reflected when JS Vue component tests are run.
3. Apply section heads to test output for clarity. This is just to make the test output more user-friendly. I'm not tied to any color or format as long as it clearly delineates the output sections.

These changes aren't inextricably tied together, so I'm happy to break this into multiple PRs if we want to debate them separately.

### DIFF
--- a/script/include/helper_functions.inc.sh
+++ b/script/include/helper_functions.inc.sh
@@ -53,3 +53,19 @@ reset_db() {
   # Seed database data
   seed_db
 }
+
+output_divider() {
+  if tty -s; then
+    echo "$( tput bold )$( tput smul )$( tput setaf 6 )${1}$( tput sgr 0)"
+  else
+    echo $1
+  fi
+}
+
+warning() {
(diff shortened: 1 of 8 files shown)


script/include/helper_functions.inc.sh
```diff
@@ -53,3 +53,19 @@ reset_db() {
...
+output_divider() {
+  if tty -s; then
+    echo "$( tput bold )$( tput smul )$( tput setaf 6 )${1}$( tput sgr 0)"
+  else
+    echo $1
+  fi
+}
+
+warning() {
+  if tty -s; then
```
leigh-mil: where is the function being used? I don't see it called anywhere in `run_test`

PR #1097, opened 2019-09-26
Styling error page

leigh-mil:
## Description
Update the styling of the error page. I was unable to find an icon on Font Awesome similar to the designs, so I used the existing cloud icon for now.

## Pivotal
https://www.pivotaltracker.com/story/show/168545107

## Screenshots
Screen Shot 2019-09-30 at 11 29 45 AM

### DIFF
--- a/atst/routes/errors.py
+++ b/atst/routes/errors.py
@@ -26,25 +26,10 @@ def notify(e, message, code):
         current_app.notification_sender.send(message)
 
 
-def handle_error(
-    e,
-    message=translate("errors.not_found"),
-    code=404,
-    submessage=translate("errors.not_found_sub"),
-    more_info=translate("common.lorem"),
-):
+def handle_error(e, message=translate("errors.not_found"), code=404):
     log_error(e)
     notify(e, message, code)
-    return (
-        render_template(
-            "error.html",
-            message=message,
-            code=code,
-            submessage=submessage,
-            more_info=more_info,
-        ),
-        code,
-    )
+    return (render_template("error.html", message=message, code=code), code)
 
 
 def make_error_pages(app):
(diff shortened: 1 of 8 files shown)


styles/components/_error_page.scss
```diff
@@ -0,0 +1,33 @@
+.error-page {
+  max-width: 475px;
+  margin: auto;
+
+  .panel__heading {
```
graham-dds: with this and `.panel__body`, you can use the `&` alias to make these rules slightly more concise. It looks like we use that here: https://github.com/dod-ccpo/atst/blob/master/styles/elements/_panels.scss#L55-L77

So this would be something like
```
.panel {
  &__heading {
    ...
  }
  &__body {
    ...
  }
}
```

atst/routes/errors.py
```diff
@@ -25,10 +26,25 @@ def notify(e, message, code):
...
 
 
-def handle_error(e, message="Not Found", code=404):
+def handle_error(
+    e,
+    message=translate("errors.not_found"),
+    code=404,
+    submessage=translate("errors.not_found_sub"),
+    more_info=translate("common.lorem"),
+):
```
dandds: We talked IRL, but I think we might be getting too granular with the kwargs here. Let's try switching on the error code and assume that the text for every error type will be static.

templates/error.html
```diff
@@ -1,19 +1,25 @@
...
-  <p>Return <a href="{{ url_for("atst.home") }}">home</a>.</p>
-{% endif %}
+{% block content %}
 
+<main class="usa-section usa-content error-page">
+    <div class="panel__heading">
+      {{ Icon('cloud', classes="icon--red icon--large")}}
+      <hr>
+      <h1>{{ code }} - {{ message }}</h1>
+      <p>{{ "errors.not_found_sub" | translate }}</p>
```
dandds: This is where I would switch on `code` so that it renders a different message for something that's not a 404. It can be generic for now:

```
<p>
{{ if code == 404 }}
{{ "errors.not_found_sub" | translate }}
{{ else }}
{{ "errors.a_generic_error" | translate }}
{{ endif }}
</p>
```

PR #1098, opened 2019-09-30
Add members to app provisioning flow

graham-dds:
Related to: https://www.pivotaltracker.com/story/show/168412308

This PR adds inviting members to the mutli-step application provisioning flow. 
 - add step 3 templates
 - move the member-invite / management screen into a reusable macro

Coming up in the next/ final PR for this story:
 - rename routes so that they're all /applications/<application_id>/new/step_<#>.html
 - add / change copy & styles according to Think designs

### DIFF
--- a/atst/routes/applications/new.py
+++ b/atst/routes/applications/new.py
@@ -7,6 +7,7 @@ from atst.forms.application import NameAndDescriptionForm, EnvironmentsForm
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.models.permissions import Permissions
 from atst.utils.flash import formatted_flash as flash
+from atst.routes.applications.settings import get_members_data, get_new_member_form
 
 
 def get_new_application_form(form_data, form_class, application_id=None):
@@ -130,3 +131,19 @@ def update_new_application_step_2(portfolio_id, application_id):
             ),
             400,
         )
+
+
+@applications_bp.route("/applications/<application_id>/step_3")
+@user_can(Permissions.CREATE_APPLICATION, message="view create new application form")
+def view_new_application_step_3(application_id):
+    application = Applications.get(application_id)
+    members = get_members_data(application)
+    new_member_form = get_new_member_form(application)
+
+    return render_template(
+        "portfolios/applications/new/step_3.html",
+        application_id=application_id,
+        application=application,
+        members=members,
+        new_member_form=new_member_form,
+    )
--- a/atst/routes/applications/new.py
+++ b/atst/routes/applications/new.py
@@ -29,6 +29,7 @@ def render_new_application_form(
     if application_id:
         application = Applications.get(application_id)
         render_args["form"] = form or form_class(obj=application)
+        render_args["application"] = application
     else:
         render_args["form"] = form or form_class()
 
--- a/atst/routes/applications/new.py
+++ b/atst/routes/applications/new.py
@@ -98,12 +98,19 @@ def create_or_update_new_application_step_1(portfolio_id, application_id=None):
 )
 @user_can(Permissions.CREATE_APPLICATION, message="view create new application form")
 def view_new_application_step_2(portfolio_id, application_id):
-    return render_new_application_form(
-        "applications/new/step_2.html",
-        EnvironmentsForm,
-        portfolio_id=portfolio_id,
-        application_id=application_id,
-    )
+    application = Applications.get(application_id)
+    render_args = {
+        "form": EnvironmentsForm(
+            data={
+                "environment_names": [
+                    environment.name for environment in application.environments
+                ]
+            }
+        ),
+        "application": application,
+    }
+
+    return render_template("portfolios/applications/new/step_2.html", **render_args)
 
 
 @applications_bp.route(
@@ -120,8 +127,8 @@ def update_new_application_step_2(portfolio_id, application_id):
         flash("application_created", application_name=application.name)
         return redirect(
             url_for(
-                "applications.portfolio_applications",
-                portfolio_id=application.portfolio_id,
+                "applications.update_new_application_step_3",
+                application_id=application_id,
             )
         )
     else:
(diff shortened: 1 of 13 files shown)


tests/routes/applications/test_new.py
```diff
@@ -94,3 +96,66 @@ def test_post_environments(client, session, user_session):
...
+    user_session(application.portfolio.owner)
+    response = client.get(
+        url_for(
+            "applications.view_new_application_step_3", application_id=application.id
+        )
+    )
+    assert response.status_code == 200
+
+
+def test_post_member(monkeypatch, client, user_session, session):
```
graham-dds: This test is nearly identical to https://github.com/dod-ccpo/atst/blob/ae57baf455d906121754f79941dbffbdedce0fe2/tests/routes/applications/test_settings.py#L358. The business logic they are testing is the same, but since that logic is related two routes, I included that test here as well. I'm wondering if there's a better way to accomplish this, though 🤔

atst/routes/applications/new.py
```diff
@@ -7,6 +7,11 @@
 from atst.domain.authz.decorator import user_can_access_decorator as user_can
 from atst.models.permissions import Permissions
 from atst.utils.flash import formatted_flash as flash
+from atst.routes.applications.settings import (
```
graham-dds: Since these functions are being used here and in settings, they should probably moved to a more general module in the next PR, yes?

atst/routes/applications/new.py
```diff
@@ -130,3 +143,32 @@ def update_new_application_step_2(portfolio_id, application_id):
             ),
             400,
         )
+
+
+@applications_bp.route("/applications/<application_id>/step_3")
+@user_can(Permissions.CREATE_APPLICATION, message="view create new application form")
```
leigh-mil: The routes for the other steps are nested under `/portfolios/<portfolio_id>/`, is there a reason that the route is not nested here too?

graham-dds: I'll address this in the next PR! 

> Coming up in the next/ final PR for this story:
> 
>     * rename routes so that they're all /applications/<application_id>/new/step_<#>.html
> 
>     * add / change copy & styles according to Think designs

graham-dds: Unless you think it's important to change in this one?

atst/routes/applications/settings.py
```diff
@@ -73,8 +73,9 @@ def filter_env_roles_form_data(member, environments):
             "environment_name": env.name,
             "role": NO_ACCESS,
         }
-        env_role = EnvironmentRoles.get_by_user_and_environment(member.user_id, env.id)
-        if env_role:
+        env_roles_set = set(env.roles).intersection(set(member.environment_roles))
+        if len(env_roles_set) == 1:
```
leigh-mil: 😍 
Would it make sense to replace `get_by_user_and_environment()` with `set(env.roles).intersection(set(member.environment_roles))`?

atst/routes/applications/settings.py
```diff
@@ -73,8 +73,9 @@ def filter_env_roles_form_data(member, environments):
             "environment_name": env.name,
             "role": NO_ACCESS,
         }
-        env_role = EnvironmentRoles.get_by_user_and_environment(member.user_id, env.id)
-        if env_role:
+        env_roles_set = set(env.roles).intersection(set(member.environment_roles))
```
leigh-mil: 😍
Would it make sense to replace get_by_user_and_environment() with set(env.roles).intersection(set(member.environment_roles)) everywhere?

translations.yaml
```diff
@@ -302,8 +302,10 @@ portfolios:
     add_member: Add a New Team Member
     add_another_environment: Add another environment
     app_settings_text: App settings
-    create_button_text: Create
-    next_button_text: "Next: Environments"
+    new:
+      step_1_button_text: "Save and add Environments"
+      step_2_button_text: "Save and add Members"
```
leigh-mil: `Add` should be capitalized here

atst/routes/applications/settings.py
dandds: Oh yeesh, I forgot we were using that in other places. This is actually a more complicated question. There are some places where it's a valid way to do the lookup. For instance, [here](https://github.com/dod-ccpo/atst/blob/master/atst/routes/applications/__init__.py#L18-L27) it's being used to check whether the current user has access to an environment. Each use needs to be assessed in context. I think we need a separate PT task to assess them and make adjustments as necessary.

dandds: Let me amend the above a little: I looked closer, and the instance I mentioned is the only other use of that method, so it's fine. Just don't use it anywhere else!

atst/routes/applications/new.py
```diff
@@ -93,12 +99,19 @@ def create_or_update_new_application_step_1(portfolio_id, application_id=None):
...
-        EnvironmentsForm,
-        portfolio_id=portfolio_id,
-        application_id=application_id,
-    )
+    application = Applications.get(application_id)
+    render_args = {
+        "form": EnvironmentsForm(
+            data={
+                "environment_names": [
+                    environment.name for environment in application.environments
```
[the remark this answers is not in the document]
graham-dds: That's true, but things might get tricky if I have to implement a "previous" button. at that point, the application will have environments, but the provisioning won't be "done". Maybe I can reuse the environments component from the settings page 🤔

translations.yaml
graham-dds: Updated in https://github.com/dod-ccpo/atst/pull/1098/commits/00547a9b34947fc8f4bea6390b2fd3d470e85651

atst/routes/applications/settings.py
```diff
@@ -153,6 +154,41 @@ def send_application_invitation(invitee_email, inviter_name, token):
     )
 
 
+def handle_create_member(application_id, http_request):
```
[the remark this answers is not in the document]
graham-dds: Good call -- updated in dfcdee391048e82b2074cbe741a537aa539e39ba

PR #1099, opened 2019-09-30
Resend invite

leigh-mil:
## Description
Adds in the functionality to resend an application invitation. I did not make the styling match the Invision designs perfectly since I know there will be changes to the styling of all modals and I did not update the resend/revoke buttons because I know designs may be changing for those as well. 

## Pivotal
https://www.pivotaltracker.com/story/show/168459552

## Screenshot
Screen Shot 2019-09-30 at 4 58 56 PM


templates/applications/settings.html
```diff
@@ -75,11 +74,14 @@
...
-    application, 
-    members, 
+    application,
+    members,
     new_member_form,
     "applications.create_member",
-    user_can(permissions.CREATE_APPLICATION_MEMBER)) }}
+    user_can(permissions.CREATE_APPLICATION_MEMBER),
+    user_can(permissions.EDIT_APPLICATION_MEMBER),
+    user_can(permissions.DELETE_APPLICATION_MEMBER),
```
dandds: I would use kwargs when passing booleans like this; it makes things more readable.

tests/routes/applications/test_settings.py
```diff
@@ -591,3 +592,33 @@ def test_filter_environment_roles():
...
+                "dod_id": user.dod_id,
+                "email": "an_email@example.com",
+            },
+        )
+
+        session.refresh(app_role)
+        assert response.status_code == 302
+        assert invite.is_revoked
+        assert app_role.status == ApplicationRoleStatus.PENDING
+        assert app_role.latest_invitation.email == "an_email@example.com"
```
dandds: Any time you want a test to show that you changed a value (`email` here) I like to hard-code the original value so that it's obvious to a reader that it changed. In this case, you'd do that when you initialize the app invite factory.

PR #1106, opened 2019-10-03
PoP Range error message

leigh-mil:
## Description
Fixes issue with the error messages that the PoP dates display. Since the error messages were based off of whether or not the date was within a min or max value, which were set based on the other date in the range. For example the start date's max value is the end date, and if the start date was after its max, the message displayed referenced the end date. However, on a new form if a user entered a start date that was after the contract end, the end date error message would appear. 
To fix this, I added a computed property to the date selector to see if the date is within the contract range, so we can use that to show a different error message.

## Pivotal
https://www.pivotaltracker.com/story/show/168408495

## Screenshots
Screen Shot 2019-10-03 at 4 10 03 PM
Screen Shot 2019-10-03 at 4 10 30 PM

### DIFF
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -38,6 +38,22 @@ def validate_funding(form, field):
         )
 
 
+def validate_date_in_range(form, field):
+    contract_start = app.config.get("CONTRACT_START_DATE")
+    contract_end = app.config.get("CONTRACT_END_DATE")
+
+    if field.data and (field.data < contract_start or field.data > contract_end):
+        raise ValidationError(
+            translate(
+                "forms.task_order.pop_errors.range",
+                {
+                    "start": contract_start.strftime("%b %d, %Y"),
+                    "end": contract_end.strftime("%b %d, %Y"),
+                },
+            )
+        )
+
+
 class CLINForm(FlaskForm):
     jedi_clin_type = SelectField(
         translate("task_orders.form.clin_type_label"),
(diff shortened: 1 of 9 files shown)


atst/forms/task_order.py
```diff
@@ -109,12 +111,30 @@ def validate(self, *args, **kwargs):
...
+            valid = False
+
+        if self.end_date.data and self.end_date.data < contract_start:
+            self.end_date.errors.append(
+                translate(
+                    "forms.task_order.pop_errors.end_pre_contract",
+                    {"date": contract_start.strftime("%b %d, %Y")},
+                )
+            )
+            valid = False
```
dandds: I should have caught this on previous PRs, but we should move these validations to be field-specific (like `validate_funding`, used on line 76. You would add a function for each of these checks and then raise `ValidationError` with the appropriate error message if they fail. That's the preferred WTForms pattern for validation.

templates/components/pop_date_range.html
```diff
@@ -142,11 +149,14 @@
                   {{ 'task_orders.form.pop_example' | translate }}
                 </p>
 
-                <div v-if='minError' class="usa-input-error-message">
+                <div v-if='outsideRange && !maxError' class="usa-input-error-message">
+                  PoP end date must be on or after {{ mindate | formattedDate(formatter="%B %d, %Y") }}.
```
dandds: Translations file 😬 ?

PR #1107, opened 2019-10-04
Multistep app provisioning design tweaks

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/168902811

This PR mostly reorganizes markup and tweaks styles so that major building blocks in the multi-step app provisioning flow are in the right place. It also:
 - makes application descriptions optional
 - adds a "context" option to the CTAHeader, so that I could add the `step {#} of 3` text
 - makes the routes for the different provisioning steps more consistent
 - adds action buttons where appropriate

To come in a future PR (or PRs):
 - pair with Think to get copy consistent across the steps. For example, sometimes they reference `Projects`, and other times they use `Applications`. Are they referring to the same thing? Or are those two separate entities?
- refactor the environments component. As it is now, this flow and the app settings flow use separate components to manage creating environments. We should make the behavior similar for both. 
- make it so that users can only go through this flow once. After, they should use the app settings page.

Screen Shot 2019-10-04 at 11 58 33 AM
Screen Shot 2019-10-04 at 12 07 36 PM
Screen Shot 2019-10-04 at 12 04 57 PM
Screen Shot 2019-10-04 at 12 03 07 PM


styles/components/_portfolio_layout.scss
```diff
@@ -138,85 +138,22 @@
       color: $color-gray;
     }
 
-    table {
```
graham-dds: this felt good.

translations.yaml
```diff
@@ -308,8 +309,38 @@ portfolios:
     add_another_environment: Add another environment
     app_settings_text: App settings
     new:
+      step_1_header: Name and Describe New Project
       step_1_button_text: "Save and Add Environments"
+      step_1_form_help_text:
+        name: |
+          <div style="margin-top: -3rem;">
```
graham-dds: that inline style is . . . ugly. But, in a future PR, I want to be able to add this to the text component as the `description` kwarg so that it goes in the right place. That would be a part of a bigger PR to align our help text / field description placement with USWDSv1 
image.

templates/fragments/members.html
```diff
@@ -15,10 +15,16 @@
 ) %}
 
   
-  {% include "fragments/flash.html" %}
+  {% if g.matchesPath("application-members") %}
+      {% include "fragments/flash.html" %}
+  {% endif %}
 
+  <div class="subheading">
```
graham-dds: I did my best to ensure all of the small headings above a main component within the app provisioning flow used a `subheading` element. That way, in a future PR, we'll be able to make the styling of all subheading consistently inline with Think's designs.

templates/fragments/members.html
```diff
@@ -78,7 +72,11 @@ <h1>{{ Icon('avatar') }} {{ member.user_name }}</h1>
             </base-form>
           {% endcall %}
         {% endfor %}
-        <table>
+
+    
+    <section class="member-list application-list" id="application-members">
+      <div class='responsive-table-wrapper'>
+        <table class="atat-table">
```
graham-dds: There were a lot of custom styles on kind of a per-page basis. It was nice to be able to use this component class here.  It would be nice to roll back some more "one-off" styles in favor of more component-level classes.

atst/routes/applications/new.py
```diff
@@ -72,12 +69,14 @@ def create_or_update_new_application_step_1(portfolio_id, application_id=None):
         if application_id:
             application = Applications.get(application_id)
             application = Applications.update(application, form.data)
+            flash("application_updated", application_name=application.name)
         else:
+            portfolio = Portfolios.get_for_update(portfolio_id)
             application = Applications.create(g.current_user, portfolio, **form.data)
```
dandds: Now that `description` is optional in the form, passing the form data this way will create an application with an empty string for the description if the user didn't enter one. We want the value to be `None` in those cases.

atst/routes/applications/new.py
```diff
@@ -135,17 +130,13 @@ def update_new_application_step_2(portfolio_id, application_id):
     else:
         return (
             render_new_application_form(
-                "applications/new/step_2.html",
-                EnvironmentsForm,
-                portfolio_id,
-                application_id,
-                form,
+                "applications/new/step_2.html", EnvironmentsForm, application_id, form
```
dandds: I think you need to specify the kwargs here since you're skipping `portfolio_id` in the positional list.

graham-dds:
@dandds  and I discussed

> Do we have designs for adding an environment on the app settings page?">

and

> I don't know that they'd hurt anything by using [the first time flow] twice, though, so this might not be high priority.

Offline. Plan is to not touch environment creation in this PR, and instead sync with Think to nail down what different parts of the environment creation / editing components do, and translate that into another PR.

atst/routes/applications/new.py
graham-dds: Added a filter to handle this here: https://github.com/dod-ccpo/atst/pull/1107/commits/237123d474407289de51cb0d1337cb95c264947e

atst/models/application.py
```diff
@@ -15,7 +15,7 @@ class Application(
 
     id = Id()
     name = Column(String, nullable=False)
-    description = Column(String, nullable=False)
+    description = Column(String)
```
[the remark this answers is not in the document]
graham-dds: Added here: https://github.com/dod-ccpo/atst/pull/1107/commits/237123d474407289de51cb0d1337cb95c264947e

PR #1108, opened 2019-10-04
Delete CSP users

richard-dds:
https://www.pivotaltracker.com/story/show/168180264

Implement the task which deletes a user in the CSP when a user's `EnvironmentRole` is deleted, as well as wiring it up to the frontend.

I did some manual testing with the mock cloud, and it looks like things are working beautifully:
Screen Shot 2019-10-09 at 11 10 26 AM

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -8,8 +8,6 @@ from flask_session import Session
 import redis
 from unipath import Path
 from flask_wtf.csrf import CSRFProtect
-import json
-from enum import Enum
 
 from atst.database import db
 from atst.assets import environment as assets_environment
@@ -31,7 +29,7 @@ from atst.models.permissions import Permissions
 from atst.queue import celery, update_celery
 from atst.utils import mailer
 from atst.utils.form_cache import FormCache
-from atst.utils.json import CustomJSONEncoder
+from atst.utils.json import CustomJSONEncoder, sqlalchemy_dumps
 from atst.utils.notification_sender import NotificationSender
 from atst.utils.session_limiter import SessionLimiter
 
@@ -150,15 +148,6 @@ def set_default_headers(app):  # pragma: no cover
 
 
 def map_config(config):
-    def sqlalchemy_dumps(dct):
-        def _default(obj):
-            if isinstance(obj, Enum):
-                return obj.name
-            else:
-                raise TypeError()
-
-        return json.dumps(dct, default=_default)
-
     return {
         **config["default"],
         "ENV": config["default"]["ENVIRONMENT"],
--- a/atst/app.py
+++ b/atst/app.py
@@ -9,6 +9,7 @@ import redis
 from unipath import Path
 from flask_wtf.csrf import CSRFProtect
 import json
+from enum import Enum
 
 from atst.database import db
 from atst.assets import environment as assets_environment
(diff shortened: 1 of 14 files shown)


atst/app.py
```diff
@@ -148,6 +150,15 @@ def _set_security_headers(response):
 
 
 def map_config(config):
+    def sqlalchemy_dumps(dct):
```
richard-dds: I had to provide a custom json serializer because the `AuditableMixin` couldn't serialize our `EnvironmentRole.Status` enum. I'm not sure exactly where to put this.

dandds: We have a `CustomJSONEncoder` and we modify the app's default encoder with it elsewhere in this module. This is slightly different since we need to modify the SQLAlchemy encoder. I'd probably still put it in `atst.utils.json` with the other one, though.

atst/domain/environment_roles.py
```diff
@@ -87,3 +87,17 @@ def get_environment_roles_pending_creation(cls) -> List[UUID]:
...
+    @classmethod
+    def get_environment_roles_pending_deletion(cls) -> List[UUID]:
+        results = (
+            db.session.query(EnvironmentRole.id)
+            .join(Environment)
+            .join(ApplicationRole)
+            .filter(Environment.deleted == False)
+            .filter(Environment.baseline_info != None)
+            .filter(EnvironmentRole.status == EnvironmentRole.Status.PENDING_DELETE)
+            .filter(ApplicationRole.status == ApplicationRoleStatus.ACTIVE)
```
dandds: This reminds me of two things:
- Do you plan to implement the status switch on this PR? (i.e., user deletes an env role in the interface and we toggle the role to `PENDING_DELETE`). We had also talked about revoking all their CSP-side perms inline in the same route function to that they can't do anything while the deletion is pending.
- We'd need similar functionality to flip every env role associated with an app role if an app role is deleted.

atst/jobs.py
```diff
@@ -112,6 +116,19 @@ def do_provision_user(csp: CloudProviderInterface, environment_role_id=None):
...
+
+
+def do_delete_user(csp: CloudProviderInterface, environment_role_id=None):
+    environment_role = EnvironmentRoles.get_by_id(environment_role_id)
+
+    with claim_for_update(environment_role) as environment_role:
+        credentials = environment_role.environment.csp_credentials
+
+        csp.delete_user(credentials, environment_role.csp_user_id)
+        environment_role.status = EnvironmentRole.Status.DELETED
```
dandds: Do we need to soft-delete `environment_role` here as well? That might be happening in one of the associated domain methods that the route function calls, but it seems cleaner to set all the real "deleted" state here. (And do we need both the soft delete and the DELETED status? Maybe, might be worth talking through.)

atst/app.py
richard-dds: I can't believe I didn't see `utils.json`!

atst/domain/environment_roles.py
richard-dds: So I did end up implementing the status thing in this PR. Also I think that that `ApplicationRole.delete` is already doing the work of calling `EnvironmentRole.delete`, which is what's triggering the status update that results in the user being deleted in the CSP.

PR #1111, opened 2019-10-09
More multistep app provisioning design tweaks

graham-dds:
This PR just tweaks a few things that I missed in my last PR for styling multi-app provisioning.
This PR will close out https://www.pivotaltracker.com/story/show/168902811

The styling and functionality of the [environments component](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382268619) and the [members component](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382268620) will happen in future cards.

### DIFF
--- a/js/components/forms/new_application/name_and_description.js
+++ b/js/components/forms/new_application/name_and_description.js
@@ -10,6 +10,10 @@ export default {
   components: {
     textinput,
   },
+  created: function() {
+    this.$root.$on('field-change', this.handleFieldChange)
+    if (this.initialData) this.changed = true
+  },
 
   data: function() {
     return {
--- a/styles/elements/_action_group.scss
+++ b/styles/elements/_action_group.scss
@@ -29,3 +29,24 @@
     }
   }
 }
+
+.action-group-footer {
+  @extend .action-group;
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+  margin-top: 0;
+  margin-bottom: 0;
+  padding-top: $gap;
+  padding-bottom: $gap;
+
+  position: fixed;
+  bottom: $footer-height;
+  background: white;
+  right: 0;
+  padding-right: $gap * 4;
+  border-top: 1px solid $color-gray-light;
+  width: 100%;
+  z-index: 1;
+}
--- a/templates/applications/new/step_1.html
+++ b/templates/applications/new/step_1.html
@@ -28,7 +28,7 @@
           <div class="form-row">
             <div class="form-col form-col--two-thirds">
               {{ TextInput(form.name, optional=False) }}
-              {{ ('portfolios.applications.new.step_1_form_help_text.description' | translate | safe) }}
+              {{ ('portfolios.applications.new.step_1_form_help_text.name' | translate | safe) }}
             </div>
           </div>
           <hr>
--- a/templates/applications/new/step_1.html
+++ b/templates/applications/new/step_1.html
@@ -45,7 +45,8 @@
         {% block next_button %}
           {{ SaveButton(text=('portfolios.applications.new.step_1_button_text' | translate)) }}
         {% endblock %}
-        <a class="usa-button usa-button-secondary" href="{{ url_for('applications.portfolio_applications', portfolio_id=portfolio.id) }}">
+        <button disabled class="usa-button usa-button-secondary">Previous</button>
+        <a href="{{ url_for('applications.portfolio_applications', portfolio_id=portfolio.id) }}">
           Cancel
         </a>
       </span>
--- a/templates/applications/new/step_1.html
+++ b/templates/applications/new/step_1.html
@@ -38,7 +38,8 @@
         </div>
       </div>
 
-      <span class="action-group">
+
+      <span class="action-group-footer">
         {% block next_button %}
           {{ SaveButton(text=('portfolios.applications.new.step_1_button_text' | translate)) }}
         {% endblock %}
--- a/templates/applications/new/step_1.html
+++ b/templates/applications/new/step_1.html
@@ -23,23 +23,20 @@
 
   <application-name-and-description inline-template v-bind:initial-data='{{ form.data|tojson }}'>
     <form method="POST" action="{{ action }}" v-on:submit="handleSubmit">
-        <div class="panel__content">
-          {{ form.csrf_token }}
-          <div class="form-row">
-            <div class="form-col form-col--two-thirds">
-              {{ TextInput(form.name, optional=False) }}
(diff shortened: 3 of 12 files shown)


leigh-mil:
Should users be able to make an app without a description? I got a sqlalchemy error when I didn't add one! (I have a screenshot of the error if you want it)

I also noticed a few discrepancies from the designs that Think gave us.
Overall differences: 
- Task Order Number is missing at the top of each form step.
- Should the buttons on each page be in a sticky footer? I remember this being mentioned, but I'm not sure what the final design decision was. 
- `.panel__content` should have 0 padding on the left

Step 1:
- Label text needs to be updated and normal font weight.

Step 2: 
- flash message text needs to be updated

Step 3: 
- Should there be a flash message on this page? There isn't one in the designs.
- Text in sticky CTA needs to be updated

graham-dds:
> Should users be able to make an app without a description? I got a sqlalchemy error when I didn't add one!

That was the cue I took from the optional label in the mockups. There was a migration to make the description optional in this PR: https://github.com/dod-ccpo/atst/pull/1107/files -- do you still get that error even after applying the migration?

graham-dds:
> Task Order Number is missing at the top of each form step.

@dandds and I were talking about this, and I think we came to the conclusion that it doesn't quite make sense? There's no explicit linking of applications and TOs, so it's not really clear which TO number supposed to appear there

leigh-mil:
> > Should users be able to make an app without a description? I got a sqlalchemy error when I didn't add one!
> 
> That was the cue I took from the optional label in the mockups. There was a migration to make the description optional in this PR: https://github.com/dod-ccpo/atst/pull/1107/files -- do you still get that error even after applying the migration?

Works now! Guess I hadn't ran migrations in a while!

graham-dds:
@leigh-mil pushed some commits that should take care of everything except the sticky nav

templates/applications/new/step_1.html
```diff
@@ -23,29 +23,27 @@
...
-          <hr class="panel__break">
-          <div class="form-row">
-            <div class="form-col form-col--two-thirds">
-              {{ TextInput(form.description, paragraph=True, optional=True) }}
-              {{ ('portfolios.applications.new.step_1_form_help_text.description' | translate | safe) }}
-            </div>
-          </div>
+      {{ form.csrf_token }}
+      <div class="form-row">
+        <div class="form-col form-col--two-thirds">
```
leigh-mil: I don't think we need these two divs here since we are not using the classes to create columns.

styles/elements/_action_group.scss
```diff
@@ -29,3 +29,24 @@
...
+  margin-bottom: 0;
+  padding-top: $gap;
+  padding-bottom: $gap;
+
+  position: fixed;
+  bottom: $footer-height;
+  background: white;
+  right: 0;
+  padding-right: $gap * 4;
+  border-top: 1px solid $color-gray;
```
leigh-mil: I think this should be one of the lighter gray colors

templates/applications/new/step_3.html
```diff
@@ -11,31 +11,30 @@
...
-    {{ MemberManagementTemplate(
-      application, 
-      members, 
-      new_member_form,
-      "applications.update_new_application_step_3",
-      user_can(permissions.CREATE_APPLICATION_MEMBER)) }}
-      
+  <span class="action-group-footer">
+    <a class="usa-button" href="{{ url_for('applications.settings', application_id=application_id) }}">  
+      Return to Application Settings
```
leigh-mil: This button should say `Save Application`

graham-dds: I went with "Return to . . . " because that button really doesn't "save" anything. At this point, the application is saved, and all this button does is link back to the application settings page.

js/components/forms/new_application/name_and_description.js
```diff
@@ -10,6 +10,11 @@ export default {
   components: {
     textinput,
   },
+  created: function() {
+    this.$root.$on('field-change', this.handleFieldChange)
+    console.log(this.initialData)
```
dandds: Can we delete the logger here?

styles/core/_spacing.scss
```diff
@@ -0,0 +1,40 @@
...
+    $posKey: nth($position, 1);
+    $posValue: nth($position, 2);
+    .#{$marginKey}#{$separator}#{$posKey}#{$separator}#{$sizeKey} {
+      margin-#{$posValue}: sizeValue($sizeKey, $sizeValue);
+    }
+    .#{$paddingKey}#{$separator}#{$posKey}#{$separator}#{$sizeKey} {
+      padding-#{$posValue}: sizeValue($sizeKey, $sizeValue);
+    }
+  }
+}
```
dandds: Feels like functional CSS. Is `m-t-6` the only use case for this so far?

I'm a little ambivalent on adding something like this because it creates a divergence in the stylesheet: pre-existing entities have classes applied based on what they _are_, not how they _look_. The latter isn't a wrong approach, it's just not what we have anywhere else. We should talk about it. If you have a strong case ("I'm going to use it to implement XYZ components and can refactor these three pages") we could consider this.

js/components/forms/new_application/name_and_description.js
graham-dds: Oops!

styles/core/_spacing.scss
graham-dds: > Is m-t-6 the only use case for this so far?

So far, but I think @leigh-mil had plans to use it in a few places?

I don't have a strong case at this point (and I don't know how much time we actually have to do any refactoring), but I do feel like this is a nicer way to apply a little padding here and a little margin there, rather than adding style via a hyper-specifc selector in our stylesheets.

Also, a more robust version of these spacing helpers [appear](https://designsystem.digital.gov/utilities/margin-and-padding/) in the USWDSv2.

Definitely down to discuss this more though.

leigh-mil: @graham-dds @dandds I may use some of these, but so far I have been able to standardize styles across the site using the existing classes and deleting the hyper specific stuff.

dandds: @graham-dds Ah, if there is a way to mirror the USWDSv2 class names _or_ just pull those in directly from v2, I would rather do that. That way we're paving the way for an upgrade path instead of rolling our own. (They don't exist in any form in v1, I take it?)

graham-dds: Welp, I removed those spacing classes just to see what things look like without them, and it turns out they're not doing anything visually now. I was able to safely remove `_spacing.scss` and those classes without too much hassle, so I'll just keep them out of this PR 🎉

PR #1112, opened 2019-10-10
App settings styling (part 1)

leigh-mil:
## Description
Updates the styling for the `Name and description` section of the app settings page and the portfolio header. Also makes some global changes to inputs based on the designs and conversations with Think:
- Removes the span that was a placeholder for error messages
- Updated the margin around `usa-input` elements to be aligned with the new designs and fixed any side effects from this change

## Pivotal
https://www.pivotaltracker.com/story/show/168511969

## Screenshot
Screen Shot 2019-10-10 at 1 01 40 PM


translations.yaml
```diff
@@ -283,8 +283,8 @@ login:
 navigation:
   portfolio_navigation:
     breadcrumbs:
-      admin: Admin
-      funding: Funding
+      admin: Settings
+      funding: Task Orders
```
graham-dds: I know that one of the advantages of this "translations" model is that we can easily change/ update copy, but it's kind of confusing to me when I see something like `funding: Task Orders`.  Would it be worth it to change the key to more closely align to the value here? Or maybe it's better to name the breadcrumb keys something more generic (like `breadcrumb_{n}`), so that if / when we change copy like this again, we won't have this same issue?

translations.yaml
```diff
@@ -314,7 +314,7 @@ portfolios:
       step_1_button_text: "Save and Add Environments"
       step_1_form_help_text:
         name: |
-          <div style="margin-top: -3rem;">
```
graham-dds: 🎉 thank you!

templates/applications/settings.html
```diff
@@ -19,58 +19,40 @@
...
-          </div>
-          <div class="panel__footer">
-            <div class="action-group">
-              {{ SaveButton('common.save_changes'|translate) }}
-            </div>
-          </div>
+      <form method="POST" action="{{ url_for('applications.update', application_id=application.id) }}" class="col col--half">
+        {{ application_form.csrf_token }}
+        {{ TextInput(application_form.name, optional=False) }}
+        {{ TextInput(application_form.description, paragraph=True, optional=True, showOptional=False) }}
```
graham-dds: I know that this isn't how it's shown in the mockups, but since this field is indeed optional, should we have `showOptional` be `True`? It's how that field [appears](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382268615) in the new app flow. I know they're slightly different contexts, though, so I'm not sure if this was an intentional design decision by Think or not.

templates/applications/settings.html
```diff
@@ -19,58 +19,40 @@
...
-          <div class="form-col">
-            <div class="usa-input usa-input__title__view-only">
-              {{ application_form.description.label() }}
-            </div>
-            <p>
-              {{ application_form.description.data }}
-            </p>
-          </div>
-        </div>
+    <!-- TODO: use new spacing styling to add in bottom margin here -->
```
graham-dds: I removed these guys from my PR!

templates/applications/settings.html
leigh-mil: @graham-dds I assumed this was intentional since we are on the settings page and not in a form, but I'll double check with them

templates/applications/settings.html
leigh-mil: @graham-dds Thats fine, I'll fix the spacing here a different way!

translations.yaml
leigh-mil: We have a chore for cleaning up the translations, and I added this to it!

PR #1114, opened 2019-10-14
App settings styling - environments table

leigh-mil:
## Description
Update styling for environments table on the app settings page.

## Pivotal
https://www.pivotaltracker.com/story/show/168511969

## Screenshots
Screen Shot 2019-10-14 at 1 45 12 PM
Screen Shot 2019-10-14 at 1 58 47 PM
Screen Shot 2019-10-14 at 1 59 00 PM
Screen Shot 2019-10-14 at 1 59 14 PM
Screen Shot 2019-10-14 at 1 59 18 PM


templates/fragments/environments.html
```diff
@@ -0,0 +1,101 @@
+{% from "components/icon.html" import Icon %}
```
dandds: If a shared fragment is specific to a domain (like applications) I think it should go in the fragments subdir for that domain, i.e. "templates/applications/fragments".

leigh-mil: 👍 Do you mind if I move that in a follow up PR? My plan is to have a follow up pr after I do each section to make sure everything is standardized across the sections since it is not just me working on styling these sections!

PR #1119, opened 2019-10-15
Resend invite fixes

leigh-mil:
## Description
Fixes 2 bugs with the resend invite flow:
1. Updates copy
2. Makes Resend button active when the modal first opens

## Pivotal
https://www.pivotaltracker.com/story/show/168459552

### DIFF
--- a/templates/fragments/members.html
+++ b/templates/fragments/members.html
@@ -17,13 +17,16 @@
 ) %}
 
 
+
+
+  <div class="subheading"  id="application-members">
+    {{ 'portfolios.applications.settings.team_members' | translate }}
+  </div>
+
   {% if g.matchesPath("application-members") %}
       {% include "fragments/flash.html" %}
   {% endif %}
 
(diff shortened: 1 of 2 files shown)


templates/fragments/members.html
```diff
@@ -87,7 +90,7 @@ <h1>Verify Member Information</h1>
                 {{ member.update_invite_form.csrf_token }}
                 {{ member_fields.InfoFields(member.update_invite_form) }}
                 <div class="action-group">
-                  {{ SaveButton(text='Resend Invite', element='input', additional_classes='action-group__action') }}
+                  <input type="submit" class="usa-button usa-button-primary action-group__action" tabindex="0" value="Resend Invite" />
```
dandds: I think the problem is a little more nuanced, unfortunately: the button should be enabled when the modal opens, but become disabled if I enter invalid data. For example, if I delete the users last name, I can still submit the form.

PR #1121, opened 2019-10-16
Make Postgres SSL connection configurable.

dandds:
This will allow us to force SSL connections to the database in
production by setting two values:

- PGSSLMODE should be set to "verify-full". This forces the client to
  verify the server against a known CA: https://www.postgresql.org/docs/10/libpq-ssl.html
- PGSSLROOTCERT should be set to the path of the public cert for the
  relevant CA.

When the database connection is made, these values are passed to the
adapter. For local development, PGSSLMODE is set to "prefer" and
PGSSLROOTCERT is left unset.

Kubernetes config has been added to maintain the root CAs for both Azure
and AWS as k8s ConfigMap objects. These are mounted into the containers
and referenced by PGSSLROOTCERT in the container environment.


config/base.ini
```diff
@@ -19,6 +19,7 @@ PGDATABASE = atat
 PGHOST = localhost
 PGPASSWORD = postgres
 PGPORT = 5432
+PGSSLMODE = prefer
```
graham-dds: Do we need a corresponding value in `config/prod.ini` to be `PGSSLMODE = require`? Or is that set somewhere else?

dandds: Right now, it would be set in a special override file in the Kubernetes cluster.

Somewhat confusingly, the two staging sites right now run using the dev environment (so the `/login-dev` backdoor and other things are available). We need `require` or `verify-full` set there, but we can't have that set when we use the dev environment locally. Basically, I don't know if the idea of environment-specific config is as useful to us as it could be. Might be something to sort out on another PR.

dandds:
Changing this to WIP because I'm going to do most of the upfront config for SSL here.

dandds:
Hmm, I need to check that any additional instances of the container that need a database connection get the updated config.

PR #1123, opened 2019-10-16
Ignore js/test_templates

leigh-mil:
Add js/test_templates to .gitignore file and removed the newly ignored folder


script/include/run_test
```diff
@@ -49,6 +49,7 @@ fi
 
 if [ "${RUN_JS_TESTS}" = "true" ]; then
   output_divider "Render Vue components"
+  mkdir js/test_templates | true
```
graham-dds: Could we have one of the scripts used in `script/test` nuke the `js/test_templates` directory when it's finished? Or is that something that we don't want to happen?

dandds: Since it's git-ignored I don't think it especially matters. If you feel strongly it should get cleaned up, we could.

graham-dds: Definitely don't feel strongly about it

PR #1127, opened 2019-10-18
Audit log feature flag

graham-dds:
This PR puts the audit event log that appears in the app behind a feature flag -- `AUDIT_LOG_FEATURE_TOGGLE` in `base.ini`.

Addresses https://www.pivotaltracker.com/story/show/169164096

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -150,6 +150,9 @@ def set_default_headers(app):  # pragma: no cover
 def map_config(config):
     return {
         **config["default"],
+        "AUDIT_LOG_FEATURE_TOGGLE": config["default"].getboolean(
+            "AUDIT_LOG_FEATURE_TOGGLE"
+        ),
         "ENV": config["default"]["ENVIRONMENT"],
         "BROKER_URL": config["default"]["REDIS_URI"],
         "DEBUG": config["default"].getboolean("DEBUG"),
--- a/atst/app.py
+++ b/atst/app.py
@@ -150,9 +150,7 @@ def set_default_headers(app):  # pragma: no cover
 def map_config(config):
     return {
         **config["default"],
-        "AUDIT_LOG_FEATURE_TOGGLE": config["default"].getboolean(
-            "AUDIT_LOG_FEATURE_TOGGLE"
-        ),
+        "USE_AUDIT_LOG": config["default"].getboolean("USE_AUDIT_LOG"),
         "ENV": config["default"]["ENVIRONMENT"],
         "BROKER_URL": config["default"]["REDIS_URI"],
         "DEBUG": config["default"].getboolean("DEBUG"),
--- a/atst/models/mixins/auditable.py
+++ b/atst/models/mixins/auditable.py
@@ -17,24 +17,29 @@ class AuditableMixin(object):
         if changed_state is None:
             changed_state = resource.history if action == ACTION_UPDATE else None
 
-        audit_event = AuditEvent(
-            user_id=user_id,
-            portfolio_id=resource.portfolio_id,
-            application_id=resource.application_id,
-            resource_type=resource.resource_type,
-            resource_id=resource.id,
-            display_name=resource.displayname,
-            action=action,
-            changed_state=changed_state,
-            event_details=resource.event_details,
-        )
+        log_data = {
+            "user_id": user_id,
+            "portfolio_id": resource.portfolio_id,
+            "application_id": resource.application_id,
+            "resource_type": resource.resource_type,
+            "resource_id": resource.id,
+            "display_name": resource.displayname,
+            "action": action,
+            "changed_state": changed_state,
+            "event_details": resource.event_details,
+        }
 
         app.logger.info(
             "Audit Event {}".format(action),
-            extra={"audit_event": audit_event.log, "tags": ["audit_event", action]},
+            extra={
+                "audit_event": {key: str(value) for key, value in log_data.items()},
+                "tags": ["audit_event", action],
+            },
         )
-        audit_event.save(connection)
-        return audit_event
+
+        if app.config.get("AUDIT_LOG_FEATURE_TOGGLE", False):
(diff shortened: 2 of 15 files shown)


atst/models/mixins/auditable.py
```diff
@@ -17,24 +17,30 @@ def create_audit_event(connection, resource, action, changed_state=None):
...
+                "tags": ["audit_event", action],
+            },
         )
-        audit_event.save(connection)
-        return audit_event
+
+        if app.config.get("AUDIT_LOG_FEATURE_TOGGLE", False):
+            audit_event = AuditEvent(**log_data)
+            audit_event.save(connection)
+            return audit_event
```
graham-dds: The current version of this function returns `audit_event`, but I don't think it goes on to get used anywhere.

tests/conftest.py
```diff
@@ -38,6 +38,14 @@ def app(request):
     ctx.pop()
 
 
+@pytest.fixture(autouse=True)
+def skip_audit_log(request):
```
graham-dds: With this fixture, any test marked with `@pytest.mark.audit_log` will be skipped

graham-dds:
I marked any test that used an `AuditEvent` or one of of its methods with `@pytest.mark.audit_log`. These tests automatically are skipped when `AUDIT_LOG_FEATURE_TOGGLE` is set to `false`. The test suite is passing, but I don't know if the flag skips too many tests.

templates/portfolios/admin.html
```diff
@@ -63,8 +63,8 @@
     {% if user_can(permissions.VIEW_PORTFOLIO_USERS) %}
       {% include "portfolios/fragments/portfolio_members.html" %}
     {% endif %}
-
-    {% if user_can(permissions.VIEW_PORTFOLIO_ACTIVITY_LOG) %}
+    
+    {% if user_can(permissions.VIEW_PORTFOLIO_ACTIVITY_LOG) and config.get("AUDIT_LOG_FEATURE_TOGGLE", False) %}
```
dandds: Nit-picky, but this supplies a default to `get` and the previous use in the app settings template doesn't.

atst/app.py
```diff
@@ -150,6 +150,9 @@ def _set_security_headers(response):
 def map_config(config):
     return {
         **config["default"],
+        "AUDIT_LOG_FEATURE_TOGGLE": config["default"].getboolean(
+            "AUDIT_LOG_FEATURE_TOGGLE"
+        ),
```
dandds: I'd suggest making the setting name more explicit. I'm not sure just reading it whether setting `AUDIT_LOG_FEATURE_TOGGLE` to be `True` turns it on or off. Maybe just `AUDIT_LOG` or `USE_AUDIT_LOG`?

tests/conftest.py
```diff
@@ -38,6 +38,14 @@ def app(request):
...
+    config = make_config()
+    if request.node.get_closest_marker("audit_log"):
+        if not config.get("AUDIT_LOG_FEATURE_TOGGLE", False):
```
dandds: Is there a more straightforward way to phrase this condition? It reads as `if not false` which always confuses me.

graham-dds: > if not false which always confuses me

Agreed, but I'm not sure about the best way to improve this.

I want to keep the meaning of the config variable such that `True` enables the log and `Flase` disables it.


I could use a short-circut:
```python
not config.get("USE_AUDIT_LOG", False) and pytest.skip(
            "audit log feature flag disabled"
        )
```

or maybe add an else clause, and just `pass` in the `True` case
```python
if config.get("USE_AUDIT_LOG", False):
    pass
else:
    pytest.skip("audit log feature flag disabled")
```

Any preference, or can you think of another way?

dandds: Maybe more readable would be to stash the config lookup result in a variable:
```
use_audit_log = config.get("USE_AUDIT_LOG", False)
```
Then structure the condition the way you have it. What do you think?

atst/models/mixins/auditable.py
leigh-mil: Would it make sense to have this return `audit_event.log` when `USE_AUDIT_LOG` is true, and when it is false return `log_data`? That way the same type of data is being returned no matter if the config is set to true to false

templates/applications/settings.html
```diff
@@ -117,7 +117,7 @@ <h1>{{ "portfolios.applications.delete.header" | translate }}</h1>
 
   <hr>
 
-  {% if user_can(permissions.VIEW_APPLICATION_ACTIVITY_LOG) %}
+  {% if user_can(permissions.VIEW_APPLICATION_ACTIVITY_LOG) and config.get("USE_AUDIT_LOG", False) %}
```
leigh-mil: should this be `config.get("USE_AUDIT_LOG", True)`?

atst/models/mixins/auditable.py
graham-dds: @leigh-mil I was thinking of even removing the return statement, because I don't think the return value is used in either case

leigh-mil: that works too!

atst/models/mixins/auditable.py
```diff
@@ -17,24 +17,29 @@ def create_audit_event(connection, resource, action, changed_state=None):
...
+            "action": action,
+            "changed_state": changed_state,
+            "event_details": resource.event_details,
+        }
 
         app.logger.info(
             "Audit Event {}".format(action),
-            extra={"audit_event": audit_event.log, "tags": ["audit_event", action]},
+            extra={
+                "audit_event": {key: str(value) for key, value in log_data.items()},
```
dandds: Don't do anything on this PR, but can you sync up with @richard-dds about getting something in the backlog to handle serializing UUIDs by default in our JSON serializer? It's annoying that we have to do this kind of stuff.

PR #1138, opened 2019-10-25
App settings bugfixes

leigh-mil:
## Description
Removes the horizontal scroll bar from modals.
Fixes the app members table so that there is a link to revoke invite for members with pending invites.

## Pivotal
https://www.pivotaltracker.com/story/show/169189658
https://www.pivotaltracker.com/story/show/169325622

### DIFF
--- a/styles/components/_modal.scss
+++ b/styles/components/_modal.scss
@@ -53,6 +53,7 @@ body {
     padding: $gap * 1.5;
     width: 100%;
     overflow-y: auto;
+    overflow-x: hidden;
     -ms-overflow-style: scrollbar;
 
     @include ie-only {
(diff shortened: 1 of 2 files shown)


styles/components/_modal.scss
```diff
@@ -53,6 +53,7 @@ body {
     padding: $gap * 1.5;
     width: 100%;
     overflow-y: auto;
+    overflow-x: hidden;
```
graham-dds: As you and I talked about on Friday, the offending element in the modals is the `<hr>` -- it's getting a negative right and left margin added to it by `_normalize.scss`. So, this works, but it could also make sense to limit the width of the HR element.

PR #1139, opened 2019-10-25
App invitation modal step 2 changes

graham-dds:
Updates some text in step 2 of inviting a member to an application.
Addresses https://www.pivotaltracker.com/story/show/169343220

| Before | After |
|--------|-------|
| | |

Note: The text in the dropdowns still says `"no access"` by default, even though the mockups have the default value as `"Select Access Level"`. I talked with JT on Slack, and he said that we could keep the default option as is.

### DIFF
--- a/translations.yaml
+++ b/translations.yaml
@@ -393,18 +393,21 @@ portfolios:
       blank_slate: This Application has no members
       form:
         env_mgmt:
-          label: Manage Environments
-          description: Add and rename project environments, assign environment access roles to team members.
+          label: Manage Application Environments
+          description: Add and rename Environments, assign Environment access roles to team members.
         team_mgmt:
-          label: Edit Team
+          label: Edit Application Team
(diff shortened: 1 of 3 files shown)


translations.yaml
```diff
@@ -393,18 +393,21 @@ portfolios:
...
-        env_access: Environment Access
+        env_access: 
+          title: Environment Roles
+          description: Additional role controls are available in the CSP console. <a href="#"> Learn More </a>
         next_button: "Next: Permissions"
-        project_perms: Project Permissions
-        step_2_title: Set Permissions and Roles
+        app_perms:
+          title: Application Permissions
+          description: Application permissions allow users to provision and modify applications and teams. <a href="#"> Learn More </a>
```
leigh-mil: You don't need to fix it in this PR, but I know in the past we avoided putting in dead links since they were confusing for the user.

PR #1142, opened 2019-10-28
Bugfix: Invited but disabled member shouldn't be able to reactivate self

graham-dds:
Fixes: https://www.pivotaltracker.com/story/show/169047971

When removing an member with a pending invitation from a portfolio, it was previously possible for that member to follow the invitation link and join the portfolio because the invitation was never actually revoked. This PR adds the invitation revoking logic to the `remove_member()` route. 

Also when fixing this PR, since I used a similar method that was used for Application invitation revoking, I moved that logic from the child classes to the base class.


atst/routes/portfolios/admin.py
```diff
@@ -183,10 +185,15 @@ def remove_member(portfolio_id, portfolio_role_id):
...
-    # roles they might have?
-    PortfolioRoles.disable(portfolio_role=portfolio_role)
+    if (
+        portfolio_role.latest_invitation
+        and portfolio_role.status == PortfolioRoleStatus.PENDING
+    ):
+        PortfolioInvitations.revoke(portfolio_role.latest_invitation.token)
+    else:
+        # TODO: should this cascade and disable any application and environment
+        # roles they might have?
```
dandds: Not necessarily. They're qualitatively different kinds of access. You might be removed as a portfolio member (someone who cares about money) but retain access as an application member (someone who cares about administering environments or does work in the CSP).

dandds: Oh, I see this isn't your TODO originally. We can get rid of it, I think.

PR #1146, opened 2019-10-28
Add revoke access to app members perms/env roles form

leigh-mil:
## Description
Adds in the functionality to revoke access to an environment.

## Pivotal
https://www.pivotaltracker.com/story/show/169343479
https://www.pivotaltracker.com/story/show/169343526

## Screenshot
Screen Shot 2019-10-31 at 10 15 49 AM
Screen Shot 2019-10-31 at 10 20 21 AM


atst/domain/environments.py
```diff
@@ -52,24 +52,24 @@ def get(cls, environment_id):
...
+            env_role = EnvironmentRoles.create(
+                application_role=application_role,
+                environment=environment,
+                role=new_role,
+            )
+            updated = True
+
+        if env_role and not new_role:
+            env_role.role = None
             updated = EnvironmentRoles.delete(application_role.id, environment.id)
```
dandds: This should call `EnvironmentRoles.disable` instead. That class method uses the CSP interface to suspend the user in the CSP. We should probably get rid of `EnvironmentRoles.delete` and use `disable` instead everywhere. @richard-dds Did we decide whether an environment role should be soft-deleted when a user's access to an environment is removed?

atst/domain/environments.py
```diff
@@ -52,24 +59,28 @@ def get(cls, environment_id):
...
+        elif not env_role and new_role:
+            env_role = EnvironmentRoles.create(
+                application_role=application_role,
+                environment=environment,
+                role=new_role,
+            )
+            updated = True
+
+        if env_role and not new_role:
+            env_role.role = None
```
leigh-mil: env roles get updated in EnvironmentRoles.disable(), so this isn't needed

atst/routes/applications/settings.py
```diff
@@ -73,11 +77,16 @@ def filter_env_roles_form_data(member, environments):
             "environment_id": str(env.id),
             "environment_name": env.name,
             "role": NO_ACCESS,
+            "disabled": False,
         }
-        env_roles_set = set(env.roles).intersection(set(member.environment_roles))
+        env_roles_set = set(env.roles).intersection(
+            set(EnvironmentRoles.get_all_for_application_member(member.id))
+        )
+
```
graham-dds: Curious, why was the change needed here? It looks like in both cases, the argument that gets passed to `intersection()` is a list of environment roles that belong to a given member.

atst/routes/applications/settings.py
```diff
@@ -367,14 +376,21 @@ def remove_member(application_id, application_role_id):
 @user_can(Permissions.EDIT_APPLICATION_MEMBER, message="update application member")
 def update_member(application_id, application_role_id):
     app_role = ApplicationRoles.get_by_id(application_role_id)
-    form = UpdateMemberForm(http_request.form)
+    application = Applications.get(application_id)
+    existing_env_roles_data = filter_env_roles_form_data(
+        app_role, application.environments
+    )
+    form = UpdateMemberForm(
+        formdata=http_request.form, environment_roles=existing_env_roles_data
```
graham-dds: Just asking because I'm unsure, does this initialize the form with `existing_env_roles_data` and then update the values with `http_request.form`?

templates/applications/fragments/environments.html
```diff
@@ -64,8 +64,9 @@ <h3>{{ "portfolios.applications.settings.environments" | translate }}</h3>
                   {% call ToggleSection(section_name="members") %}
                     <ul>
                       {% for member in env['members'] %}
```
graham-dds: It might be a little neater to make member a dict instead of a tuple -- if I had to change something here in the future, it would take me a second to figure out what `member[0]` was.

atst/routes/applications/settings.py
leigh-mil: @graham-dds This is because `application_role.environment_roles` does not include env roles that have a disabled status, but they are needed for displaying the form. 
However, we wanted to change it so that disabled roles were included. Good catch!

atst/routes/applications/settings.py
leigh-mil: @graham-dds Exactly! I had to use `filter_env_roles_form_data` because roles where the user was never assigned access were not being populated in the form

templates/applications/fragments/member_form_fields.html
```diff
@@ -1,7 +1,88 @@
...
+      <fieldset data-ally-disabled="true" v-on:change="onInput" class="usa-input__choices revoke-button">
+        {% set id = "{}-{}".format(field.disabled.name, member_role_id) %}
+        <div class="form-row" v-if="!isChecked">
+          <div class="form-col form-col--two-thirds">
+            <div class="usa-input__title-inline">
+              {{ field.environment_name.data }}
+            </div>
+            <p class="usa-input__help">
+              {{ role }}
+            </p>
```
dandds: NBD, but `div` and `p` as siblings is a little weird.

templates/applications/fragments/member_form_fields.html
```diff
@@ -1,7 +1,88 @@
+{% from "components/alert.html" import Alert %}
 {% from "components/checkbox_input.html" import CheckboxInput %}
 {% from "components/text_input.html" import TextInput %}
 {% from "components/phone_input.html" import PhoneInput %}
 
+{% macro EnvRoleInput(field, member_role_id=None) %}
```
dandds: The `field` arg here is actually a form, right? Is it clearer to call it `form`?

atst/domain/environments.py
```diff
@@ -52,24 +59,28 @@ def get(cls, environment_id):
...
+            env_role = EnvironmentRoles.create(
+                application_role=application_role,
+                environment=environment,
+                role=new_role,
+            )
+            updated = True
+
+        if env_role and not new_role:
+            EnvironmentRoles.disable(env_role.id)
+            updated = True
```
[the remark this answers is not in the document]
leigh-mil: @dandds I think we still need to get all of the roles, even if they have been disabled, because on line 70 if there is no env_role, one is created (An example is this test: `test_update_env_role_deleted_role`). Here is where we check to make sure that a disabled environment isn't being updated with a new role

PR #1155, opened 2019-10-31
Ghost Inspector to GitHub

JRNPW:
This PR addresses PT card https://www.pivotaltracker.com/story/show/169489660

I have exported the Ghost Inspector "AT-AT CI" suite of tests (the tests which run in the CircleCI workflow) to files that are compatible with the "old" Selenium IDE. The files are very readable HTML which could be imported into Selenium. They could also serve as a guideline for creating test scripts in other frameworks.

I also updated the README to indicate why this directory exists and expanded the information provided about Ghost Inspector.

It is my plan to update these files regularly (hopefully when I create new or update existing tests and ensure that they pass at least once.


JRNPW:
Bleargh! Oops! That reminds me why I usually don't do `git commit -am` when I have changed more than one file. Broke my own darn rule.

PR #1162, opened 2019-11-04
Env name bugfix

leigh-mil:
## Description
Fixes a bug where a user could create an env with spaces as the name. The changes here were made to the validators and filters, so it _should_ prevent this from happening in other places as well. 

The validator ListItemRequired() was only checking for None and an empty string, not for strings that were multiple whitespace characters. This fixes this issue by checking each item with regex to make sure it contains non whitespace characters

The filter remove_empty_string() also was not checking for strings that were multiple whitespace characters. This was also fixed by using regex to make sure that the string contains non whitespace characters, and also clips any trailing whitespace.

## Pivotal
https://www.pivotaltracker.com/story/show/169368323

### DIFF
--- a/atst/forms/forms.py
+++ b/atst/forms/forms.py
@@ -1,5 +1,6 @@
 from flask_wtf import FlaskForm
 from flask import current_app, request as http_request
+import re
 
 from atst.utils.flash import formatted_flash as flash
 
@@ -38,4 +39,8 @@ class BaseForm(FlaskForm):
 
     @classmethod
     def remove_empty_string(cls, value):
-        return value or None
+        # only return strings that contain non whitespace characters
+        if value and re.search(r"\S", value):
+            return value.strip()
+        else:
+            return None
--- a/atst/forms/forms.py
+++ b/atst/forms/forms.py
@@ -7,6 +7,14 @@ from atst.utils.flash import formatted_flash as flash
 EMPTY_LIST_FIELD = ["", None]
 
 
+def remove_empty_string(value):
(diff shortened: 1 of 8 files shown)


JRNPW:
@leigh-mil Please pull the latest changes from `master` into this branch. It is failing Ghost Inspector tests because it isn't up-to-date.

graham-dds:
Found [this](https://github.com/alphagov/notifications-admin/pull/1701/files#diff-e618e3e19b02f84ab65f8d0cef43212bR165-R186) in a PR in another repo. Might be nice to go that route and make a custom field instead of having to add the filter in multiple places.

atst/forms/forms.py
```diff
@@ -38,4 +39,8 @@ def validate(self, *args, flash_invalid=True, **kwargs):
 
     @classmethod
     def remove_empty_string(cls, value):
-        return value or None
+        # only return strings that contain non whitespace characters
+        if value and re.search(r"\S", value):
+            return value.strip()
+        else:
+            return None
```
dandds: I'm not sure there's a strong reason to prefer this as a class method instead of a plain utility function you import as-needed.

tests/forms/test_base_form.py
```diff
@@ -16,6 +16,12 @@ class FormWithChoices(BaseForm):
     )
 
 
+class FormWithList(BaseForm):
+    list = FieldList(
```
dandds: Small thing, but `list` is default class name in Python.

PR #1163, opened 2019-11-04
Fix data['fields'] in Vue component forms

leigh-mil:
## Description
When investigating this bug, I discovered that the issue was that when there are multiple forms on a page they capture all of the form fields on the page, so the form was never 'valid' because of the extra fields associated with the form. This issue was fixed by checking on form Vue components that the parent_id passed in the field-mount event is the forms uid or the uid of one of the forms child components.

## Pivotal
https://www.pivotaltracker.com/story/show/169327255

### DIFF
--- a/js/components/forms/multi_step_modal_form.js
+++ b/js/components/forms/multi_step_modal_form.js
@@ -67,8 +67,10 @@ export default {
       return valid
     },
     handleFieldMount: function(event) {
-      const { name, optional } = event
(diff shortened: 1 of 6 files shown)


dandds:
@leigh-mil This is the one we're shelving in favor of tackling the technical debt, right?

leigh-mil:
Closing this in favor of doing tech debt!

PR #1164, opened 2019-11-05
Environment bugfixes

leigh-mil:
## Description
Fixes two bugs:
1.) When an app invite is revoked, the user should not appear in the env members drop down list. Fixed by updating the relationship between Environment and EnvironmentRole so only environment roles that are not deleted are included.
2.) Removes the optional tag from the env name edit form.

## Pivotal
https://www.pivotaltracker.com/story/show/169183145
https://www.pivotaltracker.com/story/show/169368053

### DIFF
--- a/atst/models/environment.py
+++ b/atst/models/environment.py
@@ -32,6 +32,12 @@ class Environment(
 
     job_failures = relationship("EnvironmentJobFailure")
 
+    roles = relationship(
+        "EnvironmentRole",
+        back_populates="environment",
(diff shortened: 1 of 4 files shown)


dandds:
Also it would be great to have a test of the `Environment` model that confirms it loads the roles we expect!

leigh-mil:
> Also it would be great to have a test of the `Environment` model that confirms it loads the roles we expect!

@dandds I am going to hold off on adding a test until you, Richard, and I can talk about the issue you brought up on a comment on #1146

PR #1175, opened 2019-11-08
Standardize member and env name ordering

graham-dds:
There were a few bugs that pointed out inconsistencies / lack of ordering. This PR fixes those.

| Story  | Screenshot |
| ------------- | ------------- |
| [Portfolio applications- Environments dropdown](https://www.pivotaltracker.com/story/show/169607538)  |   |
| [App settings -> Application Team](https://www.pivotaltracker.com/story/show/169605561)  |   |
| [App Settings -> Environment members ](https://www.pivotaltracker.com/story/show/169607374)  |   |
| [Portfolio members](https://www.pivotaltracker.com/story/show/169607752) -- note: for this one, the ordering is PPOC, then other portfolio members in alphabetical order.  |   |
| Application Members: [New](https://www.pivotaltracker.com/story/show/169607265) / [Edit](https://www.pivotaltracker.com/story/show/169607235) member modal |   |
| [New Application - Step 2](https://www.pivotaltracker.com/story/show/169606042) |  |
| [New Application - Step 3](https://www.pivotaltracker.com/story/show/169503971) |  |

### DIFF
--- a/atst/models/application.py
+++ b/atst/models/application.py
@@ -25,6 +25,7 @@ class Application(
         primaryjoin=and_(
             Environment.application_id == id, Environment.deleted == False
         ),
+        order_by="Environment.name",
     )
     roles = relationship(
         "ApplicationRole",
--- a/atst/routes/applications/settings.py
+++ b/atst/routes/applications/settings.py
@@ -30,9 +30,9 @@ def get_environments_obj_for_app(application):
                 "pending": env.is_pending,
                 "edit_form": EditEnvironmentForm(obj=env),
                 "member_count": len(env.roles),
-                "members": [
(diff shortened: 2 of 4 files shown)


graham-dds:
>  I think we should be consistent about where we sort. Either it's a display concern that happens when we prepare data for the template, or its a data concern and should be applied with SQL. (I don't know that there's a right answer, but preferably not both.)

True. I like the idea of keeping sorting as close to the SQL as possible -- if we have to sort an application's environment names in three different places, it feels like we should get it that way by default. I went with more `sorted()`s in this PR to avoid SQLAlchemy rabbit holes, though.

>   We might want to sort humans by last name instead.

@JRNPW, any thoughts here?

PR #1179, opened 2019-11-13
Vue emitters fix - TO form

leigh-mil:
## Description
The first of what will probably be many PRS to update the Vue components to properly emit events up the component tree instead of emitting events from the root component. This PR updates the TO form and nested components. Since many of the nested components are used in other forms, I couldn't completely remove the root emitters yet. 

## Pivotal
https://www.pivotaltracker.com/story/show/169708877

### DIFF
--- a/js/components/checkbox_input.js
+++ b/js/components/checkbox_input.js
@@ -25,11 +25,18 @@ export default {
 
   methods: {
     onInput: function(e) {
+      this.$parent.$emit('field-change')
       emitEvent('field-change', this, {
         value: e.target.checked,
         name: this.name,
         valid: this.optional || this.isChecked,
       })
+    },
+  },
+
+  computed: {
+    valid: function() {
+      return this.optional || this.isChecked
     },
   },
 }
--- a/js/components/clin_fields.js
+++ b/js/components/clin_fields.js
@@ -1,5 +1,4 @@
 import { emitEvent } from '../lib/emitters'
-import Modal from '../mixins/modal'
 import optionsinput from './options_input'
 import textinput from './text_input'
 import clindollaramount from './clin_dollar_amount'
@@ -18,8 +17,6 @@ export default {
     clindollaramount,
     PopDateRange,
   },
-
-  mixins: [Modal],
 
   props: {
     initialClinIndex: Number,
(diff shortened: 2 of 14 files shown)


js/mixins/form.js
```diff
@@ -13,10 +13,12 @@ export default {
 
   mounted: function() {
     this.$root.$on('field-change', this.handleFieldChange)
```
dandds: Are these root calls removed in the follow-up PR?

js/components/checkbox_input.js
```diff
@@ -25,11 +25,18 @@ export default {
 
   methods: {
     onInput: function(e) {
+      this.$parent.$emit('field-change')
       emitEvent('field-change', this, {
         value: e.target.checked,
         name: this.name,
         valid: this.optional || this.isChecked,
       })
```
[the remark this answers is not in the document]
leigh-mil: In this PR I only added the new functionality of emitting from the parent in the TO form only. I didn't remove the old emitters because the components are used in several places, and I didn't want them to break in other places. There will be 3 follow up PRs to fix all of the other forms and then delete out the old emitters!

js/mixins/form.js
leigh-mil: Yes, they will be!

js/components/clin_fields.js
```diff
@@ -90,24 +89,25 @@ export default {
...
+        this.clinNumber = event.value
       }
+      this.$parent.$emit('field-change')
     },
 
     removeClin: function() {
       this.showClin = false
       emitEvent('remove-clin', this, {
         clinIndex: this.clinIndex,
       })
```
dandds: I missed this the first time, but shouldn't this also be a `$parent` call?

js/components/forms/to_form.js
```diff
@@ -58,8 +52,6 @@ export default {
           delete this.fields[field]
         }
       }
-
-      this.validateForm()
```
dandds: Can more stuff be removed from this component? For instance, it doesn't need to updates the fields in `handleRemoveClin`, does it?

js/components/clin_fields.js
```diff
@@ -140,5 +140,15 @@ export default {
     removeModalId: function() {
       return `remove-clin-${this.clinIndex}`
     },
+
+    valid: function() {
+      if (this.removed) {
+        // the nested component is still mounted, so valid needs to be true or the
+        // save button will never become active
```
dandds: 😬 We're using a `v-for`loop in the parent `to-form`. I though `v-for` was for looping a a list, but it looks like we treat `clins` as an integer on the `to-form`. The reason I bring this up is because if it were a genuine list of CLIN components, I would expect Vue to unmount any that were no long in the list.

PR #1183, opened 2019-11-14
Set up a CDN

richard-dds:
This adds the config and some overrides necessary to serve assets via a CDN. It entails some other changes:

- The CircleCI config needed to be updated. Our build splits right at the beginning depending on whether we're building an image for staging or master, so the workflows need to be updated to reflect this.
- It adds a yarn command and updates the Dockerfile so that we can specify the CDN url for the frontend assets.
- It adds a default CORS header ("Access-Control-Allow-Origin"). The Azure CDN will reference this the first time it loads assets from the a Flask and use it as its own ACAO header on subsequent CORS requests to the CDN. For reasons explained in the relevant commit, we added this as its own configuration value.
- This overrides `url_for` in the Jinja templates so that it uses the CDN URL whenever the `static` endpoint is referenced. All of the references to static assets in the template have been updated to use `url_for`.
- The K8s config for environment variables on both sites have been updated.


.circleci/config.yml
```diff
@@ -194,6 +209,19 @@ jobs:
...
+            docker pull curlimages/curl:latest
+            docker run --network atat \
+              curlimages/curl:latest \
+              curl --connect-timeout 3 \
+              --max-time 5 \
+              --retry 120 \
+              --retry-connrefused \
+              --retry-delay 1 \
+              --retry-max-time 120 \
+              test-atat:8000
```
tomdds: Can we give this a more descriptive name? It appears that this is using curl to poll something?

dandds: @tomdds done!

PR #1188, opened 2019-11-15
Edit member perms in new app form bugfix

leigh-mil:
## Description
Previously when updating an app members perms through the new app form, it was submitting to the same route as updating an app member through the settings page, which was causing it to redirect to the wrong page after update. This was fixed by posting to the update route for step_3 of the new app form instead.

## Pivotal
https://www.pivotaltracker.com/story/show/169682731


dandds:
And a standalone test for the new `handle` function would be good too.

PR #1196, opened 2019-11-19
Catch CSP exceptions and display a flash message with the error

leigh-mil:
## Description
Catches errors when revoking an env role fails and displays a flash message to the user. Until we get copy for this, I am using the message property on the error for the flash message.

## Pivotal
https://www.pivotaltracker.com/story/show/169852180

## Screenshot
Screen Shot 2019-11-19 at 4 38 49 PM

### DIFF
--- a/atst/routes/applications/settings.py
+++ b/atst/routes/applications/settings.py
@@ -6,6 +6,7 @@ from atst.domain.environments import Environments
 from atst.domain.applications import Applications
 from atst.domain.application_roles import ApplicationRoles
 from atst.domain.audit_log import AuditLog
+from atst.domain.csp.cloud import GeneralCSPException
 from atst.domain.common import Paginator
 from atst.domain.environment_roles import EnvironmentRoles
 from atst.domain.invitations import ApplicationInvitations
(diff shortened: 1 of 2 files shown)


atst/routes/applications/settings.py
```diff
@@ -221,14 +237,24 @@ def handle_update_member(application_id, application_role_id, form_data):
...
 
-        flash("application_member_updated", user_name=app_role.user_name)
+            flash("application_member_updated", user_name=app_role.user_name)
+
+        except CSP_ERRORS as error:
+            flash(
+                "application_member_update_error",
+                user_name=app_role.user_name,
+                message=error.message,
+            )
```
dandds: You can just import and catch `GeneralCSPException`. All of the CSP exceptions are subclasses of that.

atst/utils/flash.py
```diff
@@ -54,6 +54,11 @@
         "message_template": "You have successfully deleted {{ user_name }} from {{ application_name }}",
         "category": "success",
     },
+    "application_member_update_error": {
+        "title_template": "{{ user_name }} could not be updated",
+        "message_template": "{{ message }}",
```
dandds: I would hard-code a general error message. I don't think the real error we raise will be useful information to the user (it will be a problem with ATAT's own CSP creds, or a network failure, etc.). Something like "An unexpected problem occurred with your request. Please try again. If the problem persists, contact an administrator." Maybe ping Ben to massage the initial content.

atst/utils/flash.py
```diff
@@ -54,6 +54,11 @@
         "message_template": "You have successfully deleted {{ user_name }} from {{ application_name }}",
         "category": "success",
     },
+    "application_member_update_error": {
+        "title_template": "{{ user_name }} could not be updated",
+        "message_template": "An unexpected problem occurred with your request, please try again. If the problem persists, contact an adminsitrator.",
```
dandds: Typo. :/

PR #1197, opened 2019-11-19
Remove csp alert

leigh-mil:
## Pivotal
https://www.pivotaltracker.com/story/show/169629795


JRNPW:
I have pulled this branch and run it locally. I can see that the message is no longer there. Code change looks fine, too, but should probably be reviewed by a developer.

PR #1198, opened 2019-11-20
Exclude audit log from coverage

leigh-mil:
## Description
This PR is the result of a research spike into excluding audit log related functions from the pytest coverage report. I updated the `.coveragerc` config file so it now excludes the audit log files, and functions that are part of the AuditableMixin. This results in a coverage bump from 91.52% to 91.82%. I didn't have time to look into excluding the files and functions based on the `USE_AUDIT_LOG` config variable, so everything is just hard coded in for now. 
Here are the documentation on excluding code from coverage: https://coverage.readthedocs.io/en/coverage-4.3.4/excluding.html and https://coverage.readthedocs.io/en/coverage-4.4.2/config.html

## Pivotal
https://www.pivotaltracker.com/story/show/169291416

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,17 @@
 [run]
-omit = atst/routes/dev.py
+omit =
+  atst/routes/dev.py
+  atst/domain/audit_log.py
+  atst/models/mixins/auditable.py
+  atst/models/audit_event.py
 branch = True
+
+[report]
+exclude_lines =
+  pragma: no cover
+  if app.config.get("USE_AUDIT_LOG", False)
+  def event_details
+  def history
+  def resource_type
+  def renderAuditEvent
+  def activity_history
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -1,3 +1,16 @@
...
+
+[report]
+exclude_lines =
+  pragma: no cover
+  if app.config.get("USE_AUDIT_LOG", False)
+  event_details
+  history
+  resource_type
+  renderAuditEvent
+  activity_history
```
dandds: Do we want these exclusion lines to be `def event_details`, etc? There might be other legitimate uses of `history` in the codebase, for example, so we probably want to limit it to just the auditable implementation methods.

leigh-mil: I thought these properties might be included in if statements, which is why I did them this way initially. I just checked though and they are only really used in the property declarations, so I updated it!

PR #1199, opened 2019-11-20
Reporting refactor part 1

graham-dds:
This PR modifies backend logic and does some lofi markup / template reorganization to bring the reporting page closer to [the new designs](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/392385087)

Things to review in this PR:
- new / modified functions and associated tests in the `reports` and `csp/reports` domain classes
- high level changes to the markup in the portfolio summary section and applications / environment spending table

Things not included:
- styling
- final markup for the obligated funds and expired TOs sections. These sections changed as I was working on this PR, so they'll be modified.


templates/components/accordion.html
```diff
@@ -0,0 +1,36 @@
...
+        {{ title }}
+      </button>
+    </template>
+    <template v-else>
+      <button
+        v-on:click="toggle($event)" 
+        class="usa-accordion-button"
+        aria-expanded="false"
+        aria-controls="{{ id }}"
+      >
```
graham-dds: I have to repeat myself here because of the USWDS 1.0 way `v-bind` works with attributes. 

The expand / collapse styling design choices on the [USWDS 1.0 accordion](https://v1.designsystem.digital.gov/components/accordions/) are a little wonky. I would prefer to use v-bind and a conditional for the aria expanded property, but if the [value passed to an attribute set with v-bind is falsey, it will be omitted](https://vuejs.org/v2/guide/syntax.html#Attributes). And ommitted property and `false` don't mean the same thing for this component, so this is the solution I came up with.

atst/domain/csp/reports.py
```diff
@@ -35,14 +36,16 @@ def generate_sample_dates(_max=8):
     current = pendulum.now()
     sample_dates = []
     for _i in range(_max):
```
leigh-mil: I don't think we need to have the max here anymore since the new reports page only shows 2 months of data. This is a holdover from when reporting would show several months of data.

atst/domain/csp/reports.py
```diff
@@ -270,7 +233,14 @@ def monthly_totals_for_environment(self, environment_id):
...
         """
-        return self.MONTHLY_SPEND_BY_ENVIRONMENT.get(environment_id, {})
+        environment_monthly_totals = self.MONTHLY_SPEND_BY_ENVIRONMENT.get(
+            environment_id, {}
+        ).copy()
+
+        environment_monthly_totals["total_spend_to_date"] = sum(
+            monthly_total for monthly_total in environment_monthly_totals.values()
+        )
+        return environment_monthly_totals
```
leigh-mil: I think we need to decide how we are going to fake the expended amounts and apply it consistently across the reporting page. Here you are totaling the faked spending over a series of months, but in `get_obligated_funds_by_JEDI_clin()` the expended amount is a percentage of the obligated amount.

atst/domain/reports.py
```diff
@@ -2,16 +2,48 @@
...
-
     @classmethod
     def monthly_totals(cls, portfolio):
         return current_app.csp.reports.monthly_totals(portfolio)
 
     @classmethod
-    def cumulative_budget(cls, portfolio):
-        return current_app.csp.reports.cumulative_budget(portfolio)
+    def current_obligated_funds(cls, portfolio):
+        return current_app.csp.reports.get_current_obligated_funds(portfolio)
```
leigh-mil: Where is `get_current_obligated_funds` defined? 
Also, should this property or function be defined on the Portfolio?

atst/domain/reports.py
```diff
@@ -2,16 +2,48 @@
...
+            return "{} - {}".format(
+                earliest_pop_start_date.strftime("%B %d, %Y"),
+                latest_pop_end_date.strftime("%B %d, %Y"),
+            )
+        return "-"
+
+    @classmethod
+    def days_remaining(cls, portfolio):
+        days_remaining = portfolio.days_to_funding_expiration
+        return days_remaining if days_remaining else "-"
```
leigh-mil: Should this maybe default to `0` or `"No"`, that way in the UI it will display `0 days` or `No days` which I think is a more readable.

templates/portfolios/reports/expired_task_orders.html
```diff
@@ -0,0 +1,34 @@
+{% from "components/accordion.html" import Accordion %}
+
+
+<div>
+  <ul class="usa-accordion">      
+    {% call Accordion("Expired Task Orders", "expired_task_orders") %}
```
leigh-mil: Is there a reason you did not use the existing `ToggleButton` and `ToggleSection` macros here?

tests/domain/test_reports.py
```diff
@@ -1,27 +1,111 @@
...
-    portfolio = PortfolioFactory.create()
-    monthly = Reports.monthly_totals(portfolio)
 
-    assert not monthly["environments"]
-    assert not monthly["applications"]
-    assert not monthly["portfolio"]
+def test_active_task_orders(session):
+    portfolio = PortfolioFactory()
+    TaskOrderFactory.create(
+        portfolio=portfolio,
```
leigh-mil: I think you need to add `signed_at` as a kwarg here because if a TO is not signed, it can never be active.

tests/domain/test_reports.py
```diff
@@ -1,27 +1,111 @@
...
+        funding_start_date.strftime("%B %d, %Y"),
+        funding_end_date.strftime("%B %d, %Y"),
+    )
+
+    # empty portfolio
+    empty_portfolio = PortfolioFactory()
+    assert Reports.funding_duration(empty_portfolio) == "-"
+
+
+def test_portfolio_with_task_orders(session):
```
leigh-mil: Can you update the name of the test to more accurately reflect what you are testing, since you check both a portfolio with TOs and one without any TOs.

atst/domain/csp/reports.py
leigh-mil: When I made this comment, I didn't realize you were using the previous month's data to calculate the expended amounts

atst/domain/csp/reports.py
leigh-mil: We discussed it in irl and decided that faking this data here is potentially not worth the effort for this PR

templates/portfolios/reports/expired_task_orders.html
leigh-mil: Discussed irl -- existing macros were not used because of the styling for them did not work here.

templates/components/accordion.html
dandds: I'd like to play with this a little. You're probably right but I want to take a little time and see if there's a more minimal way to implement this.

templates/portfolios/reports/application_and_env_spending.html
```diff
@@ -0,0 +1,82 @@
...
+    {{ EmptyState(
+      'Nothing to report',
+      action_label='Add a new application' if can_create_applications else None,
+      action_href=url_for('applications.create_new_application_step_1', portfolio_id=portfolio.id) if can_create_applications else None,
+      icon='chart',
+      sub_message=message,
+      add_perms=can_create_applications
+    ) }}
+  {% else %}
+    <spend-table
```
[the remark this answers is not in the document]
dandds: Nevermind, I see we do that so that we can track applicationState as a series of objects. I don't love this pattern, and we're only doing it because of the table issue I mention below.

templates/portfolios/reports/application_and_env_spending.html
```diff
@@ -0,0 +1,82 @@
...
+                  <span v-html='formatDollars(application[currentMonthIndex] || 0)'></span>
+                </td>
+                <td class="table-cell--align-right">
+                  <span v-html='formatDollars(application[prevMonthIndex] || 0)'></span>
+                </td>
+                <td class="table-cell--align-right">
+                  <span v-html='formatDollars(application["total_spend_to_date"])'></span>
+                </td>
+              </tr>
+              <tr v-for='(environment, envName) in environments[name]' v-show='application.isVisible'>
```
dandds: I have mixed feelings about using a flat table for this, as opposed to a nested table or list. My first reaction is _of course it's a table_ but these rows are qualitatively different kinds of data (environments) than the primary rows (applications). It also requires that we track state in a weird way so that we can append more rows when we need to; we track a list of applications in a parent component, as opposed to having smaller components that track their own state. I know this was the existing pattern here, but let's talk about whether we can come up with something better.

templates/portfolios/reports/expired_task_orders.html
```diff
@@ -0,0 +1,34 @@
...
+      {% for task_order in expired_task_orders %}
+        <a href="{{ url_for("task_orders.review_task_order", task_order_id=task_order["id"]) }}">
+          Task Order {{ task_order["number"] }}
+        </a>
```
dandds: Should this maybe get wrapped in a block or inline-block element?

templates/portfolios/reports/obligated_funds.html
```diff
@@ -0,0 +1,29 @@
...
+        <p>{{ funds["expended_funds"] | dollars }}</p>
+      </div>
+      <hr>
+      {% endfor %}
+      {% for task_order in portfolio.active_task_orders %}
+        <a href="{{ url_for("task_orders.review_task_order", task_order_id=task_order.id) }}">{{ task_order.number }}</a>
+      {% endfor %}
+    </div>
+  </div>
+</div>
```
dandds: It seems like this content should be in a section or something together. It's hard to discern, just reading the template code, why these two top-level divs share a partial. In general, I would consider organizing the page into `section` tags (probably delimited by the `hr` we have now, which could possibly just be a bottom border on each section).

templates/portfolios/reports/application_and_env_spending.html
```diff
@@ -0,0 +1,82 @@
...
+  <h2>Funds Expended per Application and Environment</h2>
+  {% set current_month_index = current_month.strftime('%m/%Y') %}
+  {% set prev_month_index = prev_month.strftime('%m/%Y') %}
+
+  {% if not portfolio.applications %}
+
+    {% set can_create_applications = user_can(permissions.CREATE_APPLICATION) %}
+    {% set message = 'This portfolio has no cloud environments set up, so there is no spending data to report. Create an application with some cloud environments to get started.'
+      if can_create_applications
+      else 'This portfolio has no cloud environments set up, so there is no spending data to report. Contact the portfolio owner to set up some cloud environments.'
```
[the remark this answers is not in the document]
graham-dds: Got it into this PR 👍

templates/portfolios/reports/application_and_env_spending.html
graham-dds: I tried a few different approaches, but gave up for this PR 🙃

templates/portfolios/reports/expired_task_orders.html
graham-dds: This all gets wrapped in a div by the accordion macro

templates/portfolios/reports/obligated_funds.html
```diff
@@ -0,0 +1,31 @@
+
+<section>
+  <header>
+    <h2>Current Obligated funds</h2>
+    <span>As of DATE</span>
+  </header>
+  <div class='panel'>
```
graham-dds: @dandds does this fix https://github.com/dod-ccpo/atst/pull/1199#discussion_r349548718?

PR #1200, opened 2019-11-20
Add CCPO user script

leigh-mil:
## Description
Add script for giving a user CCPO permissions from the command line.

## Pivotal
https://www.pivotaltracker.com/story/show/169814115

## Screenshots
Screen Shot 2019-11-20 at 4 29 01 PM
Screen Shot 2019-11-20 at 4 29 13 PM

### DIFF
--- a/script/add_ccpo_user.py
+++ b/script/add_ccpo_user.py
@@ -1,0 +1,35 @@
+import os
+import sys
+
+parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.append(parent_dir)
+
+from atst.app import make_config, make_app
+
+from atst.domain.exceptions import NotFoundError
+from atst.domain.users import Users
+
+
+def grant_ccpo_perms(dod_id):
+    try:
+        user = Users.get_by_dod_id(dod_id)
+        if user.permission_sets:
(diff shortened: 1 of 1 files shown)


script/add_ccpo_user.py
```diff
@@ -0,0 +1,34 @@
...
+
+    with app.app_context():
+        dod_id = sys.argv[1]
+
+        try:
+            user = Users.get_by_dod_id(dod_id)
+            if user in Users.get_ccpo_users():
+                print(
+                    "%s (DoD ID: %s) already CCPO user." % (user.full_name, user.dod_id)
+                )
```
dandds: Rather than loading all the CCPO users, you could just check that `user.permission_sets != None`. All `get_ccpo_users` does is query for all the users with that permission.

script/add_ccpo_user.py
```diff
@@ -0,0 +1,34 @@
...
+            else:
+                Users.give_ccpo_perms(user)
+                print(
+                    "CCPO permissions successfully granted to %s (DoD ID: %s)."
+                    % (user.full_name, user.dod_id)
+                )
+
+        except NotFoundError:
+            print("User not found.")
```
dandds: I would put lines 20-34 in a standalone function that takes a DOD ID as an argument. It's useful for debugging purposes and lets us compose scripts later if we have to.

PR #1208, opened 2019-11-25
App members edit menu fixes

leigh-mil:
## Description
Fixes two app members edit menu styling bugs: 
1 - Adds a hover state to the edit menu
2 - Menu now closes when focus changes

## Pivotal
https://www.pivotaltracker.com/story/show/169708341
https://www.pivotaltracker.com/story/show/169708287

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -253,6 +253,11 @@
         border-radius: 3px;
         cursor: pointer;
 
+        &:hover,
+        &--active {
+          background-color: $color-aqua-lightest;
+        }
+
         .icon {
           margin: $gap / 2;
         }
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -283,6 +283,10 @@
           &:last-child {
             border-bottom: 0;
           }
+
+          &:hover {
+            background-color: $color-aqua-lightest;
+          }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


JRNPW:
@leigh-mil This is failing `integration-tests` because nothing happens when one clicks the "Return to Application Settings" button after adding a team member during the Create a New Application flow. You can watch the interactions of Ghost Inspector [here](https://app.ghostinspector.com/tests/5daf6473dd63d511764df27a).

I verified what's happening by pulling your branch and trying the same steps locally.

styles/components/_portfolio_layout.scss
```diff
@@ -253,6 +253,11 @@
         border-radius: 3px;
         cursor: pointer;
 
+        &:hover,
+        &--active {
+          background-color: $color-aqua-lightest;
+        }
```
dandds: When I brought this up originally, I actually meant a hover state for the links within the menu. I think both would be great. What do you think?

templates/applications/fragments/members.html
```diff
@@ -94,83 +94,78 @@ <h1>{{ "invites.revoke" | translate }}</h1>
...
+                      <span class="env-role__environment">
+                        {{ env.environment_name }}
+                      </span>
+                      <span class="env-role__role">
+                        : {{ env.role }}
+                      </span>
+                    </div>
+                  {% endfor %}
+                  {% if user_can(permissions.EDIT_APPLICATION_MEMBER) -%}
+                    <toggle-menu inline-template>
```
dandds: Can you throw `v-cloak` on here? I think it's better than seeing each menu flash open and then shut, at least in FF.

styles/components/_portfolio_layout.scss
leigh-mil: 🤦‍♀ I agree! I'll check in with design about it!

PR #1206, opened 2019-11-22
Initial Setup for Consuming FlexVol in AKS

tomdds:
## Description

This PR includes the necessary k8s configs to set up FlexVol in an AKS cluster. This only sets up the necessary services and information, but does not include consuming it. That will be part of a later PR.

## Pivotal

https://www.pivotaltracker.com/story/show/169464333

PR #1210, opened 2019-11-25
Nginx Config Update

dandds:
## Description

* Adds JSON logging to NGINX containers - Azure's log aggregation is a little clunky when it comes to traditional logging, since it reports each line of the log as a separate log entry, even for stack traces and the like. This will help make the logs more searchable/parsable.
* Expands the configuration for SSL, making it more strictly enforced and thoroughly implemented.
* FlexVol (set up in https://github.com/dod-ccpo/atst/pull/1206) to mount the dhparam.pem file from the Azure KeyVault to the nginx containers, and allow future mounting of other nginx secrets.
    * Updates deploy README and kustomize script to include the need for TENANT_ID


deploy/azure/azure.yml
```diff
@@ -124,6 +126,19 @@ spec:
...
             defaultMode: 0666
+        - name: snippets
+          configMap:
+            name: nginx-snippets
+        - name: nginx-secret
+          flexVolume:
+            driver: "azure/kv"
+            options:
+              usepodidentity: "true"
+              keyvaultname: "atat-vault-test"
```
tomdds: Should we leave this static for now? Choose a name that we decide is standard or also template it in?

tomdds: Added an overlay for this value.

deploy/azure/azure.yml
```diff
@@ -75,54 +75,56 @@ spec:
...
-              mode: 0644
-            - key: tls.key
-              path: atat.key
-              mode: 0640
+              - key: tls.crt
+                path: atat.crt
+                mode: 0644
+              - key: tls.key
+                path: atat.key
+                mode: 0640
```
dandds: We should be able to remove this secret from the config in favor the KV-controlled cert.

deploy/azure/azure.yml
```diff
@@ -132,9 +134,22 @@ spec:
...
+        - name: snippets
+          configMap:
+            name: nginx-snippets
+        - name: nginx-secret
+          flexVolume:
+            driver: "azure/kv"
+            options:
+              usepodidentity: "true"
+              keyvaultname: "atat-vault-test"
+              keyvaultobjectnames: "dhparam4096;staging-cert"
```
dandds: I would make this one `master-cert` (or whatever it is) and change to staging with the overlay.

deploy/azure/azure.yml
```diff
@@ -132,9 +134,22 @@ spec:
...
+              keyvaultobjectaliases: "dhparam.pem;atat.crt"
+              keyvaultobjecttypes: secret
```
dandds: Does this need to be `"secret,certificate"`?

deploy/azure/azure.yml
```diff
@@ -161,19 +176,20 @@ spec:
...
-          ]
+          args:
+            [
+              "/opt/atat/atst/.venv/bin/python",
+              "/opt/atat/atst/.venv/bin/celery",
+              "-A",
+              "celery_worker.celery",
+              "worker",
+              "--loglevel=info",
+            ]
```
dandds: Is your editor reformatting YAML on the fly? 🤔 I don't have strong feelings about YAML format, I just prefer to not see it in the diff every time. Let's tackle that separately.

deploy/azure/azure.yml
tomdds: Yep, `"secret;cert"` I'll fix it.

deploy/azure/azure.yml
tomdds: Yeah, I have auto-format on to help with the fact that we enforce Black formatting for python, must be on for all file types. This seems to have gotten inserted as part of a rebase, so I'll see about rolling it back.

PR #1212, opened 2019-11-25
Configuration Guide

richard-dds:
https://www.pivotaltracker.com/story/show/166097924

Add explanations for all of our configuration settings in the README.


README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
     yarn test:coverage
 
+## Configuration
+
+- `ASSETS_URL`: URL to host which serves static assets (such as a CDN).
+- `BLOB_STORAGE_URL`: URL to Azure blob storage container.
+- `CAC_URL`: URL for the CAC authentication route.
+- `CA_CHAIN`: Path to the CA chain file.
+- `CDN_ORIGIN`: URL for the origin host for asset files.
+- `CELERY_DEFAULT_QUEUE`:
```
dandds: I think we can link to the Celery docs here and say something like "The name of the queue background tasks will be added to" 

https://docs.celeryproject.org/en/latest/userguide/configuration.html#task-default-queue

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `CLASSIFIED`: Boolean specifying if we're running in a classified environment.
```
dandds: I think we could drop this from here and the mapping in `atst.app`. We're not actually using it in the app.

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `CONTRACT_END_DATE`: String specifying the end date of the JEDI contract. Used for task order validation. Example: 2019-09-14
+- `CONTRACT_START_DATE`: String specifying the start date of the JEDI contract. Used for task order validation. Example: 2019-09-14.
+- `CRL_FAIL_OPEN`: Boolean specifying if expired CRLs should fail open, rather than closed.
+- `CRL_STORAGE_CONTAINER`: Path to a directory where the CRL cache will be stored.
+- `CSP`: String specifying the cloud service provider to use. Acceptable values: "azure", "mock", "mock-csp".
+- `DEBUG`: Boolean. A truthy value enables Flask's debug mode. https://flask.palletsprojects.com/en/1.1.x/config/#DEBUG
+- `DISABLE_CRL_CHECK`: Boolean specifying if CRL check should be bypassed.
```
dandds: Also "Useful for instances of application container that are not serving HTTP requests, such as Celery workers" or something like that, maybe.

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `ENVIRONMENT`: String specifying the current environment. Acceptable values: "dev", "prod".
+- `LIMIT_CONCURRENT_SESSIONS`: Boolean specifying if users should be allowed only one active session at a time.
+- `LOG_JSON`: Boolean specifying whether app should log in a json format.
+- `PERMANENT_SESSION_LIFETIME`: Integer specifying how many seconds a user's session can stay valid for. https://flask.palletsprojects.com/en/1.1.x/config/#PERMANENT_SESSION_LIFETIME
+- `PE_NUMBER_CSV_URL`:
+- `PGAPPNAME`:
```
dandds: Do you mind dropping these two from the base.ini?

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `PGDATABASE`: String specifying the name of the postgres database.
+- `PGHOST`: String specifying the hostname of the postgres database.
+- `PGPASSWORD`: String specifying the password of the postgres database.
+- `PGPORT`: Integer specifying the port number of the postgres database.
+- `PGSSLMODE`: String specifying the ssl mode to use when connecting to the postgres database. https://www.postgresql.org/docs/9.1/libpq-ssl.html
+- `PGSSLROOTCERT`: Path to the root SSL certificate for the postgres database.
+- `PGUSER`: String specifying the username to use when connecting to the postgres database.
+- `PORT`: Integer specifying the port to bind to when running the flask server.
```
dandds: Maybe also "Used only for local development" or something.

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `REDIS_URI`: URI for the redis server.
+- `RQ_QUEUES`:
```
dandds: Do you mind dropping this?

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `SECRET_KEY`: String key which will be used to sign the session cookie. Should be a long string of random bytes. https://flask.palletsprojects.com/en/1.1.x/config/#SECRET_KEY
+- `SERVER_NAME`: Hostname for ATAT.
```
dandds: Also "Only needs to be specified in contexts where the hostname cannot be inferred from the request, such as Celery workers". (It automatically makes it so that the app can only serve requests from the specified hostname, which would break our auth, so I want it to be clear that it's only set in certain contexts.)

README.md
```diff
@@ -220,6 +220,48 @@ To generate coverage reports for the Javascript tests:
...
+- `SESSION_COOKIE_NAME`: String value specifying the name to use for the session cookie. https://flask.palletsprojects.com/en/1.1.x/config/#SESSION_COOKIE_NAME
+- `SESSION_TYPE`: String value specifying the cookie storage backend. https://pythonhosted.org/Flask-Session/
+- `SESSION_USE_SIGNER`: Boolean value specifying if the cookie sid should be signed.
+- `SQLALCHEMY_ECHO`: Boolean value specifying if SQLAlchemy should log queries to stdout.
+- `STATIC_URL`: URL specifying where static assets are hosted.
+- `USE_AUDIT_LOG`: Boolean value describing if ATAT should write to the audit log. Set to "false" by default for performance reasons.
```
dandds: "write to a audit table in the database". We're still issuing audit events as regular app logs. (Maybe that setting name could be clearer.)

README.md
dandds: https://flask.palletsprojects.com/en/1.1.x/config/#SERVER_NAME

PR #1215, opened 2019-11-26
Schema review

leigh-mil:
## Description
Updated non nullable constraints on database columns to match business logic.  
Currently this is done through multiple migrations and commits, but I will combine it all before merging. 

## Pivotal
https://www.pivotaltracker.com/story/show/169739993


alembic/versions/24f8a90623a7_clins_table_update.py
```diff
@@ -0,0 +1,61 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('clins', 'end_date',
+               existing_type=sa.DATE(),
+               nullable=False)
+    op.alter_column('clins', 'jedi_clin_type',
+               existing_type=sa.VARCHAR(length=11),
+               nullable=False)
+    op.alter_column('clins', 'number',
+               existing_type=sa.VARCHAR(),
+               nullable=False)
```
dandds: This is optional in `CLINForm` for some reason. 🤔

alembic/versions/f09ae38c0920_portfolio_to_and_user_table_updates.py
```diff
@@ -0,0 +1,68 @@
...
+               nullable=False)
+    op.alter_column('portfolios', 'name',
+               existing_type=sa.VARCHAR(),
+               nullable=False)
+    op.alter_column('task_orders', 'portfolio_id',
+               existing_type=postgresql.UUID(),
+               nullable=False)
+    op.alter_column('task_orders', 'user_id',
+               existing_type=postgresql.UUID(),
+               nullable=False)
```
dandds: Who is this? The user who made the original task order? Is there a reason we're even recording that?

alembic/versions/24f8a90623a7_clins_table_update.py
leigh-mil: Good catch! I'll double check the other forms too

alembic/versions/f09ae38c0920_portfolio_to_and_user_table_updates.py
leigh-mil: @dandds  The user_id is for the TO's creator. It was added in this commit: https://github.com/dod-ccpo/atst/commit/d0bfa16f1755a0f2b6e72bc725a63fed7a479765 
It looks like it is a holdover from when TOs had officers. I'll do a bit more investigating, but from a quick review it looks like we can delete that column!

dandds: Lol, so it was me. But great that we can remove it!

PR #1219, opened 2019-11-27
TO next button activation bugfix

leigh-mil:
## Description
When running the Ghost Inspector "AT-AT Walker" tests as a suite, there was an issue where the next button on the TO form step 1 was inconsistently taking a long time to become active. Previously the field change emitter, which triggers the button to become active was after an async function, so I moved the event emitter to be before the async function call so it will hopefully trigger the change event on the parent component earlier.
I also moved when `this.changed` is updated to be inside if statement because it should only be
updated if the file is successfully uploaded.

## Pivotal
https://www.pivotaltracker.com/story/show/169577065


JRNPW:
I understand Leigh's logic as a potential solution to the problem. It needs proper review by a developer as well.
These tests always pass when run in CI. They also pass when each test is run individually. It's *only* when they're run on the Staging or Prod site as a suite that they fail. We will only be sure that this is a fix for the issue after several iterations of the "Walker" suite without these failures.

js/components/upload_input.js
```diff
@@ -58,6 +58,8 @@ export default {
         this.$refs.attachmentFilename.value = file.name
         this.$refs.attachmentObjectName.value = response.objectName
         this.$refs.attachmentInput.disabled = true
+        emitFieldChange(this)
+        this.changed = true
```
dandds: Do we want to emit the change on error, too? (That's a real question, not rhetorical.) Does it matter?

leigh-mil: I think we do still want to emit because if the form was valid, then becomes invalid because the user uploads a new file we will want the button to become disabled.

dandds: Riiight. Can you add those lines to the `else` branch, then?

PR #1220, opened 2019-11-27
Styling for expired funding and spend table sections of reporting

graham-dds:
Styling and some slight refactoring for:

### Expired Funds 

image

### Spend Table

image

### DIFF
--- a/templates/portfolios/reports/expired_task_orders.html
+++ b/templates/portfolios/reports/expired_task_orders.html
@@ -1,34 +1,51 @@
 {% from "components/accordion.html" import Accordion %}
+{% from "components/icon.html" import Icon %}
 
 
 <section>
   <div class="usa-accordion">      
-    {% call Accordion("Expired Task Orders", "expired_task_orders", "h3") %}
-      {% for task_order in expired_task_orders %}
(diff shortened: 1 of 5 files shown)


templates/portfolios/reports/expired_task_orders.html
```diff
@@ -1,34 +1,51 @@
...
-        <div>
-          <p>Total Expended</p>
-          <p>{{ task_order["expended_funds"] | dollars }}</p>
-        </div>
-        <div>
-          <p>Total Unused</p>
-          <p>{{ (task_order["total_obligated_funds"] - task_order["expended_funds"]) | dollars }}</p>
-        </div>
-      {% endfor %}
+    {% call Accordion(
```
dandds: It looks like this is the only instance of the `Accordion` component so far. I think it's hard to justify the level of abstraction in the macro template for just one instance. I would say either: 
- Just inline the Vue code here and get rid of the macro, or
- figure out how to make a couple more global decisions in the macro, e.g.: Do you need `content_tag`? Can that just be a generic `div` and the caller can make decisions about styling in its own top-level structure?

PR #1223, opened 2019-12-02
Update Current Obligated funds stacked bar graph

graham-dds:
image

### DIFF
--- a/atst/domain/csp/reports.py
+++ b/atst/domain/csp/reports.py
@@ -44,7 +44,8 @@ def generate_sample_dates(_max=8):
 
 
 class MockReportingProvider(ReportingInterface):
-    MOCK_PERCENT_EXPENDED_FUNDS = 0.75
+    MOCK_PERCENT_INVOICED_EXPENDED_FUNDS = 0.60
+    MOCK_PERCENT_ESTIMATED_EXPENDED_FUNDS = 0.15
 
     FIXTURE_MONTHS = generate_sample_dates()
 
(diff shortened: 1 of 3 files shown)


atst/domain/csp/reports.py
```diff
@@ -298,7 +299,12 @@ def get_obligated_funds_by_JEDI_clin(self, portfolio):
                 return_dict[jedi_clin.value] = {
                     "obligated_funds": obligated_funds,
                     "expended_funds": (
-                        obligated_funds * Decimal(self.MOCK_PERCENT_EXPENDED_FUNDS)
+                        obligated_funds
+                        * Decimal(self.MOCK_PERCENT_INVOICED_EXPENDED_FUNDS)
+                    ),
+                    "estimated_expended_funds": (
+                        obligated_funds
+                        * Decimal(self.MOCK_PERCENT_ESTIMATED_EXPENDED_FUNDS)
```
leigh-mil: This is 100% optional if it is too difficult, but I was thinking that it would be nice to mock the invoiced and expended funds as variable percentages. That way each of the bar graphs on the page won't look identical.

dandds: Yeah, agreed. Folks always get hung up on details like that in demos.

templates/portfolios/reports/obligated_funds.html
```diff
@@ -8,22 +8,38 @@ <h2 class="reporting-section-header__header">Current Obligated funds</h2>
...
           </h3>
           <p class="jedi-clin-funding__subheader">Total obligated amount: {{ funds["obligated_funds"] | dollars }}</p>
-          <meter class="jedi-clin-funding__meter" value='{{remaining_funds}}' min='0' max='{{ funds["obligated_funds"] }}' title='{{ JEDI_clin }}'>
-            <div class='jedi-clin-funding__meter-fallback' style='width:{{ (funds["expended_funds"] / funds["obligated_funds"]) * 100 }}%;'></div>
-          </meter>
-          <div class="jedi-clin-funding__meter-values">
+          <div class="jedi-clin-funding__graph">
+            <span style="width:{{(funds.expended_funds / funds.obligated_funds) * 100 | int}}%" class="jedi-clin-funding__graph-bar jedi-clin-funding__graph-bar--invoiced"></span>
+            <span style="width:{{(funds.estimated_expended_funds / funds.obligated_funds) * 100 | int}}%" class="jedi-clin-funding__graph-bar jedi-clin-funding__graph-bar--estimated"></span>
+            <span style="width:{{(remaining_funds / funds.obligated_funds) * 100 | int}}%" class="jedi-clin-funding__graph-bar jedi-clin-funding__graph-bar--remaining"></span>
```
dandds: It might be nice to have a filter function that takes two numerical args and returns a percentage, instead of having all the math in the template.

graham-dds:
Moving these changes into the PR where I refactor reporting data

PR #1225, opened 2019-12-02
Revoke env access before env role has been provisioned

leigh-mil:
## Description
Previously we were disabling an env role in the CSP every time env access was revoked, which was causing issues when suspending the role for an user that had not accepted their invite yet. I added in a check to see if the environment has been provisioned before disabling the env role in the csp.

## Pivotal
https://www.pivotaltracker.com/story/show/170004953

### DIFF
--- a/atst/domain/environment_roles.py
+++ b/atst/domain/environment_roles.py
@@ -105,8 +105,9 @@ class EnvironmentRoles(object):
     def disable(cls, environment_role_id):
         environment_role = EnvironmentRoles.get_by_id(environment_role_id)
 
-        credentials = environment_role.environment.csp_credentials
-        app.csp.cloud.disable_user(credentials, environment_role.csp_user_id)
+        if not environment_role.environment.is_pending:
+            credentials = environment_role.environment.csp_credentials
+            app.csp.cloud.disable_user(credentials, environment_role.csp_user_id)
 
         environment_role.status = EnvironmentRole.Status.DISABLED
         db.session.add(environment_role)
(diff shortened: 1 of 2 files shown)


atst/domain/environment_roles.py
```diff
@@ -105,8 +105,9 @@ def get_environment_roles_pending_creation(cls) -> List[UUID]:
     def disable(cls, environment_role_id):
         environment_role = EnvironmentRoles.get_by_id(environment_role_id)
 
-        credentials = environment_role.environment.csp_credentials
-        app.csp.cloud.disable_user(credentials, environment_role.csp_user_id)
+        if not environment_role.environment.is_pending:
```
dandds: I think we also might see errors under these conditions:
- the environment has been provisioned
- the user has not accepted the application role invitation

Under those circumstances, `environment_role.csp_user_id` will not have a value. You _should_ just be able to check whether that column is null or not. For safety, it might be wise to keep your current condition and add the other one.

dandds: For reference, you can see the CSP-side user being provisioned here: https://github.com/dod-ccpo/atst/blob/staging/atst/jobs.py#L106-L118

atst/domain/environment_roles.py
```diff
@@ -105,8 +105,9 @@ def get_environment_roles_pending_creation(cls) -> List[UUID]:
     def disable(cls, environment_role_id):
         environment_role = EnvironmentRoles.get_by_id(environment_role_id)
 
-        credentials = environment_role.environment.csp_credentials
-        app.csp.cloud.disable_user(credentials, environment_role.csp_user_id)
+        if environment_role.csp_user_id or not environment_role.environment.is_pending:
```
dandds: I think this should be `and`, not `or`.

PR #1232, opened 2019-12-04
TO and App index pages redesign (Part 1)

leigh-mil:
## Description
First PR for restyling the TO and App index pages. This PR only covers the styling, the logic for sorting the TOs by status, faking the expended funds from the CSP, and the button for collapsing all lists will be in a follow up PR. 
I removed the styling for the deprecated `accordion` class and references to it, however this means that the application index page is temporarily wonky until it is updated.

## Pivotal
https://www.pivotaltracker.com/story/show/168907682
https://www.pivotaltracker.com/story/show/169715584

## Screenshot
Screen Shot 2019-12-04 at 2 46 20 PM
Screen Shot 2019-12-06 at 3 35 35 PM


templates/applications/index.html
```diff
@@ -39,20 +39,20 @@
...
+            <header class='row'>
               <div class='col col-grow'>
-                <h3 class='icon-link accordion__title' v-on:click="toggleSection('{{ section_name }}')">{{ application.name }}</h3>
-                <p class='accordion__description'>
+                <h3 class='icon-link' v-on:click="toggleSection('{{ section_name }}')">{{ application.name }}</h3>
+                <p class=''>
                   {{ application.description }}
                 </p>
-                <div class='accordion__actions'>
+                <div class=''>
```
dandds: I don't think we should keep empty class attrs.

templates/applications/index.html
```diff
@@ -61,15 +61,15 @@ <h3 class='icon-link accordion__title' v-on:click="toggleSection('{{ section_nam
                       <span v-else>
                         {{ Icon('caret_down') }}
                       </span>
-                      <div class="triangle-up" v-if="selectedSection === '{{ section_name }}'"></div>
+                      <div class="" v-if="selectedSection === '{{ section_name }}'"></div>
```
dandds: So this is just an empty `div` if the conditional evaluates to true?

styles/elements/_accordions.scss
```diff
@@ -1,148 +1,51 @@
...
-  &__title {
-    @include block-list__title;
-
-    color: $color-blue;
+    padding: $gap * 2 $gap * 3;
 
-    @include h3;
-
-    &.icon-link {
+    h3,
```
graham-dds: Is it possible to apply these same styles in a class rather than targeting these specific elements? For example, what if I wanted the copy in my `accordion__header` class to utilize an `h4` or `h5` instead of an `h3`? Or a `span` instead of a `p`?

templates/task_orders/index.html
```diff
@@ -6,89 +7,45 @@
...
-        <div class="card__header">
-          <h3>Task Order #{{ task_order.number }}</h3>
-        </div>
-        <div class="card__body">
-          <b>Total amount: </b>{{ task_order.total_contract_amount | dollars }}
-        </div>
-        <div class="card__body">
-          <b>Obligated amount: </b>{{ task_order.total_obligated_funds | dollars }}
+{% macro TaskOrderList(task_orders, status) %}
+  {% set status = "All Task Orders" %}
```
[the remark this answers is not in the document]
leigh-mil: Yup! Part 2 is hopefully coming this afternoon

atst/routes/applications/index.py
```diff
@@ -23,4 +24,11 @@ def has_portfolio_applications(_user, portfolio=None, **_kwargs):
     message="view portfolio applications",
 )
 def portfolio_applications(portfolio_id):
-    return render_template("applications/index.html")
+    user_env_roles = EnvironmentRoles.for_user(g.current_user.id, portfolio_id)
+    environment_access = {}
+    for env_role in user_env_roles:
+        environment_access[env_role.environment_id] = env_role.role
```
dandds: Little thing, but I think it would be more Pythonic to just use a dictionary comprehension: `environment_access = {env_role.environment_id: env_role.role for env_role in env_roles}`.

PR #1233, opened 2019-12-05
Add script for running Ghost Inspector tests locally.

dandds:
Eventually, this should replace the CircleCI config for running the
integration tests to avoid duplication. In the interest of time so that
I don't have to debug broken builds, I'm only adding it as a utility
script.

### DIFF
--- a/uitests/README.md
+++ b/uitests/README.md
@@ -30,7 +30,7 @@ QA lead monitors all failures on the CI suite and determines that such a change 
 the Ghost Inspector UI.
 
 3. If the (potentially) failing test will need to be reworked to account for functional changes, the test is moved from the CI
-suite to the "Holding" suite until the PR is merged. Then the test can be edited and returned to the CI suite. 
+suite to the "Holding" suite until the PR is merged. Then the test can be edited and returned to the CI suite.
 
 ## Running Ghost Inspector tests locally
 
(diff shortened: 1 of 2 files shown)


uitests/README.md
```diff
@@ -30,34 +30,32 @@ QA lead monitors all failures on the CI suite and determines that such a change
...
 
 ```
-circleci local execute -e GI_SUITE=<SUITE_ID> -e GI_API_KEY=<API KEY> -e NGROK_TOKEN=<NGROK TOKEN> --job integration-tests -c local-ci.yml
+NGROK_TOKEN=<token> GI_API_KEY=<api key> GI_SUITE=<suite> CONTAINER_IMAGE=atat:build ./script/integration_tests
 ```
 
-If the job fails and you want to re-run it, you may receive errors
-about running docker containers or the network already existing.
-Some version of the following should reset your local docker state:
+### Troubleshooting
```
JRNPW: You might want to create some comments about failure messages that are (apparently) okay. E.g., I received messages like
- `No such container: test-atat` after the `postgres` command was run
- `curl: (7) Failed to connect to test-atat port 8000: Connection refused` because it seems to take a while to connect

And perhaps even note that a large blob of HTML will be thrown into the Terminal right after `curl` apparently connects successfully.

PR #1237, opened 2019-12-09
Management group sketches

tomdds:
Sketch in an implementation of Management Groups. This is a _very_ incomplete sketch, mostly to serve as a template for the api calls we need to use. This can also serve as a jumping off point for figuring out what we need to store at each entity level for CSP details.

### DIFF
--- a/atst/domain/csp/cloud.py
+++ b/atst/domain/csp/cloud.py
@@ -3,6 +3,7 @@ import re
 from uuid import uuid4
 
 from atst.models.user import User
+from atst.models.application import Application
 from atst.models.environment import Environment
 from atst.models.environment_role import EnvironmentRole
 
@@ -399,13 +400,14 @@ REMOTE_ROOT_ROLE_DEF_ID = "/providers/Microsoft.Authorization/roleDefinitions/00
 
 class AzureSDKProvider(object):
     def __init__(self):
-        from azure.mgmt import subscription, authorization
+        from azure.mgmt import subscription, authorization, managementgroups
         import azure.graphrbac as graphrbac
         import azure.common.credentials as credentials
         from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD
 
         self.subscription = subscription
         self.authorization = authorization
+        self.managementgroups = managementgroups
         self.graphrbac = graphrbac
         self.credentials = credentials
         # may change to a JEDI cloud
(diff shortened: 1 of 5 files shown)


atst/domain/csp/cloud.py
```diff
@@ -428,42 +430,23 @@ def __init__(self, config, azure_sdk_provider=None):
     def create_environment(
         self, auth_credentials: Dict, user: User, environment: Environment
     ):
+        # since this operation would only occur within a tenant, should we source the tenant
+        # via lookup from environment once we've created the portfolio csp data schema
+        # something like this:
+        # environment_tenant = environment.application.portfolio.csp_data.get('tenant_id', None)
```
dandds: Do we want this class to know so many details of the schema, or do we want things to work by dependency injection? Not rhetorical, I could see either way.

atst/domain/csp/cloud.py
```diff
@@ -502,6 +485,83 @@ def create_atat_admin_user(
             "role_name": role_assignment_id,
         }
 
+    def _create_application(self, auth_credentials: Dict, application: Application):
```
dandds: When do you think this would be called? Do we tend to think creating an application will be a standalone task that a beat worker will run a SQL check to kick off, or will it happen when the first environment in the application is created?

atst/domain/csp/cloud.py
tomdds: I think I agree that most of the necessary details should be sourced by the task that calls into this, rather than deeply derived from the passed entities. A lot of these comments are intended to be referenced when we revisit the structure of this class and determine what the new interface should expose and which are implementation details. Despite writing it off previously, I could see us choosing to have a `CSPDetails` data class that gets populated by the task handler and passed into this, containing all the info it needs to carry out it's various functions (including KV sourced credentials and the like). Since we're considering an initial DB schema that normalizes the details out of our hierarchy and prepares for potential multi-cloud, this data-prep layer would probably end up being necessary.

atst/domain/csp/cloud.py
tomdds: I think this may need to be a standalone prerequisite for environment creation rather than lazily called by environment creation. Environments are created in batches as a bunch of tasks that can run in parallel. I think it'd be a bigger lift to coordinate the lazy creation of the owning application management group across the environment creation tasks, than to have a standalone task to create the application that then allows the environment creation tasks to be completed. Probably adding a query for the existence/completeness of the csp details for the application the environment(s) belong to.

dandds:
@tomdds I think this is ready to merge if you want to take it out of draft.

dandds:
So uuuhhh I merged this but the PR is still here? Some kind of bug with GH. It's in staging, so I'm closing.

dandds:
So uuuhhh I merged this but the PR is still here? Some kind of bug with GH. It's in staging, so I'm closing.

dandds:
What is happening Github

PR #1239, opened 2019-12-10
TO and Application index pages -- part 2

leigh-mil:
## Description
- Sort task orders by status on index page.
- Create AccordionList macro and Vue component to encapsulate the `Collapse All` logic.
- Add `Collapse All` toggle to TO and App index pages.
- Fake amount invoiced to a TO.

## Pivotal
https://www.pivotaltracker.com/story/show/168907682
https://www.pivotaltracker.com/story/show/169715584

### DIFF
--- a/atst/models/task_order.py
+++ b/atst/models/task_order.py
@@ -20,12 +20,7 @@ class Status(Enum):
     UNSIGNED = "Not signed"
 
 
-SORT_ORDERING = {
-    status: order
-    for (order, status) in enumerate(
-        [Status.DRAFT, Status.ACTIVE, Status.UPCOMING, Status.EXPIRED, Status.UNSIGNED]
-    )
-}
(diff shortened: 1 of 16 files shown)


atst/models/task_order.py
```diff
@@ -170,6 +169,11 @@ def balance(self):
         # Faked for display purposes
         return 50
 
+    @property
+    def invoiced_funds(self):
+        # TODO: implement this using reporting data from the CSP
+        return self.total_obligated_funds * 0.75
```
dandds: Hmm I didn't really think this one through. I'm not sure when or how we'll get this data per task order. I'm not suggesting you change it on this PR, just flagging it as a larger concern we have to run down.

dandds: I'm also getting an error from this method with the sample data: `TypeError: unsupported operand type(s) for *: 'decimal.Decimal' and 'float'`

PR #1242, opened 2019-12-11
Styling / UI changes for new portfolio page

graham-dds:
image

Addresses https://www.pivotaltracker.com/story/show/168907648

This PR is a step in the direction of redesigning the New Portfolio flow according to Think's [designs](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382268615). It mainly removes many (now) unnecessary form input options from the creation page. 

Any styling that looks off from the screen linked above:
 - refers to a (now) non-existent step 2
 - seems like a change that needs to be made on a larger scale, e.g. standardizing element widths across different screens


js/components/multi_checkbox_input.js
```diff
@@ -13,40 +13,30 @@ export default {
...
+      this.showError = !this.valid
+      this.showValid = !this.showError
+      this.validationError = 'This field is required.'
     },
   },
 
   computed: {
     valid: function() {
-      return this.optional || this.showValid
+      return this.optional || this.selections.length > 0
```
dandds: I notice we don't have any tests for component. Do you feel up to adding a couple to cover the basic valid/invalid states?

atst/models/portfolio.py
```diff
@@ -19,19 +18,11 @@ class Portfolio(
 
     id = types.Id()
     name = Column(String, nullable=False)
+    description = Column(String)
```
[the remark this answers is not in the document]
graham-dds: Yep, I noticed that too. I asked in the design channel a while back, but I can ask them again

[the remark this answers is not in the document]
graham-dds: Looks like this will appear in the portfolio settings page now.

js/components/multi_checkbox_input.js
graham-dds: https://github.com/dod-ccpo/atst/pull/1242/commits/2c20429c0d760e0f1aee6a5e78ab519d8d75395c

tests/render_vue_component.py
```diff
@@ -82,6 +89,20 @@ def test_make_checkbox_input_template(checkbox_input_macro, initial_value_form):
...
 
 
+def test_make_multi_checkbox_input_template(
+    multi_checkbox_input_macro, initial_value_form
+):
+    initial_value_form.datafield.widget = ListWidget()
+    initial_value_form.datafield.option_widget = CheckboxInput()
+    initial_value_form.datafield.choices = [("a", "A"), ("b", "B")]
+    rendered_multi_checkbox_input_macro = multi_checkbox_input_macro(
+        initial_value_form.datafield, optional=Markup("'optional'")
```
graham-dds: We were on the right track with destructuring initialData. I had to use [Markup](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=markup#flask.Markup) to get `'optional'` to show up in the generated test template though

JRNPW:
@graham-dds One comment: Think/we are following USWDS 1.0, which specifies labeling optional fields as such. I note that the new "Defense Component" checkbox group isn't labeled as Optional. Is it now a required field?

graham-dds:
> @graham-dds One comment: Think/we are following USWDS 1.0, which specifies labeling optional fields as such. I note that the new "Defense Component" checkbox group isn't labeled as Optional. Is it now a required field?

It looks to be based on [this](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382267360) screen

PR #1244, opened 2019-12-12
K8s cronjob for resetting the database on staging.

dandds:
This K8s CronJob will run the scripts for resetting and seeding sample
data to the database. It will only be applied to the staging site.


deploy/overlays/staging/reset-cron-job.yml
```diff
@@ -0,0 +1,49 @@
...
+            image: $CONTAINER_IMAGE
+            command: [
+              "/bin/sh", "-c"
+            ]
+            args: [
+              "/opt/atat/atst/.venv/bin/python",
+              "/opt/atat/atst/script/reset_database.py",
+              "&&",
+              "/opt/atat/atst/.venv/bin/python",
+              "/opt/atat/atst/script/seed_sample.py",
```
dandds: Running the seed sample script won't work because we don't include the dev dependencies in the container image. Need to think about whether we really need to do this and if so, how?

PR #1245, opened 2019-12-12
Enforce TO number uniqueness

leigh-mil:
## Description
Enforces uniqueness of TO numbers in the database and fixes the update and create TaskOrder domain methods to reflect the database constraint. Also updates the UI to display an flash error message when a user tries to put in a duplicate TO number.

## Pivotal
https://www.pivotaltracker.com/story/show/170112925

## Screenshots
Screen Shot 2019-12-12 at 2 32 57 PM
Screen Shot 2019-12-12 at 2 33 04 PM


atst/forms/task_order.py
```diff
@@ -142,6 +142,14 @@ class TaskOrderForm(BaseForm):
     )
     clins = FieldList(FormField(CLINForm))
 
+    @property
+    def data(self):
+        _data = super().data
+        if _data["number"] == "":
+            _data["number"] = None
```
dandds: I think we have the `remove_empty_string` filter for situations like this.

atst/routes/task_orders/new.py
```diff
@@ -176,12 +189,8 @@ def cancel_edit(task_order_id=None, portfolio_id=None):
         else:
             form = TaskOrderForm(form_data)
 
-        if form.validate(flash_invalid=False):
-            task_order = None
-            if task_order_id:
-                task_order = TaskOrders.update(task_order_id, **form.data)
-            else:
-                task_order = TaskOrders.create(portfolio_id, **form.data)
+        update_task_order(form, portfolio_id, task_order_id, flash_invalid=False)
```
dandds: Did we decide to just swallow the error if they enter a redundant TO number and not save their data?

leigh-mil: I talked it over with Ben, and what we decided to do was to show the flash error on the TO index page, and save all of the users changes except the dupe TO number.  `flash_invalid=False` here prevents the flash error from the BaseForm validator from showing up in the case of other update errors. https://github.com/dod-ccpo/atst/blob/2edfecdd4581586cfccc30545bac2b72248b6669/atst/forms/forms.py#L45

PR #1249, opened 2019-12-13
App name uniqueness

leigh-mil:
## Description
Added a constraint to the applications table that requires the application name to be unique within its portfolio relationship. 

## Pivotal
https://www.pivotaltracker.com/story/show/170266779

## Screenshot
Screen Shot 2019-12-13 at 4 10 54 PM


atst/models/application.py
```diff
@@ -34,6 +34,11 @@ class Application(
         ),
     )
     members = synonym("roles")
+    __table_args__ = (
+        UniqueConstraint(
+            "name", "portfolio_id", name="name_portfolio_id_unique_constraint"
```
dandds: Nit-picky, but we probably want the table in there for clarity: `applications_name_portfolio_id_key` would work.

atst/domain/applications.py
```diff
@@ -28,7 +31,12 @@ def create(cls, user, portfolio, name, description, environment_names=None):
         if environment_names:
             Environments.create_many(user, application, environment_names)
 
-        db.session.commit()
+        try:
+            db.session.commit()
+        except IntegrityError:
+            db.session.rollback()
+            raise AlreadyExistsError("application")
+
```
dandds: It's probably worth creating a utility function for this pattern that takes the message for `AlreadyExistsError` as an argument. You can use it here and in the TO create/update methods.

PR #1257, opened 2019-12-17
Env and App name uniqueness

leigh-mil:
## Description
Enforce environment name uniqueness within an application and catch dupe name error when updating an application through the settings page. 

## Pivotal
https://www.pivotaltracker.com/story/show/170266784
https://www.pivotaltracker.com/story/show/170266779

## Screenshot
Screen Shot 2019-12-17 at 12 26 22 PM


atst/domain/environments.py
```diff
@@ -12,6 +12,7 @@
     CLIN,
 )
 from atst.domain.environment_roles import EnvironmentRoles
+from atst.utils import update_or_raise_already_exists_error
```
graham-dds: I know this is super nit-picky, but could we rename this function to like `commit_or_raise_error`? It feels weird to have a function named `update_...` called in a `create` function.

PR #1258, opened 2019-12-17
Form action bars

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/170290282

Moves all action buttons that are a part of a form flow from the sticky CTA to a floating bottom footer.

### Example:
Before:
image

After:
image

This might also close https://www.pivotaltracker.com/story/show/168907648 since it doesn't sound like we're doing a multi-step flow for creating portfolios for MVP.

### DIFF
--- a/styles/components/_forms.scss
+++ b/styles/components/_forms.scss
@@ -1,7 +1,8 @@
 // Form Grid
 .form-row {
   margin: ($gap * 4) 0;
-  &--separated {
+
+  &--bordered {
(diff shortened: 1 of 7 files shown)


JRNPW:
@graham-dds This causes Ghost Inspector's Task Order creation tests to fail because the Step X of Y moved from Sticky Header Text to Sticky Header Context. We'll need to work together to update Ghost Inspector before this merges to `staging`.

PR #1264, opened 2019-12-18
Update home page

leigh-mil:
## Description
Update styling and copy on home page and side nav.
Some global changes were made to the `hr` element:
- Previously `hr`s were globally styled to extend into the margin of their containing element. Since this is only necessary in a few places, I changed the default to have the `hr` fill the element and not extend into the margin, and added a util class `.full-width` for the few cases were a `hr` needs to extend into the margins of its container.
- Deleted unused class `panel__break` from `hr` elements.

## Pivotal
https://www.pivotaltracker.com/story/show/168907390
https://www.pivotaltracker.com/story/show/168907361

## Screenshots
Screen Shot 2019-12-18 at 4 57 35 PM
Screen Shot 2019-12-18 at 4 57 38 PM
Screen Shot 2019-12-18 at 4 58 38 PM

### DIFF
--- a/templates/applications/new/step_3.html
+++ b/templates/applications/new/step_3.html
@@ -15,7 +15,7 @@
   <p>
     {{ ('portfolios.applications.new.step_3_description' | translate) }}
   </p>
-  <hr class="panel__break">
+  <hr>
 
   {{ MemberManagementTemplate(
     application,
--- a/styles/core/_util.scss
+++ b/styles/core/_util.scss
@@ -88,5 +88,9 @@ p {
 hr {
   border: 0;
   border-bottom: 1px solid $color-gray-light;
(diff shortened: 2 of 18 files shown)


templates/applications/new/step_3.html
```diff
@@ -15,7 +15,7 @@
   <p>
     {{ ('portfolios.applications.new.step_3_description' | translate) }}
   </p>
-  <hr class="panel__break">
+  <hr>
```
dandds: Why did HR have to change everywhere? (It would be helpful to note this kind of thing in the PR description!)

leigh-mil: I explained these changes in the commit message, but I'll add it to the PR description too for posterity!

PR #1275, opened 2019-12-23
To view review pages

leigh-mil:
## Description
Update styling of the TO view page and the review page (step 4 of TO builder) according to new designs. I also renamed the view TO route and templates (previously `review`) to better reflect the purpose of the route and template.

## Pivotal
https://www.pivotaltracker.com/story/show/170383868
https://www.pivotaltracker.com/story/show/170383849

## Screenshots
Step 4 of TO Builder:
Screen Shot 2019-12-23 at 11 48 08 AM
Screen Shot 2019-12-23 at 11 48 20 AM

View TO:
Screen Shot 2019-12-23 at 11 48 31 AM
Screen Shot 2019-12-23 at 11 48 34 AM


templates/task_orders/fragments/task_order_view.html
```diff
@@ -0,0 +1,91 @@
...
+{% from "components/tooltip.html" import Tooltip %}
+
+{% macro TaskOrderView(task_order, portfolio, builder_mode=False) %}
+  {% set obligated_funds = task_order.total_obligated_funds %}
+  {% set contract_amount = task_order.total_contract_amount %}
+  {% set expended_funds = task_order.invoiced_funds %}
+
+  <div>
+    <section class="row">
+      <div class='col col--grow reporting-summary-item'>
```
graham-dds: I love that we were able to reuse most of this CSS component (with a few tweaks) in the task order view as well 😍 . But now that it's doing double duty, we should probably rename to something more general so that it's not as reporting-specific. Maybe just take off reporting and make it `summary-item`?

templates/task_orders/fragments/task_order_view.html
```diff
@@ -0,0 +1,91 @@
...
+        <tbody>
+          {% for clin in task_order.sorted_clins %}
+          <tr>
+            <td>{{ clin.number }}</td>
+            <td>{{ clin.type }}</td>
+            <td>{{ "{}".format(clin.jedi_clin_type) | translate}}</td>
+            <td>
+             {{ clin.start_date | formattedDate }} -  {{ clin.end_date | formattedDate }}
+            </td>
+            {# TODO: Swap in total CLIN amount #}
```
graham-dds: This was left over from when I was working on this section 🤦‍♂. I think the real total CLIN amount is here now, so we can remove this comment.

PR #1281, opened 2019-12-27
Better incorporate fixture data into reporting

graham-dds:
image

Before this commit, if a portfolio wasn't present in the spending fixture data, the reporting screen would be empty -- even if the portfolio had applications and environments associated with it on the database. Now, 0s appear if an application and / or environment isn't present in the fixture data.

Addresses: https://www.pivotaltracker.com/story/show/170325621

### DIFF
--- a/atst/domain/csp/reports.py
+++ b/atst/domain/csp/reports.py
@@ -34,16 +34,24 @@ class MockReportingProvider:
             }
         ]
         """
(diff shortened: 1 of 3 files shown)


JRNPW:
Quick functional testing looks good. It also passes all tests and has no negative impact on Ghost Inspector tests. This deserves Code Review, so this is my polite nod but not a formal approval.

PR #1282, opened 2019-12-27
Tidy sidenav and topbar

graham-dds:
Removes some crufty styles from the topbar and sidenav.

I wanted to go further -- theoretically we should be able to apply the `.sidenav-container` class to the same `div` that has the `.global-navigation` class, but there were some layout issues on some pages.


JRNPW:
Ghost Inspector is happy with this branch, so that means I am. :-) It deserves Code Review before merging. This gets my nod of approval.

PR #1284, opened 2019-12-28
State machine poc

philip-dds:
@tomdds This is my latest work from my fork.
can't seem to push to origin branch.

Some notes on the work I've done so far.

There are a bunch of alembic revision files in there that should run but are not necessary since they were created before I finalized the db schema. I was not able to find a clean way of squashing these migrations. I will try to figure that out later.
 
I introduced a depenency https://github.com/edelooff/sqlalchemy-json that helps setup the sqlalchemy boilerplate for a mutable json field. `pipenv install sqlalchemy-json` upgraded a bunch of dependencies. Not sure how you guys handle that. Or it is ok to just merge the new Pipfile?

The tenant creation process should work end to end.

there are 4 `triggers` tied to transitions
 
sm.init()
sm.start()
sm.create_tenant()
sm.finish_create_tenant()

here is an example of the calls needed to push the tenant creation from `UNSTARTED` to `TENANT_CREATED`

p = db.session.query(Portfolio).first() 
sm = Portfolios.create_state_machine(p)
In [4]: sm.state                                                                                                                                                                              
Out[4]: <FSMStates.UNSTARTED: 'unstarted'>
sm.init()
 sm.start()
sm.create_tenant()

In [9]: sm.portfolio.csp_data                                                                                                                                                                 
Out[9]: 
{'tenant_data': {'tenant_id': 'string',
  'user_id': 'string',
  'user_object_id': 'string'}}
sm.finish_create_tenant()
In [11]: sm.state                                                                                                                                                                             
Out[11]: <FSMStates.TENANT_CREATED: 'tenant created'>

`Portfolios.get_portfolios_pending_provisioning` is a method that the celery beat task uses to filter for portfolios that need to be picked up in the current iteration. Right now I have the query pick up porfolios that are either unstarted or failed. The idea there is to skip the "in progress" states since they may be doing work at the moment and let them handle the errors. The "in progress" states will transition to "failed" if they run out of attemps.
For example:
     The `TENANT_CREATION_IN_PROGRESS` state, `AzureFSMMixin.after_create_tenant` is where the api call is made.
     It is wrapped in simple retry logic that will result in `TENANT_CREATION_FAILED` state after running out of attempts to make the call, otherwise portfolio.csp_data is a json field that hold the collected data from csp.

let me know what you think of how I set this up and I will extend this to the rest of the provision tasks (create/update billing profile, admin subscription, etc)

I have some unit tests in there that I am continuously updating.

Philip


atst/models/portfolio_state_machine.py
```diff
@@ -0,0 +1,332 @@
...
+                    password_recovery_email_address='password@email.com')
+
+        csp = event.kwargs.get('csp')
+
+        if csp is not None:
+            self.csp = AzureCSP(app).cloud
+        else:
+            self.csp = MockCSP(app).cloud
+
+        for attempt in range(5):
```
dandds: @tomdds @philip-dds So we're going to do our own retries within the context of the running Celery task, instead of just relying on the Celery retry mechanism? Is this just to avoid cluttering the task queue?

alembic/versions/735e7528b723_portfolio_job_failure_table.py
```diff
@@ -0,0 +1,60 @@
...
+    sa.UniqueConstraint('task_id', name='celery_taskmeta_task_id_key')
+    )
+    op.create_table('celery_tasksetmeta',
+    sa.Column('id', sa.INTEGER(), autoincrement=False, nullable=False),
+    sa.Column('taskset_id', sa.VARCHAR(length=155), autoincrement=False, nullable=True),
+    sa.Column('result', postgresql.BYTEA(), autoincrement=False, nullable=True),
+    sa.Column('date_done', postgresql.TIMESTAMP(), autoincrement=False, nullable=True),
+    sa.PrimaryKeyConstraint('id', name='celery_tasksetmeta_pkey'),
+    sa.UniqueConstraint('taskset_id', name='celery_tasksetmeta_taskset_id_key')
+    )
```
dandds: Can we remove the Celery stuff from the downgrade too?

atst/jobs.py
```diff
@@ -125,6 +141,42 @@ def do_work(fn, task, csp, **kwargs):
...
+    print("running provision for portfolio <%s>" % (portfolio_id))
+    #portfolios = db.session.query(Portfolio).\
+    #        filter_by(portfolio_id=portfolio_id, deleted=False).all()
+    portfolio = Portfolios.get_for_update(portfolio_id)
+    if  portfolio.state_machine is None:
+        fsm = Portfolios.create_state_machine(portfolio)
+        db.session.add(fsm)
+        db.session.commit()
+    else:
+        fsm = PortfolioStateMachinesQuery.get(portfolio.state_machine.id)
```
dandds: I don't think you need to fetch the state machine again, do you? Should just be `portfolio.state_machine`.

atst/jobs.py
```diff
@@ -125,6 +141,42 @@ def do_work(fn, task, csp, **kwargs):
...
+        fsm.init()
+
+    if fsm.state == FSMStates.STARTING:
+        fsm.start()
+
+    if fsm.state == FSMStates.STARTED:
+        fsm.create_tenant(csp=csp.cloud)
+
+    if fsm.state == FSMStates.TENANT_CREATION_IN_PROGRESS:
+        fsm.finish_create_tenant()
```
dandds: @tomdds @philip-dds Does this kind of state-checking have to live in the Celery tasks? I thought maybe it would be internal to the state machine implementation, in which case this function would just call something like `fsm.do_next()` and the state machine would figure out what it had to do. I'm not very familiar with a typical implementation for this, though, so I might be off-base.

dandds:
Oh also, please let's make sure we don't have any new LGTM warnings by the time this is ready to merge. Thanks!

atst/domain/csp/cloud.py
```diff
@@ -323,6 +323,78 @@ def create_atat_admin_user(self, auth_credentials, csp_environment_id):
...
+                "startDate": "2019-12-18T16:47:40.909Z",
+                "endDate": "2019-12-18T16:47:40.909Z",
+                "externalReferenceId": "string"
+            }
+        }
+        """
+        # response will be mostly the same as the body, but we only really care about the id
+        self._maybe_raise(self.NETWORK_FAILURE_PCT, self.NETWORK_EXCEPTION)
+        self._maybe_raise(self.SERVER_FAILURE_PCT, self.SERVER_EXCEPTION)
+        self._maybe_raise(self.UNAUTHORIZED_RATE, self.AUTHORIZATION_EXCEPTION)
```
tomdds: We should finalize the decision between throwing exceptions from the CSP interface (as the mock one and these do) or using the `_ok` and `_error` methods to create response bodies that the state machine will inspect for determining success or retry/fail.

tomdds:
Closed in lieu of #1303

PR #1287, opened 2020-01-02
Portfolio defense component bug

leigh-mil:
## Description
When the form for portfolios changed, the column for `defense_component` was never update to an array type to reflect the change, this PR fixes this issue. 
Until we get guidance from design, each defense component is displayed on its own line on the portfolio settings page. 

## Pivotal
https://www.pivotaltracker.com/story/show/170293397


alembic/versions/02ac8bdcf16f_update_portfolios_defense_component_.py
```diff
@@ -0,0 +1,36 @@
...
+               nullable=False)
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('portfolios', 'defense_component',
+               type_=sa.VARCHAR(),
+               existing_type=postgresql.ARRAY(sa.VARCHAR()),
+               postgresql_using="defense_component[1]::character varying",
```
dandds: 💡

alembic/versions/02ac8bdcf16f_update_portfolios_defense_component_.py
```diff
@@ -0,0 +1,36 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('portfolios', 'defense_component',
+               type_=postgresql.ARRAY(sa.VARCHAR()),
+               existing_type=sa.VARCHAR(),
+               postgresql_using="string_to_array(defense_component, ',')::character varying[]",
```
dandds: I don't think this conversion is working the way we want. After I ran the migration, it turned any existing defense component selections in lists split at each character, like `["a", "r", "m", "y"]`.

leigh-mil: As discussed irl, the string gets split into characters here because the column type was not updated from `Array` to `String` on the Portfolio model.

leigh-mil:
> Looks most of the way there. I'm getting the following error when I try to navigate to the portfolio settings page now, after running the migrations:
> image

As discussed irl, this is caused by bad data that resulted from not updating the column type to an array when the portfolio form was updated, so the array was being stored as a string (ie `'{army, navy}'`), so when it is split into an array it is including the brackets (ie `['{army', 'navy}']`)

PR #1290, opened 2020-01-03
Fix translations 

leigh-mil:
## Description
Another small fix to get the CI tests running again.  The translations for all the JEDI clin texts were generalized, but they should not have been because the labels for CLINs are different on the reporting and TO view/review pages.


translations.yaml
```diff
@@ -554,10 +559,10 @@ task_orders:
   status_empty_state: 'This Portfolio has no {status} Task Orders.'
   status_list_title: '{status} Task Orders'
 JEDICLINType:
-  JEDI_CLIN_1: 'Unclassified IaaS and PaaS (IDIQ CLIN 0001)'
-  JEDI_CLIN_2: 'Classified IaaS and PaaS (IDIQ CLIN 0002)'
-  JEDI_CLIN_3: 'Unclassified Cloud Support Package (IDIQ CLIN 0003)'
-  JEDI_CLIN_4: 'Classified Cloud Support Package (IDIQ CLIN 0004)'
+  JEDI_CLIN_1: 'IDIQ CLIN 0001 Unclassified IaaS/PaaS'
```
graham-dds: It looks like we use this version of the `JEDICLINType` on the [reports](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/395405310) page, but the designs use `JEDI CLIN ...` instead of `IDIQ CLIN ...`

leigh-mil:
> This looks pretty straightforward, but instead of changing the translations file, I'm wondering if it makes more sense to look back at the designs and unify the way we present this information. These are so close to each other that I bet the case could be made to make them the same.

@graham-dds I agree that we could probably generalize the labels on the two pages, but is that something that we can fix later as we implement the copy changes? This PR is intended as a quick fix to get the ghost CI tests running again.

PR #1294, opened 2020-01-07
Portfolio admin styling - Managers table

leigh-mil:
## Description
Updates the portfolio managers table and add new manager form to be in line with Think's designs. 
The drop down menu to edit members and revoke/resend invites will be in a follow up PR. 

## Pivotal
https://www.pivotaltracker.com/story/show/170388182

## Screenshots
Screen Shot 2020-01-07 at 10 09 51 AM
Screen Shot 2020-01-07 at 10 10 11 AM
Screen Shot 2020-01-07 at 10 10 19 AM
Screen Shot 2020-01-07 at 10 10 27 AM
Screen Shot 2020-01-07 at 10 10 31 AM


atst/domain/portfolios/portfolios.py
```diff
@@ -75,10 +75,10 @@ def invite(cls, portfolio, inviter, member_data):
         permission_sets = PortfolioRoles._permission_sets_for_names(
             member_data.get("permission_sets", [])
         )
-        role = PortfolioRole(portfolio_id=portfolio.id, permission_sets=permission_sets)
+        role = PortfolioRole(portfolio=portfolio, permission_sets=permission_sets)
 
         invitation = PortfolioInvitations.create(
-            inviter=inviter, role=role, member_data=member_data
+            inviter=inviter, role=role, member_data=member_data["user_data"]
```
dandds: This change is breaking the seed_sample script.

atst/routes/portfolios/admin.py
```diff
@@ -17,63 +17,52 @@
...
 def render_admin_page(portfolio, form=None):
     pagination_opts = Paginator.get_pagination_opts(http_request)
     audit_events = AuditLog.get_portfolio_events(portfolio, pagination_opts)
-    members_data = get_members_data(portfolio)
     portfolio_form = PortfolioForm(obj=portfolio)
-    member_perms_form = member_forms.MembersPermissionsForm(
-        data={"members_permissions": members_data}
-    )
-
+    ppoc = filter_members_data([portfolio.owner_role])[0]
```
dandds: I don't think we should filter the whole list twice. Can you follow up with the designers about whether the PPoC needs to be the first member of the table? If they don't I think you could just just add a boolean key/value to your dictionary for each member to denote which one needs the PPoC tag.

atst/routes/portfolios/admin.py
```diff
@@ -106,34 +95,6 @@ def admin(portfolio_id):
...
-        return redirect(
-            url_for(
-                "portfolios.admin",
-                portfolio_id=portfolio_id,
-                fragment="portfolio-members",
-                _anchor="portfolio-members",
-            )
-        )
-    else:
-        return render_admin_page(portfolio)
```
dandds: Are we deleting this because the new designs have you edit users individually?

leigh-mil: @dandds  Yes! In the follow up PR (#1297) I add in the toggle menu and forms to edit the portfolio managers perms, resend invites, revoke invites, and remove managers.

atst/routes/portfolios/admin.py
```diff
@@ -17,63 +17,51 @@
...
+def filter_members_data(members_list, portfolio):
+    members_data = []
+    for member in members_list:
+        members_data.append(
+            {
+                "role_id": member.id,
+                "user_name": member.user_name,
+                "permission_sets": filter_perm_sets_data(member),
+                "status": member.display_status,
+                "ppoc": (True if portfolio.owner_role == member else False),
```
dandds: I think it might be faster to just do `PermissonSet.PORTFOLIO_POC in member.permission_sets` here. The `owner_role` method has to loop every portfolio member.

templates/components/member_form.html
```diff
@@ -0,0 +1,65 @@
...
+  <div class='action-group'>
+    {{ next_button }}
+    {% if previous %}
+      <input
+          type='button'
+          v-on:click="previous()"
+          class='action-group__action usa-button usa-button-secondary'
+          value='{{ "common.previous" | translate }}'>
+    {% endif %}
+    <a class='action-group__action' v-on:click="closeModal('{{ modal }}')">{{ "common.cancel" | translate }}</a>
```
dandds: Little thing, but you might want `cursor: pointer` on this element.

templates/components/member_form.html
```diff
@@ -0,0 +1,65 @@
+<!-- Layout macro -->
```
dandds: Is this used exclusively for portfolio manages? If so, I think we should call it something more specific than "member_form".

leigh-mil: @dandds Right now it is being used for app members and portfolio managers. However, when we aren't up against a deadline I would like to refactor and make a new modal form macro since they are all consistent design wise.

PR #1296, opened 2020-01-07
Remove an XSS vulnerability in CLIN title form input

graham-dds:
This PR escapes the title of a CLIN (which comes from the Clin number input) so that the page is no longer vulnerable to XSS.

### DIFF
--- a/js/components/clin_fields.js
+++ b/js/components/clin_fields.js
@@ -1,4 +1,5 @@
 import { emitFieldChange } from '../lib/emitters'
+import escape from '../lib/escape'
 import optionsinput from './options_input'
 import textinput from './text_input'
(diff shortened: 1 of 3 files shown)


graham-dds:
I was surprised I had to resort to this non-Vue solution. I looked at how we were displaying the text in the template, and I noticed that we were using [`v-html`](https://vuejs.org/v2/guide/syntax.html#Raw-HTML) which as the name suggests, displays un-escaped html. I thought simply switching `v-html` to [`v-text`](https://vuejs.org/v2/api/#v-text) would do the trick since Vue's text interpolation escapes HTML by default, but this didn't seem to have any effect when I tested it.

PR #1300, opened 2020-01-09
Update environment roles list

leigh-mil:
## Description
Update environment roles list with new roles list. 
Update roles column type from string to enum, and fixed all tests and functions impacted by the updated.

## Pivotal
https://www.pivotaltracker.com/story/show/170606132


alembic/versions/828d8c188dce_update_environment_roles_enum_list.py
```diff
@@ -0,0 +1,59 @@
...
+        "role",
+        type_=sa.Enum(
+            "ADMIN",
+            "BILLING_READ",
+            "CONTRIBUTOR",
+            name="role",
+            native_enum=False,
+        ),
+        existing_type=sa.VARCHAR(),
+        nullable=True,
```
dandds: Do we actually want this to be nullable? Genuine question. Is there a legit case where we might have an env role that corresponds to nothing? (If we opt to make it non-null I would make all the existing users contributors.)

leigh-mil: @dandds If the user does not have a role for that environment, the role is `None`

dandds: Where are we creating `environment_role` entries for users who don't have any real environment role?

leigh-mil: When a role becomes disabled then the role is None. If the user isn't assigned a role in the first place then we don't create an environment_role for them.

PR #1302, opened 2020-01-09
uWSGI plugin config changes.

dandds:
uWGI was generating warnings about being unable to find plugin files we
specify. To fix this, I've added uwsgi-python3 to the list of Alpine
packages we install in the container specified the plugins directory in
the uWSGI config. The updated uWSGI ConfigMap has been applied to the
staging cluster, which eliminated the warning about the logfile plugin.
The remaining warning about the python3 plugin will be eliminated once
the new container built by this branch is deployed.

PT: https://www.pivotaltracker.com/story/show/170505212


whit1206:
I have a question, will this help with the timestamp not being evaluated in the log output or is that a separate issue?

dandds:
@whit1206 I don't think so; I'll follow up with you on Slack.

PR #1305, opened 2020-01-10
Record users' certificate serial numbers.

dandds:
In order to eventually disable users based on the CRLs, ATAT needs to
record users' certificate serial numbers. We record whatever serial
number is presented when they log in. If they present a different
certificate on a subsequent login, we record the new serial number. This
is based around two assumptions validated with CCPO:

- DISA will never unrevoke revoked certificate serial numbers.
- When a user's certificate it revoked, DISA will revoke all their
  certificates, so all of their certificate serial numbers will be
  published to CRLs. This means recording any one of their serial
  numbers is sufficient.

Also stores issuer hash for user's certificates.

Storing the serial number alone for a user's certificate is not enough
to positively recognize a cert. A serial number is only unique with
reference to a given CA. We need to also store the issuer hash so that
we can correctly identify revoked users based on their serial number and
the hash of the issuer.

I chose to implement this by making the parsed certificate available
from the CRLCache. This is slightly awkward but is the path of least
resistance for getting that bit of information.


dandds:
RFC 5280 says serial numbers must be unique for the CA, so we need to store the CA somehow. https://www.ietf.org/rfc/rfc5280.txt (4.1.2.2.  Serial Number)

dandds:
Closing; de-prioritizing this work. We'll wait for the DISA identity provider to become available.

PR #1308, opened 2020-01-12
Additional validation and escaping for file names.

dandds:
This adds additional front and backend validations for task order file
names. Additionally, it escapes HTML characters in file names anywhere
they are used on the frontend. We are now restricting file names to a
whitelist regex of [A-Za-z0-9\-_ \.] for simplicity.

Note:
On the frontend, the filename string must have at least one character.
This is not true in the backend validation; because of the way the
entire task order form is validated, requiring input would break the
business logic currently implemented.

PT: https://www.pivotaltracker.com/story/show/170636650

### DIFF
--- a/atst/forms/task_order.py
+++ b/atst/forms/task_order.py
@@ -7,7 +7,7 @@ from wtforms.fields import (
     HiddenField,
 )
 from wtforms.fields.html5 import DateField
-from wtforms.validators import Required, Length, NumberRange, ValidationError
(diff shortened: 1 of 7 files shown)


graham-dds:
https://github.com/dod-ccpo/atst/pull/1314 I think using `v-text` instead of `v-html` will remove the need to use `escape()`

dandds:
@graham-dds Yeah, I'll adjust this once that's merged. I think we still want to be very restrictive with file names since they get written to our database and sent to the CSP.

PR #1311, opened 2020-01-14
170614119 - Adds initial connection lists and architecture doc

robgil-dds:
This adds the initial lists of transient connections (protocols, ports,
etc). Also adds listening services and hosts used.

### DIFF
--- a/docs/EdgeControls.md
+++ b/docs/EdgeControls.md
@@ -2,19 +2,19 @@
 This document describes the expected connections and listening services.
 
 ## Transient Connections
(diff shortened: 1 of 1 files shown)


robgil-dds:
@whit1206 Updated with protocols. I figure application protocol would be appropriate here vs. just tcp/udp.

For inbound/outbound, I have this documented as Ingress/Egress.

The CRL service is listed under the "Transient Connections" section. 

We'll have to continually update this as we learn more about the apps/services. I'm sure we're missing things in here, but if this looks good to you for a start, lets merge and keep iterating on it.

PR #1312, opened 2020-01-14
Ghost Inspector updates 1/14/2019

JRNPW:
It has been a while since I updated GitHub from Ghost Inspector. During that time, many tests were updated because we've been doing a good bit of work to match designs more accurately. This PR represents all those changes.


JRNPW:
All dates should read 2020, not 2019. Oops.

PR #1313, opened 2020-01-14
Typography updates

Hmbrink:
- Adjusted H4 font size to 1.9rem
- Added semibold font weight


dandds:
@Hmbrink thanks, this is great!

@JRNPW Your Ghost Inspector screenshots are all option, right? So this shouldn't fail 👻 ?

PR #1314, opened 2020-01-14
Use v-text instead of v-html to mitigate XSS attacks

graham-dds:
In [this PR](https://github.com/dod-ccpo/atst/pull/1296), I was surprised that using [`v-text`](https://vuejs.org/v2/api/#v-text) instead of [`v-html`](https://vuejs.org/v2/guide/syntax.html#Raw-HTML) didn't mitigate XSS attacks. Well, I gave it another go, and found that it worked?

I swapped `v-text` for `v-html` in most instances. This PR addresses:
https://www.pivotaltracker.com/story/show/170565459
image

https://www.pivotaltracker.com/story/show/170708088
image

https://www.pivotaltracker.com/story/show/170562838 *
image

* even though this was already done, instead of using `esacpe()` in some places and `v-text` in others, I opted to refactor this bit to be consistent.

I left `v-html` for `validationError`s and the `title` of the `Alert()` macro. We should double check that the validation errors and alerts that we surface don't rely on user input, possibly in another story / PR.

### DIFF
--- a/js/components/clin_fields.js
+++ b/js/components/clin_fields.js
@@ -1,5 +1,4 @@
 import { emitFieldChange } from '../lib/emitters'
-import escape from '../lib/escape'
 import optionsinput from './options_input'
 import textinput from './text_input'
 import clindollaramount from './clin_dollar_amount'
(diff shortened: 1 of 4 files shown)


JRNPW:
For posterity, and for others reviewing this Pull Request, it's important to note why this works.

In the description for [v-html](https://vuejs.org/v2/guide/syntax.html#Raw-HTML):

>     The double mustaches interprets the data as plain text, not HTML.

And in the explanation for [v-text](https://vuejs.org/v2/api/#v-text):

```
<span v-text="msg"></span>
<!-- same as -->
<span>{{msg}}</span>
```

PR #1317, opened 2020-01-14
First Pass Azure Tenant Creation Integration + Orchestration

tomdds:
## Description

This PR is the integration with the custom tenant setup endpoints. It currently is just a series of separate functions that each do a step of the integration. Next step for this PR is to integrate the functions with the Portfolio State Machine.


atst/domain/csp/cloud.py
```diff
@@ -657,102 +879,188 @@ def create_atat_admin_user(
...
+                "Could not resolve token for billing profile creation"
+            )
+
+        create_billing_account_body = payload.dict(by_alias=True)
+
+        create_billing_account_headers = {
+            "Authorization": f"Bearer {sp_token}",
+        }
+
+        # TODO: unsure if this is a static value or needs to be constructed/configurable
```
dandds: Sounds like no? Or is it the billing account name of whatever the current whitelisted app registration is associated with?

PR #1319, opened 2020-01-15
Remove the remaining uses of v-html

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/170636527

No more `v-html` anywhere!

I was able to do this safely because the remaining `v-html` calls were being used on `validationError`s from [`input_validations.js`](https://github.com/dod-ccpo/atst/blob/bugfix/use-v-text-where-possible/js/lib/input_validations.js), and a few ([1](https://github.com/dod-ccpo/atst/blob/bugfix/use-v-text-where-possible/translations.yaml#L317)) places ([2](https://github.com/dod-ccpo/atst/blob/bugfix/use-v-text-where-possible/translations.yaml#L110)) in `translations.yaml`. None of these entries use markup, so I switched them to `v-text`. This has benefits that were previously explored in https://github.com/dod-ccpo/atst/pull/1314

### DIFF
--- a/templates/components/alert.html
+++ b/templates/components/alert.html
@@ -25,7 +25,7 @@
 
     <div class='usa-alert-body'>
       {% if vue_template %}
-        <h3 class='usa-alert-heading' v-html='title'></h3>
+        <h3 class='usa-alert-heading' v-text='title'></h3>
       {% elif title %}
         <h3 class='usa-alert-heading'>{{ title | safe }}</h3>
       {% endif %}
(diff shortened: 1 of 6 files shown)


graham-dds:
This PR also addresses the comment that I left at the bottom of https://github.com/dod-ccpo/atst/pull/1314. @JRNPW. Since v-html is no longer being used anywhere, `Alerts()` are no longer susceptible like I thought they were.

PR #1322, opened 2020-01-16
Operator KeyVault

robgil-dds:
This is a new keyvault specifically for storing operator secrets and
things that would not be accessible to applications. The primary use
case for this is for launching things like postgres (root postgres
creds) and other services which would require secrets to be added to the
terraform configuration. This approach avoids adding secrets to
terraform.

An accompanying script will be added to populate the new keyvault.


robgil-dds:
I merged this with the secrets tool PR https://github.com/dod-ccpo/atst/pull/1321. Closing this in favor of the other PR so I'm not blocked waiting for merges.

PR #1327, opened 2020-01-16
Text input validation

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/170601644

Adds a little more text validation for a few `StringFields()` in our forms.

It also:
 - alters the `IsNumber()` validator to allow falsey values. We should use WTForm's built-in `Requried()` validator if we want input
- Makes CLIN numbers a max of 4 digits, and alters some test data to reflect this fact

TODO in another PR: Make sure all of our text input fields have a character limit in the models

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -222,7 +222,7 @@ def make_config(direct_config=None):
         config.read_dict({"default": direct_config})
 
     # Assemble DATABASE_URI value
-    database_uri = "postgres://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
+    database_uri = "postgresql://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
         config.get("default", "PGUSER"),
         config.get("default", "PGPASSWORD"),
         config.get("default", "PGHOST"),
--- a/atst/forms/application.py
+++ b/atst/forms/application.py
@@ -1,14 +1,14 @@
 from .forms import BaseForm, remove_empty_string
 from wtforms.fields import StringField, TextAreaField, FieldList
-from wtforms.validators import Required, Optional
-from atst.forms.validators import ListItemRequired, ListItemsUnique
+from wtforms.validators import Required, Optional, Length
+from atst.forms.validators import ListItemRequired, ListItemsUnique, Name, AlphaNumeric
 from atst.utils.localization import translate
 
 
 class EditEnvironmentForm(BaseForm):
     name = StringField(
         label=translate("forms.environments.name_label"),
-        validators=[Required()],
+        validators=[Required(), Name(), Length(max=100)],
         filters=[remove_empty_string],
     )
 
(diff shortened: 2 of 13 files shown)


atst/app.py
```diff
@@ -222,7 +222,7 @@ def make_config(direct_config=None):
         config.read_dict({"default": direct_config})
 
     # Assemble DATABASE_URI value
-    database_uri = "postgres://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
+    database_uri = "postgresql://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
```
graham-dds: Screen Shot 2020-01-16 at 9 55 25 AM

atst/forms/application.py
```diff
@@ -31,6 +31,7 @@ class EnvironmentsForm(BaseForm):
         StringField(
             label=translate("forms.application.environment_names_label"),
             filters=[remove_empty_string],
+            validators=[AlphaNumeric(), Length(max=100)],
```
dandds: This is good, but I'm reluctant to enforce backend validations that we don't also have on the frontend. That's a surprise for the user when they submit the form. (Or at least I don't see where we're enforcing this on the frontend.)

atst/forms/portfolio.py
```diff
@@ -20,10 +21,11 @@ class PortfolioForm(BaseForm):
                 min=4,
                 max=100,
                 message=translate("forms.portfolio.name.length_validation_message"),
-            )
+            ),
+            Name(),
         ],
     )
-    description = TextAreaField(translate("forms.portfolio.description.label"),)
+    description = TextAreaField(translate("forms.portfolio.description.label"))
```
dandds: Do we want a length validator? (I don't know if the intent of this PR is to cover every string and hidden field.)

tests/factories.py
```diff
@@ -327,7 +331,7 @@ class Meta:
         model = CLIN
 
     task_order = factory.SubFactory(TaskOrderFactory)
-    number = factory.LazyFunction(random_task_order_number)
+    number = factory.LazyFunction(random_clin_number)
```
dandds: I was wondering why I was seeing 13-digit numbers in my test database.

atst/forms/task_order.py
```diff
@@ -129,7 +137,7 @@ class AttachmentForm(BaseForm):
             Length(
                 max=40, message=translate("forms.attachment.object_name.length_error")
             ),
-            Regexp(regex=REGEX_ALPHA_NUMERIC),
+            AlphaNumeric(),
         ],
     )
```
dandds: There are also `HiddenField` instance in `atst.forms.application_member.EnvironmentForm`; can you add this validator there, too?

graham-dds:
@dandds I'm going to reopen this PR to add:
- frontend validation where necessary
- remaining `Length()` validations to form fields
- character limits on our models, which will add a migration

I was initially thinking of doing this in two PRs, but one probably makes sense for this

graham-dds:
@dandds I'm going to reopen this PR to add:
- frontend validation where necessary
- remaining `Length()` validations to form fields
- character limits on our models, which will add a migration

I was initially thinking of doing this in two PRs, but one probably makes sense for this

PR #1328, opened 2020-01-17
Update _topbar.scss

dandds:
-Cleaned up spacing, margins, and padding
-Refined font styling for JEDI logo and added semibold weight to nav links


dandds:
@Hmbrink thanks for this! I'm going to merge now so that we don't need to rebase it again.

PR #1334, opened 2020-01-17
Portfolio Layout & Spacing

Hmbrink:
Portfolio adjustments:
-Cleaned up container margins/padding
-Added max-width to header and Settings section
-Refined Portfolio header and nav links

Other adjustments:
-Reduced font size of side nav links

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -4,16 +4,19 @@
     min-height: 500px;
   }
 
-  margin-left: 2 * $gap;
+  .col.col--grow {
+    padding: 0;
+  }
 }
 
 .portfolio-header {
   flex-direction: column;
+  align-items: center;
+  margin: $spacing-medium 0;
+  max-width: 90rem;
   @include media($small-screen) {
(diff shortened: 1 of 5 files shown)


styles/components/_portfolio_layout.scss
```diff
@@ -4,32 +4,38 @@
...
-  margin-left: 2 * $gap;
+  .col.col--grow {
+    padding: 0;
+  }
+
 }
 
 .portfolio-header {
   flex-direction: column;
+  align-items: center;
```
dandds: We have to support IE11. [Caniuse.com says `align-items` isn't supported in 11](https://caniuse.com/#search=align-items), but it looks fine when I load the page in 11 🤷‍♂ :

Screen Shot 2020-01-19 at 11 04 08 AM

If that screenshot looks right to you, I think it's okay to use.

dandds: Not a problem! Like I said, it doesn't actually look wrong to me in IE11, but I'll leave it to you to check compatibility. Thanks!

PR #1335, opened 2020-01-18
Display client source IP correctly in logs.

dandds:
In order to display the client source IP, we need to make a
configuration change to the cluster. Setting externalTrafficPolicy to
"Local" preserves the client IP, per:

https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

### DIFF
--- a/deploy/azure/azure.yml
+++ b/deploy/azure/azure.yml
@@ -309,6 +309,7 @@ metadata:
   namespace: atat
 spec:
   loadBalancerIP: 13.92.235.6
+  externalTrafficPolicy: Local
   ports:
     - port: 80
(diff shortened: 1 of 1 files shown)


raydds:
> Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type services, but risks potentially imbalanced traffic spreading.

I'm sure you saw this, but this seems like an acceptable trade-off to me.

dandds:
@raydds Yeah, I forgot; I meant to bring that to your attention in case it affects any of the load testing. Thanks!

PR #1336, opened 2020-01-19
Copy updates

leigh-mil:
## Description
Various copy updates:
- Remove lorem text
- New portfolio form
- TO builder
- New application form
- App settings page

## Pivotal
https://www.pivotaltracker.com/story/show/170609885

### DIFF
--- a/atst/utils/flash.py
+++ b/atst/utils/flash.py
@@ -29,7 +29,7 @@ MESSAGES = {
         "category": "error",
     },
     "application_invite_resent": {
-        "title": "flash.application_invite.resent.title",
+        "title": "",
         "message": "flash.application_invite.resent.message",
         "category": "success",
     },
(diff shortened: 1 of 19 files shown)


atst/utils/flash.py
```diff
@@ -29,7 +29,7 @@
         "category": "error",
     },
     "application_invite_resent": {
-        "title": "flash.application_invite.resent.title",
+        "title": "",
```
dandds: I think we should use `None` in these cases to be consistent.

translations.yaml
```diff
@@ -98,8 +98,7 @@ flash:
       title: Application invitation error
       message: There was an error processing the invitation for {user_name} from {application_name}
     resent:
-      title: Application invitation resent
-      message: You have successfully resent the invite for {user_name} from {application_name}
+      message: "{email} has been sent an invitation to access this Application"
```
dandds: Should we update `flash.new_portfolio_member` to match this?

dandds: In fact, probably all the portfolio/app invite flash content should match.

PR #1337, opened 2020-01-19
Forcibly destroy existing session on logout.

dandds:
To comply with security guidelines, we need to destroy the session when
a user logs out. This means that the session's key in the Redis cache
needs to be deleted. Flask expects to _always_ have a session object. If
the current session object does not exist in the Redis cache, Flask will
reserialize and store it at the end of the request. In order for
session deletion to work, we need to delete the key for the existing
session and then replace the session object with a new, empty one.

This also updates the SessionLimiter class so that the session prefix is
configurable.

PT: https://www.pivotaltracker.com/story/show/170753153

Additionally, this adds configuration to that `LIMIT_CONCURRENT_SESSIONS` is set to true by default for deployed environments.

PT: https://www.pivotaltracker.com/story/show/170635453


dandds:
@whit1206 @JRNPW One nuance I didn't realize: Flask, the web framework, _always_ updates a session implicitly at the end of every web request. On this PR, we destroy the old session here:

https://github.com/dod-ccpo/atst/pull/1337/files#diff-d6241e5702eb50643ca00d89c0e0fb0aR71

and then override the in-memory session object with a new, empty session:

https://github.com/dod-ccpo/atst/pull/1337/files#diff-d6241e5702eb50643ca00d89c0e0fb0aR73

This means that the user's existing session is forcibly deleted when they log out and that we replace the old session with a new, empty one. This satisfies Flask's requirement that there be a session object to update at the end of the request.

whit1206:
@dandds This looks really good. I have a question for you about the new empty session object that is created. With this empty session, what could someone do with it if they got it? What is the difference between a logged in user with a session? Can these empty session expire? Will it add extra work for redis to do? Thanks

dandds:
@whit1206 Thanks!

> With this empty session, what could someone do with it if they got it? What is the difference between a logged-in user with a session and an empty one? Can these empty sessions expire and will it add extra workload on redis that we should be concerned about? How does the empty session effect logins at all? If there is am empty session vs an existing session object vs no session object?

if someone got ahold of the empty session, they couldn't do anything. Every used has a session whether they are logged in or not. When a user logs in, we populate a `user_id` field in the server-side session. On every subsequent request they make, we can tell whether they're logged in or not by the presence of that `user_id` field in the server-side session. An empty session means that that field is not populated, so they're logged out. (An empty session only has some basic metadata associated with it.)

The empty sessions expire just like regular ones, but we don't incur much penalty/extra workload. Redis's performance is very good; I'd guess we'd need many thousands of concurrent users before it would start to matter.

whit1206:
@dandds Thanks for the additional info. 

One last question, do we have an automated test to expire sessions or do we have to test this manually?

dandds:
@whit1206 No, we don't have anything automated to test session expiry; that's been manual so far.

PR #1338, opened 2020-01-20
Configure processes, threads, and increase CPU/memory limits

raydds:
Based on [this benchmarking](https://gist.github.com/raydds/8802da6ff6a79bb4dbdd4351be0ed7d8), 4 processes and 2 threads seems to produce the highest number of requests per second.

However, running with 2 processes and 2 threads allows us to fit two replicas on a node with 2vcpus. Therefore, I'm advocating for running with 2 processes and 2 threads. Keep in mind, that each node has 1930m CPU available. That means, in order to fit two replicas on a single node, we need to limit the web process and nginx to 965m CPU (combined).

Increasing the number of processes and threads did increase CPU/memory usage, so I had to bump the limits.

[Here's what it looks like when we're really hitting the app with some load](https://gist.github.com/raydds/ac45ff5e6fff127316ef9ba4d824a60b). Previously, we were getting ~2.5RPS, but with these changes, I saw the app get up to 9.5RPS.

### DIFF
--- a/deploy/azure/autoscaling.yml
+++ b/deploy/azure/autoscaling.yml
@@ -17,7 +17,7 @@ spec:
     - type: Resource
       resource:
         name: cpu
-        targetAverageUtilization: 70
+        targetAverageUtilization: 60
 ---
 apiVersion: autoscaling/v2beta1
 kind: HorizontalPodAutoscaler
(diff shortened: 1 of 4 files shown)


rzane:
Yeah, the env vars were extremely handy! Otherwise, I would have had to rebuild the container before each round of experiments, which would have taken much longer!

raydds:
Yeah, if it weren't for the env vars, I would have had to rebuild the containers before each round of experiments!

PR #1342, opened 2020-01-21
Validation on route parameters

leigh-mil:
## Description
Removes the outdated and unused `/help` route.
Adds in a function that validates a url path, which addresses the security vulnerability of rendering templates with params passed in by the user without validation. 

## Pivotal
https://www.pivotaltracker.com/story/show/170603379


graham-dds:
Didn't dig into the code yet, but we should check whether or not we should fully delete the `/help` route:
image

leigh-mil:
@graham-dds I checked in with Ben about deleting it since it is undetermined where help content will go. That link will be to an email address that is TBD.

atst/routes/__init__.py
```diff
@@ -42,26 +43,21 @@ def root():
...
-        return render_template("{}.html".format(path))
-    except TemplateNotFound:
+    path_valid = re.match(r"(/[-a-zA-Z]+|/\d+)+", path)
+
+    if path_valid:
+        try:
+            return render_template("{}.html".format(path))
+        except TemplateNotFound:
+            raise NotFound()
+    else:
```
dandds: Are there any valid use cases for this route right now? If every page in the app has an explicit route specified elsewhere, I'd say we can just get rid of this.

tests/test_auth.py
```diff
@@ -94,32 +94,6 @@ def test_protected_routes_redirect_to_login(client, app):
...
-    user = UserFactory.create()
-
-    resp = client.get(url_for("atst.helpdocs"))
-    assert resp.status_code == 200
-    assert user.full_name not in resp.data.decode()
-
-    user_session(user)
-    resp = client.get(url_for("atst.helpdocs"))
-    assert resp.status_code == 200
-    assert user.full_name in resp.data.decode()
```
dandds: It definitely shouldn't be dup'd, but I think this is still a valuable test. It's really checking that a user's name is displayed when they log in, even on a page that doesn't require authentication. You could just substitute one of the other routes here for `atst.helpdocs`: https://github.com/dod-ccpo/atst/blob/staging/atst/domain/auth.py#L6-L16

dandds: In fact you should also remove `atst.helpdocs` from that whitelist I linked.

PR #1343, opened 2020-01-21
Form text input validation

graham-dds:
Addresses https://www.pivotaltracker.com/story/show/170601644

Adds missing text input validation for text input to forms on the frontend and backend.

A schema migration will come in another PR, because that would add noise to an already (somewhat) noisy PR.

Speaking of noise -- There are a lot of single quoute -> double quote changes in `translations.yaml` -- an autoformatting plugin that I have on VSCode keeps `"helping"` me when I edit that file, so I thought I might as well leave the changes in this time if it's all the same.

### DIFF
--- a/atst/app.py
+++ b/atst/app.py
@@ -222,7 +222,7 @@ def make_config(direct_config=None):
         config.read_dict({"default": direct_config})
 
     # Assemble DATABASE_URI value
-    database_uri = "postgres://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
+    database_uri = "postgresql://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
         config.get("default", "PGUSER"),
         config.get("default", "PGPASSWORD"),
         config.get("default", "PGHOST"),
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -36,10 +36,11 @@ def DateRange(lower_bound=None, upper_bound=None, message=None):
 
 def IsNumber(message=translate("forms.validators.is_number_message")):
     def _is_number(form, field):
-        try:
-            int(field.data)
-        except (ValueError, TypeError):
-            raise ValidationError(message)
+        if field.data:
+            try:
+                int(field.data)
+            except (ValueError, TypeError):
+                raise ValidationError(message)
 
     return _is_number
 
--- a/atst/forms/validators.py
+++ b/atst/forms/validators.py
@@ -2,13 +2,10 @@ from datetime import datetime
 import re
 
 from werkzeug.datastructures import FileStorage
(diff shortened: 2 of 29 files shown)


atst/app.py
```diff
@@ -222,7 +222,7 @@ def make_config(direct_config=None):
         config.read_dict({"default": direct_config})
 
     # Assemble DATABASE_URI value
-    database_uri = "postgres://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
+    database_uri = "postgresql://{}:{}@{}:{}/{}".format(  # pragma: allowlist secret
```
graham-dds: image

atst/forms/validators.py
```diff
@@ -36,10 +33,11 @@ def _date_range(form, field):
 
 def IsNumber(message=translate("forms.validators.is_number_message")):
```
leigh-mil: Can we change this to just `Number` to make it consistent with the older validator names?

graham-dds: Good call!

js/lib/input_validations.js
```diff
@@ -53,6 +73,13 @@ export default {
     unmask: [','],
     validationError: 'Please enter a number',
   },
+  name: {
+    mask: false,
+    match: /[\w \,\.\'\-]{1,100}/,
```
dandds: This is sort of a pain, but checking the docs I don't think `\w` means the same thing in Python in JS. In most regex implementations that's used as a character class meaning "any valid alphanumeric character". The question is whether that includes Unicode word characters. It looks like in Python it does by default:

https://docs.python.org/3/library/re.html#regular-expression-syntax

and in JS it does not:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Character_Classes

This means someone would get a frontend validation error for a name with non-latin characters but not a backend validation error. Maybe this answer will help? https://stackoverflow.com/questions/280712/javascript-unicode-regexes

I wouldn't spend too much time on this; if it's a pain to get it right in the JS, I would just loosen that regex for now to allow most characters and document the problem with a PT ticket and a comment there near the validator. We still have the protection of the backend validation; the frontend validation is for UX.

translations.yaml
```diff
@@ -1,17 +1,17 @@
...
+# `{{ "login.title" | translate | safe }}`
 audit_log:
   events:
     default:
-      change: '{from} to {to}'
-      changes: 'Changes:'
-      details: 'Details:'
+      change: "{from} to {to}"
+      changes: "Changes:"
+      details: "Details:"
```
dandds: I don't have a horse in the quote race, but be careful rebasing since there are a lot of changes happening to this file.

PR #1347, opened 2020-01-22
Tf db user

dandds:
WIP

### DIFF
--- a/terraform/modules/keyvault/main.tf
+++ b/terraform/modules/keyvault/main.tf
@@ -70,3 +70,9 @@ resource "azurerm_key_vault_access_policy" "keyvault_admin_policy" {
     "update",
   ]
 }
+
+resource "azurerm_key_vault_secret" "secret-pgpassword" {
+  name         = var.secret_pgpassword_name
+  value        = var.secret_pgpassword_value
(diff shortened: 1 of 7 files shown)


dandds:
@robgil-dds Two questions, one simple and one probably less so. I've got some kind of logical TF error with the way I'm trying to mirror the secret from the operator vault to the app secrets vault in the second commit. I get errors like:

```
Error: Missing required argument

  on secrets.tf line 1, in module "operator_keyvault":
   1: module "operator_keyvault" {

The argument "secret_pgpassword_value" is required, but no definition was
found.
```

from `terraform plan`.

Second question: the postgres provider can't connect to the database:

```
Error: Error initializing PostgreSQL client: error detecting capabilities: error PostgreSQL version: pq: no pg_hba.conf entry for host "
108.16.207.173", user "BxILtasSLhzeTGFRWxzysguIWktFfm", database "postgres", SSL on

  on ../../modules/postgres/main.tf line 47, in provider "postgresql":
  47: provider "postgresql" {
```

Do I need to update the "pg_hba.conf" file, probably with that Postgres settings resource? Would I have to whitelist the current operator's IP 😬 ?

dandds:
Closing in favor of another approach

PR #1351, opened 2020-01-23
Spacing & Layout, Content Max-Width, Nav/Footer refinements, Portfolio Header alignment

Hmbrink:


styles/components/_empty_state.scss
```diff
@@ -1,8 +1,6 @@
 .empty-state {
-  padding: $gap * 3;
-  max-width: 100%;
-  background-color: $color-gray-lightest;
-  margin-top: $gap * 5;
+  max-width: $max-panel-width;
+  background-color: #f6f6f7;
```
leigh-mil: In the `_variables.scss` file there is a variable defined `$color-offwhite: #fbfbfd;`, which is very close in color to `#f6f6f7`. 
`$color-offwhite` is what is used across the site for the background, can you use that variable here? Or can we update the variable to use the hex code you used here?

templates/portfolios/new/step_1.html
```diff
@@ -10,10 +10,11 @@
...
 <main class="usa-section usa-content">
   {% include "fragments/flash.html" %}
-  <div class='portfolio-header__name'>
-    <p>{{ "portfolios.header" | translate }}</p>
-    <h1>{{ "portfolios.new.title" | translate }}</h1>
-  </div>
+  <div class="portfolio-header-new">
+    <div class='portfolio-header__name'>
+      <p>{{ "portfolios.header" | translate }}</p>
+      <h1>{{ "New Portfolio" }}</h1>
```
leigh-mil: Can you change line 16 to be `<h1>{{ 'portfolios.new.title' | translate }}</h1>`

styles/components/_empty_state.scss
Hmbrink: Let's stick with what exists! I'll adjust to `$color-offwhite: #fbfbfd`.

Hmbrink: After testing, the off-white blending in with the background for certain states isn't the best, so I'm going use the existing variable `$color-gray-lightest` to keep content contained.

PR #1354, opened 2020-01-23
New TO Number Rules

leigh-mil:
## Description
Updates TO validation and coersion to account for alpha numeric TO numbers, 13-17 characters, filtering out dashes, and coercing TO numbers to be all caps.

## Pivotal
https://www.pivotaltracker.com/story/show/170846902


atst/forms/task_order.py
```diff
@@ -60,6 +62,20 @@ def validate_date_in_range(form, field):
...
+        return value.replace("-", "")
+    else:
+        return None
+
+
+def coerce_upper(value):
+    if value:
+        return value.upper()
+    else:
+        return None
```
graham-dds: To tighten these functions up, you could:
- use a ternary
```python
def remove_dashes(value):
    return value.replace("-", "") if value else None
```
- drop the else clause, since a function without a return statement will return `None` by default
```python
def coerce_upper(value):
    if value:
        return value.upper()
```

Also, if these filters are only used by TOs, it might be worth just defining them as `lambda` functions in the `TaskOrderForm`

js/lib/input_validations.js
```diff
@@ -106,9 +106,9 @@ export default {
   },
   taskOrderNumber: {
     mask: false,
-    match: /^.{13}$/,
+    match: /^.{13,17}[^-]$/,
```
dandds: The story says:

> No less than 13 chars, no more than 17
> Some TOs include dashes but these don't 'count' as part of the TO, ie, ABC-DE-F == ABCDEF.

Does this mean that a user can enter _more_ than 17 characters, but the TO number has to a maximum of 17 characters when the dashes are stripped? I think that's different than what the validation here permits.

atst/forms/task_order.py
dandds: I remember seeing a comment from @graham-dds that these might be nicer as ternaries, but now I can't find it.

js/lib/input_validations.js
dandds: I'm also not sure about this regex. It looks like it says "13 to 17 of any character with no trailing hyphen."

atst/forms/task_order.py
graham-dds: Yeah @dandds, I had mentioned this, but I deleted the comment because I clicked "Add single comment" when I wanted to use "Start a review". But this ended up being the only comment I had 🤷‍♂

PR #1359, opened 2020-01-26
Application Provisioning

dandds:
This wires up the orchestration for application provisioning. 

It adds a few incidental changes:
- Fixes import errors and missing imports.
- Collapses multiple model-specific job failure tables into one.

See commits for details.

The `create_application` method has been tested with real Azure service principal creds. See comments on PT card: https://www.pivotaltracker.com/story/show/170850018/comments/210774035

PT: https://www.pivotaltracker.com/story/show/170850018


dandds:
@tomdds Curious if the linked method is close to what you had in mind. I needed a way to get the correct creds to add to the payload dataclass for `create_application`. It seems like something the CSP interface should resolve internally for the caller:

https://github.com/dod-ccpo/atst/pull/1359/files#diff-2d0a9d6676bd8bbb124d56ef75fa30c3R1473

dandds:
Also `test_get_applications_pending_creation` passes for me locally but fails in CI every time. :/

atst/domain/applications.py
```diff
@@ -118,3 +125,21 @@ def invite(
         db.session.commit()
 
         return invitation
+
+    @classmethod
+    def get_applications_pending_creation(cls) -> List[UUID]:
+        results = (
+            db.session.query(Application.id)
+            .join(Portfolio)
+            .join(PortfolioStateMachine)
```
graham-dds: I think you can join both of these on the same line `.join(Portfolio, PortfolioStateMachine)`

atst/domain/applications.py
```diff
@@ -118,3 +125,21 @@ def invite(
...
+            .filter(PortfolioStateMachine.state == FSMStates.COMPLETED)
+            .filter(Application.deleted == False)
+            .filter(Application.cloud_id.is_(None))
+            .filter(
+                or_(
+                    Application.claimed_until.is_(None),
+                    Application.claimed_until <= func.now(),
+                )
+            )
```
graham-dds: You could also do this with one filter call instead of chaining them
```python
.filter(
    PortfolioStateMachine.state == FSMStates.COMPLETED,
    Application.deleted == False,
    Application.cloud_id.is_(None),
    or_(
        Application.claimed_until.is_(None),
        Application.claimed_until <= func.now(),
    ),
)
```
I don't know if either is actually better though 🤔 . What do you think?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -109,9 +109,7 @@ def get_secret(self, secret_key):
                 exc_info=1,
             )
 
-    def create_environment(
-        self, auth_credentials: Dict, user: User, environment: Environment
-    ):
+    def create_environment(self, auth_credentials: Dict, user, environment):
```
graham-dds: Why did we drop the `User` and `Environment` type checking from this function? Looks like this happened in a few places in this PR -- just wanted to be sure that it was intentional.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -626,3 +631,27 @@ def _root_creds(self):
...
         }
+
+    def get_credentials(self, scope="portfolio", tenant_id=None):
+        """
+        This could be implemented to determine, based on type, whether to return creds for:
+            - scope="atat": the ATAT main app registration in ATAT's home tenant
+            - scope="tenantadmin": the tenant administrator credentials
+            - scope="portfolio": the credentials for the ATAT SP in the portfolio tenant
+        """
+        if scope == "atat":
```
graham-dds: Should this function throw an error if the provided `scope` param isn't one of the branches of this `if / else` block? Or is returning `None` fine in this case?

atst/domain/csp/cloud/mock_cloud_provider.py
```diff
@@ -340,3 +343,12 @@ def _authorize(self, credentials):
         self._delay(1, 5)
         if self._with_authorization and credentials != self._auth_credentials:
             raise self.AUTHENTICATION_EXCEPTION
+
+    def create_application(self, payload: ApplicationCSPPayload):
+        self._maybe_raise(self.UNAUTHORIZED_RATE, GeneralCSPException)
+
+        id_ = f"{AZURE_MGMNT_PATH}{payload.management_group_name}"
+        return ApplicationCSPResult(id=id_)
```
graham-dds: Any reason that we define this variable above rather than just sticking it in this line? i.e.
`return ApplicationCSPResult(id=f"{AZURE_MGMNT_PATH}{payload.management_group_name}")`

atst/domain/csp/cloud/models.py
```diff
@@ -232,3 +234,57 @@ class Config:
...
+
+
+AZURE_MGMNT_PATH = "/providers/Microsoft.Management/managementGroups/"
+
+MANAGEMENT_GROUP_NAME_REGEX = "^[a-zA-Z0-9\-_\(\)\.]+$"
+
+
+class ManagementGroupCSPPayload(BaseCSPPayload):
+    """
+    :param: management_group_name: Just pass a UUID for this.
```
graham-dds: Does this expect a string that is a UUID, or python UUID object? Does it matter?

atst/jobs.py
```diff
@@ -63,6 +54,28 @@ def send_notification_mail(recipients, subject, body):
     app.mailer.send(recipients, subject, body)
 
 
+def do_create_application(csp: CloudProviderInterface, application_id=None):
+    application = Applications.get(application_id)
+
+    with claim_for_update(application) as application:
+
+        if application.cloud_id is not None:
```
graham-dds: Is this the same as `if application.cloud_id:` ?

tests/domain/cloud/test_payloads.py
```diff
@@ -0,0 +1,72 @@
...
+        payload = ManagementGroupCSPPayload(
+            creds={},
+            management_group_name="council of Naboo 1%^&",
+            display_name="Council of Naboo",
+            parent_id="Galactic_Senate",
+        )
+    # shortens management_group_name to fit
+    name = "council_of_naboo"
+    for _ in range(90):
+        name = f"{name}1"
```
graham-dds: Is this what [this](https://docs.python.org/3/library/stdtypes.html#str.ljust) for loop is doing?

tests/domain/test_applications.py
```diff
@@ -196,3 +197,20 @@ def test_update_does_not_duplicate_names_within_portfolio():
 
     with pytest.raises(AlreadyExistsError):
         Applications.update(dupe_application, {"name": name})
+
+
+def test_get_applications_pending_creation():
+    now = datetime.now()
+    later = now + timedelta(minutes=30)
```
graham-dds: Probably not for this PR, but I think it would be cool to be consitent with what library we use for datetime stuff. Sometimes we use the built in `datetime`, other times we use `pendulum`.

tests/factories.py
```diff
@@ -121,6 +122,7 @@ def _create(cls, model_class, *args, **kwargs):
         owner = kwargs.pop("owner", UserFactory.create())
         members = kwargs.pop("members", [])
         with_task_orders = kwargs.pop("task_orders", [])
+        state = kwargs.pop("state", None)
```
graham-dds: I think `.pop()` will return None by default, though I suppose this is more explicit.

atst/domain/csp/cloud/azure_cloud_provider.py
dandds: It is. It causes circular import errors when you try to import the AzureCloudInterface outside the application context, which I was trying to do in a sample script. Plus, we're moving to use those dataclass payloads everywhere for args, so this will go away anyway.

atst/domain/csp/cloud/azure_cloud_provider.py
dandds: This function is going to change a lot on a future commit.

atst/domain/csp/cloud/models.py
dandds: Doesn't matter. It will supply a default UUID if you don't pass one.

atst/domain/applications.py
dandds: For this and the previous, I think that's probably nicer in the long-run. Since all our existing queries use single-filters, though, I think I'm going to leave as-is until we can update all of them.

tests/domain/test_applications.py
leigh-mil: Since I know this affects a bunch of different files, I am going to put this as a bug!

tests/factories.py
leigh-mil: @graham-dds When used with a dictionary, `pop()` a `KeyError` is raised if the key is missing from the dict

graham-dds: Ahh, [you're right](https://docs.python.org/3.7/library/stdtypes.html#dict.pop) 👍

tests/domain/cloud/test_payloads.py
dandds: Yup, good catch! Fixed on a recent commit.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -198,6 +209,9 @@ def _create_management_group(
 
         # result is a synchronous wait, might need to do a poll instead to handle first mgmt group create
         # since we were told it could take 10+ minutes to complete, unless this handles that polling internally
+        # TODO: what to do is status is not 'Succeeded' on the
```
tomdds: Yeah, the SDK throws it's own exceptions for anything going wrong while resolving the result. If you pop into the source for create_or_update I think it points out the higher order exception type that it can throw in the docstring somewhere.

atst/domain/csp/cloud/models.py
```diff
@@ -232,3 +234,110 @@ class Config:
...
+            values.get("root_tenant_id"),
+            values.get("root_sp_client_id"),
+            values.get("root_sp_key"),
+        ]
+        if any(sp_creds) and not all(sp_creds):
+            raise ValueError(
+                "root_tenant_id, root_sp_client_id, and root_sp_key must all be set if any one is"
+            )
+
+        return values
```
tomdds: 🙌

atst/jobs.py
```diff
@@ -3,47 +3,38 @@
...
+        "application_id",
+        "environment_id",
+        "environment_role_id",
+    ]
 
+    def _derive_entity_info(self, kwargs):
+        matches = [e for e in self._ENTITIES if e in kwargs.keys()]
+        if matches:
+            match = matches[0]
+            return {"entity": match.replace("_id", ""), "entity_id": kwargs[match]}
```
tomdds: Don't have to change it, but a pattern I found recently for "find first" in python:
```python
# next() will the next value from a sequence, or None if sequence is empty
# Note that comprehension internally is a generator () not list []
# using a generator in conjunction with next will only search until the first match is found or sequence is exhausted
match = next((e for e in self._ENTITIES if e in kwargs.keys()))
if match is not None:
   return {"entity": match.replace("_id", ""), "entity_id": kwargs[match]}
else:
   return None
```

PR #1362, opened 2020-01-27
Azure Logging

robgil-dds:
This adds log analytics configurations for the following services
- keyvault
- acr
- postgres
- aks
- redis
- buckets
- cdn

Load balancing is blocked until we have internal load balancers and k8s applied to dev. This is just the logging configuration. Dashboards will be a separate PR once we have worked out the bugs on the azure side with lots not flowing in.


terraform/providers/dev/container_registry.tf
```diff
@@ -6,6 +6,7 @@ module "container_registry" {
   owner         = var.owner
   backup_region = var.backup_region
   policy        = "Deny"
-  subnet_ids    = []
+  subnet_ids    = [module.vpc.subnet_list["private"].id]
```
robgil-dds: This makes the container registry privately accessible in the "private" subnet.

dandds: Is this something we need to do for each service?

robgil-dds: Yes. This is the "service endpoint" configuration necessary to get each service in to the private network. This makes it accessible over the private IP space instead of going out to the public internet to access it.

PR #1363, opened 2020-01-27
new branch for terraform storage metric

hcdds:


terraform/modules/bucket/main.tf
```diff
@@ -10,7 +10,11 @@ resource "azurerm_storage_account" "bucket" {
   account_tier             = "Standard"
   account_replication_type = "LRS"
 }
-
+resource "azurerm_storage_container" "bucket" {
```
robgil-dds: I think this is old. Can you rebase?

terraform/modules/bucket/main.tf
```diff
@@ -25,8 +29,27 @@ resource "azurerm_storage_account_network_rules" "acls" {
...
-resource "azurerm_storage_container" "bucket" {
-  name                  = "content"
-  storage_account_name  = azurerm_storage_account.bucket.name
-  container_access_type = var.container_access_type
-}
+##alert rule tag on the the storage module 
+resource "azurerm_metric_alertrule" "storage" {
+  name                = "${var.service_name}-storage"
+  resource_group_name = "${azurerm_resource_group.bucket.name}"
+  location            = "${azurerm_resource_group.bucket.location}"
```
robgil-dds: This looks ok, but will throw errors in TF 0.12. Please adjust for the latest TF and latest azurerm provider.

terraform/modules/bucket/main.tf
```diff
@@ -25,8 +29,27 @@ resource "azurerm_storage_account_network_rules" "acls" {
...
+
+  description = "An alert rule to watch the metric Storage"
+
+  enabled = true
+
+  resource_id = "$(azurerm_storage_).id}"
```
robgil-dds: Is this valid? Perhaps it should be `azurerm_storage_bucket.bucket.id`

terraform/modules/bucket/variables.tf
```diff
@@ -46,3 +46,8 @@ variable "whitelist" {
   description = "A map of whitelisted IPs and CIDR ranges. For single IPs, Azure expects just the IP, NOT a /32."
   default     = {}
 }
+
+variable "custom_emails" {
+  type        = string
```
robgil-dds: Please make this a list so multiple email addresses can be supported. Map would be good also so that you have an associative array of email addresses.

terraform/modules/bucket/main.tf
```diff
@@ -25,8 +29,27 @@ resource "azurerm_storage_account_network_rules" "acls" {
...
+  operator    = "GreaterThan"
+  threshold   = 1649267441664
+  aggregation = "Maximum"
+  period      = "PT10M"
+
+  email_action {
+    send_to_service_owners = false
+
+    custom_emails = [
+      "${var.custom_emails}", # add email
```
robgil-dds: This should support multiple IPs. This should use a for loop (terraform 0.12 feature).

terraform/modules/bucket/main.tf
hcdds: updated

hcdds:
> Thanks @hcdds this is great! Could you remove the `.DS_Store` files? (I'd recommend adding a `*.DS_Store` entry to the `~/.gitignore` file on your machine.) I'll let Rob leave some TF-specific feedback when he has a moment.

done

terraform/modules/bucket/main.tf
hcdds: let's talk over in today's meeting how we want to store variables

terraform/modules/bucket/main.tf
hcdds: updated with new code

terraform/modules/bucket/main.tf
hcdds: update variable as mapped list

dandds:
Thanks, we'll re-open in the migrated repo when time allows!

PR #1364, opened 2020-01-28
Add fn to ensure a url matches an app url pattern

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/170830031

In some route functions, we redirect a user based on a parameter in the query string.  This commit adds a function that checks to see if a given url matches a url pattern of a view function in the app. This will help us ensure that the url passed as the redirect parameter isn't malicious.


atst/routes/__init__.py
```diff
@@ -61,17 +63,36 @@ def _make_authentication_context():
...
 
 def redirect_after_login_url():
-    if request.args.get("next"):
-        returl = request.args.get("next")
-        if request.args.get(app.form_cache.PARAM_NAME):
-            returl += "?" + url.urlencode(
-                {app.form_cache.PARAM_NAME: request.args.get(app.form_cache.PARAM_NAME)}
-            )
+    returl = request.args.get("next")
+    if match_url_pattern(returl):
```
dandds: I was a little worried about what would happen if `returl` were `None` in this implementation. It seems like `adapter.match(None, method="GET")` just returns `atst.root`, which is more generous than I'd expect.

PR #1372, opened 2020-01-29
Azure Admin Provisioning

tomdds:
This PR implements numerous new provisioning tasks, mostly based around setting up the administrative identities within the tenant and granting them appropriate permissions and ownership.

This PR also includes a small amount of changes to config values in support of the above as well as some refactoring of how secrets and auth are handled in the azure CSP provider.


atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -577,22 +778,42 @@ def _extract_subscription_id(self, subscription_url):
...
 
         # TODO: handle failure states here
         token_response = context.acquire_token_with_client_credentials(
-            resource, client_id, secret_key
+            self.sdk.cloud.endpoints.resource_manager, client_id, secret_key
+        )
+
+        return token_response.get("accessToken", None)
+
+    def _get_up_token_for_resource(self, username, password, tenant_id, resource):
```
dandds: What's "up" in this context?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -577,22 +778,42 @@ def _extract_subscription_id(self, subscription_url):
...
+
+        context = self.sdk.adal.AuthenticationContext(
+            f"{self.sdk.cloud.endpoints.active_directory}/{tenant_id}"
+        )
+
+        # TODO: handle failure states here
+        token_response = context.acquire_token_with_username_password(
+            resource, username, password, self.ps_client_id
```
dandds: Are we going to just fish out the client ID from a PS instance and pop it into our KV-managed config? Also, why does it have to be PS?

tomdds: My understanding is that this is the passthrough client ID that allows for acquiring this token with username and password for administrative operations.

atst/domain/csp/cloud/azure_cloud_provider.py
tomdds: Shortening for Username and Password, I can change it if it's not very clear.

atst/domain/csp/cloud/azure_cloud_provider.py
dandds: Sounds good. Do we know how we're sourcing that yet?

tomdds: In code, they're sourced via config:
```
self.ps_client_id = config["POWERSHELL_CLIENT_ID"]
self.owner_role_def_id = config["AZURE_OWNER_ROLE_DEF_ID"]
```
For the actual values, they're both publicly published:
Client ID: https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-rest-api-use?view=azs-1910#example
Owner Role Definition ID: https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles#owner

PR #1373, opened 2020-01-30
Limit file size

leigh-mil:
## TODO:
- figure out how to get the file size limit from the config to use in TaskOrderForm

## Description
Set an upper size limit for file uploads that is less than the max limit for GMail SMTP attachments.
 
## Pivotal
https://www.pivotaltracker.com/story/show/170884385


atst/forms/task_order.py
```diff
@@ -164,8 +164,14 @@ class TaskOrderForm(BaseForm):
...
-        label=translate("task_orders.form.supporting_docs_size_limit"),
-        description=translate("task_orders.form.supporting_docs_size_limit"),
+        label=translate(
+            "task_orders.form.supporting_docs_size_limit",
+            {"file_size_limit": app.config.get('FILE_SIZE_LIMIT')},
+        ),
+        description=translate(
+            "task_orders.form.supporting_docs_size_limit",
+            {"file_size_limit": app.config.get('FILE_SIZE_LIMIT')},
+        ),
```
dandds: Okay, I'm gonna tell you to do a bad thing: hard-code the description into the `templates/component/upload_input.html`. This is our only upload field, and it's not worth the time and effort to keep it generic right now. You can add something like this to the template instead of `field.description`:

```
      <p>
        {{ "task_orders.form.supporting_docs_size_limit" | translate({"file_size_limit": file_size_limit}) }}
      </p>
```

You'll have to do some math to make that a human-readable number and not bytes. Please add a comment to the template noting that this is technical debt and should be fixed.

PR #1374, opened 2020-01-30
Download and email tos

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/170606529

This PR adds functions to download PDFs from Azure and send them to a Microsoft email address TBD.


atst/domain/csp/files.py
```diff
@@ -53,7 +65,7 @@ def get_token(self):
         sas_token = bbs.generate_blob_shared_access_signature(
             self.container_name,
             object_name,
-            permission=self.BlobPermissions.CREATE,
+            permission=self.BlobSasPermissions(create=True),
```
graham-dds: We had to make this change because the `BlobPermissions` class didn't exist anymore. We'll explore this further when we dig into making uploading work again.

graham-dds:
This PR might be a little noisier than necessary because we did some class renaming. https://github.com/dod-ccpo/atst/pull/1374/commits/130b4c5b2975a107b7d573f4b068aa6cb08b259c and https://github.com/dod-ccpo/atst/pull/1374/commits/6900275b85f45f9c89a0771c31c4fa14a188cf93 represent the real "work" of the PR.

tests/domain/cloud/test_azure_file_service.py
```diff
@@ -0,0 +1,22 @@
...
+
+class MockBlockBlobService(object):
+    def __init__(self, exception=None, **kwargs):
+        self.exception = exception
+
+    def get_blob_to_bytes(self, blob_name="test.pdf", **kwargs):
+        if self.exception:
+            raise self.exception
+        else:
+            return Blob(name=blob_name, content=b"mock content")
```
dandds: I think we should pass in the mock content or have it as a default kwarg instead of hard-coding it.

atst/utils/mailer.py
```diff
@@ -83,3 +91,26 @@ def send(self, recipients, subject, body):
...
+            part = MIMEBase("application", "octet-stream")
+            part.set_payload(attachment.read())
+            encoders.encode_base64(part)
+            part.add_header(
+                "Content-Disposition",
+                f"attachment; filename= {downloaded_task_order['name']}",
+            )
+            message.attach(part)
+
+        self.connection.send(message)
```
dandds: This is where I was talking about making this behavior a top-level function in the module, like `atst.utils.mailer.send_task_order`. You can choose to make it a class method if that's easier. There are a couple things about this that I think are too tightly coupled to other parts of the application:

- I think `downloaded_task_order` should be two separate args to the method, instead of a dictionary. 
- I don't think that this module should import `app`. Ideally, things in `utils` don't know anything about the flask app. If you're going to keep this as a class method, there are two ways you could handle this. You could have the `__init__` method take a kwarg for that setting:

```
def __init__(self, connection, sender, task_order_recipient=None):
    [...]
    self.task_order_recipient = task_order_recipient
```

Or you could just make it another argument to `send_task_order`. 

If you're going to keep this as a method on the class, I think it would be better for it to have it be a generic method, like `send_with_attachment` that has a method signature that specifies the extra data it needs, something like:

`def send_with_attachment(self, recipients, subject, body, attachment_name, attachment_content, mime_type="octet-stream"):`

tests/domain/cloud/test_azure_file_service.py
dandds: Resolved IRL

config/base.ini
```diff
@@ -25,6 +25,7 @@ MAIL_PORT
 MAIL_SENDER
 MAIL_SERVER
 MAIL_TLS
+MICROSOFT_TASK_ORDER_EMAIL_ADDRESS = example@example.com
```
dandds: Can you add this to the config guide section of the README with a short explanation of what it's for?

PR #1375, opened 2020-01-30
Add functionality for creating and verifying subscriptions.

tomdds:
Currently the create call will be consumed by on-demand requests from the frontend, and the 2 stage create will be used by the enviroment management group provisioning to verify an initial subscription was created.


atst/routes/applications/settings.py
```diff
@@ -533,7 +535,20 @@ def create_subscription(environment_id):
...
+            "invoice_sections"
+        ][0]["invoice_section_name"]
+
+        payload = SubscriptionCreationCSPPayload(
+            tenant_id=csp_data.get("tenant_id"),
+            parent_group_id=parent_group_id,
+            billing_account_name=csp_data.get("billing_account_name"),
+            billing_profile_name=csp_data.get("billing_profile_name"),
+            invoice_section_name=invoice_section_name,
+        )
```
dandds: Can we break this out into a function?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -517,6 +473,59 @@ def create_billing_instruction(self, payload: BillingInstructionCSPPayload):
...
 
+    def create_subscription(self, payload: SubscriptionCreationCSPPayload):
+        sp_token = self._get_tenant_principal_token(payload.tenant_id)
+        if sp_token is None:
+            raise AuthenticationException(
+                "Could not resolve token for subscription creation"
+            )
+
+        request_body = {
+            "displayName": "Test Sub 1",
```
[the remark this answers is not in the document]
tomdds: Ha, yeah, my bad, left the test value in. I'll make it a property of the payload.

atst/routes/applications/settings.py
```diff
@@ -525,6 +527,25 @@ def resend_invite(application_id, application_role_id):
...
 
 
+def build_subscription_payload(environment) -> SubscriptionCreationCSPPayload:
+    csp_data = environment.application.portfolio.csp_data
+    parent_group_id = environment.cloud_id
+    invoice_section_name = csp_data["billing_profile_properties"]["invoice_sections"][
+        0
+    ]["invoice_section_name"]
+
+    display_name = f"{environment.application.name}-{environment.name}"
```
dandds: This needs to be unique, but I'm going to track that here:

https://www.pivotaltracker.com/story/show/171036058

PR #1376, opened 2020-01-30
Give the cluster perms to create load balancers.

dandds:
In order for the cluster app registration to create new load balancers,
it needs to have the Network Contributor role for the virtual network.
In the future, we should create a custom policy scoped to exactly the
permissions the cluster needs, per:

https://docs.microsoft.com/en-us/azure/aks/configure-azure-cni#prerequisites

Screen Shot 2020-01-30 at 4 49 19 PM

### DIFF
--- a/terraform/modules/k8s/main.tf
+++ b/terraform/modules/k8s/main.tf
@@ -81,3 +81,9 @@ resource "azurerm_monitor_diagnostic_setting" "k8s_diagnostic-1" {
     }
   }
 }
+
+resource "azurerm_role_assignment" "k8s_network_contrib" {
+  scope                = var.vnet_id
+  role_definition_name = "Network Contributor"
+  principal_id         = azurerm_kubernetes_cluster.k8s.identity[0].principal_id
+}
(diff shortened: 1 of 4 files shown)


terraform/modules/k8s/main.tf
```diff
@@ -81,3 +81,9 @@ resource "azurerm_monitor_diagnostic_setting" "k8s_diagnostic-1" {
     }
   }
 }
+
+resource "azurerm_role_assignment" "k8s_network_contrib" {
+  scope                = var.vnet_id
```
robgil-dds: We already have the vnet id configured in this module. 

https://github.com/dod-ccpo/atst/blob/ca1d26cc672434d7c6ddff3917737fd8e72548a5/terraform/modules/k8s/variables.tf#L32-L35

terraform/modules/k8s/variables.tf
```diff
@@ -66,4 +66,9 @@ variable "client_secret" {
 variable "workspace_id" {
   description = "Log Analytics workspace for this resource to log to"
   type        = string
-}
\ No newline at end of file
+}
+
+variable "vnet_id" {
```
robgil-dds: See above

terraform/modules/k8s/main.tf
dandds: It looks like this is the ID for the private subnet; I need the ID for the whole VNET for the role assignment.

robgil-dds: Yup this makes sense. I missed that. :+1:

PR #1382, opened 2020-02-03
Add celery job for sending Task Orders

leigh-mil:
## Description
Adds a query for finding all Task Order PDFs that need to be sent to Microsoft, creates a celery beat job for sending the PDFs.

## Pivotal
https://www.pivotaltracker.com/story/show/170984890


atst/jobs.py
```diff
@@ -233,3 +255,24 @@ def dispatch_create_atat_admin_user(self):
...
+        body = translate("email.task_order_sent.body", {"to_number": task_order.number})
+
+        file = app.csp.files.download_task_order(task_order.pdf.object_name)
+        file["maintype"] = "application"
+        file["subtype"] = "pdf"
+
+        send_with_attachment.delay(
+            recipients=recipients, subject=subject, body=body, attachments=[file]
+        )
+        TaskOrders.update_pdf_last_sent_at(task_order)
```
dandds: The problem with committing database changes here is that you don't know for a fact that sending the email (`send_with_attachment.delay`) worked. I'm wondering if it would be fine to do batch emails directly inside this `dispatch_send_task_order_files` without bothering to enqueue sub-tasks. You would just have to be careful to make sure all the emails are sent before updating the database. Pseudocode:

```
for task_order in task_orders:
    send_this_task_order(task_order)
    task_order.pdf_last_sent_at = datetime.now()
    db.session.add(task_order)

db.session.commit()
```
The important bit is that the commit doesn't happen until all the emails are sent. If the job raises an error and is retried, we'll resend some of the emails, but I don't think that's a big deal.

atst/jobs.py
```diff
@@ -233,3 +255,24 @@ def dispatch_create_atat_admin_user(self):
         pendulum.now()
     ):
         create_atat_admin_user.delay(environment_id=environment_id)
+
+
+@celery.task(bind=True)
+def dispatch_send_task_order_files(self):
+    task_orders = TaskOrders.get_for_send_task_order_files()
+    recipients = app.config.get("MICROSOFT_TASK_ORDER_EMAIL_ADDRESS")
```
graham-dds: From what I can tell, `recipients` gets set to `app.config.get("MICROSOFT_TASK_ORDER_EMAIL_ADDRESS")` here, makes it's way through a few functions, and eventually gets passed to `app.mailer._build_message()`. Looks like that method expects `recipients` [to be a list](https://github.com/dod-ccpo/atst/blob/b20df2811601144302a375ea1cad511b6b2644eb/atst/utils/mailer.py#L75). Maybe we should change that method to handle non-lists in some way 🤔

atst/jobs.py
tchandler: You could do something like this, where you just `continue` after exceptions and only update last_sent_at where send didn't throw an exception?
```python
for task_order in task_orders:
    try:
        send_this_task_order(task_order)
    except TaskSendError as exc:
         log.error(exc)
         continue
     task_order.pdf_last_sent_at = datetime.now()
     db.session.add(task_order)

db.session.commit()
```

tomdds: You could do something like this, where you just `continue` after exceptions and only update last_sent_at where send didn't throw an exception?
```python
for task_order in task_orders:
    try:
        send_this_task_order(task_order)
    except TaskSendError as exc:
         log.error(exc)
         continue
     task_order.pdf_last_sent_at = datetime.now()
     db.session.add(task_order)

db.session.commit()
```

atst/jobs.py
```diff
@@ -233,3 +236,30 @@ def dispatch_create_atat_admin_user(self):
...
+
+        try:
+            file = app.csp.files.download_task_order(task_order.pdf.object_name)
+            file["maintype"] = "application"
+            file["subtype"] = "pdf"
+            send_mail(
+                recipients=recipients, subject=subject, body=body, attachments=[file]
+            )
+        except (AzureError, SMTPException):
+            continue
```
dandds: Definitely log the error here:

```
except (MyException) as err:
    app.logger.exception(err)
    continue
```

atst/utils/mailer.py
```diff
@@ -72,7 +72,7 @@ def _build_message(self, recipients, subject, body):
         msg = EmailMessage()
         msg.set_content(body)
         msg["From"] = self.sender
-        msg["To"] = ", ".join(recipients)
+        msg["To"] = ", ".join(recipients) if type(recipients) == list else recipients
```
dandds: This seems less intuitive to me than just requiring the caller to wrap single-recipients in a list, but that's fine.

tests/test_jobs.py
```diff
@@ -320,3 +325,82 @@ def test_provision_portfolio_create_tenant(
...
+
+    # Check that send_with_attachment was called once for each task order
+    assert mock.call_count == 3
+    mock.reset_mock()
+
+    # Create new TO
+    task_order = TaskOrderFactory.create(create_clins=[{"number": "0001"}])
+    assert not task_order.pdf_last_sent_at
+
+    dispatch_send_task_order_files.run()
```
dandds: I'm not going to be nitpicky about this here, but generally if you're exercising the function under test more than once with different conditions, it should be multiple tests.

tests/test_jobs.py
```diff
@@ -320,3 +325,82 @@ def test_provision_portfolio_create_tenant(
...
+    monkeypatch.setattr(
+        "atst.domain.csp.files.MockFileService.download_task_order",
+        _download_task_order,
+    )
+
+    task_order = TaskOrderFactory.create(create_clins=[{"number": "0002"}])
+    dispatch_send_task_order_files.run()
+
+    # Check that pdf_last_sent_at has not been updated
+    assert not task_order.pdf_last_sent_at
```
dandds: A pattern I like that I think I'm going to encourage in the future (again, not now) is using test classes when you have multiple tests of a single method. That way they're namespaced together and you can add utility methods to the class, like:

```
class Test_dispatchSend_task_order_files():
    def util_method(self):
        [...]

    def test_download_failure(self):
       [...]
```

PR #1385, opened 2020-02-03
Alert for revoking an invite

leigh-mil:
## Description
Add success flash message for revoked portfolio invite. Updates the flash messages and translations so the same message can be used for revoking a portfolio or application invite

## Pivotal
https://www.pivotaltracker.com/story/show/170831810

## Screenshot
Screen Shot 2020-02-03 at 4 34 20 PM

### DIFF
--- a/translations.yaml
+++ b/translations.yaml
@@ -128,9 +128,6 @@ flash:
       message: There was an error processing the invitation for {user_name} from {application_name}
     resent:
       message: "{email} has been sent an invitation to access this Application"
-    revoked:
-      title: Application invitation revoked
-      message: You have successfully revoked the invite for {user_name} from {application_name}
   application_member:
     removed:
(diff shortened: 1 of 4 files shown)


translations.yaml
```diff
@@ -166,6 +163,9 @@ flash:
     errors:
       title: There were some errors
       message: Please see below.
+  invite_revoked:
+    title: Application invitation revoked
+    message: You have successfully revoked the invite for {user_name} from {application_name}
```
graham-dds: As a way to mitigate https://github.com/dod-ccpo/atst/pull/1385#pullrequestreview-353114200, maybe we could extend this translation to handle applications and portfolios

```yaml
   title: "{thing} invitation revoked"
   message: "You have successfully revoked the invite for {user_name} from {thing_name}"
```

PR #1386, opened 2020-02-04
Move Terraform config to its own repo.

dandds:
The TF config will live in its own private repo in an Azure DevOps
workspace for now. While it doesn't contain secrets, it does contain
plenty of information we wouldn't like to advertise, like IP addresses
and CIDR ranges. Once the TF config is further evolved, we can pull
the modules back into this repo. For now, it will all live in one place
for ease of use.


whit1206:
Should we work to truncate the history of this branch now then or do it after we merge TF scripts back into this branch?

dandds:
@whit1206 I think I want to wait until after go-live to truncate so that I have time to think through the consequences. I also want staging merged into master right before we do that so that I can just make a fresh staging branch.

dandds:
@whit1206 yep, we should move `docs` too.

PR #1388, opened 2020-02-04
Update SQL query to find pending portfolios.

dandds:
The query to find portfolios that are pending provisioning is updated to
check for:

- a period of performance that has started
- a portfolio state machine that has an UNSTARTED or one of the CREATED
  states

I left several TODOs to ensure that the orchestration functions
correctly for portfolio.

### DIFF
--- a/atst/models/portfolio_state_machine.py
+++ b/atst/models/portfolio_state_machine.py
@@ -175,11 +175,14 @@ class PortfolioStateMachine(
             app.logger.info(exc.json())
             print(exc.json())
             app.logger.info(payload_data)
+            # TODO: Ensure that failing the stage does not preclude a Celery retry
             self.fail_stage(stage)
+        # TODO: catch and handle general CSP exception here
         except (ConnectionException, UnknownServerException) as exc:
             app.logger.error(
                 f"CSP api call. Caught exception for {self.__repr__()}.", exc_info=1,
             )
+            # TODO: Ensure that failing the stage does not preclude a Celery retry
             self.fail_stage(stage)
 
         self.finish_stage(stage)
--- a/atst/domain/portfolios/portfolios.py
+++ b/atst/domain/portfolios/portfolios.py
@@ -15,6 +15,8 @@ from atst.models import (
     Permissions,
     PortfolioRole,
     PortfolioRoleStatus,
+    TaskOrder,
+    CLIN,
 )
 
 from .query import PortfoliosQuery, PortfolioStateMachinesQuery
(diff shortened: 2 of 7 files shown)


atst/models/portfolio_state_machine.py
```diff
@@ -175,11 +175,14 @@ def after_in_progress_callback(self, event):
             app.logger.info(exc.json())
             print(exc.json())
             app.logger.info(payload_data)
+            # TODO: Ensure that failing the stage does not preclude a Celery retry
             self.fail_stage(stage)
+        # TODO: catch and handle general CSP exception here
         except (ConnectionException, UnknownServerException) as exc:
```
dandds: Is this correct @tomdds ? I want add this to our kanban board if so.

atst/models/portfolio_state_machine.py
```diff
@@ -175,11 +175,14 @@ def after_in_progress_callback(self, event):
             app.logger.info(exc.json())
             print(exc.json())
             app.logger.info(payload_data)
+            # TODO: Ensure that failing the stage does not preclude a Celery retry
```
dandds: I want to think through this one @philip-dds and @tomdds . Ideally, retry should be handled by Celery, but my understanding is that `fail_stage` puts the machine in a failed state. When the Celery task is retried, the job will see that the the machine is failed, and none of the failure states have a transition.

 I'm wondering how we solve this. Simplest might be to have a transition that allows FAILED to go to IN_PROGRESS for every step. When Celery exhausts its retries, the machine will rest in the FAILED state.

atst/models/portfolio_state_machine.py
tomdds: Yes, I think there's already a post-it there for that.

atst/models/portfolio_state_machine.py
tomdds: Yeah, I didn't realize the `X_FAILED -> X_IN_PROGRESS` transitions weren't set up yet, that should be fairly easy to implement though. That was the intended idea from the beginning, just need to put transitions in to handle it. We'll probably also need to add a logic branch in `trigger_next_transition` to push the job back on track.

philip-dds: @dandds I've discussed how we would handle exceptions with @tomdds just now.
We talked about catching sdk exceptions such as [ErrorResponseException](https://docs.microsoft.com/en-us/python/api/azure-mgmt-managementgroups/azure.mgmt.managementgroups.models.errorresponseexception?view=azure-python) as well as `requests` network exceptions. Catching as much as we can in azure_cloud_provider.py, reraising then as our own domain.csp.cloud.exceptions ones. This way the state machine handler can catch our own exceptions and decide if the state should fail or get picked up by the celery beat again.

PR #1389, opened 2020-02-04
state machine triggers for resuming progress from a failed state

philip-dds:
@dandds @tomdds I've added triggers for each stage for transitioning from FAILED to *_IN_PROGRESS states.

there is a method `is_ready_resume_progress` where we can put the logic that determines if progress can be resumed. 

Using this setup we can call `fail_stage` from anywhere to stop progress.

Now the questions is how to store the state such as the particular exception that was caught when errors happen and then based on that figure out if processing can be resumed. Seems like something clever is required here. Let me know if you have any ideas provided you decide to accept this implementation.

This PR also contains a sample of how to handle exceptions in azure_cloud_provider `requests` library calls.
Seems like a lot of exception catching to replicate for each of the numerous `requests` based calls


atst/models/portfolio_state_machine.py
```diff
@@ -162,27 +179,34 @@ def after_in_progress_callback(self, event):
...
             app.logger.error(
                 f"CSP api call. Caught exception for {self.__repr__()}.", exc_info=1,
             )
             self.fail_stage(stage)
 
-        self.finish_stage(stage)
+        if response.get("status") == "error":
+            self.fail_stage(stage)
+
+        elif response.get("status") == "ok":
```
tomdds: This response object is always going to be a *CSPResult data class, so it won't have a status on it, nor will it have `.get`.  The intention was that if this line ran `response = getattr(self.csp, func_name)(payload_data)` then the response was valid, otherwise it'd throw an exception from the csp function, you'd add more `except` clauses above for the different kinds of failures coming from within the csp function.

atst/models/portfolio_state_machine.py
```diff
@@ -162,27 +179,34 @@ def after_in_progress_callback(self, event):
...
+                app.logger.error(
+                    f"Failed to cast response to valid result class {self.__repr__()}:",
+                    exc_info=1,
+                )
+                app.logger.info(exc.json())
+                print(exc.json())
+                app.logger.info(payload_data)
+                self.fail_stage(stage)
+
+            self.finish_stage(stage)
```
tomdds: This is going to get called even if self.fail_stage was previously called. Not sure what kind of weird effects it might have, but it might make sense to do either do an early return whenever a failure is triggered or wrap this in some kind of conditional.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -83,7 +89,8 @@ def __init__(self):
         self.graphrbac = graphrbac
         self.credentials = credentials
         self.identity = identity
-        self.exceptions = exceptions
+        self.azure_exceptions = exceptions
+        self.requests_exceptions = requests.exceptions
```
tomdds: I don't think we need to add `requests_exceptions` here, since we already have `requests` below and can just namespace off of it.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -83,7 +89,8 @@ def __init__(self):
         self.graphrbac = graphrbac
         self.credentials = credentials
         self.identity = identity
-        self.exceptions = exceptions
+        self.azure_exceptions = exceptions
```
tomdds: Make sure too add this to the provider in `mock_azure.py` as well.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -116,7 +123,7 @@ def set_secret(self, secret_key, secret_value):
         )
         try:
             return secret_client.set_secret(secret_key, secret_value)
-        except self.exceptions.HttpResponseError:
+        except self.azure_exceptions.HttpResponseError:
```
tomdds: Anything set on `self` in the Provider is accessed within the CSP interface on `self.sdk`, so this would be `self.sdk.azure_exceptions`

atst/models/portfolio_state_machine.py
philip-dds: hmm. what about the _ok and _error methods? they return the dict with the `status` and `body` keys.

atst/models/portfolio_state_machine.py
philip-dds: oh. I was under the assumption that `fail_stage` which is really a call to `trigger` machine method, will stop the processing. Need to test this more closely.

atst/models/portfolio_state_machine.py
tomdds: They created a blurry distinction between who owned transforming the response into a format our business logic cared about and would've put a lot more "magic" in the state machine that would've needed to be replicated in the other provisioning jobs. Going with either a response type that is successfully populated OR an exception felt like a better solution. Do you have thoughts either way?

philip-dds: in case we go with your proposal we would need to get rid of the _ok and _error methods and just return the *CSPResult data class. 
I guess if we needed more data returned from csp functions we could serialize to json like the _ok, _error methods do.
So since all we need is the data class then lets go with your solution.
Should I go and remove the _ok and _error calls and just return the *CSPResult data classes from all the csp functions?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -150,7 +153,7 @@ def get_secret(self, secret_key):
         )
         try:
             return secret_client.get_secret(secret_key).value
-        except self.sdk.exceptions.HttpResponseError:
+        except self.azure_exceptions.HttpResponseError:
```
tomdds: Same here, `self.sdk.azure_exceptions`

atst/models/portfolio_state_machine.py
```diff
@@ -122,7 +142,23 @@ def trigger_next_transition(self, **kwargs):
                     app.logger.info(
                         f"could not locate 'create trigger' for {self.__repr__()}"
                     )
-                    self.fail_stage(stage)
+                    self.trigger("fail")
+
+            elif self.current_state == FSMStates.FAILED:
+                # get the first trigger that starts with 'create_'
```
tomdds: Can you update or remove this comment?

tests/domain/test_portfolio_state_machine.py
```diff
@@ -37,21 +57,103 @@ def test_state_machine_trigger_next_transition(portfolio):
...
+def test_fail_stage(portfolio):
+    sm = PortfolioStateMachineFactory.create(portfolio=portfolio)
+    sm.state = FSMStates.TENANT_IN_PROGRESS
+    sm.fail_stage("tenant")
+    assert sm.state == FSMStates.TENANT_FAILED
+    # import ipdb;ipdb.set_trace()
+
+
+def test_stage_state_to_stage_name():
+    # sm = PortfolioStateMachineFactory.create(portfolio=portfolio)
```
tomdds: Comments here and above should be remove.

tests/mock_azure.py
```diff
@@ -66,6 +66,12 @@ def mock_policy():
     return Mock(spec=policy)
 
 
+def mock_azure_exceptions():
+    from azure.core import exceptions
+
+    return Mock(spec=exceptions)
```
tomdds: You can use the real exceptions here, no need to mock them, otherwise we'd probably run into the same issues as the requests problem

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -914,12 +1158,28 @@ def _create_active_directory_user(self, graph_token, payload: UserCSPPayload):
...
+                f"Could not create tenant. Connection Error", exc_info=1,
+            )
+            raise ConnectionException("connection error creating tenant")
+        except self.sdk.requests.exceptions.Timeout:
+            app.logger.error(
+                f"Could not create tenant. Request timed out.", exc_info=1,
+            )
+            raise ConnectionException("timout error creating tenant")
+        except self.sdk.requests.exceptions.HTTPError:
+            raise UnknownServerException("azure application error creating tenant")
```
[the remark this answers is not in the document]
philip-dds: @tomdds I've added status code to UnknownServerException and mocked up requests in such a way that raise_for_status can be tested.
test coverage for ~1800 lines in azure_cloud_provider.py is at 81%

PR #1390, opened 2020-02-04
Add cloud method to get reporting data

graham-dds:
First PR for https://www.pivotaltracker.com/story/show/170268176

Adds a method to `azure_cloud_provider` to query the Cost Management API for usage data per invoice.  For now, this query is relatively static. We're always calling the API at the billing invoice section scope, with the widest timeframe possible (one year), and with the same requested dataset. As the scope of the application's reporting needs changes, this function may change to be more general and/or revert back to the SDK.

### DIFF
--- a/atst/domain/csp/cloud/azure_cloud_provider.py
+++ b/atst/domain/csp/cloud/azure_cloud_provider.py
@@ -24,6 +24,7 @@ from .models import (
     BillingProfileTenantAccessCSPResult,
     BillingProfileVerificationCSPPayload,
     BillingProfileVerificationCSPResult,
+    CostManagementQueryCSPResult,
     KeyVaultCredentials,
     ManagementGroupCSPResponse,
     ProductPurchaseCSPPayload,
@@ -32,6 +33,7 @@ from .models import (
     ProductPurchaseVerificationCSPResult,
     PrincipalAdminRoleCSPPayload,
     PrincipalAdminRoleCSPResult,
+    ReportingCSPPayload,
     TaskOrderBillingCreationCSPPayload,
     TaskOrderBillingCreationCSPResult,
     TaskOrderBillingVerificationCSPPayload,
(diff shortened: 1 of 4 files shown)


atst/domain/csp/cloud/mock_cloud_provider.py
```diff
@@ -487,3 +490,25 @@ def get_credentials(self, scope="portfolio", tenant_id=None):
...
+
+        properties = CostManagementQueryProperties(
+            **dict(
+                columns=[
+                    {"name": "PreTaxCost", "type": "Number"},
+                    {"name": "UsageDate", "type": "Number"},
+                    {"name": "InvoiceId", "type": "String"},
+                    {"name": "Currency", "type": "String"},
+                ],
+                rows=[],
```
dandds: Do you want to have some fake data here, or is that a concern for another PR?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1070,3 +1072,38 @@ def _source_tenant_creds(self, tenant_id) -> KeyVaultCredentials:
         hashed = sha256_hex(tenant_id)
         raw_creds = self.get_secret(hashed)
         return KeyVaultCredentials(**json.loads(raw_creds))
+
+    def get_reporting_data(self, payload: ReportingCSPPayload):
+        """
+        Queries the Cost Management API for a invoice section's raw reporting data
+        """
```
dandds: Can you add a sentence about how we specify the scope via the `invoice_section_id`? I missed that at a glance.

atst/domain/csp/cloud/models.py
```diff
@@ -499,3 +499,31 @@ def supply_password_default(cls, password):
...
+class ReportingCSPPayload(BaseCSPPayload):
+    invoice_section_id: str
+    from_date: str
+    to_date: str
+
+    @root_validator(pre=True)
+    def extract_invoice_section(cls, values):
+        values["invoice_section_id"] = values["billing_profile_properties"][
+            "invoice_sections"
+        ][0]["invoice_section_id"]
```
dandds: If there's an index or key error for this lookup, what gets raised here? Is it caught and raised as a Pydantic validation error? I just want to make sure that it will be intelligible in the logs, so we can say "oh the portfolio was missing billing info we expected to have".

atst/domain/csp/cloud/mock_cloud_provider.py
graham-dds: That'll be in the next PR 👍

atst/domain/csp/cloud/models.py
graham-dds: Good call. Do you think something like
```python
def extract_invoice_section(cls, values):
    try:
        values["invoice_section_id"] = values["billing_profile_properties"][
            "invoice_sections"
        ][0]["invoice_section_id"]
        return values
    except (KeyError, IndexError):
        raise ValueError("Invoice section ID not present in payload")
```
would handle that a little better?

dandds: Yup, that's great. Maybe just include the invoice section ID in the error text and add a couple unit tests for the error path.

tests/domain/cloud/test_azure_csp.py
```diff
@@ -718,3 +722,77 @@ def test_create_subscription_verification(mock_azure: AzureCloudProvider):
...
+        "billing_profile_properties": {"invoice_sections": [{}],},
+    }
+
+    for malformed_payload in [key_error, index_error]:
+        with pytest.raises(pydantic.ValidationError):
+            assert mock_azure.get_reporting_data(
+                ReportingCSPPayload(
+                    from_date="foo", to_date="bar", **malformed_payload,
+                )
+            )
```
dandds: You could put a simpler unit test in `test_models` in this module that just exercises `ReportingCSPPayload` without the mock interface, but this is fine.

PR #1392, opened 2020-02-05
Bugfix - 404 errors for non-existent pages

leigh-mil:
## Description
Previously, if a user tried to access a page that didn't exist they would get a 500 error (ie `/foo`). This was because `get_resources_from_context()` expected a dictionary to be passed as `view_args`, so if `view_args` was `None` a KeyError would be raised, which would cause the 500 error page. 

## Pivotal
https://www.pivotaltracker.com/story/show/170897424


atst/utils/context_processors.py
```diff
@@ -19,7 +19,7 @@
 def get_resources_from_context(view_args):
     query = None
```
dandds: It might be simpler to just set a default at the start of the function:

```
if view_args is None:
    view_args = {}
```

PR #1395, opened 2020-02-06
Fix spacing at the bottom of forms

leigh-mil:
## Description
Because of the `action-group-footer` that was moved to the bottom of form pages, the text was being cut off on some pages. This PR adds  the `form` class which adds margin-bottom to offset the action-group-footer height. I applied this to all forms, but the pages where it was an issue were:
- Portfolio form
- Application builder step 1
- TO builder step 4 with multiple CLINs
- TO builder step 5

## Pivotal
https://www.pivotaltracker.com/story/show/171009912


graham-dds:
I'll approve this when it gets updated to staging, but in the meantime:
- Are there other, existing classes for these elements that we could use to apply this margin bottom? 
- If not, could we give this utility class a more descriptive name? If I were reading this for the first time and just saw a class called `form`, I'd have no idea what it did.

PR #1396, opened 2020-02-06
Reporting integration - Use cloud data in reporting view

graham-dds:
For https://www.pivotaltracker.com/story/show/170268176

This PR uses the new methods for querying Azure for reporting data and integrates it into the [MVP reporting view](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/403059546).

This gets us closer to what reporting will actually look like, as it makes a call to the mock cloud provider or Azure using information in the CSP data (theoretically) available as a property on a portfolio. Right now, that CSP data isn't available in development or staging environments, so I don't know to what extent we can visually QA the bar graph portion of the reporting page for this PR. 

image

Until we start populating portfolios with CSP data in different environments, this is always how the reporting page will look, i.e. invoiced and estimated funds will always be 0. Maybe we can actually insert mock CSP into profile objects to test this in a future PR.


atst/domain/csp/reports.py
```diff
@@ -11,128 +11,25 @@ def load_fixture_data():
...
+        estimated: Decimal
+    }
+    """
+
+    estimated = []
+    while rows:
+        if pendulum.parse(rows[-1][1]) >= pendulum.now(tz="utc").start_of("month"):
+            estimated.append(rows.pop())
+        else:
+            break
```
dandds: Wouldn't "estimated" always be the last row in `rows`? Or is that not the case?

atst/domain/reports.py
```diff
@@ -1,44 +1,32 @@
...
-            active_clins, key=lambda clin: clin.jedi_clin_type
-        ):
-            if not clin_spending.get(jedi_clin.name):
-                clin_spending[jedi_clin.name] = {}
-            clin_spending[jedi_clin.name]["obligated"] = sum(
-                clin.obligated_amount for clin in clins
-            )
+    def get_portfolio_spending(cls, portfolio):
+        from_date = pendulum.now().subtract(years=1).add(days=1).format("YYYY-MM-DD")
+        to_date = pendulum.now().format("YYYY-MM-DD")
```
dandds: Not going to ask you to change it here, but maybe leave a TODO: it would be nice to be able to pass in these values or make them configurable somehow, since eventually we'll have to represent more than a year. (But not for a year!)

tests/domain/cloud/reports/test_reports.py
```diff
@@ -1,58 +1,47 @@
...
-
-    totals = MockReportingProvider._get_application_monthly_totals(
-        portfolio, application
-    )
-    assert totals["name"] == "Test Application"
-    assert totals["this_month"] == 300
-    assert totals["last_month"] == 700
-    assert totals["total"] == 2500
-    assert [env["name"] for env in totals["environments"]] == ["A", "Z"]
+class TestPrepareAzureData:
```
dandds: yesssss

atst/domain/csp/reports.py
graham-dds: I'm not sure. That's what I thought too, but I wasn't sure if there would be any weird corner cases where, e.g. if there could be two or more rows for the current billing period (it looks like this can happen for past billing periods).

atst/domain/csp/reports.py
```diff
@@ -11,128 +11,25 @@ def load_fixture_data():
...
-                for environment in application["environments"]:
-                    for clin, spend in environment["spending"]["this_month"].items():
-                        CLIN_spend_dict[clin]["estimated"] += Decimal(spend)
-                    for clin, spend in environment["spending"]["total"].items():
-                        CLIN_spend_dict[clin]["invoiced"] += Decimal(spend)
-            return CLIN_spend_dict
-        return {}
+def prepare_azure_reporting_data(rows: list):
+    """
+    Returns a dict representing invoiced and expended funds for a portfolio given
```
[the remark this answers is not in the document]
graham-dds: Thank you!

PR #1398, opened 2020-02-06
Fix sidenav

leigh-mil:
## Description
In the `ExpandSidenav` mixin, `cookieName` needed to be outside of the mixin's props in order for it to work properly. 

## Pivotal
https://www.pivotaltracker.com/story/show/171115854

### DIFF
--- a/js/components/sidenav_toggler.js
+++ b/js/components/sidenav_toggler.js
@@ -1,5 +1,6 @@
 import ExpandSidenavMixin from '../mixins/expand_sidenav'
 import ToggleMixin from '../mixins/toggle'
+import { sidenavCookieName } from '../lib/constants'
 
 export default {
   name: 'sidenav-toggler',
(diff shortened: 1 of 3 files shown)


js/components/sidenav_toggler.js
```diff
@@ -1,6 +1,8 @@
 import ExpandSidenavMixin from '../mixins/expand_sidenav'
 import ToggleMixin from '../mixins/toggle'
 
+const cookieName = 'expandSidenav'
```
graham-dds: Looks like we have the same definition for this cookie's name in `expand_sidenav.js`. Since we're using this cookie in two places, I think moving this value to a variable is the right call, but we should probably define this as a constant and import it where needed. Maybe that could happen in `index.js`, or a new file like `constants.js`.  And because you'd be moving the cookie name outside of the context that it has now, you might want to name the cookie something a little more descriptive, like `sidenavCookieName` or something.

PR #1400, opened 2020-02-07
class method to export portfolio data as dictionary as an arg to stat…

philip-dds:
…e machine


atst/models/portfolio.py
```diff
@@ -153,6 +156,48 @@ def portfolio_id(self):
...
+            "password_recovery_email_address": ppoc.email,
+            "address": {  # TODO: TBD if we're sourcing this from data or config
+                "company_name": "",
+                "address_line_1": "",
+                "city": "",
+                "region": "",
+                "country": "",
+                "postal_code": "",
+            },
+            "billing_profile_display_name": "My Billing Profile",
```
graham-dds: I'm guessing this won't be "My Billing Profile" for each portfolio, but I don't know where it _will_ come from 🤔.

atst/models/portfolio.py
```diff
@@ -153,6 +156,48 @@ def portfolio_id(self):
     def application_id(self):
         return None
 
+    def to_dictionary(self):
+        ppoc = self.owner
+        user_id = f"{ppoc.first_name[0]}{ppoc.last_name}".lower()
+        domain_name = re.sub("[^0-9a-zA-Z]+", "", self.name).lower()
+        portfolio_data = {
+            "user_id": user_id,
+            "password": "jklfsdNCVD83nklds2#202",  # pragma: allowlist secret
```
graham-dds: I saw this in the tests [here](https://github.com/dod-ccpo/atst/blob/da283e4615358156febee9a9135893a7f0607156/tests/domain/test_portfolio_state_machine.py#L132). I'm guessing this won't be a static value? Will it be sourced from some kind of key store?

atst/models/portfolio.py
tomdds: I think we settled on just calling it `ATAT Billing Profile`. The name is purely cosmetic, but required.

atst/models/portfolio.py
tomdds: Good catch, we're actually generating this password within `create_tenant`, so this can be omitted now. It'll need to be passed into tests though.

atst/models/portfolio.py
```diff
@@ -153,6 +156,48 @@ def portfolio_id(self):
     def application_id(self):
         return None
 
+    def to_dictionary(self):
+        ppoc = self.owner
+        user_id = f"{ppoc.first_name[0]}{ppoc.last_name}".lower()
+        domain_name = re.sub("[^0-9a-zA-Z]+", "", self.name).lower()
```
tomdds: Can you add a random set of characters to the end of this, to help prevent potential name collisions? Something like:
```python
from string import  ascii_lowercase, digits
from random import choices
''.join(choices(ascii_lowercase + digits, k=4)) 
```

atst/models/portfolio.py
```diff
@@ -153,6 +156,48 @@ def portfolio_id(self):
...
+            initial_clin = initial_task_order.sorted_clins[0]
+            portfolio_data.update(
+                {
+                    "initial_clin_amount": initial_clin.obligated_amount,
+                    "initial_clin_start_date": initial_clin.start_date.strftime(
+                        "%Y/%m/%d"
+                    ),
+                    "initial_clin_end_date": initial_clin.end_date.strftime("%Y/%m/%d"),
+                    "initial_clin_type": initial_clin.number,
+                    "initial_task_order_id": initial_task_order.number,
```
tomdds: @dandds @graham-dds @leigh-mil Do any of you remember if this is the pre-parsed TO number or does it already have the non-alphanumerics stripped out?

dandds: It should have the non-alphanumerics stripped. Also, does it matter what CLIN we use here?

PR #1404, opened 2020-02-07
New portfolio styling

Hmbrink:
- Refined `action-group-footer`
- Added `max-form-width` to form
- Added bottom margin to form so `action-group-footer` doesn't overlap it


styles/core/_grid.scss
```diff
@@ -32,6 +32,11 @@
   &.row--max {
     max-width: $site-max-width;
   }
+
+  &--portfolio-create {
+    max-width: $max-form-width;
```
leigh-mil: Should this max-width be applied to all forms? 
If so, maybe it should be applied to the `.form` class https://github.com/dod-ccpo/atst/blob/35e33f94c64c6a8a333cbb153b917dd8cd86557a/styles/core/_util.scss#L102-L104, since that class is already applied to all forms and sets the bottom margin.

leigh-mil:
Closing this PR because of the weirdness that happened during the merge commits and because some of the changes here were already implemented in recent bug-fixes.

PR #1407, opened 2020-02-10
TO index page fixes

leigh-mil:
## Description
Updates TO index page so that unsigned TOs are included in the 'Draft' category. Also updates page so that categories that have TOs default to being open. 

## Pivotal
https://www.pivotaltracker.com/story/show/171164394
https://www.pivotaltracker.com/story/show/171164356


tests/domain/test_task_orders.py
```diff
@@ -149,11 +149,12 @@ def test_task_order_sort_by_status():
...
 
     sorted_by_status = TaskOrders.sort_by_status(initial_to_list)
-    assert len(sorted_by_status["Draft"]) == 3
+    assert len(sorted_by_status["Draft"]) == 4
     assert len(sorted_by_status["Active"]) == 1
     assert len(sorted_by_status["Upcoming"]) == 1
     assert len(sorted_by_status["Expired"]) == 2
-    assert len(sorted_by_status["Unsigned"]) == 1
+    with pytest.raises(KeyError):
+        sorted_by_status["Unsigned"]
```
dandds: It would be simpler to just do `assert "Unsigned" not in sorted_by_status`, but I'm not sure that you even need the assertion. As long as the sum of the other states equals the total number of TOs, we can assume there aren't any extra states.

PR #1409, opened 2020-02-10
DoD help text

leigh-mil:
## Description
Add DoD ID tooltip to app member and portfolio member forms.

## Pivotal
https://www.pivotaltracker.com/story/show/171163756

### DIFF
--- a/styles/elements/_tooltip.scss
+++ b/styles/elements/_tooltip.scss
@@ -95,4 +95,9 @@
   .icon {
     @include icon-size(16);
   }
+
+  &--tight {
+    margin: 0;
+    padding: 0;
+  }
 }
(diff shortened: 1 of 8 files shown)


styles/elements/_tooltip.scss
```diff
@@ -95,4 +95,9 @@
   .icon {
     @include icon-size(16);
   }
+
+  &--tight {
```
graham-dds: I like this rule. We have a similar definition [here](https://github.com/dod-ccpo/atst/blob/staging/styles/components/_portfolio_layout.scss#L516) when we use the `summary-item` component, but I think it makes more sense as a variant on the `tooltip` component like you have here. Would you mind replacing the `summary-item__header-icon` classes with this new class?

PR #1414, opened 2020-02-11
Fix portfolio header in FF

leigh-mil:
## Description
Fix styling of the portfolio header because it was not right justified in Firefox. 

## Pivotal
https://www.pivotaltracker.com/story/show/171008310

### DIFF
--- a/styles/components/_portfolio_layout.scss
+++ b/styles/components/_portfolio_layout.scss
@@ -20,7 +20,6 @@
 
   .col--grow {
     overflow: inherit;
-    display: table;
     min-height: 10rem;
   }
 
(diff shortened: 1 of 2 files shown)


styles/components/_portfolio_layout.scss
```diff
@@ -20,14 +20,15 @@
 
   .col--grow {
```
graham-dds: If what we're trying to do is center the portfolio's name in the header, try this:

```scss
  .col--grow {
    overflow: inherit;
    align-self: center; // centers flex item on cross-axis
  }

  &__name {
    @include h1;   
    ...
```

PR #1416, opened 2020-02-11
Bugfix: Accurately report Total Portfolio Value

graham-dds:
Addresses: https://www.pivotaltracker.com/n/projects/2160940/stories/171198912

The `Total Portfolio Value` figure in the portfolio summary was just simply returning the current obligated funds for the portfolio. It now displays obligated funds for current and future task orders. The tooltip text was also changed to be more clear.

image

image

This PR also added some missing tests and fixed a small bug in the `seed_sample.py` script.

### DIFF
--- a/tests/models/test_portfolio.py
+++ b/tests/models/test_portfolio.py
@@ -132,6 +132,31 @@ def test_active_task_orders(session):
     assert len(portfolio.active_task_orders) == 1
 
 
+class TestCurrentObligatedFunds:
+    """
+    Tests the current_obligated_funds property
+    """
+
+    def test_no_task_orders(self):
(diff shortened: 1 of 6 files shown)


tests/models/test_portfolio.py
```diff
@@ -7,6 +7,51 @@
...
+import pytest
+
+
+@pytest.fixture(scope="function")
+def upcoming_task_order():
+    return dict(
+        signed_at=pendulum.today().subtract(days=3),
+        create_clins=[
+            dict(
+                start_date=pendulum.today().add(days=2),
```
leigh-mil: Can you use the random date functions that were already imported from the factories here?

PR #1420, opened 2020-02-12
Update CLINs

leigh-mil:
## Description
Adds a query method to find all CLINs that are active and un-sent to the TaskOrder model. Also adds celery beat worker to send CLIN information to MS using `create_billing_instruction` and mark CLINs as sent. 

## Pivotal 
https://www.pivotaltracker.com/story/show/170930490

### DIFF
--- a/atst/jobs.py
+++ b/atst/jobs.py
@@ -10,6 +10,7 @@ from atst.domain.csp.cloud import CloudProviderInterface
 from atst.domain.csp.cloud.exceptions import GeneralCSPException
 from atst.domain.csp.cloud.models import (
     ApplicationCSPPayload,
+    BillingInstructionCSPPayload,
     EnvironmentCSPPayload,
     UserCSPPayload,
     UserRoleCSPPayload,
@@ -315,5 +316,33 @@ def send_task_order_files(self):
 
         task_order.pdf_last_sent_at = pendulum.now(tz="UTC")
         db.session.add(task_order)
+
+    db.session.commit()
+
+
+@celery.task(bind=True)
+def create_billing_instruction(self):
+    clins = TaskOrders.get_clins_for_create_billing_instructions()
+    for clin in clins:
+        portfolio = clin.task_order.portfolio
+        clin_data = clin.to_dictionary()
+        portfolio_data = portfolio.to_dictionary()
+
+        payload = BillingInstructionCSPPayload(
+            tenant_id=portfolio.csp_data.get("tenant_id"),
+            billing_account_name=portfolio.csp_data.get("billing_account_name"),
+            billing_profile_name=portfolio.csp_data.get("billing_profile_name"),
+            **clin_data,
+            **portfolio_data,
+        )
+
+        try:
+            app.csp.cloud.create_billing_instruction(payload)
+        except (AzureError) as err:
+            app.logger.exception(err)
+            continue
+
+        clin.last_sent_at = pendulum.now(tz="UTC")
+        db.session.add(clin)
 
     db.session.commit()
--- a/atst/jobs.py
+++ b/atst/jobs.py
@@ -325,15 +325,16 @@ def create_billing_instruction(self):
     clins = TaskOrders.get_clins_for_create_billing_instructions()
     for clin in clins:
         portfolio = clin.task_order.portfolio
-        clin_data = clin.to_dictionary()
-        portfolio_data = portfolio.to_dictionary()
 
         payload = BillingInstructionCSPPayload(
             tenant_id=portfolio.csp_data.get("tenant_id"),
             billing_account_name=portfolio.csp_data.get("billing_account_name"),
             billing_profile_name=portfolio.csp_data.get("billing_profile_name"),
-            **clin_data,
-            **portfolio_data,
+            initial_clin_amount=clin.obligated_amount,
+            initial_clin_start_date=str(clin.start_date),
+            initial_clin_end_date=str(clin.end_date),
+            initial_clin_type=clin.number,
+            initial_task_order_id=str(clin.task_order_id),
         )
 
         try:
--- a/atst/queue.py
+++ b/atst/queue.py
@@ -31,6 +31,10 @@ def update_celery(celery, app):
             "task": "atst.jobs.send_task_order_files",
             "schedule": 60,
         },
+        "beat-create_billing_instruction": {
+            "task": "atst.jobs.create_billing_instruction",
+            "schedule": 60,
+        },
     }
 
     class ContextTask(celery.Task):
--- a/tests/domain/test_task_orders.py
+++ b/tests/domain/test_task_orders.py
@@ -9,6 +9,27 @@ from atst.models.task_order import TaskOrder, SORT_ORDERING, Status
 from tests.factories import TaskOrderFactory, CLINFactory, PortfolioFactory
 
 
+@pytest.fixture
+def new_task_order():
+    return TaskOrderFactory.create(create_clins=[{}])
+
+
+@pytest.fixture
+def updated_task_order():
+    return TaskOrderFactory.create(
+        create_clins=[{"last_sent_at": pendulum.date(2020, 2, 1)}],
+        pdf_last_sent_at=pendulum.date(2020, 1, 1),
+    )
+
+
+@pytest.fixture
+def sent_task_order():
+    return TaskOrderFactory.create(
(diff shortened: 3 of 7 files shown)


atst/queue.py
```diff
@@ -27,6 +27,10 @@ def update_celery(celery, app):
             "task": "atst.jobs.dispatch_create_environment_role",
             "schedule": 60,
         },
+        "beat-dispatch_create_billing_instruction": {
+            "task": "atsdt.jobs.dispatch_create_billing_instruction",
+            "schedule": 60,
+        },
```
leigh-mil: Should `dispatch_send_task_order_files` also be here? It was not added in PR #1382

graham-dds: That seems right? @tomdds or @dandds what do you think?

tests/domain/test_task_orders.py
```diff
@@ -181,19 +181,39 @@ def test_allows_alphanumeric_number():
         assert TaskOrders.create(portfolio.id, number, [], None)
 
 
-def test_get_for_send_task_order_files():
-    new_to = TaskOrderFactory.create(create_clins=[{}])
-    updated_to = TaskOrderFactory.create(
+@pytest.fixture
```
graham-dds: I like the use of these fixtures, but it looks like the pattern in other test modules is to define fixtures at the top of the file.

atst/jobs.py
```diff
@@ -284,3 +285,31 @@ def dispatch_send_task_order_files(self):
...
+            **portfolio_data,
+        )
+
+        try:
+            app.csp.cloud.create_billing_instruction(payload)
+        except (AzureError) as err:
+            app.logger.exception(err)
+            continue
+
+        clin.last_sent_at = pendulum.now()
```
graham-dds: I think we want to set `tz="UTC"` in pendulum.now(), since it defaults to the local TZ by default.

atst/jobs.py
```diff
@@ -284,3 +285,31 @@ def dispatch_send_task_order_files(self):
         db.session.add(task_order)
 
     db.session.commit()
+
+
+@celery.task(bind=True)
+def dispatch_create_billing_instruction(self):
```
graham-dds: This function looks solid, but is there a reason we strayed away from the pattern of some of our other celery tasks? Admittedly, I'm no celery expert, but I see this pattern of each big task goal having three functions: `task()`, `do_create_task()`, and `dispatch_task()`. I haven't looked too deeply into _why_ that pattern is in place, just that it's there.

For example, with creating users:
```python
def do_create_user(csp: CloudProviderInterface, application_role_ids=None):
    ...

@celery.task(bind=True, base=RecordFailure)
def create_user(self, application_role_ids=None):
    do_work(
        do_create_user, self, app.csp.cloud, application_role_ids=application_role_ids
    )

@celery.task(bind=True)
def dispatch_create_user(self):
    for application_role_ids in ApplicationRoles.get_pending_creation():
        create_user.delay(application_role_ids=application_role_ids)
```

tests/test_jobs.py
```diff
@@ -412,3 +413,57 @@ def _download_task_order(MockFileService, object_name):
...
+        # The session needs to be nested to prevent detached SQLAlchemy instance
+        session.begin_nested()
+        dispatch_create_billing_instruction()
+        session.rollback()
+
+        # check that last_sent_at has been updated
+        assert unsent_clin.last_sent_at
+
+    def test_failure(self, monkeypatch, session):
+        def _create_billing_instruction(MockFileService, object_name):
```
graham-dds: What does the `MockFileService` do here? I didn't thing creating billing instructions had anything to do with files.

atst/queue.py
dandds: Yup, that needs to be scheduled too. Good catch.

atst/jobs.py
leigh-mil: AFAIK, we don't want to use the `do_work()` wrapper function here, because that will continue to retry the task. Here, if `create_billing_instruction()` fails for one clin, we want to continue `create_billing_instruction` and updating the rest of the clins. Since the `last_sent_at` property won't be updated on the clin, it will be pulled with the query next time this celery task is done, so we don't need to worry about retrying the task here. 
@dandds please correct me if this is wrong!

atst/queue.py
leigh-mil: I put that in a separate PR!

tests/test_jobs.py
leigh-mil: woops! that was a copy-paste mistake!

atst/jobs.py
leigh-mil: @graham-dds is this something that should be applied across the whole codebase after your PR to convert everything to use `pendulum`?

atst/jobs.py
dandds: This is a decision similar to the one you made with the emails; do we want every CLIN to be handled by a separate background task? The pattern is:

- a `dispatch` function finds a list of work that needs to be done, and enqueues a separate celery task for each list item
- a celery task that has the `do_work` wrapper in order to ensure retries happen and that failures are recorded
- a function that actually does the work for a single item

In this case I think it's fine to do them all in one function if that's what you want. The change I'd suggest is calling it something other than `dispatch`; we use that name to imply that it's enqueuing multiple tasks.

atst/jobs.py
graham-dds: Good question. I think there should certainly be an audit, but I was thinking that there could be some cases where we might _not_ want to use `"UTC"` 🤔

atst/jobs.py
leigh-mil: I'll fix the similar task for sending TOs in #1430

atst/jobs.py
```diff
@@ -305,3 +306,31 @@ def dispatch_send_task_order_files(self):
...
+        clin_data = clin.to_dictionary()
+        portfolio_data = portfolio.to_dictionary()
+
+        payload = BillingInstructionCSPPayload(
+            tenant_id=portfolio.csp_data.get("tenant_id"),
+            billing_account_name=portfolio.csp_data.get("billing_account_name"),
+            billing_profile_name=portfolio.csp_data.get("billing_profile_name"),
+            **clin_data,
+            **portfolio_data,
+        )
```
tomdds: There's a lot of somewhat brittle conditions that are making this work correctly. It might be worth being more explicit about the data you're passing in. My recommendation would be to actually drop the `**portfolio_data` entirely, since `BillingInstructionCSPPayload` isn't reading anything from it. If you want to be even more careful, you could map the `clin_data` fields to the appropriate `BillingInstructionCSPPayload` arguments directly. That, sadly, removes some of the niceness of just splatting in the data, but helps to decouple the CLIN model from the CSPPayload object.

Here's some details about pydantic issues:

Lets use a sample test model
```python
class TestModel(BaseModel):
    my_field: str
    class Config:
        fields = { "my_field": "myField" }
```

Condition 1: Pydantic field aliases take precedent over the original field name and `clin_data` happens to use the aliases.
```python
foo = { "my_field": "foo" }
bar = { "myField": "bar" }

# because bar uses the alias, both of these will have a my_field of "bar"
TestModel(**foo, **bar) == TestModel(**bar, **foo)
```

Condition 2: Python will throw an exception if you repeat keys across different splat arguments, clin_data and portfolio_data happen to use different keys for their clin info
```python
foo = { "my_field": "foo" }
bar = { "my_field": "bar" }

TestModel(**foo, **bar)
# TypeError: ModelMetaclass object got multiple values for keyword argument 'my_field'
```

tests/test_jobs.py
```diff
@@ -459,3 +461,57 @@ def _download_task_order(MockFileService, object_name):
...
+        )
+        unsent_clin = portfolio.task_orders[0].clins[0]
+
+        # The session needs to be nested to prevent detached SQLAlchemy instance
+        session.begin_nested()
+        create_billing_instruction()
+        session.rollback()
+
+        # check that last_sent_at has not been updated
+        assert not unsent_clin.last_sent_at
```
tomdds: When running this test as part of the total suite, this works fine, but when running the test on it's own, I get the following error:
```
pipenv run python -m pytest --no-cov -s tests/test_jobs.py -k "test_update_clin_last_sent_at"

sqlalchemy.orm.exc.ObjectDeletedError: Instance '<CLIN at 0x115665a20>' has been deleted, or its row is otherwise not present.
```

To fix it, I moved the assert to before `session.rollback()`

PR #1421, opened 2020-02-12
Remove references to expended funds from task orders

leigh-mil:
## Description
Removed all references to expended or invoiced funds at a Task Order level. 

## Pivotal
https://www.pivotaltracker.com/story/show/171210456


templates/task_orders/fragments/task_order_view.html
```diff
@@ -8,7 +8,7 @@
   {% set expended_funds = task_order.invoiced_funds %}
 
   <div>
-    <section class="row">
+    <section class="row task-order-summary">
```
graham-dds: Instead of this one-off style rule, we might be able to use the USWDS grid system here:
```html
<section class="usa-grid">
      <div class='usa-width-one-third summary-item'>...</div>
      <div class='usa-width-one-third summary-item'>...</div>
</section>
```

templates/task_orders/index.html
```diff
@@ -34,7 +34,7 @@
           <div class="accordion__content--list-item">
             <h4><a href="{{ url_for('task_orders.view_task_order', task_order_id=task_order.id) }}">{{ to_number }} {{ Icon("caret_right", classes="icon--tiny icon--primary" ) }}</a></h4>
             {% if status != 'Expired' -%}
-              <div class="row">
+              <div class="row to-index--list-item">
```
graham-dds: Might be able to use `usa-grid` here too:
```html
<div class="usa-grid">
  <div class="usa-width-one-third">...(PoP)</div> 
  <div class="usa-width-one-fourth">...(Total)</div>
  <div class="usa-width-one-fourth">...(Obligated)</div>
</div>
```
the nice thing about this too is that the three values are vertically aligned throughout the page, which is what it looks like the [designs](https://thinkcompany.invisionapp.com/share/CUTRLB32MB7#/screens/382858361) are doing.

templates/task_orders/fragments/task_order_view.html
leigh-mil: 💡 My thought process in making new classes was that this may be temporary styling because the `expended` data may be added back in eventually, but your suggestion is better!

PR #1423, opened 2020-02-12
Notify PPOC when portfolio is provisioned

graham-dds:
Addresses: https://www.pivotaltracker.com/story/show/171218188


atst/jobs.py
```diff
@@ -177,10 +178,30 @@ def do_work(fn, task, csp, **kwargs):
...
+    domain_name = portfolio_dict.get("domain_name")
+
+    send_mail(
+        recipients=[ppoc_email],
+        subject=translate("email.portfolio_ready.subject"),
+        body=translate(
+            "email.portfolio_ready.body",
+            {
+                "password_reset_address": app.config.get("AZURE_LOGIN_URL"),
+                "username": f"{user_id}@{domain_name}.onmicrosoft.com",
```
leigh-mil: Would it make sense to put `.onmicrosoft.com` in a config variable too? It's only used here and in the tests, so I'm not sure if it is worth it

leigh-mil: I'm revising my earlier statement -- I think it is worth it to put this into a config variable, we will be sending other emails that will using the same domain name

config/base.ini
```diff
@@ -5,6 +5,7 @@ AZURE_ACCOUNT_NAME
 AZURE_CLIENT_ID
 AZURE_GRAPH_RESOURCE="https://graph.microsoft.com/"
 AZURE_POLICY_LOCATION=policies
+AZURE_LOGIN_URL="https://portal.azure.com/"
```
leigh-mil: Since you are already going in to make some changes, would you mind putting this in alpha order?

atst/jobs.py
graham-dds: Good call, it was used in some other non-test parts of the app too. I put it as a constant, but not a config variable -- similar to how we set the contract start and end dates.

PR #1425, opened 2020-02-13
Email notification for new app role

leigh-mil:
## Description
Send an email to a user after their application role is created. 

## Pivotal
https://www.pivotaltracker.com/story/show/171336631

### DIFF
--- a/tests/test_jobs.py
+++ b/tests/test_jobs.py
@@ -135,62 +135,63 @@ def test_create_application_job_is_idempotent(csp):
     csp.create_application.assert_not_called()
 
 
-def test_create_user_job(session, csp, app):
-    portfolio = PortfolioFactory.create(
-        csp_data={
-            "tenant_id": str(uuid4()),
-            "domain_name": f"rebelalliance.{app.config.get('OFFICE_365_DOMAIN')}",
-        }
-    )
-    application = ApplicationFactory.create(portfolio=portfolio, cloud_id="321")
-    user = UserFactory.create(
-        first_name="Han", last_name="Solo", email="han@example.com"
-    )
-    app_role = ApplicationRoleFactory.create(
-        application=application,
-        user=user,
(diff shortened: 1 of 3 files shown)


translations.yaml
```diff
@@ -82,7 +82,10 @@ errors:
   not_found_sub: This page does not exist.
 email:
   application_invite: "{inviter_name} has invited you to a JEDI cloud application"
-  portfolio_invite: "{inviter_name} has invited you to a JEDI cloud portfolio"
+  app_role_created:
+    subject: Application Role Created
+    body: "Your application role has been created.\nVisit {url}, and use your username, {username}, to log in."
```
dandds: I don't think we want to say "Application Role" here. The app role is just the excuse ATAT uses to provision an identity in the portfolio AAD tenant, and I don't think we want to make those concepts appear linked to the user. I know we don't have formal content for this, I'm just flagging it as something that needs a little bit of design/content.

tests/test_jobs.py
```diff
@@ -135,28 +135,61 @@ def test_create_application_job_is_idempotent(csp):
...
+        )
 
-    do_create_user(csp, [app_role.id])
-    session.refresh(app_role)
+    def test_create_user_job(self, session, csp, app_role_1):
+        session.begin_nested()
+        do_create_user(csp, [app_role_1.id])
+        session.rollback()
+
+        assert app_role_1.cloud_id
```
dandds: This structure is much better, but one thing I like for tests like this is an assertion to ensure there is no `cloud_id` before the function under test is called. That way it's very clear to the reader what changed. (I only feel this way because our method for making app roles is a little indirect.)

translations.yaml
leigh-mil: @dandds Could we just have a generic email that says something along the lines of 'You have access to something new'? I'll check in with Ben and Think about it!

PR #1430, opened 2020-02-18
Schedule celery task for dispatch_send_task_order_files

leigh-mil:
## Description
In #1382, I forgot to add in the new celery task to the schedule! 
Also takes care of some tech debt from the original PR of moving the tests into their own class, and updates the name of the celery task to better represent what the task does.

## Pivotal
https://www.pivotaltracker.com/story/show/170984890

### DIFF
--- a/atst/queue.py
+++ b/atst/queue.py
@@ -27,6 +27,10 @@ def update_celery(celery, app):
             "task": "atst.jobs.dispatch_create_environment_role",
             "schedule": 60,
         },
+        "beat-dispatch_send_task_order_files": {
+            "task": "atst.jobs.dispatch_send_task_order_files",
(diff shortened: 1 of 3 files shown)


atst/queue.py
```diff
@@ -27,6 +27,10 @@ def update_celery(celery, app):
             "task": "atst.jobs.dispatch_create_environment_role",
             "schedule": 60,
         },
+        "beat-send_task_order_files": {
+            "task": "atst.jobs.send_task_order_files",
+            "schedule": 60,
+        },
```
dandds: I'm not telling you to change this now, but I just want to say out loud that we should think through the schedule for these tasks. Probably not all of them need to be every minute.

PR #1435, opened 2020-02-19
Email new env role

leigh-mil:
## Description
Email user when their env role has been created. Also refactored tests relating to the celery task to be inside a test class

## Pivotal
https://www.pivotaltracker.com/story/show/171336647

### DIFF
--- a/atst/jobs.py
+++ b/atst/jobs.py
@@ -6,7 +6,7 @@ from azure.core.exceptions import AzureError
 from atst.database import db
 from atst.domain.application_roles import ApplicationRoles
 from atst.domain.applications import Applications
-from atst.domain.csp.cloud import CloudProviderInterface
+from atst.domain.csp.cloud import CloudProviderInterface, generate_user_principal_name
 from atst.domain.csp.cloud.exceptions import GeneralCSPException
 from atst.domain.csp.cloud.models import (
     ApplicationCSPPayload,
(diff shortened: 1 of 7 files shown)


atst/jobs.py
```diff
@@ -164,7 +164,20 @@ def do_create_environment_role(csp: CloudProviderInterface, environment_role_id=
         env_role.cloud_id = result.id
         db.session.add(env_role)
         db.session.commit()
-        # TODO: should send notification email to the user, maybe with their portal login name
+        user = env_role.application_role.user
+        mail_name = user.full_name.replace(" ", ".").lower()
+        username = f"{mail_name}@{csp_details.get('tennant_id')}.{app.config.get('OFFICE_365_DOMAIN')}"
```
dandds: This implementation also exists in `atst.domain.csp.cloud.models.UserMixin`. I'm wondering if there should be a set of shared utility functions for things like this. If I were to do it I'd put it in `atst.domain.csp.cloud.utils` and then barrel the functions to they can be imported directly from `atst.domain.csp.cloud`.

dandds: Oh, and you also want "domain_name", not "tenant_id" from `csp_details`.

dandds: @tomdds I'm just calling this out, but we have a slight problem where we might need to occasionally reference data we sent to Azure that could change in ATAT. In this case, if we were to rebuild the user's Azure username and they've changed their first or last name fields in the meantime, our derived username would be wrong.

atst/jobs.py
```diff
@@ -164,7 +164,20 @@ def do_create_environment_role(csp: CloudProviderInterface, environment_role_id=
...
+        send_mail(
+            recipients=[user.email],
+            subject=translate("email.azure_account_update.subject"),
+            body=translate(
+                "email.azure_account_update.body",
+                {"url": app.config.get("AZURE_LOGIN_URL"), "username": username},
+            ),
+        )
+        app.logger.info(
+            f"Notification email sent for enivornment role creation. User id: {user.id}"
```
dandds: small typo: environment

PR #1436, opened 2020-02-20
azure disable user call

philip-dds:


atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -311,6 +311,41 @@ def _create_policy_definition(
             management_group_id=management_group_id,
         )
 
+    def disable_user(self, tenant_id, cloud_id):
```
dandds: I would make the name `cloud_id` specific to this context; maybe `role_assignment_id` or something like that?

PR #1439, opened 2020-02-21
Billing account name config and refactor portfolio to_dictionary method

graham-dds:
https://www.pivotaltracker.com/story/show/171364906

Add a `AZURE_BILLING_ACCOUNT_NAME` config variable to be included in initial CSP data. When adding this config variable, I also refactored how that data is generated in the `portfolio.to_dictionary()` method.


atst/models/portfolio.py
```diff
@@ -1,22 +1,22 @@
 import re
-from string import ascii_lowercase, digits
```
graham-dds: A lot of noise up here, but this is just the result of using VSCodes `Python Refactor: Sort Import` command.

tests/domain/test_portfolio_state_machine.py
```diff
@@ -319,7 +318,11 @@ def test_fsm_transition_start(
     config = {"billing_account_name": "billing_account_name"}
 
     assert state_machine.state == FSMStates.UNSTARTED
-    portfolio_data = get_portfolio_csp_data(portfolio)
+    portfolio_data = {
+        **portfolio.to_dictionary(),
+        "display_name": "mgmt group display name",
+        "management_group_name": "mgmt-group-uuid",
```
graham-dds: Not including these entries caused this test to fail. Should they be moved into the `to_dictionary` method?

leigh-mil: Is there anyplace we use `portfolio.to_dictionary()` where we don't need those two key/value pairs? 
If this is the only place we use `to_dictionary` I think we can move them into the method, otherwise I think makes more sense to leave it as is.

graham-dds: That seems right, and I think this is the only place where we include these k/v pairs like this. From what I understand though, this method is really only supposed to be used to "seed" some data to send to the CSP at the beginning of the provisioning process. After that, we should get data from `portfolio.csp_data`. So I'm not sure if the method should include these things, or we should explore how this test interacts with this method.

Issue #130, opened 2018-08-06
Should we rename this repo to just be "at-at"?


jordangov:
Seems like this is going to be the main codebase, and the project is "AT-AT" ... so?

ddsdevon:
Maybe... though we will probably spin at least a couple of microservices back out.
If we are voting on names, I wouldn't mind putting atat- as the first part of all the repo names, just so if the CCPO puts other stuff in the org it is easy to see whats what.

jordangov:
That's not a bad idea...

PR #1442, opened 2020-02-24
change name to ATAT

leigh-mil:
## Note:
Leaving this in a WIP state for now, since it will be easier to merge when it is the only PR up!

## Description
Change `atst` to `atat` across the site! 

image

## Issue
https://github.com/dod-ccpo/atst/issues/130


.circleci/config.yml
```diff
@@ -135,9 +135,9 @@ commands:
       - run:
           name: Update Kubernetes cluster
           command: |
-            kubectl set image deployment.apps/atst atst=<< parameters.tag >> --namespace=<< parameters.namespace >>
-            kubectl set image deployment.apps/atst-worker atst-worker=<< parameters.tag >> --namespace=<< parameters.namespace >>
-            kubectl set image deployment.apps/atst-beat atst-beat=<< parameters.tag >> --namespace=<< parameters.namespace >>
+            kubectl set image deployment.apps/atat atat=<< parameters.tag >> --namespace=<< parameters.namespace >>
+            kubectl set image deployment.apps/atat-worker atat-worker=<< parameters.tag >> --namespace=<< parameters.namespace >>
+            kubectl set image deployment.apps/atat-beat atat-beat=<< parameters.tag >> --namespace=<< parameters.namespace >>
```
dandds: Hmm I don't know if we can do a global find and replace for this. This, for example, is changing the names associated with resources in the k8s cluster, which would require deleting and redeploying the whole thing. That's good and should happen but should probably happen on its own PR.

leigh-mil: @dandds I didn't think about the config files when I did the global replace! 
Are there any other config files I should revert back to `atst`?

leigh-mil:
Going to close this PR out in favor of opening a new one that changes just the atst module and tests.

PR #1444, opened 2020-02-26
Validate tenant name

leigh-mil:
## Description
Add function for validating and generating a unique tenant name. 

## Pivotal
https://www.pivotaltracker.com/story/show/171469381

### DIFF
--- a/atst/domain/csp/cloud/azure_cloud_provider.py
+++ b/atst/domain/csp/cloud/azure_cloud_provider.py
@@ -1,5 +1,5 @@
 import json
-from secrets import token_urlsafe
+from secrets import token_hex, token_urlsafe
 from uuid import uuid4
 from flask import current_app as app
 
@@ -12,6 +12,7 @@ from .exceptions import (
     ConnectionException,
     UnknownServerException,
     SecretException,
+    DomainNameException,
 )
 from .models import (
     AdminRoleDefinitionCSPPayload,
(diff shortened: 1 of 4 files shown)


atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -311,12 +312,32 @@ def _create_policy_definition(
...
+        # If an existing tenant with name cannot be found, 'error' will be in the response
+        return "error" in response.text
+
+    def generate_valid_domain_name(self, name, suffix="", max_tries=6):
+        if max_tries > 0:
+            domain_name = name + suffix
+            if self.validate_domain_name(domain_name):
+                return domain_name
+            else:
+                suffix = token_hex(6)
```
tomdds: Just as a note, `token_hex(N)` generates a string of length N*2 because N is number of "bytes" and a hex byte is 2 characters. Might want to drop this to 2 or 3.

tomdds: https://docs.python.org/3/library/secrets.html#secrets.token_hex

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -311,12 +312,32 @@ def _create_policy_definition(
             management_group_id=management_group_id,
         )
 
+    def validate_domain_name(self, name):
+        response = self.sdk.requests.get(
+            f"https://login.microsoftonline.com/{name}.onmicrosoft.com/.well-known/openid-configuration",
```
tomdds: Similar to other URLs in this file, format in `https://login.microsoft.online.com` using self.sdk.cloud.endpoints.active_directory, see `_get_sp_token` for an example.

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -311,12 +312,32 @@ def _create_policy_definition(
...
+        # If an existing tenant with name cannot be found, 'error' will be in the response
+        return "error" in response.text
+
+    def generate_valid_domain_name(self, name, suffix="", max_tries=6):
+        if max_tries > 0:
+            domain_name = name + suffix
+            if self.validate_domain_name(domain_name):
+                return domain_name
+            else:
+                suffix = token_hex(3)
```
dandds: What's the advantage to `token_hex` over `token_urlsafe` here?

atst/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -311,12 +312,32 @@ def _create_policy_definition(
             management_group_id=management_group_id,
         )
 
+    def validate_domain_name(self, name):
+        response = self.sdk.requests.get(
+            f"{self.sdk.cloud.endpoints.active_directory}/{name}.onmicrosoft.com/.well-known/openid-configuration",
+            timeout=30,
+        )
+        # If an existing tenant with name cannot be found, 'error' will be in the response
+        return "error" in response.text
```
dandds: I wonder if checking `response.json()` would be better here.

atst/domain/csp/cloud/azure_cloud_provider.py
leigh-mil: `token_urlsafe` includes dashes and underscores, which are not allowed in the tenant name!

atst/domain/csp/cloud/azure_cloud_provider.py
leigh-mil: Both worked! 
Is there a reason using `.json()` would be preferable?

dandds: I think that checking if `"error"` is a key in the dictionary is a little safer than checking that the string "error" isn't in the response anywhere. With the latter, MS could add a property like `"message": "no errors!"` and we'd interpret that as an error.

PR #1448, opened 2020-02-28
Change atst to atat

leigh-mil:
## Description
Change name from atst to atat in the modules and tests. I tested out most of the scripts to make sure they still work, but let me know if I missed anything.

## Pivotal
https://www.pivotaltracker.com/story/show/171601739


Dockerfile
```diff
@@ -80,7 +80,7 @@ COPY --from=builder /install/.venv/ ./.venv/
 COPY --from=builder /install/alembic/ ./alembic/
 COPY --from=builder /install/alembic.ini .
 COPY --from=builder /install/app.py .
-COPY --from=builder /install/atst/ ./atst/
+COPY --from=builder /install/atat/ ./atst/
```
dandds: I'd prefer we not change the Dockerfile just yet if possible. This will necessitate changes all over the K8s config too, b/c we reference directory paths in the container for a bunch of commands. We should change those, but probably on an isolated PR.

PR #1452, opened 2020-03-05
step in state machine to reset tenant admin password to random value

philip-dds:


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1581,6 +1599,62 @@ def _update_active_directory_user_email(self, graph_token, user_id, payload):
                 f"azure application error updating active directory user email. {str(exc)}",
             )
 
+    def _update_active_directory_user_password_profile(self, graph_token, payload):
+        request_body = {
+            "passwordProfile": {
+                "forceChangePasswordNextSignIn": True,
+                "forceChangePasswordNextSignInWithMfa": False,
+                "password": payload.new_password
+                or "".join(random.choice(string.ascii_letters) for i in range(15)),
```
graham-dds: How `"secure"` does this implementation of a random password have to be? I don't know if `random` is random enough and if we should rely on [`secrets`](https://docs.python.org/3/library/secrets.html#module-secrets) instead.

philip-dds: @graham-dds from what I understand the purpose of this call is to set the password for the tenant admin user to a value that is not stored anywhere. 
as Dan put it: you're essentially making ATAT "forget" the human admin's login creds because it's done using them

dandds: @philip-dds You're right, but I think @graham-dds is suggesting you use `token_urlsafe(32)` to generate the password instead of `random`, is all. (That's been our pattern for other passwords.)

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -933,6 +936,20 @@ def create_product_purchase_verification(
...
+
+        graph_token = self._get_tenant_admin_token(
+            payload.tenant_id, self.graph_resource
+        )
+        if graph_token is None:
+            raise AuthenticationException(
+                "Could not resolve graph token for tenant admin"
+            )
+
+        self._update_active_directory_user_password_profile(graph_token, payload)
```
dandds: Does this need to return a `TenantAdminCredentialResetCSPResult` object?

alembic/versions/18b29ab247fb_tenant_admin_password_update.py
```diff
@@ -0,0 +1,297 @@
...
+            "TENANT_ADMIN_OWNERSHIP_FAILED",
+            "TENANT_ADMIN_CREDENTIAL_RESET_CREATED",
+            "TENANT_ADMIN_CREDENTIAL_RESET_IN_PROGRESS",
+            "TENANT_ADMIN_CREDENTIAL_RESET_FAILED",
+            "TENANT_PRINCIPAL_OWNERSHIP_CREATED",
+            "TENANT_PRINCIPAL_OWNERSHIP_IN_PROGRESS",
+            "TENANT_PRINCIPAL_OWNERSHIP_FAILED",
+            name="fsmstates",
+            native_enum=False,
+        ),
```
dandds: Are you sure this list is complete for both? I might be missing it, but I don't see the `BILLING_OWNER` states in the `existing_type` list.

atat/models/mixins/state_machines.py
```diff
@@ -29,6 +29,7 @@ class AzureStages(Enum):
     TENANT_ADMIN_OWNERSHIP = "tenant admin ownership"
     TENANT_PRINCIPAL_OWNERSHIP = "tenant principial ownership"
     BILLING_OWNER = "billing owner"
+    TENANT_ADMIN_CREDENTIAL_RESET = "tenant admin credential reset"
```
dandds: You have the tests looking for this step in a different place. Should it go before BILLING_OWNER?

philip-dds: you are absolutely right.I changed the order.

atat/domain/csp/cloud/azure_cloud_provider.py
philip-dds: the result data class has to exist but does not need to be returned. Since the csp call returns a 204 with no data I will not be returning anything here either.

dandds: I think we need to at least return an empty dataclass, because `after_in_progress_callback` expects to be able to convert the result to a dictionary to merge with the portfolio's `csp_data`: https://circleci.com/gh/dod-ccpo/atst/18510?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #1455, opened 2020-03-06
Ensure single user tenant

leigh-mil:
## Description
Adds in checks so that a user isn't created twice in the AAD at both the route function for creating a new user and the create user job. The check sees if a user already has a cloud_id on any existing application_roles, and if so assigns the cloud_id to the new application role when it is created or the create user job is run. 

## Pivotal
https://www.pivotaltracker.com/story/show/171466035


atat/domain/users.py
```diff
@@ -120,3 +120,18 @@ def update_last_session_id(cls, user, session_id):
...
+            return next(
+                (
+                    app_role.cloud_id
+                    for app_role in user.application_roles
+                    if app_role.cloud_id is not None
+                ),
+                None,
+            )
+        except NoResultFound:
+            return None
```
dandds: I think this needs to be scoped to a portfolio. Try adding a couple tests for this scenario:
1. I am a developer with application access in Portfolio A. All my Azure accounts are provisioned.
2. I am given new access to an application in Portfolio B.
3. I am also given a new application role in Portfolio A.

For 2, my existing application roles should not prevent ATAT from creating a new identity for Portfolio B, because I don't have an existing identity in that tenant. For 3, your current implementation should work; the new application role is still tied to my existing identity in the Portfolio A tenant.

atat/domain/users.py
```diff
@@ -120,3 +120,19 @@ def update_last_session_id(cls, user, session_id):
...
+            user = db.session.query(User).filter_by(dod_id=dod_id).one()
+            return next(
+                (
+                    app_role.cloud_id
+                    for app_role in user.application_roles
+                    if app_role.cloud_id is not None
+                    and app_role.portfolio_id == portfolio_id
+                ),
+                None,
+            )
```
dandds: This could be done with just SQL, which would be a little more efficient. Basically, join Portfolio, Application, and ApplicationRole and filter where the portfolio ID and user ID equal the supplied values. Then you can call `one_or_none` on the query instead of catching the error. I'd also consider putting this method on ApplicationRoles, since it returns an app role.

https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.one_or_none

PR #1460, opened 2020-03-11
Make config adjustments to fix TO uploads on the new sites.

dandds:
- Make AZURE_ACCOUNT_NAME configurable in the write_dotenv script that
  generates an environment variables file for the JS build process.
- Add an additional build arg to the docker build to specify the Azure
  storage account name for the client JS bundle. Pass the value in
  CircleCI.
- Specify the STATIC_URL as an FQDN for the staging and master sites so
  that it's valid in the Content-Security-Policy header.
- Update BLOB_STORAGE_URL setting for staging and master sites.

The client side bundle feels very tightly coupled to the backend
configuration in a way that makes it painful to keep everything updated.
We should look to change the JS uploader so that it can receive all the
config it needs from the backend REST endpoint.

### DIFF
--- a/script/write_dotenv
+++ b/script/write_dotenv
@@ -1,7 +1,11 @@
 #!/bin/sh
 
-if [ -z "${CSP+is_set}" ]; then
+if [ -z "${CSP:+is_set}" ]; then
   CSP=mock
+fi
+
+if [ -z "${AZURE_ACCOUNT_NAME:+is_set}" ]; then
+  AZURE_ACCOUNT_NAME=atat
 fi
 
 if [ $CSP = "aws" ]; then
@@ -9,7 +13,7 @@ if [ $CSP = "aws" ]; then
 elif [ $CSP = "azure" ]; then
   cat << EOF > .env
 CLOUD_PROVIDER=azure
-AZURE_ACCOUNT_NAME=atat
+AZURE_ACCOUNT_NAME=${AZURE_ACCOUNT_NAME}
 AZURE_CONTAINER_NAME=task-order-pdfs
 EOF
 else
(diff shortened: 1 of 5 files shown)


script/write_dotenv
```diff
@@ -4,12 +4,16 @@ if [ -z "${CSP+is_set}" ]; then
   CSP=mock
 fi
 
+if [ -z "${AZURE_ACCOUNT_NAME+is_set}" ]; then
```
jesse-dds: Are you sure you want `+` and not `:+`? We want to the value to be null if the parameter is set but null right? So situations like this?

```
PARAM=1
AZURE_ACCOUNT_NAME
PARAM=2
```

Based on this chart https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
```
                                     | parameterSet and Not Null | parameterSet But Null | parameterUnset
${parameter:+word} | substitute word | substitute null | substitute null
${parameter+word}  | substitute word | substitute word | substitute null
```

jesse-dds: Closing, don't know enough about bash to say if this is the behavior we want.

dandds: You're right, good catch! Fixed 8a47070c7917c5cacfb5688dded2c4c56a26ed7e

script/write_dotenv
```diff
@@ -1,15 +1,19 @@
 #!/bin/sh
 
-if [ -z "${CSP+is_set}" ]; then
+if [ -z "${CSPi:+is_set}" ]; then
```
jesse-dds: Is this a typo?

dandds: amended!

PR #1466, opened 2020-03-19
Implement Hybrid Mock Service provider

jesse-dds:
## What

Create a hybrid mock service provider to facilitate end-to-end testing.

## Why

To enable end-to-end testing where we call as many live API endpoints as possible.

https://www.pivotaltracker.com/story/show/171586540


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -308,11 +308,11 @@ def _create_policy_definition(self, session, root_management_group_id, policy):
     def _create_policy_set(
         self,
         session,
-        root_management_group_id,
+        root_management_group_name,
         policy_set_definition_name,
         definition_references,
     ):
-        create_policy_set_uri = f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementgroups/{root_management_group_id}/providers/Microsoft.Authorization/policySetDefinitions/{policy_set_definition_name}?api-version=2019-09-01"
+        create_policy_set_uri = f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementGroups/{root_management_group_name}/providers/Microsoft.Authorization/policySetDefinitions/{policy_set_definition_name}?api-version=2019-09-01"
```
dandds: Curious about these case changes: did you find that this mattered, or was incorrect?

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1083,7 +1087,7 @@ def create_tenant_admin_ownership(self, payload: TenantAdminOwnershipCSPPayload)
 
         assignment_guid = str(uuid4())
 
-        url = f"{self.sdk.cloud.endpoints.resource_manager}/providers/Microsoft.Management/managementGroups/{payload.tenant_id}/providers/Microsoft.Authorization/roleAssignments/{assignment_guid}?api-version=2015-07-01"
+        url = f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementGroups/{payload.tenant_id}/providers/Microsoft.Authorization/roleAssignments/{assignment_guid}?api-version=2015-07-01"
```
dandds: I've seen a couple changes like this. Should we just use a standard lib URL function so that we don't have to care whether there's a trailing slash? Seems like it would be less buggy. If so, we should document it as a small piece of tech debt to address.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1184,7 +1188,9 @@ def create_tenant_principal_app(self, payload: TenantPrincipalAppCSPPayload):
                 "Could not resolve graph token for tenant admin"
             )
 
-        request_body = {"displayName": "ATAT Remote Admin"}
+        request_body = {
+            "displayName": f"ATAT Remote Admin :: Portfolio {payload.display_name}"
+        }
```
dandds: Small thing to note, but if we're allowing user input here we should be sure that we never exceed the character limits for that field in Azure.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1726,7 +1732,7 @@ def _update_active_directory_user_password_profile(self, graph_token, payload):
             "Authorization": f"Bearer {graph_token}",
         }
 
-        url = f"{self.graph_resource}v1.0/users/{payload.user_id}"
+        url = f"{self.graph_resource}/v1.0/users/{payload.user_object_id}"
```
dandds: This is a good change. Is it distinct from the type of user ID used above in `_update_active_directory_user_email`, or is that also an object ID?

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+        created through API calls to Azure, however for the hybrid mock we
+        will generate and store a tenant with a simple generated UUID string
+        instead.
+        """
+
+        # This is the username and password for the primary point of contact
+        # for a portfolio. These are normally generated for the tenant, but
+        # but here we just inject existing and valid credentials.
+        tenant_admin_username = os.environ.get("TENANT_ADMIN_USERNAME")
+        tenant_admin_password = os.environ.get("TENANT_ADMIN_PASSWORD")
```
dandds: I want to think about this some more; ideally we would have a single source of truth for config and wouldn't have to have classes rooting around in the runtime environment. This is a special case though because it's information that is only used by the hybrid.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.tenant_id = str(uuid4())
+
+        # This is our mocked result of what would have been the API call to
+        # create a tenant.
+        result_dict = {
+            "tenant_id": self.tenant_id,
+            "user_id": "HybridCSPIntegrationTestUser",
+            "user_object_id": os.environ.get("USER_OBJECT_ID"),
```
dandds: Should this be sourced as a variable first in the same place as `tenant_admin_username` and `tenant_admin_password`?

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+            KeyVault at the end of this stage.
+            """
+            creds.tenant_id = self.azure.tenant_id
+            original_update_tenant_creds(tenant_id, creds)
+
+        with monkeypatched(self.azure, "_get_tenant_admin_token", lambda *_a: token):
+            with monkeypatched(
+                self.azure, "update_tenant_creds", mocked_update_tenant_creds
+            ):
+                return self.azure.create_tenant_principal_credential(payload)
```
dandds: Interesting strategy, I like it. I think this is fine for right now. I think this a little painful because the Azure interface has too many responsibilities right now. I wonder if, in the future, the real Azure interface should be purer and not be responsible for managing credential updates or even know about the Key Vault. That could be the caller's responsibility for negotiate, and then every method on the Azure interface is passed the credentials it requires.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def create_admin_role_definition(self, payload):
+        return self.azure.create_admin_role_definition(payload)
```
dandds: Are you finding that things like this are idempotent, or are you getting multiple role definitions? That might be a helpful thing to note in doc strings or comments for each one.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,170 @@
...
+        token = self.azure._get_elevated_management_token(payload.tenant_id)
+        payload.tenant_id = self.azure.tenant_id
+        with monkeypatched(
+            self.azure, "_get_elevated_management_token", lambda _: token
+        ):
+            try:
+                return self.azure.create_tenant_admin_ownership(payload)
+            except UnknownServerException:
+                return TenantAdminOwnershipCSPResult(
+                    id="/providers/Microsoft.Management/managementGroups/b5ab0e1e-09f8-4258-afb7-fb17654bc5b3/providers/Microsoft.Authorization/roleAssignments/4a6c2abb-6a58-407e-8bc0-084a9731aed5"
```
dandds: This probably shouldn't be hard-coded.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -0,0 +1,67 @@
...
+            + list(config.items())
+        )
+
+        state_machine.trigger_next_transition(csp_data=collected_data)
+        assert (
+            "created" in state_machine.state.value
+            or state_machine.state == FSMStates.COMPLETED
+        )
+
+        csp_data = state_machine.portfolio.csp_data
```
dandds: Another worthwhile test to have (that could also be run with real credentials locally) would be a Celery task test that, given a portfolio instance set up like this, calls `dispatch_provision_portfolio` until the portfolio is done provisioning. That's a very close model to how this will work when deployed.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -0,0 +1,67 @@
...
+@pytest.fixture(scope="function")
+def state_machine(app, portfolio):
+    return PortfolioStateMachineFactory.create(
+        portfolio=portfolio, cloud=HybridCSP(app, test_mode=True).cloud
+    )
+
+
+def test_hybrid_cloud(pytestconfig, state_machine: PortfolioStateMachine):
+    if pytestconfig.getoption("--hybrid-test") != "True":
+        pytest.skip("Skip test_hybrid_cloud")
```
dandds: Is it possible to use pytest markers or [`skipif`](https://docs.pytest.org/en/latest/skipping.html#id1) to skip this by default instead?

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.fixture(scope="function")
+def state_machine(app, portfolio):
+    return PortfolioStateMachineFactory.create(
+        portfolio=portfolio, cloud=HybridCSP(app, test_mode=True).cloud
+    )
+
+
+def test_hybrid_cloud(pytestconfig, state_machine: PortfolioStateMachine):
```
dandds: Would be great to have info in the README about how to run this (i.e., here is a list of the config values you need to set, here is how to run the test, here is what to check for in the portfolio's csp_data and the Azure portal afterward).

atat/domain/csp/cloud/azure_cloud_provider.py
graham-dds: Zoom discussion: Move this to hybrid implementation?

atat/domain/csp/cloud/hybrid_cloud_provider.py
graham-dds: Zoom discussion: Move these config values to the base.ini file and explain them

atat/domain/csp/cloud/hybrid_cloud_provider.py
graham-dds: Zoom convo: Pull this into config

atat/domain/csp/cloud/azure_cloud_provider.py
graham-dds: It is distinct. This `user_object_id` is sourced from the csp data, but the `user_id` arg in `_update_active_directory_user_email` is a fully pathed "Azure ID" as a result of calling `_create_active_directory_user`, and then accessing the `.id` property.

atat/domain/csp/cloud/azure_cloud_provider.py
graham-dds: Discussed in Zoom call

atat/domain/csp/cloud/azure_cloud_provider.py
graham-dds: https://www.pivotaltracker.com/story/show/171825640

tests/domain/cloud/test_hybrid_csp.py
graham-dds: We're now have a marker `@pytest.mark.hybrid` that is skipped by default, unless you add a --hybrid flag when running pytest (i.e. `pytest --hybrid`). This is also now documented in the README.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,322 @@
...
+
+        return self.azure.create_principal_admin_role(payload)
+
+    def create_initial_mgmt_group(
+        self, payload: InitialMgmtGroupCSPPayload
+    ) -> InitialMgmtGroupCSPResult:
+        """Create a management group.
+        
+        A management group is an group of subscriptions to which "governance 
+        conditions" conditions can be applied.
```
dandds: I think it's a little broader than a group of subscriptions; it can also contain nested management groups. (You can also remove one of the instances of "conditions"). The important thing here is that it creates the root management group for the tenant, which is the parent to all future management groups (applications, environments).

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,322 @@
...
+        payload.display_name = f"Hybrid :: {payload.display_name}"
+        return self.azure.create_initial_mgmt_group(payload)
+
+    def create_initial_mgmt_group_verification(
+        self, payload: InitialMgmtGroupVerificationCSPPayload
+    ) -> InitialMgmtGroupVerificationCSPResult:
+        """Verify the creation of a management group.
+        
+        A management group is an group of subscriptions to which "governance 
+        conditions" conditions can be applied.
```
dandds: See above!

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,322 @@
...
+
+        return self.azure.create_initial_mgmt_group_verification(payload)
+
+    def create_tenant_admin_ownership(
+        self, payload: TenantAdminOwnershipCSPPayload
+    ) -> TenantAdminOwnershipCSPResult:
+        """For this step, we needed to be able to retrieve the elevated management 
+        token from KeyVault with the original tenant id, but make the role assignment 
+        request with the root credentials.
+        """
```
dandds: Do you want to describe what this step is supposed to do? It gives the tenant admin (human) the Owner role on the root management group.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def create_tenant_principal_ownership(
+        self, payload: TenantPrincipalOwnershipCSPPayload
+    ) -> TenantPrincipalOwnershipCSPResult:
+        """For this step, we needed to be able to retrieve the elevated 
+        management token from KeyVault with the original tenant id, but make 
+        the tenant principal ownership request with the root credentials.
+        
+        The security principal object defines the access policy and permissions 
+        for the user/application in the Azure AD tenant.
```
dandds: Similar to the previous method, this step is giving the SP the owner role over the root management group.

PR #1467, opened 2020-03-24
Use poetry

graham-dds:
https://www.pivotaltracker.com/n/projects/2160940

Since [Pipenv](https://github.com/pypa/pipenv/issues/4058) is no longer being maintained, this PR replaces it with [poetry](https://python-poetry.org/).


Dockerfile
```diff
@@ -34,12 +33,14 @@ RUN apk update && \
         util-linux \
         yarn
 
+RUN python -m venv /.venv
```
dandds: I don't think you actually need this addition. If you look at line 81, we're coping `/install/.venv/` to the location in the final build stage, which means that `poetry` went ahead and generated `/install/.venv` by itself. Try running the build again without this line.

script/include/run_setup
```diff
@@ -35,9 +35,9 @@ fi
 rm -rf ./node_modules
 
 if [ "${CREATE_VENV}" = "true" ]; then
-  # Ensure pipenv is installed
-  if ! pipenv --version >/dev/null 2>&1 ; then
-    echo "ERROR: pipenv is malfunctioning or not present"
+  # Ensure poetry is installed
+  if ! poetry --version >/dev/null 2>&1 ; then
+    echo "ERROR: poetry is malfunctioning or not present"
```
dandds: The script tries to check the Pipfile on line 44 below.

Dockerfile
graham-dds: I think you're right! Removed

script/include/run_setup
graham-dds: Good call --changed here: https://github.com/dod-ccpo/atst/pull/1467/files#diff-0092d782907521aa0f88bd9fc620d585R44

PR #1468, opened 2020-03-25
Allow the hybrid cloud service provider to create applications

jesse-dds:
It's a pretty small change, it just passes the `create_application` call through the azure instance. 

### Other Changes
- Added a test for `create_application` that runs with the `--hybrid` flag
- Use `["key"]` access because `.get` is only appropriate when the `None` case is explicitly handled/expected
- Add a `logger.warning` when attempting to create an application which already exists


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -33,10 +40,13 @@ def portfolio():
 
 
 @pytest.fixture(scope="function")
-def state_machine(app, portfolio):
-    return PortfolioStateMachineFactory.create(
-        portfolio=portfolio, cloud=HybridCSP(app, test_mode=True).cloud
-    )
+def csp(app):
+    return HybridCSP(app, test_mode=True).cloud
+
```
graham-dds: 👍

README.md
```diff
@@ -225,6 +225,7 @@ To generate coverage reports for the Javascript tests:
 - `AZURE_CALC_SECRET`: The secret key used to generate a token for the Azure pricing calculator
 - `AZURE_CALC_URL`: The redirect URL for the Azure pricing calculator
 - `AZURE_LOGIN_URL`: The URL used to login for an Azure instance.
+- `AZURE_ROOT_MGMT_GROUP_ID`: The fully pathed root management group ID for the Hybrid Cloud.
```
dandds: This is tricky. I can see why it's useful to have this available for one-off testing of application creation, but if a portfolio is provisioned and managed by the Hybrid interface end-to-end, it should have `root_management_group_id` available as a result of the `create_initial_mgmt_group` step. In the case of your test, you're not provisioning a whole portfolio and so you need to supply this value. You also don't want running the entire portfolio provisioning process to be part of your test setup. My hesitation is that config settings are generally reserved for things we have to set in order for the app to run in a given environment, and this one is not. Can we talk about how to handle this?

dandds: We talked it through and we're going to let this stand for now and just make clear that this is only for running the tests locally.

dandds:
@jesse-dds I just noticed: are you PRing against a branch other than staging on purposes?

PR #1470, opened 2020-03-27
Bugfix -- Portfolio provisioning query

graham-dds:
This PR:
 - Fixes a bug in the domain query to check for portfolios that need provisioning. The query would miss portfolios that didn't have a state machine attached yet.
- Moves state machine stage setting to the `__init__` method of the PortfolioStateMachine(). Previously, stages for a state machine were set when calling attach_machine(), but since that function is decorated with [@reconstructor](https://docs.sqlalchemy.org/en/13/orm/constructors.html#sqlalchemy.orm.reconstructor), it shouldn't take arguments.

### DIFF
--- a/atat/domain/portfolios/portfolios.py
+++ b/atat/domain/portfolios/portfolios.py
@@ -156,7 +156,7 @@ class Portfolios(object):
 
         results = (
             db.session.query(Portfolio.id)
-            .join(PortfolioStateMachine)
+            .outerjoin(PortfolioStateMachine)
             .join(TaskOrder)
             .join(CLIN)
             .filter(Portfolio.deleted == False)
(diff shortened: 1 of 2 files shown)


atat/domain/portfolios/portfolios.py
```diff
@@ -156,14 +156,15 @@ def get_portfolios_pending_provisioning(cls, now) -> List[UUID]:
...
-            .join(PortfolioStateMachine)
+            .outerjoin(PortfolioStateMachine)
             .join(TaskOrder)
             .join(CLIN)
             .filter(Portfolio.deleted == False)
             .filter(CLIN.start_date <= now)
             .filter(CLIN.end_date > now)
             .filter(
                 or_(
+                    Portfolio.state_machine == None,
```
graham-dds: LGTM yells at this, but I think this is necessary for [SQLAlchemy purposes](https://docs.sqlalchemy.org/en/13/orm/tutorial.html#common-relationship-operators)

PR #1471, opened 2020-03-30
Hybrid cleanup

dandds:
This makes some small fixes to the state machine implementation and other parts of the orchestration based on my experiences running it locally. Please see the commit messages for details on each one.


atat/domain/csp/__init__.py
```diff
@@ -44,6 +44,6 @@ def make_csp_provider(app, csp=None):
     elif csp == "mock-test":
         app.csp = MockCSP(app, test_mode=True)
     elif csp == "hybrid":
-        app.csp = HybridCSP(app, test_mode=True)
+        app.csp = HybridCSP(app, test_mode=False)
```
graham-dds: I think the `test_mode` kwarg is a little confusing.

```python
class MockCSP:
    def __init__(self, app, test_mode=False):
        self.cloud = MockCloudProvider(
            app.config,
            with_delay=(not test_mode),
            with_failure=(not test_mode),
            with_authorization=(not test_mode),
        )
```
It looks like if `test_mode` is `False`, then all of the `with_` kawargs would be `True`. I'd assume that `with_delay=True` would mean `induce random delays`.

Maybe we can rename `test_mode` to like `simulate_failures` or something. Then, the value of that kwarg could be the same as all of the `with_` kwargs.

tests/domain/test_portfolios.py
```diff
@@ -261,7 +261,7 @@ def test_create_state_machine(portfolio):
 class TestGetPortfoliosPendingCreate(EnvQueryTest):
     """Portfolios.get_portfolios_pending_provisioning should return portfolios that
        are in their period of performace AND with a state machine is in one of the
-       following states: 
+       following states:
```
graham-dds: Should we update this docstring to say that this query should find `STARTING` and `STARTED`? Or maybe this actually belongs as the docstring of `Portfolios.get_portfolios_pending_provisioning`

atat/jobs.py
```diff
@@ -232,8 +232,10 @@ def send_PPOC_email(portfolio_dict):
 
 
 def make_initial_csp_data(portfolio):
+    csp_data = portfolio.csp_data or {}
     return {
         **portfolio.to_dictionary(),
+        **csp_data,
         "billing_account_name": app.config.get("AZURE_BILLING_ACCOUNT_NAME"),
     }
```
graham-dds: We have to repeat this process here and in the tests -- should logic be moved to a `Portfolio` domain method?

atat/domain/csp/__init__.py
dandds: That change makes sense. I saw fake network errors when I ran the Hybrid locally, which made me think I had to flip the switch.

dandds: I decided to make it configurable via a setting: 3699ed9cf7d964a9da9b997b8b97c5b1dc6dea3a

tests/domain/test_portfolios.py
dandds: Fixed 868c1a2d6d70f955d56ef3d12d17e3462b26bead

atat/jobs.py
dandds: I'm not sure the best way to resolve this because I don't think we should be importing the Flask app into the domain classes. I know we already do in a few cases, but I think it's a pattern we should fix. It's one of the reasons its hard for the Celery workers to run without an app context, going back to the point @jesse-dds raised in a separate conversation.

atat/domain/portfolios/portfolios.py
```diff
@@ -148,10 +148,15 @@ def base_provision_query(cls):
     @classmethod
     def get_portfolios_pending_provisioning(cls, now) -> List[UUID]:
         """
-        Any portfolio with a corresponding State Machine that is either:
-            not started yet,
-            failed in creating a tenant
-            failed
```
graham-dds: Based on the old docstring, is this query supposed to pick up state machines that failed a step too?

tests/domain/test_portfolios.py
```diff
@@ -277,6 +279,28 @@ def test_finds_unstarted(self):
         # Then the query will return the portfolio
         assert len(portfolios_pending) == 1
 
+    def test_finds_starting(self):
+        # Given: A portfolio is in its period of performance
+        # Given: The portfolio's state machine is in its "STARTING" stage
+        self.create_portfolio_with_clins(
+            [(self.YESTERDAY, self.TOMORROW)], state_machine_status="STARTING"
```
jesse-dds: Minor and nitpicky, but would it be better to create the fixture with `FSMStates.STARTING`? In the completely ridiculous chance that the value of `FSMStates.STARTING` changes.

tests/domain/test_portfolios.py
```diff
@@ -277,6 +279,28 @@ def test_finds_unstarted(self):
...
+        # When I query for portfolios pending provisioning
+        portfolios_pending = Portfolios.get_portfolios_pending_provisioning(self.NOW)
+        # Then the query will return the portfolio
+        assert len(portfolios_pending) == 1
+
+    def test_finds_started(self):
+        # Given: A portfolio is in its period of performance
+        # Given: The portfolio's state machine is in its "STARTED" stage
+        self.create_portfolio_with_clins(
+            [(self.YESTERDAY, self.TOMORROW)], state_machine_status="STARTED"
```
jesse-dds: Ditto with the `STARTING` comment.

atat/domain/portfolios/portfolios.py
dandds: It should do that but currently isn't written to do so, so I took out that reference until the SQL is correct.

tests/domain/test_portfolios.py
dandds: Yeah, the `PortfolioFactory` should actually be updated to take the enum and not a string, but we can use the enum here. Updates here: f52d69fb3e06fbcc1c9de9e9db3d77f387e410df

PR #1473, opened 2020-03-30
Environment creation hybrid

jesse-dds:
This pull request really just allows the hybrid to respond to create environment jobs by delegating to the azure instance.

https://www.pivotaltracker.com/n/projects/2160940


atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -228,3 +228,8 @@ def create_policies(self, payload: PoliciesCSPPayload) -> PoliciesCSPResult:
 
     def create_application(self, payload):
         return self.azure.create_application(payload)
+
+    def create_environment(
+        self, payload: EnvironmentCSPPayload
+    ) -> EnvironmentCSPResult:
+        return self.azure.create_environment(payload)
```
graham-dds, line 235: Should we somehow mutate this payload so that we know it was created with the hybrid -- similar to what we did with Portfolio provisioning?

jesse-dds: I think that's a good idea, I'll go ahead and do that.

PR #1476, opened 2020-03-31
Bandaid docker network fails on build

mikepw:
for the sake of unblocking dev, runs docker network rm (defensively) to test isolation of parallel tests phases and hopefully for the short term remove errors in the build.


mikepw:
removed!

PR #1479, opened 2020-03-31
Less random factory data

graham-dds:
Addresses [AT-4893](https://ccpo.atlassian.net/browse/AT-4893?atlOrigin=eyJpIjoiNzUyM2I2MjMwZjRjNDVjNzliNzA5NDM3NDQ2ZmViZjQiLCJwIjoiaiJ9)

We use [`factory_boy`](https://factoryboy.readthedocs.io/en/latest/) to generate fixture objects for tests. In these factories, we sometimes utilize randomly generated data via [`Faker`](https://faker.readthedocs.io/en/master/), which is a dependency of `factory_boy` .

Because some areas of our database schema have uniqueness constraints, coupled with the fact that sometimes this random data isn't so random, tests sometimes fail, even though the logic in them is sound.

This PR aims to eliminate those false negatives by replacing most instances of `Faker` with [`Sequence`es](https://factoryboy.readthedocs.io/en/latest/reference.html#sequence).

Below is a summary of each database entity with unique constraints, along with a description of how factories were modified to generate data for that value, if they used factories at all:

##### Attachment > object_name 
```python
object_name = Column(String, unique=True, nullable=False)
```
Factory: generates a UUID as a hex, which is highly unlikely to create a dupe

##### Permission Set > name
```python
name = Column(String, index=True, unique=True, nullable=False)
```
Factory: These values are solely controlled by `atat.domain.permission_sets.PermissionSets`

##### Task Order > number
```python
number = Column(String, unique=True,)
```
Factory: Switch from random number picker to `factory.Sequence`. For each iteration of the sequence, the value is left padded with 0s so that the resulting string is 10 characters long.

##### User > dod_id
```python
dod_id = Column(String, unique=True, nullable=False)
```
Factory: Switch from random number picker to `factory.Sequence`. For each iteration of the sequence, the value is left padded with 0s so that the resulting string is 10 characters long.

##### Application > Unique Constraint
```python
__table_args__ = (
    UniqueConstraint(
        "name", "portfolio_id", name="applications_name_portfolio_id_key"
    ),
)
```
Factory: Switch from factory.Faker to factory.Sequence. The resulting Application name will be `Application Name {n}`, where `n` is the current value of the sequence.

##### Environment > Unique Constraint
```python
__table_args__ = (
        UniqueConstraint(
            "name", "application_id", name="environments_name_application_id_key"
        ),
    )
```
Factory: Switch from factory.Faker to factory.Sequence. The resulting Application name will be `Environment Name {n}`, where `n` is the current value of the sequence.

##### Application Role > Index
```python
Index(
    "application_role_user_application",
    ApplicationRole.user_id,
    ApplicationRole.application_id,
    unique=True,
)
```
Factory: Not involved here

##### Environment Role > Index
```python
Index(
    "environments_role_user_environment",
    EnvironmentRole.application_role_id,
    EnvironmentRole.environment_id,
    unique=True,
)
```
Factory: Not involved here

##### Portfolio Role > Index
```python
Index(
    "portfolio_role_user_portfolio",
    PortfolioRole.user_id,
    PortfolioRole.portfolio_id,
    unique=True,
)
```
Factory: Not involved here

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,6 +27,18 @@ dictConfig({"version": 1, "handlers": {"wsgi": {"class": "logging.NullHandler"}}
 
 def pytest_addoption(parser):
     parser.addoption("--hybrid", action="store_true", default=False)
+    parser.addoption(
+        "--repeat", action="store", help="Number of times to repeat each test"
+    )
+
+
+def pytest_generate_tests(metafunc):
+    if metafunc.config.option.repeat is not None:
+        count = int(metafunc.config.option.repeat)
+
+        # Append fixture that accepts how many times each test should be repeated
+        metafunc.fixturenames.append("tmp_ct")
+        metafunc.parametrize("tmp_ct", range(count))
 
 
 def pytest_collection_modifyitems(config, items):
--- a/tests/factories.py
+++ b/tests/factories.py
@@ -1,18 +1,18 @@
-import re
 import operator
 import random
 import string
-import factory
+from typing import Callable
 from uuid import uuid4
-import pendulum
 
-from atat.forms import data
-from atat.models import *
-from atat.models.mixins.state_machines import FSMStates
+import factory
+import pendulum
 
 from atat.domain.invitations import PortfolioInvitations
 from atat.domain.permission_sets import PermissionSets
 from atat.domain.portfolio_roles import PortfolioRoles
+from atat.forms import data
+from atat.models import *
+from atat.models.mixins.state_machines import FSMStates
 
 
 def random_choice(choices):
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -27,6 +27,18 @@
 
 def pytest_addoption(parser):
     parser.addoption("--hybrid", action="store_true", default=False)
+    parser.addoption(
+        "--repeat", action="store", help="Number of times to repeat each test"
+    )
+
+
+def pytest_generate_tests(metafunc):
+    if metafunc.config.option.repeat is not None:
```
jesse-dds: I believe you can simply shorten this to `metafunc.config.option.repeat` since any string is inherently truthy.

jesse-dds: Also this is a pedantic comment, shorten it if you want, `is not None` is more expressive.

graham-dds: 👍 I think I'll keep `is not None` then, if it's all the same

PR #1483, opened 2020-04-01
User creation hybrid mock

carleton-dds:
[Jira issue](https://ccpo.atlassian.net/browse/AT-4883?atlOrigin=eyJpIjoiMjVkNzI4MWYxYjI1NDQwY2I0MjNjZGEyYWU3ZDNhYTgiLCJwIjoiaiJ9)

Adds `create_user` to hybrid cloud which utilizes `azure.create_user`. Also included `TestHybridCreateUserJob` test suite. The 3 tests create users in AD which should eventually be cleaned up if a clean up/tear down script is ever produced. The users have the first name `f"test-user-{uuid4()}",` and last name of "Solo"

### DIFF
--- a/tests/domain/cloud/test_hybrid_csp.py
+++ b/tests/domain/cloud/test_hybrid_csp.py
@@ -1,15 +1,19 @@
 import pendulum
 import pytest
+from unittest.mock import Mock
+from uuid import uuid4
 
 from atat.domain.csp import HybridCSP
 from atat.domain.csp.cloud.models import (
     EnvironmentCSPPayload,
     KeyVaultCredentials,
 )
-from atat.jobs import do_create_application
+from atat.jobs import do_create_application, do_create_user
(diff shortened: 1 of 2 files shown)


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -128,3 +131,95 @@ def test_hybrid_create_environment_job(session, csp):
     result = csp.create_environment(payload)
 
     assert result.id
+
+
+class TestHybridCreateUserJob:
```
dandds, line 136: I think this needs the `hybrid` marker that some of the other tests have; CI is trying to run it and failing.

carleton-dds: Whoops, I had that in there at one point but it must have gotten deleted somehow. Updated.

PR #1484, opened 2020-04-01
[AT-5184] Retry failed portfolio provisioning jobs (part 1)

graham-dds:
This is part 1 (of probably 2) of [AT-5184](https://ccpo.atlassian.net/browse/AT-5184)

In this PR, I:
- Added / refactored tests to cover `provision_portfolio` and `do_provision_portfolio` in `atat/jobs.py`
- Split `PortfolioStateMachine.after_in_progress_callback` into separate methods
- Moved tests related to state machine mixins into their own module

The next PR will ensure that all of the cases where we call `PortfolioStateMachine.fail_stage` trigger the appropriate exception so that Celery knows to retry them.

### DIFF
--- a/atat/models/portfolio_state_machine.py
+++ b/atat/models/portfolio_state_machine.py
@@ -239,12 +239,18 @@ class PortfolioStateMachine(
             # TODO: Ensure that failing the stage does not preclude a Celery retry
             self.fail_stage(self.current_stage)
 
-    def _stage_state_to_stage_name(self, stage_state):
-        return self.current_state.name.split(f"_{stage_state.name}")[0].lower()
-
     @property
-    def current_stage(self):
-        return self._stage_state_to_stage_name(StageStates.IN_PROGRESS)
+    def current_stage(self) -> str:
+        """Returns the current stage of the CSP provisioning process       
+        
+        E.g. TENANT_IN_PROGRESS -> tenant
+        """
+        for stage_state in StageStates:
+            if self.current_state.name.endswith(stage_state.name):
+                stage, stage_state = self.current_state.name.split(
+                    f"_{stage_state.name}"
+                )
+                return stage.lower()
 
     def after_in_progress_callback(self, event):
         payload = event.kwargs.get("csp_data")
--- a/atat/models/portfolio_state_machine.py
+++ b/atat/models/portfolio_state_machine.py
@@ -36,10 +36,6 @@ def _stage_to_classname(stage):
     return "".join(map(lambda word: word.capitalize(), stage.split("_")))
 
 
-def _stage_state_to_stage_name(state, stage_state):
-    return state.name.split(f"_{stage_state.name}")[0].lower()
-
-
 def get_stage_csp_class(stage, class_type):
     """
     given a stage name and class_type return the class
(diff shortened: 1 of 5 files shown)


atat/models/portfolio_state_machine.py
```diff
@@ -234,16 +229,35 @@ def after_in_progress_callback(self, event):
...
         except (ConnectionException, UnknownServerException) as exc:
             app.logger.error(
                 f"CSP api call. Caught exception for {self.__repr__()}.", exc_info=1,
             )
             # TODO: Ensure that failing the stage does not preclude a Celery retry
-            self.fail_stage(current_stage)
+            self.fail_stage(self.current_stage)
+
+    def _stage_state_to_stage_name(self, stage_state):
+        return self.current_state.name.split(f"_{stage_state.name}")[0].lower()
```
dandds, line 243: A little outside your scope, but could you throw a docstring here? I don't think it's immediately obvious what this does.

atat/models/portfolio_state_machine.py
```diff
@@ -234,16 +229,35 @@ def after_in_progress_callback(self, event):
...
+
+    @property
+    def current_stage(self):
+        return self._stage_state_to_stage_name(StageStates.IN_PROGRESS)
```
dandds, line 247: Does this assume that `self.current_state` is always a variation of `*_IN_PROGRESS`?

tests/models/mixins/test_state_machines.py
```diff
@@ -0,0 +1,61 @@
...
+        ["TENANT", "IN_PROGRESS"],
+        ["TENANT", "FAILED"],
+    ]
+    assert [t.get("trigger") for t in transitions] == [
+        "complete",
+        "create_tenant",
+        "finish_tenant",
+        "fail_tenant",
+        "resume_progress_tenant",
+    ]
```
dandds, line 38: This might also be outside your scope, but I'm not sure it makes sense that these states--which are specific to portfolios--live in a mixin.

atat/models/portfolio_state_machine.py
graham-dds: It did, but I refactored it here: https://github.com/dod-ccpo/atst/pull/1484/commits/36ccb50e75bf5b794b62dfaa0dc47cdac388f577

That being said, for now, this property is only ever accessed in `after_in_progress_callback`, which only will be called in an `*_IN_PROGRESS` stage

atat/models/portfolio_state_machine.py
graham-dds: I ended up deleting this method and refactoring the only place it's used -- `current_stage`. See below

tests/models/mixins/test_state_machines.py
graham-dds: Yeah, there are a few weird ways that CSP +System states are composed. Let's chat more about this 🙏

graham-dds:
Some things that could change in future PRs:
- We only have one model, the `PortfolioStateMachine` (`PSM`) that uses a state machine, so I don't think state machine mixins make sense. We can abstract some logic to mixins if/when we need to reuse that logic, but right now everything could very well live in the `PSM` class.
- There are some other functions and constants floating around in the state machine mixins and `PSM` modules, but outside their respective classes, that I think should should also move inside the `PSM` class. Things like `_build_csp_states(csp_stages)` & `_build_transitions(csp_stages)` in the state machine mixins, and `_stage_to_classname(stage)` & `get_stage_csp_class(stage, class_type)` in the `PSM` model class.
- I could see the `FSMStates = Enum("FSMStates", _build_csp_states(AzureStages))` constant being problematic. We shouldn't have a constant like is tied to a specific set of CSP stages. We're already specifying a CSP when we initialize this model - I think we should also set some property on the class to replace this constant. Maybe the CSP could know about its stages, so we could do something like `self.FSMStates = Enum("FSMStates", _build_csp_states(self.cloud.stages))`.

PR #1487, opened 2020-04-02
Update setup instructions

okaycj-atat:
While setting up my local environment during onboarding I ran into two issues.

1. The version of `node-sass` required a different version of nodejs then what was installed on laptop.  I added the requirement to the this repos readme along with instructions on how to install for macOS.  
2. The default ulimit caused some issues while running redis.  I added a command for resolving this issue to the trouble shooting section.

I don't feel strongly about either of these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,18 @@ locally:
   This dependency is optional. If present, the queue worker process will hot
   reload in development.
 
+* `node` == 10
+  This is a requirement of [node-sass](https://github.com/sass/node-sass).  For macOS, use brew to install:
+  ```
+  brew install node@10
+  ```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,6 +107,12 @@ run:
 createdb atat
 ```
 
+If there are errors were there are too many files open, you may need to increase the machine's ulimit:
```
dandds, line 110: Small typo: "were" -> "where". Also, do you know the source of this error (i.e., parcel, postgres, etc.)?

dandds: Actually you said it right in your PR description. Can you add a description or reference link here so other devs have more context on the error?

okaycj-atat: Thanks!  It's with redis.  I'll update the text.

PR #1488, opened 2020-04-02
User role assign unassign hybrid

jesse-dds:
Enables the hybrid interface to create user role assignments and disable them.

https://ccpo.atlassian.net/browse/AT-4886
https://ccpo.atlassian.net/browse/AT-4887

### DIFF
--- a/tests/domain/cloud/test_hybrid_csp.py
+++ b/tests/domain/cloud/test_hybrid_csp.py
@@ -1,3 +1,4 @@
+import json
 import pendulum
 import pytest
 from unittest.mock import Mock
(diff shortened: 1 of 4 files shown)


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -224,3 +228,62 @@ def test_hybrid_user_has_tenant(self, session, csp, app_role_1, app_1, user):
...
+            user_object_id=app_role_1.cloud_id,
+        )
+
+        create_user_role_result = csp.azure.create_user_role(payload)
+        disable_user_result = csp.azure.disable_user(
+            csp.azure.tenant_id, create_user_role_result.id
+        )
+        assert disable_user_result["id"] == create_user_role_result.id
+
+    def test_hybrid_do_create_environment_role_job(self, session, csp, portfolio, app):
```
dandds, line 262: I missed the fact that we had some tests for Celery tasks in this file. I think those would be better `tests/test_jobs.py`. We can leave it here for now; maybe we can do a sweep when we add better integration tests.

PR #1489, opened 2020-04-03
Error when creating application

carleton-dds:
[AT-4881](https://ccpo.atlassian.net/browse/AT-4881)

This solves a bug that was identified during the creation of application. To quote the Jira ticket:

> *Steps to reproduce*
> 1. start a new application
> 2. complete steps 1 (name and description) and 2 (add environments)
> 3. once on step 3 (add members) click the previous button
> 4. click the next button on step 2

Now instead of erroring after completing the above steps you successfully move back to step 3. 

The problem was that moving from step 2 to step 3 attempted to create the environments currently listed in the add-environments form, regardless of whether they already exist. I solved this problem by skipping any names from the add-environments forms if they already exist. I also add a test for this behavior.

One thing that this ticket highlighted to me was that it's really not communicated to the user is that moving from step 2 to step 3 immediately begins the process of creating the environments. Imagine the following situation:

1. While creating a new application a user accepts the default environments (Staging, Prod, Dev) and moves from step 2 to step 3
2. The user changes their mind, hits back on step 3 to return to the create environments page (step 2)
3. They edit "Staging" to instead be "Pre-Prod"
4. They click "Next" and move to step 3 (add members)

The user has now in fact created 4 environments (Staging, Prod, Dev, Pre-Prod) when in fact that may have thought they were creating 3. 

Not sure if they has already been considered, but I wanted to highlight it.

### DIFF
--- a/tests/domain/test_environments.py
+++ b/tests/domain/test_environments.py
@@ -155,6 +155,8 @@ def test_create_many_environments_will_skip_already_created_names():
         application.portfolio.owner, application, ["Staging", "Production"]
     )
     environments_second_run = Environments.create_many(
-        application.portfolio.owner, application, ["Staging", "Production"]
+        application.portfolio.owner,
+        application,
+        ["Staging", "Production", "Development"],
     )
-    assert len(environments_second_run) == 0
+    assert len(application.environments) == 3
--- a/tests/domain/test_environments.py
+++ b/tests/domain/test_environments.py
@@ -147,3 +147,14 @@ class TestGetEnvironmentsPendingCreate(EnvQueryTest):
             app_data={"cloud_id": uuid4().hex},
         )
         assert len(Environments.get_environments_pending_creation(self.NOW)) == 0
+
+
+def test_create_many_environments_will_skip_already_created_names():
+    application = ApplicationFactory.create()
+    environments_first_run = Environments.create_many(
+        application.portfolio.owner, application, ["Staging", "Production"]
+    )
+    environments_second_run = Environments.create_many(
+        application.portfolio.owner, application, ["Staging", "Production"]
+    )
+    assert len(environments_second_run) == 0
--- a/atat/domain/environments.py
+++ b/atat/domain/environments.py
@@ -29,8 +29,11 @@ class Environments(object):
     def create_many(cls, user, application, names):
         environments = []
         for name in names:
-            environment = Environments.create(user, application, name)
-            environments.append(environment)
+            if name not in [
+                existing_envs.name for existing_envs in application.environments
+            ]:
+                environment = Environments.create(user, application, name)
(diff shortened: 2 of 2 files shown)


graham-dds:
> One thing that this ticket highlighted to me was that it's really not communicated to the user is that moving from step 2 to step 3 **immediately begins** the process of creating the environments.

Not always. Say a user creates a portfolio, application, and environments in ATAT. Say they also add a task order with CLINs that mean the period of performance starts 5 days from now. Nothing will actually start happening as far as cloud provisioning _until_ the portfolio enters its period of performance in 5 days.

This probably could happen in another PR (with another ticket), but we should have the code reflect this:
- If the portfolio the application is being created under has not started provisioning, I think we're fine to edit names in the DB. That could make make it's way into this PR?
- If the portfolio _has_ started the provisioning process, it's a little trickier. @dandds had talked about just not letting users edit things once provisioning started. That being said, it looks like we're able to [update management groups](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/createorupdate), so I could see actually allowing users to edit env names in ATAT and having that be reflected in the CSP

tests/domain/test_environments.py
```diff
@@ -147,3 +147,14 @@ def test_with_already_provisioned_env(self, session):
...
+
+def test_create_many_environments_will_skip_already_created_names():
+    application = ApplicationFactory.create()
+    environments_first_run = Environments.create_many(
+        application.portfolio.owner, application, ["Staging", "Production"]
+    )
+    environments_second_run = Environments.create_many(
+        application.portfolio.owner, application, ["Staging", "Production"]
+    )
+    assert len(environments_second_run) == 0
```
dandds, line 160: Do we want to `assert` that `application` only had two envs, or do you not want to test side effects?

carleton-dds: I can make the changes you're looking for

carleton-dds:
@graham-dds Thanks for your comments. I may have used some inaccurate language in my description due my general ignorance still about the project. Can you clarify whether or not my concern around how a user can inadvertently create a new environment when they return to the add-environments page is a problem?

dandds:
@carleton-dds Yes, the scenario you outline can happen. If you can reproduce, can you add a bug ticket?

tests/domain/test_environments.py
carleton-dds: Much better: https://github.com/dod-ccpo/atst/pull/1489/commits/8588dbee055bb4fd8e079ab9204ec7cd108b9127

carleton-dds:
> @carleton-dds Yes, the scenario you outline can happen. If you can reproduce, can you add a bug ticket?

https://ccpo.atlassian.net/browse/AT-5206

PR #1493, opened 2020-04-03
Script for removing Hybrid app registrations.

dandds:
This script will delete app registrations added by the
HybridCloudInterface. It could be extended to delete additional entities
that provider can provision, like billing users and management groups. I
opted to handle app registrations to start because those entities are
given a huge number of permissions and it seems unwise to have too many
in a tenant.

### DIFF
--- a/script/teardown_hybrid.py
+++ b/script/teardown_hybrid.py
@@ -1,0 +1,84 @@
+import adal
+import requests
+from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD
+import os
+import sys
+
+parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+sys.path.append(parent_dir)
+
+from atat.app import make_config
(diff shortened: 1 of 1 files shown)


script/teardown_hybrid.py
```diff
@@ -0,0 +1,84 @@
...
+if __name__ == "__main__":
+    """
+    This script deletes applications created by the HybridCloudProvider
+    interface. It expects that the ATAT hybrid configuration settings
+    referenced below in this function have been set.
+    """
+    config = make_config()
+
+    if config["CSP"] != "hybrid":
+        raise ValueError("CSP setting must be 'hybrid' for script to work.")
```
graham-dds, line 67: Is there a way to set this on the fly in the script so one doesn't need to remember to flip the `CSP` to `hybrid`?

dandds: That's a good point. I thought about it and decided what the script really cares about is whether the config settings it's using are set. I was just using `CSP=hybrid` as a proxy for that, since those are hybrid-specific settings. I updated it to really ensure those are set.

PR #1495, opened 2020-04-07
Update query for finding CLINs to resend to the CSP.

dandds:
The query method for finding CLINs that need to be sent to the CSP needs
to check whether a portfolio is done being provisioned (i.e., its state
machine has the final state of COMPLETED). There are two cases where we
need to send CLIN information to the CSP:

- during the initial provisioning of the portfolio.
- when CLINs are added or edited after a portfolio has been provisioned.

The query and Celery task it's used by exist to handle the second case.
Before this update, ATAT would try to resend CLINs for portfolios that
had not finished provisioning and were missing necessary information
(such as the billing profile).

This commit also updates the dictionary access method to the portfolio's
csp_data in the Celery task, since we do not handle cases where we don't
find the relevant data.

Found and fixed while working on [AT-5192](https://ccpo.atlassian.net/browse/AT-5192).

### DIFF
--- a/atat/domain/task_orders.py
+++ b/atat/domain/task_orders.py
@@ -2,8 +2,14 @@ from sqlalchemy import or_
 import pendulum
 
 from atat.database import db
-from atat.models.clin import CLIN
-from atat.models.task_order import TaskOrder, SORT_ORDERING
+from atat.models import (
+    CLIN,
+    TaskOrder,
+    Portfolio,
+    PortfolioStateMachine,
+    FSMStates as PortfolioStates,
+)
+from atat.models.task_order import SORT_ORDERING
 from . import BaseDomainClass
 from atat.utils import commit_or_raise_already_exists_error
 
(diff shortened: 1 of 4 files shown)


tests/domain/test_task_orders.py
```diff
@@ -212,8 +212,36 @@ def test_get_for_send_task_order_files(
...
+        new_clins = TaskOrders.get_clins_for_create_billing_instructions()
+        assert provisioned_task_order.clins == new_clins
+
+    def test_sent_task_order(self, sent_task_order):
+        new_clins = TaskOrders.get_clins_for_create_billing_instructions()
+        assert len(new_clins) == 0
+
+    def test_portfolio_is_completed(self, provisioned_task_order):
+        new_clins = TaskOrders.get_clins_for_create_billing_instructions()
+        assert provisioned_task_order.clins == new_clins
```
graham-dds, line 239: This is the same test as 229-231, right?

atat/domain/task_orders.py
```diff
@@ -2,8 +2,14 @@
...
 
 from atat.database import db
-from atat.models.clin import CLIN
-from atat.models.task_order import TaskOrder, SORT_ORDERING
+from atat.models import (
+    CLIN,
+    TaskOrder,
+    Portfolio,
+    PortfolioStateMachine,
+    FSMStates as PortfolioStates,
```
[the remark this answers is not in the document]
dandds: Yup! Captured as [AT-5218](https://ccpo.atlassian.net/browse/AT-5218).

tests/domain/test_task_orders.py
dandds: Oh yup, good catch.

PR #1498, opened 2020-04-07
Require `get_portfolios_pending_provisioning` to return signed TOs

jesse-dds:
When a user enters a task order, and enters a start date for a CLIN POP that is on or before the current date, but does NOT sign the TO, provisioning will begin immediately. Only fully signed TOs should ‘count’ for the purposes of determining whether a portfolio is funded.

https://ccpo.atlassian.net/browse/AT-5182

### DIFF
--- a/atat/domain/portfolios/portfolios.py
+++ b/atat/domain/portfolios/portfolios.py
@@ -178,6 +178,6 @@ class Portfolios(object):
                     PortfolioStateMachine.state.like("%CREATED"),
                 )
             )
-            .filter(TaskOrder.signed_at != None)
+            .filter(TaskOrder.signed_at.isnot(None))
         )
         return [id_ for id_, in results]
--- a/atat/domain/portfolios/portfolios.py
+++ b/atat/domain/portfolios/portfolios.py
@@ -1,3 +1,5 @@
+import pendulum
+
 from sqlalchemy import or_
 from typing import List
 from uuid import UUID
@@ -146,9 +148,8 @@ class Portfolios(object):
         return db.session.query(Portfolio.id)
 
     @classmethod
-    def get_portfolios_pending_provisioning(cls, now) -> List[UUID]:
-        """
-        Any portfolio with
+    def get_portfolios_pending_provisioning(cls, now: pendulum.DateTime) -> List[UUID]:
+        """Retrieve UUIDs for any portfolio with the following properties:
             - Active CLINs (within the period of performance)
             - Not soft-deleted
             - No state machine attached OR
@@ -157,6 +158,7 @@ class Portfolios(object):
               - STARTING
               - STARTED
               - any that string-match /*CREATED$/
+            - A signed task order
         """
 
         results = (
@@ -176,5 +178,6 @@ class Portfolios(object):
                     PortfolioStateMachine.state.like("%CREATED"),
                 )
             )
+            .filter(TaskOrder.signed_at != None)
         )
         return [id_ for id_, in results]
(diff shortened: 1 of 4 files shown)


atat/domain/portfolios/portfolios.py
```diff
@@ -176,5 +178,6 @@ def get_portfolios_pending_provisioning(cls, now) -> List[UUID]:
                     PortfolioStateMachine.state.like("%CREATED"),
                 )
             )
+            .filter(TaskOrder.signed_at != None)
```
okaycj-atat, line 181: What do you think of using [`isnot()`](https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.isnot) instead?

tests/test_jobs.py
```diff
@@ -332,7 +332,8 @@ def test_dispatch_provision_portfolio(csp, monkeypatch):
                         "start_date": pendulum.now().subtract(days=1),
                         "end_date": pendulum.now().add(days=1),
                     }
-                ]
+                ],
+                "signed_at": pendulum.now()
```
okaycj-atat, line 336: In other tests, there are class properties have `self.YESTERDAY`, `self.TOMORROW`, etc.  Should that be implemented here?  Or is that too much refactoring?

atat/domain/portfolios/portfolios.py
okaycj-atat, line 181: What do you think of using [`isnot()`](https://docs.sqlalchemy.org/en/13/core/sqlelement.html#sqlalchemy.sql.expression.ColumnElement.isnot) here instead?

tests/test_jobs.py
jesse-dds: Interestingly, it looks like there were global `now`, `yesterday` and `tomorrow` constants already defined. I refactored the code in jobs to make use of them.

tests/domain/test_portfolios.py
```diff
@@ -389,3 +373,17 @@ def test_with_active_clins_and_state_machine(self):
...
         # Then the query will return the pending portfolio
         assert len(portfolios_pending) == 1
+
+    def test_with_unsigned_task_order(self):
+        # Given: A portfolio is in its period of performance
+        # Given: The portfolio has begun the provisioning process
+        # Given: Portfolio is associated with an unsigned task order
+        self.create_portfolio_with_clins(
+            [(self.YESTERDAY, self.TOMORROW)],
+            state_machine_status=FSMStates.TENANT_CREATED.name,
```
dandds, line 383: Very nitpicky, but a more realistic example would be UNSTARTED, since a portfolio wouldn't make it to the TENANT_CREATED state with an unsigned task order (we hope).

tests/utils.py
```diff
@@ -1,15 +1,13 @@
...
+from contextlib import contextmanager
 from cryptography.hazmat.backends import default_backend
 from flask import template_rendered
-import pendulum
-
-from atat.utils.notification_sender import NotificationSender
-
-import tests.factories as factories
+from OpenSSL import crypto
+from unittest.mock import Mock
```
dandds, line 10: What's the logic to the rearrangement here? I can see that the second block of imports is alphabetical, but the first block is not. My own preference for python import order is 1) stdlib imports, 2) dependency imports, 3) application import, and 4) test imports, each separated by a linebreak and alphabetical within its own block. (This isn't consistent in the code, though.)

jesse-dds: Ah, yeah I wanted to (failingly) rearrange them alphabetically haha. I'll revert this now that I see the structure.

PR #1499, opened 2020-04-08
Service desk url

okaycj-atat:
The service link appears on most pages in the top bar.  The solution I favored was to create a global constant within Jinja that stores service desk url from the base.ini config file and makes it available in the template.  

See [AT-5181](https://ccpo.atlassian.net/browse/AT-5181)


templates/navigation/topbar.html
```diff
@@ -2,33 +2,34 @@
...
-        </a>
-        <a href="#" class="topbar__link">
-          {{ Icon('question', classes='topbar__link-icon') }}
-          <span class="topbar__link-label">Support</span>
-        </a>
+      <a href="{{ url_for('users.user') }}" class="topbar__link">
+        {{ Icon('user', classes='topbar__link-icon') }}
+        <span class="topbar__link-label">{{ g.current_user.first_name + " " + g.current_user.last_name }}</span>
+      </a>
+      <a href="{{ service_desk_url() or '#' }}" rel="noopener noreferrer" target="_blank" class="topbar__link">
```
graham-dds, line 17: I guess this is just being extra defensive, but since you `app.config.get("SERVICE_DESK_URL", "#")` when you build `_service_desk_url()`, you could probably get away with the `or '#'` here.

config/base.ini
```diff
@@ -75,3 +75,4 @@ SQLALCHEMY_ECHO = False
 STATIC_URL=/static/
 USE_AUDIT_LOG = false
 WTF_CSRF_ENABLED = true
+SERVICE_DESK_URL
```
graham-dds, line 78: Can you reorder this in the list so that the config values are in alphabetical order? Also, could you add a description of this config value to [this section](https://github.com/dod-ccpo/atst/blob/staging/README.md#configuration) of the `README`?

PR #1503, opened 2020-04-10
Log portfolio and application ids in jobs

jesse-dds:
https://ccpo.atlassian.net/browse/AT-5185

**Print any relevant errors/the response body for 400 responses from Azure (these are basically input validation errors in many cases)**

This seemed to be already covered. All of our API calls to azure seemed to follow this pattern:

```python
       try:
            sdk.apicall(...)
        except self.sdk.requests.exceptions.HTTPError:
            app.logger.error(f"...", exc_info=1)
            raise ConnectionException("connection error creating tenant")
```

The `exc_info` handles printing out the traceback, which includes info about the HTTP response.

**Stop printing full HTTP exchanges from Key Vault API calls**

This will be handled by this task https://ccpo.atlassian.net/browse/AT-5221 which will give us full control over what logging information we output. My attempts to retrieve the SDK's logger by name seemed to break the SDK's setup of its own logger and no information would output.

To reproduce this, set a test to `DEBUG` mode like so

```python
@pytest.mark.hybrid
def test_hybrid_create_application_job(session, csp, caplog):
    import logging

    caplog.set_level(logging.DEBUG)
    ...
```

And add `logging.getLogger("azure.core.pipeline.policies.http_logging_policy")` to our application's setup. I chose `make_app`. This line will cause the sdk to stop logging messages with "azure.core.pipeline.policies.http_logging_policy".

**Ensure linebreaks do not print as multiple log statements**

This will also be fixed by task https://ccpo.atlassian.net/browse/AT-5221

The only culprit I could find was the `azure.core.pipeline.policies.http_logging_policy` which seems to intentionally log each line of the HTTP response as a separate log message. Linebreaks do not necessarily print as multiple log statements. 

Here's a library doing it the way we want:
```
DEBUG [msal.token_cache] event={
    "response": {
        "access_token": "********",
        "expires_in": 3599,
        "expires_on": 1586534711,
        "ext_expires_in": 3599,
        "token_type": "Bearer"
    },
    "scope": [
        "https://vault.azure.net/.default"
    ]
}
```

Here are multiple logs per line:
```
INFO  [azure.core.pipeline.policies.http_logging_policy] Request method: 'GET'
INFO  [azure.core.pipeline.policies.http_logging_policy] Request headers:
INFO  [azure.core.pipeline.policies.http_logging_policy]     'Accept': 'application/json'
```

### DIFF
--- a/atat/jobs.py
+++ b/atat/jobs.py
@@ -81,6 +81,12 @@ def do_create_application(csp: CloudProviderInterface, application_id=None):
         csp_details = application.portfolio.csp_data
         parent_id = csp_details["root_management_group_id"]
         tenant_id = csp_details["tenant_id"]
+
+        app.logger.debug(f"application.id = {application.id}")
+        app.logger.debug(f"application.portfolio.id = {application.portfolio.id}")
+        app.logger.debug(f"tenant_id = {tenant_id}")
+        app.logger.debug(f"parent_id = {parent_id}")
+
         payload = ApplicationCSPPayload(
             tenant_id=tenant_id, display_name=application.name, parent_id=parent_id
         )
@@ -100,6 +106,9 @@ def do_create_user(csp: CloudProviderInterface, application_role_ids=None):
     with claim_many_for_update(app_roles) as app_roles:
 
         if any([ar.cloud_id for ar in app_roles]):
+            app.logger.warning(
+                f"Application role cloud ID {ar.cloud_id} already present."
+            )
             return
 
         csp_details = app_roles[0].application.portfolio.csp_data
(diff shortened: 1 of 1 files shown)


jesse-dds:
I ignored tasks explicitly marked with `ignore_result=True`.

atat/jobs.py
```diff
@@ -247,80 +271,101 @@ def make_initial_csp_data(portfolio):
...
-def dispatch_provision_portfolio(self):
+def dispatch_provision_portfolio(self: Task):
     """
     Iterate over portfolios with a corresponding State Machine that have not completed.
     """
+    portfolio_ids = []
     for portfolio_id in Portfolios.get_portfolios_pending_provisioning(pendulum.now()):
         provision_portfolio.delay(portfolio_id=portfolio_id)
+        portfolio_ids.append(portfolio_id)
+    return portfolio_ids
```
dandds, line 326: Is it simpler/more readable to do:

```
portfolio_ids = Portfolios.get_portfolios_pending_provisioning(pendulum.now())
for portfolio_id in portfolio_ids:
   [etc]
return portfolio_ids
```
Genuine question. Probably 6 of one, half dozen of the other.

atat/jobs.py
```diff
@@ -247,80 +271,101 @@ def make_initial_csp_data(portfolio):
...
     for application_id in Applications.get_applications_pending_creation():
         create_application.delay(application_id=application_id)
+        application_ids.append(application_id)
+    return application_ids
 
 
 @celery.task(bind=True)
-def dispatch_create_user(self):
+def dispatch_create_user(self: Task):
+    application_role_idss = []
```
dandds, line 340: Maybe `application_role_id_groups`? 🤷‍♂️

atat/jobs.py
jesse-dds: That works, saves the append line too.

PR #1504, opened 2020-04-10
[AT-5216] Set default config value for billing account name

graham-dds:
Closes: [AT-5216](https://ccpo.atlassian.net/browse/AT-5216)
There was a bug where portfolios would fail to provision with the mock
CSP because a value wasn't set for the `AZURE_BILLING_ACCOUNT_NAME` config
var, and we'd hit a validation error in the `BILLING_PROFILE_CREATION`
step. In patching that bug:
- In reporting, instead of guarding against calling a reporting query
  based on the mere presence of CSP data on the portfolio, ensure that a
  state machine is present and completed.
- Get billing account name by index (not .get()) so that an error is
  thrown if it isn't present.


atat/domain/reports.py
```diff
@@ -1,10 +1,12 @@
+import pendulum
 from flask import current_app
+
 from atat.domain.csp.cloud.models import (
     CostManagementQueryCSPPayload,
     CostManagementQueryCSPResult,
 )
 from atat.domain.csp.reports import prepare_azure_reporting_data
-import pendulum
+from atat.models.mixins.state_machines import FSMStates
```
dandds, line 9: Can we import this from `atat.models`?

atat/domain/reports.py
```diff
@@ -20,8 +22,10 @@ def get_portfolio_spending(cls, portfolio):
         from_date = pendulum.now().subtract(years=1).add(days=1).format("YYYY-MM-DD")
         to_date = pendulum.now().format("YYYY-MM-DD")
         rows = []
-
-        if portfolio.csp_data:
+        if (
+            portfolio.state_machine
+            and portfolio.state_machine.state_str == FSMStates.COMPLETED.name
+        ):
```
dandds, line 28: This seems to make `Reports` responsible for knowing a lot of implementation details. Would you consider implementing a boolean property like `is_provisioned` or something like that on the Portfolio model?

atat/domain/reports.py
graham-dds: I opted to make the `is_provisioned` property, and in doing so, I think I needed to keep this full path to `FSMStates` to prevent a circular import (moved to the Portfolio model, though)

dandds: Yeah, using the full path inside the models is fine!

tests/domain/test_reports.py
```diff
@@ -7,7 +7,7 @@
 
 @pytest.fixture(scope="function")
 def portfolio():
-    portfolio = PortfolioFactory.create()
+
     return portfolio
```
dandds, line 11: Should we just delete this fixture?

PR #1505, opened 2020-04-10
[AT-5225] Normalize management group IDs for role creation.

dandds:
There was a bug in user role creation (the `create_user_role` method on
the `AzureCloudProvider`). ATAT is storing environments' management
group IDs as full Azure paths, but that method was expecting them as
simple GUIDs. To address this, I've made it so that the corresponding
payload model (`UserRoleCSPPayload`) normalizes its management group ID
to always be a full path. That way, the user role creation method can
always expect to have a full path. This involved extracting an existing
validator function from another model so that it can be re-used, as
explained in the pydantic docs here:
https://pydantic-docs.helpmanual.io/usage/validators/#reuse-validators

[AT-5225](https://ccpo.atlassian.net/browse/AT-5225)

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1867,7 +1867,7 @@ class AzureCloudProvider(CloudProviderInterface):
             )
 
         role_guid = self.roles[payload.role]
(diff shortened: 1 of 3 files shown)


dandds:
@graham-dds Yup, agree on all points. Could you add this write-up as a tech debt ticket?

PR #1506, opened 2020-04-13
Remove starting and started stages

carleton-dds:
[AT-5183](https://ccpo.atlassian.net/browse/AT-5183)

Originally portfolios went through several system stages  (unstarted, starting, started) before beginning the CSP specific stages (tenant creation being the first of those at the moment). This PR removes the "starting" and "started" stages so now we'll go straight from "unstarted" to tenant creation. I couldn't find any reason for these system stages to exist as we simply moved through them immediately without triggering any callbacks or whatever. I mostly just stripped out the code that dealt with those stages, including some tests.

There had also been some discussion around the FAILED system stage and whether it was needed as the FAILED stage for each provisioning step (tenant_creation_failed) generally covers _most_ of our needs if a CSP stage fails . However, the system state FAILED is still helpful because  if we fail to properly find the next CSP trigger, especially at the very beginning of the process, (say we fail to find the very first step) we want to be able to put the process into a failed state, and that's not possible without a system FAILED. [test_find_and_call_stage_trigger_fails](https://github.com/dod-ccpo/atst/blob/8751658e6d63efc6e1010d0c4ade281592898cae/tests/models/mixins/test_state_machines.py#L29-L36) tests this functionality.


tests/models/test_portfolio_state_machine.py
```diff
@@ -59,22 +59,6 @@ def test_fsm_creation(portfolio):
...
-        state_machine.state = FSMStates.STARTING
-        state_machine.trigger_next_transition()
-        assert state_machine.current_state == FSMStates.STARTED
-
-    def test_started(self, state_machine):
-        state_machine.state = FSMStates.STARTED
-        state_machine.trigger_next_transition(
-            csp_data=state_machine.portfolio.to_dictionary()
-        )
-        assert state_machine.current_state == FSMStates.TENANT_CREATED
```
graham-dds, line 76: Instead of fully deleting this test, could you modify it so that it asserts that, given a state machine that is `UNSTARTED`, `trigger_next_transition()` transitions to `FSMStates.TENANT_CREATED`?

PortfolioProvision.md
```diff
@@ -38,7 +38,7 @@ A Finite State Machine `atat.models.portoflio_state_machine.PortfolioStateMachin
 
 Aach time the FSM is created/accessed it will generate a list of States and Transitions between the states.
 
-There is a set of "system" states such as UNSTARTED, STARTING, STARTED, COMPLETED, FAILED etc
+There is a set of three "system" states: UNSTARTED, COMPLETED and FAILED.
```
dandds, line 41: Good catch!

tests/models/test_portfolio_state_machine.py
carleton-dds: FAIL to CONFIGURATION_ERROR: https://github.com/dod-ccpo/atst/pull/1506/commits/8719861cbfd1a1af1e39f27045dafe80362337cb

carleton-dds: `test_started` test re-added: https://github.com/dod-ccpo/atst/pull/1506/commits/6c1263676b4445aa340cee28445a046211b5ab6e

Ready for re-review

carleton-dds:
Had to fix an upstream test on staging, so need a new approval https://github.com/dod-ccpo/atst/pull/1506/commits/edc060a29293f3ccbe72bfdcac04800e82bcb6a5

PR #1509, opened 2020-04-14
Delete removed envs and restore deleted envs

okaycj-atat:
During application creation, a user has the opportunity to create environments and edit them.  If a user decides to remove and/or rename an environment, these edits weren't being properly saved.

To resolve this issue, we're now marking removed environments as deleted.  Additionally, when an environment is renamed back to an environment that has been created, it is restored. 

__Open questions__

Instead of deleting environments, should we instead update the name field?  This would require the UI to know the environment's ID.


okaycj-atat:
This issue has been marked `won't fix`.  Closing PR.

PR #1512, opened 2020-04-15
[AT-4890] Abstract azure error handling

graham-dds:
This is my current solution to [AT-4890](https://ccpo.atlassian.net/browse/AT-4890). Other things I'm thinking about:
- If I were to go one step further, instead of using a decorator, I'd use a `requests` [event hook](https://2.python-requests.org/en/master/user/advanced/#event-hooks). This would make it so that `raise_for_status()` (and logging in the case of an error) would be called automatically on the response.
- If I were to go one step _further_, I'd use a [`session object`](https://2.python-requests.org/en/master/user/advanced/#session-objects) to make API calls. This way, we wouldn't have to remember to register the status checking and error logging hook for each API call -- it would be included automatically. 

That being said -- I think that would take some non-trivial refactoring of our tests to implement either of these setups. There's a lot of mocking that happens that would have to be changed.

There's also some opportunity to DRY stuff that has to do with token-getting, but that might be a better task for [AT-5195](https://ccpo.atlassian.net/browse/AT-5195)

Questions:
- would it be worth it to further explore the event hook method, with the caveat that it would bump up the story points for this ticket?
- There's probably a reason that we import requests in each SDK class class and use that in the cloud provider class (e.g. `self.sdk.requests.<method>`) instead of just importing requests in the module and using it directly in the class, but, why?

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -330,36 +330,17 @@ class AzureCloudProvider(CloudProviderInterface):
         response = mgmgt_group_client.management_groups.get(management_group_id)
         return response
 
+    @log_and_raise_exceptions
     def _create_policy_definition(self, session, root_management_group_name, policy):
         create_policy_definition_uri = f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementGroups/{root_management_group_name}/providers/Microsoft.Authorization/policyDefinitions/{policy.definition['properties']['displayName']}?api-version=2019-09-01"
         body = policy.definition
-        try:
-            result = session.put(create_policy_definition_uri, json=body, timeout=30,)
-            result.raise_for_status()
-            if result.status_code == 201:
-                return result.json()
 
-        except self.sdk.requests.exceptions.ConnectionError:
-            app.logger.error(
-                f"Could not create policy definition. Connection Error", exc_info=1,
-            )
-            raise ConnectionException("connection error creating policy definition")
-        except self.sdk.requests.exceptions.Timeout:
-            app.logger.error(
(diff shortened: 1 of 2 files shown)


graham-dds:
Oh --  [`raise_for_status()`](https://2.python-requests.org/en/master/api/#requests.Response.raise_for_status) only raises HTTPErrors. So even if we did want to go down the route of request hooks to automatically call `raise_for_status`, we'd probably want the decorator as well for all of the other exceptions that could be raised.

With that in mind, I'm going to mark this PR as ready for review, and explore the hook / session object in a separate PR.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -83,6 +84,31 @@
 DEFAULT_POLICY_SET_DEFINITION_NAME = "Default JEDI Policy Set"
 
 
+def log_and_raise_exceptions(func):
+    @wraps(func)
+    def wrapped_func(self, *args, **kwargs):
```
dandds, line 89: Little thing, but I might call `self` something else here (`cloud`?) since this is defined outside the class.

graham-dds: Updated. Yeah, I was initially unsure where to actually define this decorator. I came across [this article](https://medium.com/@vadimpushtaev/decorator-inside-python-class-1e74d23107f6) that explored other options, but ultimately I think this works just fine.

graham-dds:
> Do you want to add the decorator to `validate_domain_name`?

Good catch! Added.

> There's no way to avoid having to call `raise_for_status` every time without hooks, right?

I don't think so... unless you have any ideas?

Agreed with your lack of `else` branches

> I'm not certain that adding the hooks is going to buy us enough additional flexibility to be worth pursuing on this ticket. 

👍 https://ccpo.atlassian.net/browse/AT-5248

PR #1513, opened 2020-04-15
Add method for disabling user role to mock cloud interface.

dandds:
I noticed that the mock implementation of the cloud interface was
missing the `create_user_role` while debugging something unrelated. This
commit implements it by generating a GUID for the resulting user role.
It also alphabetizes the list of model imports in the mock module.

### DIFF
--- a/atat/domain/csp/cloud/mock_cloud_provider.py
+++ b/atat/domain/csp/cloud/mock_cloud_provider.py
@@ -26,46 +26,48 @@ from .models import (
     BillingProfileTenantAccessCSPResult,
     BillingProfileVerificationCSPPayload,
     BillingProfileVerificationCSPResult,
+    CostManagementQueryCSPPayload,
+    CostManagementQueryCSPResult,
+    EnvironmentCSPPayload,
+    EnvironmentCSPResult,
     InitialMgmtGroupCSPPayload,
     InitialMgmtGroupCSPResult,
     InitialMgmtGroupVerificationCSPPayload,
     InitialMgmtGroupVerificationCSPResult,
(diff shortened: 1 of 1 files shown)


atat/domain/csp/cloud/mock_cloud_provider.py
```diff
@@ -438,6 +440,12 @@ def disable_user(self, tenant_id, role_assignment_cloud_id):
 
         return self._maybe(12)
 
+    def create_user_role(self, payload: UserRoleCSPPayload):
```
graham-dds, line 443: It looks like since `create_user_role` involves grabbing a token, it has a chance to throw an Authentication exception. Do you want to include that here as well?

Also, to prevent bugs like this in the future, should we add a ticket to make sure all of the methods are on the Cloud Provider Interface -- and maybe rename it Azure Cloud Provider interface, since that's what it's become? Don't know how helpful this would be, and we'd still have to remember to add new methods on the interface, but maybe we could write a test that asserts that the interface and each subclass have the same public methods.

dandds: Good call, added that exception type.

I'm not sure what the best approach is to the larger problem, to be honest. Python interfaces are ad-hoc, not part of the language core. Using a test to enforce it feels a little clunky to me but is a possibility. What do you think? @jesse-dds likes type annotations; do you have any thoughts on this?

graham-dds: Would an [abstract base class](https://docs.python.org/3/library/abc.html#module-abc) be helpful here?

PR #1516, opened 2020-04-16
Add get_reporting_data to hybrid csp class

okaycj-atat:
See [AT-5212](https://ccpo.atlassian.net/browse/AT-5212)

Add `get_reporting_data()` method to Hybrid CSP class using Hybrid specific config values.


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1383,17 +1383,18 @@ def _source_tenant_creds(self, tenant_id) -> KeyVaultCredentials:
         return KeyVaultCredentials(**json.loads(raw_creds))
 
     @log_and_raise_exceptions
-    def get_reporting_data(self, payload: CostManagementQueryCSPPayload):
+    def get_reporting_data(self, payload: CostManagementQueryCSPPayload, token=None):
```
dandds, line 1386: Are there circumstances where we would pass in the token?

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1383,17 +1383,18 @@ def _source_tenant_creds(self, tenant_id) -> KeyVaultCredentials:
...
         """
-        creds = self._source_tenant_creds(payload.tenant_id)
-        token = self._get_service_principal_token(
-            payload.tenant_id, creds.tenant_sp_client_id, creds.tenant_sp_key
-        )
+        if token is None:
+            creds = self._source_tenant_creds(payload.tenant_id)
+            token = self._get_sp_token(
+                payload.tenant_id, creds.tenant_sp_client_id, creds.tenant_sp_key
+            )
```
dandds, line 1397: I might be missing some nuance, but it seems simpler to me to just update this method to use:

```
token = self.azure._get_tenant_principal_token(payload.tenant_id)
```

Then you don't need to worry about passing the token or getting a bearer token in the Hybrid version of this method.

config/base.ini
```diff
@@ -39,7 +41,10 @@ DEBUG = true
 DEBUG_MAILER = false
 DISABLE_CRL_CHECK = false
 ENVIRONMENT = dev
+FAKE_TENANT_ID
+FAKE_TENANT_IDASSETS_URL
```
dandds, line 45: Looks like a typo.

config/base.ini
```diff
@@ -25,6 +25,8 @@ AZURE_TENANT_ID
 AZURE_TO_BUCKET_NAME
 AZURE_USER_OBJECT_ID
 AZURE_VAULT_URL
+BILLING_ACCOUNT_ID
+BILLING_PROFILE_ID
```
dandds, line 29: For Azure-specific values, I think we should prefix them with `AZURE_`, so `AZURE_BILLING_ACCOUNT_ID`, etc.

config/base.ini
```diff
@@ -25,6 +25,8 @@ AZURE_TENANT_ID
 AZURE_TO_BUCKET_NAME
 AZURE_USER_OBJECT_ID
 AZURE_VAULT_URL
+BILLING_ACCOUNT_ID
```
graham-dds, line 28: Can we prefix these config vars with `AZURE_` and add them to the README?

config/base.ini
```diff
@@ -39,7 +41,10 @@ DEBUG = true
 DEBUG_MAILER = false
 DISABLE_CRL_CHECK = false
 ENVIRONMENT = dev
+FAKE_TENANT_ID
```
dandds, line 44: I don't think we need to supply this as config; that was just for demo purposes in my example script. See my note below in the test.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -132,6 +132,23 @@ def test_hybrid_create_environment_job(session, csp):
     assert result.id
 
 
+@pytest.mark.hybrid
+def test_get_reporting_data(csp):
```
dandds, line 136: Take a look at the test setup `test_hybrid_create_application_job` does. It uses this code to set up a key vault credential object for a known tenant:

```
    csp.azure.create_tenant_creds(
        csp.azure.tenant_id,
        KeyVaultCredentials(
            root_tenant_id=csp.azure.tenant_id,
            root_sp_client_id=csp.azure.client_id,
            root_sp_key=csp.azure.secret_key,
            tenant_id=csp.azure.tenant_id,
            tenant_sp_key=csp.azure.secret_key,
            tenant_sp_client_id=csp.azure.client_id,
        ),
    )
```
Right now, your test is dependent on there always being a key vault credential object that corresponds to the one supplied with FAKE_TENANT_ID. It would be better if the test handled its own setup without that dependency.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -132,6 +132,23 @@ def test_hybrid_create_environment_job(session, csp):
...
+    from_date = pendulum.now().subtract(years=1).add(days=1).format("YYYY-MM-DD")
+    to_date = pendulum.now().format("YYYY-MM-DD")
+    fake_tenant_id = csp.azure.config["FAKE_TENANT_ID"]
+
+    payload = CostManagementQueryCSPPayload(
+        tenant_id=fake_tenant_id,
```
graham-dds, line 142: Instead of setting a `FAKE_TENANT_ID` config variable, I think you can just use `csp.azure.tenant_id` here since we're querying data for the PW tenant.

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -260,3 +262,16 @@ def create_user_role(self, payload: UserRoleCSPPayload) -> UserRoleCSPResult:
 
     def disable_user(self, tenant_id: str, role_assignment_cloud_id: str) -> Dict:
         return self.azure.disable_user(tenant_id, role_assignment_cloud_id)
+
+    def get_reporting_data(self, payload: CostManagementQueryCSPPayload, token=None):
+        billing_account_id, billing_profile_id, invoice_section_id = itemgetter(
+            "BILLING_ACCOUNT_ID", "BILLING_PROFILE_ID", "INVOICE_SECTION_ID"
+        )(app.config)
```
graham-dds, line 269: Instead of pulling in the flask app, you can also get this values from `self.azure.config`

atat/domain/csp/cloud/azure_cloud_provider.py
okaycj-atat: I assumed we needed to preserve the current behavior in the Azure CSP.  I'll update this.

atat/domain/csp/cloud/azure_cloud_provider.py
okaycj-atat: No.  Taking into account the below comments, I can remove this argument.

config/base.ini
graham-dds, line 45: Is this supposed to be used anywhere?

graham-dds:
@okaycj-atat deleted some of my comments since @dandds beat me to it!

PR #1517, opened 2020-04-16
[WIP] [AT-5208] VCR Hybrid Testing

jesse-dds:
Caches most HTTP interactions performed by the hybrid tests as "cassette" (yaml) files. When the tests are run with the cassettes present, instead of issuing real requests, the VCR library instead serves the cached response. 

https://ccpo.atlassian.net/browse/AT-5208

Co-authored by @carleton-dds 

## Issues

@carleton-dds and I found that the management group SDK was using two different versions of the SDK at the same time. `2015` and `2018-preview`. We probably should investigate this in other places and standardize the API version used.

### DIFF
--- a/tests/domain/cloud/test_hybrid_csp.py
+++ b/tests/domain/cloud/test_hybrid_csp.py
@@ -33,8 +33,15 @@ from tests.factories import (
     TaskOrderFactory,
     UserFactory,
 )
-from uuid import uuid4
 from vcr import VCR
+
+
+REDACTION_STRING = "*****"
+UUID_RE = (
+    r"[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}"
+)
+HASH_RE = r"[0-9a-fA-F]{32}"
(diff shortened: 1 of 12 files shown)


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -48,15 +167,29 @@ def portfolio():
 
 @pytest.fixture(scope="function")
 def csp(app):
-    return HybridCSP(app, simulate_failures=True).cloud
+    app.config["AZURE_CLIENT_ID"] = app.config["AZURE_CLIENT_ID"] or str(uuid4())
+    app.config["AZURE_SECRET_KEY"] = app.config["AZURE_SECRET_KEY"] or str(uuid4())
+    app.config["AZURE_TENANT_ID"] = app.config["AZURE_TENANT_ID"] or str(uuid4())
+    app.config["AZURE_VAULT_URL"] = (
+        app.config["AZURE_VAULT_URL"]
+        or "https://hybridcz-pwdev-keyvault.vault.azure.net/"
```
graham-dds, line 175: Do we need to be careful about commit this Keyvault url?

jesse-dds: I think it's ok, you need credentials / an auth tokens to get to the URL

PR #1518, opened 2020-04-16
[AT-5195] Raise AuthenticationExceptions for Adal errors and DRY token getting

graham-dds:
Closes: [AT-5195](https://ccpo.atlassian.net/browse/AT-5195)

I discovered that this bug might have been squashed by https://github.com/dod-ccpo/atst/pull/1497.  That catches all exceptions raised during `after_in_progress_callback` and fails the stage. This PR:
- catches ADAL errors and re-raises them as our `AuthenticationException` error
- DRYs token checking. We would do something like:
```python
token = self.fn_that_gets_a_token()
if not token:
    raise AuthenticationException()
```

in each cloud method that needed some authentication token. Now, that logic lives in `fn_that_gets_a_token()`. This makes it so you don't have to remember to check for a token or raise an exception in each method.

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1343,29 +1343,27 @@ class AzureCloudProvider(CloudProviderInterface):
             creds.root_tenant_id, creds.root_sp_client_id, creds.root_sp_key
         )
 
+    @log_and_raise_exceptions
     def _get_service_principal_token(
         self, tenant_id, client_id, secret_key, resource=None
     ):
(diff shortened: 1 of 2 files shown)


carleton-dds:
I don't know if this is a concern at all, but I did notice that the [python adal package README](https://github.com/AzureAD/azure-activedirectory-library-for-python) states that:
>This library, ADAL for Python, will no longer receive new feature improvements. Instead, use the new library MSAL for Python

dandds:
> I don't know if this is a concern at all

@carleton-dds That's a good catch; could you add a ticket?

PR #1521, opened 2020-04-20
Database scripts

carleton-dds:
[AT-5220](https://ccpo.atlassian.net/browse/AT-5220)

The bug this PR addresses is the fact that the `atat` Postgres user in Azure does not have sufficient privileges to run `script/reset_database.py`.  The error message is:
`psycopg2.errors.InsufficientPrivilege: must be owner of function lock_dod_id`

This error is caused by the fact that `script/reset_database.py` reapplies all migrations. [One of the migrations](https://github.com/dod-ccpo/atst/blob/018229dcd8dfb4ac93ef184b8c6e8e6d53778fe3/alembic/versions/e0c6eb21771f_reset_migrations_with_new_schema.py) creates the stored procedure/function `lock_dod_id`. 

Postgres enforces strict permissions on functions, limiting the ability to edit them to their owner or superusers. Because the database migrations are originally run by a Azure admin user, the `atat` user does not have permission to re-create the function when re-running all migrations.

To resolve this problem it's been decided that we will edit the `script/database_setup.py` script that is initially run by the Azure Postgres user so that it no longer calls `reset_database` which is when migrations are initially applied. Instead `database_setup` and `reset_database` will be called separately by the deployment process so that the admin user can create the `atat` user, and then the `atat` user will be used to run `reset_database` for the first time, guaranteeing that `atat` is the owner of the `lock_dod_id` function. 

What this PR does specifically is 
1. decouple `reset_database` from `database_setup.py`. 
2. Rename `script/add_ccpo_user.py` to `script/grant_ccpo_perms.py`, to better reflect what it actually does
3. move `add_ccpo_users` functionality out of `database_setup.py` into it's own script, `add_ccpo_users.py`
4. Rename `database_setup.py` to `create_database_user.py`, to better reflect what it actually does now.

These changes make for cleaner, simpler scripts that better focus on a single task.

I've  opened separate story [AT-5254](https://ccpo.atlassian.net/browse/AT-5254) to update the Terraform/deploy process to ensure that the new `create_database_user.py` and `reset_database.py` scripts are run separately with the appropriate user info.

Existing deployments in Azure won't be affected by this change and will need to wait for the completion of AT-5254 to be redeployed. In the mean time you can simply manually alter the function so that the owner is `atat`


dandds:
@carleton-dds This looks good. Seeing it in code makes me realize that adding the list of CCPO users is probably outside the scope of work for this script too. (They're entries in the users table, not Postgres users.) That would probably be better handled in another script, possible by adding it to `script/seed_roles.py` and updating the name to something more generic. Would you want to handle that on this ticket, or should I open another?

carleton-dds:
@dandds I'll take care of it here just for expediency's sake.

carleton-dds:
@dandds Do I have to worry about Terraform configs breaking if I rename any of the scripts in this PR? Should that kind of renaming be left to the Terraform PR?

dandds:
@carleton-dds Thanks! And it will be a breaking change with the code in the TF repo, but I think it's best to handle it in the follow-up ticket. The way we call the scripts will probably change and the rename can be updated there at the same time.

carleton-dds:
@dandds I just want to clarify that you're suggesting I rename the scripts as needed here rather than in the follow-up ticket. I wasn't sure when you said:
>  I think it's best to handle it in the follow-up ticket.
If you were referring to the renaming of the scripts, or updating terraform.

dandds:
@carleton-dds Sure. I'm suggesting you rename the Python scripts as-needed here, and we'll correct the TF process to match the new names in the follow-on ticket.

PR #1522, opened 2020-04-21
Update K8s README for adding Lets Encrypt certs.

dandds:
Add correct instructions for setting new certificates.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -127,19 +127,15 @@ This will start new pods for the web service, and the new ACME challenge will be
 
 Repeat this process for the second domain. If the validation is successful, certbot will write new certificates to your host machine.
 
-### Create the TLS secret
+### Create the Key Vault certificate object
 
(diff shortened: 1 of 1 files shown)


deploy/README.md
```diff
@@ -127,19 +127,15 @@ This will start new pods for the web service, and the new ACME challenge will be
...
-kubectl -n atat delete secret azure-atat-code-mil-tls
+az keyvault certificate import --vault-name my-vault-name --name my-cert-object-name --file /path/to/my/cert.pem
 ```
 
-Then:
-
-```
-kubectl -n atat create secret tls azure-atat-code-mil-tls --key="[path to the private key]" --cert="[path to the full chain]"
-```
+Make sure that the vault name and cert name match the ones used in the FlexVol integration.
```
okaycj-atat, line 138: Do we have docs on FlexVol?  Could we link to them from here?

dandds: Sure! That's discussed further down in the READE, so I amended this sentence to make that clear.

PR #1523, opened 2020-04-21
[AT-5209] Remove unused Azure SDK packages and refactor test mocking

graham-dds:
Closes [AT-5209](https://ccpo.atlassian.net/browse/AT-5209)

This PR's diff is pretty big because it covers a lot of ground in the files that it touches. For review, I'd suggest stepping through each commit individually first. I'm also happy to split this up into separate PRs if that makes things easier.

The first part of this ticket asked to remove unused imports from the Azure SDK. This all happened here https://github.com/dod-ccpo/atst/commit/d64fff3393cf70dcad16a407781bde86c59a4370

The second part of the ticket asked to assess whether the `AzureSDKProvider` pattern was still necessary:

> The AzureSDKProvider was initially conceived of as a means to provider the Azure interface all the Azure SDK module dependencies it needs via dependency injection so that those could be mocked for testing. Since many of them will  be removed, a developer working on this issue should also assess whether the AzureSDKProvider is still necessary.

I think for now, this pattern still suits our needs. While some packages were removed, many still remain and are used frequently in our API calls -- many of the remaining packages are used for authentication / authorization. If we decide to fully retire the SDK in favor of our own API built entirely with requests, we could probably get away with removing this dependency injection pattern in favor of mocking `requests` with a library like [`responses`](https://github.com/getsentry/responses).

Additionally, I noticed that we weren't taking full advantage of our dependency injection pattern because we frequently patched the same functions in our unit tests. In https://github.com/dod-ccpo/atst/pull/1523/commits/dc8eddaf025af72ad9aa7db759512a2c99454dc3, https://github.com/dod-ccpo/atst/pull/1523/commits/82910fb4f0c3cc4d015058afd767ae7a7aee67dd, and https://github.com/dod-ccpo/atst/pull/1523/commits/b416fb80a835b46d3648d1e681fb4cd86c3c4227, I tried to move some of this patching behavior to the `mock_azure` fixture. Now, we pretty much only mock / patch when we want to simulate a failure case and raise an exception.


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -136,18 +136,15 @@ def __init__(self):
         import adal
         import requests
 
-        self.subscription = subscription
-        self.policy = policy
         self.managementgroups = managementgroups
-        self.authorization = authorization
-        self.adal = adal
         self.graphrbac = graphrbac
```
dandds, line 140: It looks like this one is only being used in an internal method that actually isn't called anywhere, `_get_management_service_principal`. Could you remove both?

graham-dds: Good call https://github.com/dod-ccpo/atst/pull/1523/commits/d64fff3393cf70dcad16a407781bde86c59a4370

PR #1524, opened 2020-04-22
[AT-5199] Add hybrid config section

graham-dds:
Closes: [AT-5199](https://ccpo.atlassian.net/browse/AT-5199)

Adds a `[hybrid]` section to the `base.ini` configuration file. This section contains configuration options that used only by the `hybrid` CSP. These configuration options should only be added when `CSP=hybrid`.


atat/app.py
```diff
@@ -262,6 +262,10 @@ def make_config(direct_config=None):
     config.set("default", "REDIS_URI", redis_uri)
     config.set("default", "BROKER_URL", celery_uri)
 
+    if config.get("default", "CSP") == "hybrid":
+        for option, value in dict(config.items("hybrid")).items():
+            config.set("default", option, value)
```
dandds, line 267: Can you just use `config.update` here instead of looping each one?

atat/app.py
```diff
@@ -262,6 +262,10 @@ def make_config(direct_config=None):
...
+
```
dandds, line 268: Larger concern: If this new code is placed at the end of the function, I don't think we'll be able to override these settings with file-based values or environment variables. If you move this to roughly line 214, I think it will work.

atat/app.py
graham-dds: I was looking for some update-like method in the [configparser API](https://docs.python.org/3.7/library/configparser.html), but couldn't find one. 

That said, I can make this slightly more readable by doing:
```
        for option, value in config.items("hybrid"):
            config.set("default", option, value)
```

atat/app.py
graham-dds: Good call. Thinking more about it, I think need to do this `hybrid -> default` in two places: once after loading the base and env files, and possibly again after applying the `direct_config`.

atat/app.py
graham-dds: I think `config.read_dict()` ended up behaving like `update`, so I ended up going with that!

graham-dds:
@jesse-dds thanks for the approval! I rebased on staging and squashed some commits.

PR #1528, opened 2020-04-28
adds ansible terraform

mikepw:


ansible/roles/azure_keyvault/lookup_plugins/azure_keyvault_list_secrets.py
```diff
@@ -0,0 +1,177 @@
...
+token_params = {
+    'api-version': '2018-02-01',
+    'resource': 'https://vault.azure.net'
+}
+token_headers = {
+    'Metadata': 'true'
+}
+token = None
+try:
+    token_res = requests.get('http://169.254.169.254/metadata/identity/oauth2/token', params=token_params, headers=token_headers)
```
dandds, line 102: What IP is this? Is this something that should be passed in as config, or is their a corresponding domain?

ansible/roles/azure_keyvault/lookup_plugins/azure_keyvault_secret.py
```diff
@@ -0,0 +1,172 @@
...
+        from azure.common.credentials import ServicePrincipalCredentials
+        from azure.keyvault import KeyVaultClient
+        from msrest.exceptions import AuthenticationError, ClientRequestError
+        from azure.keyvault.models.key_vault_error import KeyVaultErrorException
+    except ImportError:
+        raise AnsibleError('The azure_keyvault_secret lookup plugin requires azure.keyvault and azure.common.credentials to be installed.')
+
+    client_id = kwargs.pop('client_id', None)
+    secret = kwargs.pop('secret', None)
+    tenant_id = kwargs.pop('tenant_id', None)
```
dandds, line 127: Is this a preferred pattern with Ansible functions, where you pop kwargs off instead of explicitly defining them in the function signature?

ansible/README.md
```diff
@@ -0,0 +1,16 @@
+# Intro
+
+Ansible is a widely used, first class config as code tool. Its an idempotent, stateless way to make source controlled, declaritive, repeatable changes to config. yay.
+
+
+## Requirements
+
+* Python 3.7
+* pip
```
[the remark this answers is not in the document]
mikepw: I think [this link](https://github.com/python-poetry/poetry/blob/master/tests/utils/fixtures/setups/ansible/setup.py) proves its able to be managed by poetry

ansible/roles/azure_keyvault/lookup_plugins/azure_keyvault_list_secrets.py
mikepw: Good eye. This is local IP, reserved IP space that cloud providers use to collect metadata about the resource from which this request is executed.  Here are the docs:

[Azure](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service)
[AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html)

ansible/roles/azure_keyvault/lookup_plugins/azure_keyvault_secret.py
mikepw: This is a plugin I didn't write. Happy to modify further to satisfy any preferred patter we like!

ansible/README.md
```diff
@@ -0,0 +1,16 @@
...
+* ansible
+
+## Usage
+
+I won't recreate the docs, but an example ansible run of something that affects config/runs a playbook looks like:
```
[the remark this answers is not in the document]
mikepw: The updated readme has some examples.  It mentions some ops usage examples.  But in short I'll answer here, and if you feel the README isn't explicative enough I'll add more specifics.  

Config values are versioned in an encrypted-store-vendor agnostic way by using a deploy tag as the key. To update a value, one has a few choices (more as needed for convenience):
1) Add the variable naturally a local variables.tf file and then run Ansible to parse and populate vault.
2) Use the csp cli to pull down the config value for deploy tag or commit id, modify and atomically push back to encrypted store.
3) Use the Ansible task to pull variable values by deploy tag or commit id to a json file, modify and add back using the cli.

This will evolve to a more and more convenient flow/set of options.  Most importantly it supports natural TF development.

mikepw:
separating TF and ansible into two diff PRs by dan's request

PR #1529, opened 2020-04-28
[AT-5243] Add hidden input for revoking env access form

graham-dds:
Addresses [AT-5243](https://ccpo.atlassian.net/browse/AT-5243)

Revoking environment access uses a form with checkbox input. When an
environment was revoked, the "checkbox" would be checked, while the
other environments' checkboxes were left unchecked. This means that when
the form was POSTed to the app, only the revoked, i.e. checked
envionments were included in the form data, since unchecked checkboxes
aren't included in form data by default. This mismatch caused issues when
we went to build the form with the POSTed form data and existing environment
data.

To overcome this, when a checkbox is unchecked, we should include a
hidden input with a falsy value for the form field. This way, the number
of environments and input values for those environments will be equal,
and WTForms will be able to process that input accurately.

### DIFF
--- a/templates/applications/fragments/member_form_fields.html
+++ b/templates/applications/fragments/member_form_fields.html
@@ -28,6 +28,7 @@
           <div class="form-col form-col--third">
             {{ sub_form.disabled(id=id, checked=True, **{"v-model": "isChecked"}) }}
             {{ sub_form.disabled.label(for=id, class="usa-button button-danger-outline") | safe }}
+            <input type="hidden" name={{sub_form.disabled.name}} value="">
           </div>
         </div>
         <div v-else>


JRNPW:
Which ticket are you addressing with this, @graham-dds ?

graham-dds:
> Which ticket are you addressing with this, @graham-dds ?

[AT-5243](https://ccpo.atlassian.net/browse/AT-5243). Just updated the title and description to include that info.

graham-dds:
@JRNPW https://app.ghostinspector.com/tests/5deab03e5a3744018445439b should be modified to cover this bug from a testing perspective. We revoke multiple environments and/or not the first environment and assert that the correct environments were were revoked

JRNPW:
Graham and I discussed that this change, as currently implemented, will best be tested by a change to the "Revoke Environment Access" Ghost Inspector test. We may reconsider how to test if there is a change to the way to implement this fix.

PR #1530, opened 2020-04-29
Add README details about how to configure the application.

dandds:
Small change to the README to describe how config is currently applied and the various ways it can be set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,22 @@ To generate coverage reports for the Javascript tests:
 
 ## Configuration
 
+### Setting Configuration
+
+All config settings must be declared in "config/base.ini", even if they are null. Configuration is set in the following order:
+
+1. Settings from "config/base.ini" are read in and applied.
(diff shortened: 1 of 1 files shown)


dandds:
@carleton-dds Yeah, I think I would put that in the configuration section per-setting. 

One thing I'm struggling with is that some of that info (like AZURE_TENANT_ADMIN_PASSWORD) lives in our password manager _and_ Key Vault right now. We source it from the password manager when we need it to run the hybrid tests locally, but the app sources it from the Key Vault. We could transition to source it from KV, which might cut down on duplication. I am wondering if it's worthwhile to have some kind of wrapper script that can just pull the values you need and inject them into your environment when you run the hybrid tests, provided that you authenticate and have the right perms.

carleton-dds:
@dandds A wrapper script like you describe sounds like it could be really helpful. I agree with the tension around having multiple places where config values are at rest is not great and my Google doc is also kind of exacerbating that issue. I can type up an issue if you think it'd be helpful

dandds:
@carleton-dds The Google Doc is a big improvement over what we had, imo. Would you mind adding a ticket if you have time?

carleton-dds:
@dandds  Issue for wrapper script https://ccpo.atlassian.net/browse/AT-5287

PR #1536, opened 2020-05-06
Use REST calls for GETting management groups

graham-dds:


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -348,10 +339,18 @@ def _create_management_group(
...
+        sp_token = self._get_tenant_principal_token(self.tenant_id)
+        headers = {
+            "Authorization": f"Bearer {sp_token}",
+        }
+        response = self.sdk.requests.get(
+            f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementGroups/{management_group_name}?api-version=2020-02-01",
+            headers=headers,
+        )
+        response.raise_for_status()
+        return response.json()
```
dandds, line 353: Is there actually a reason to have `_get_management_group` as an internal method? Should we just collapse it into `create_initial_mgmt_group_verification` ?

graham-dds: Yeah I was on the fence about this. I guess there isn't, but I was thinking there _might_ be. But until that day comes, I suppose it's fine to just collapse it into the verification stage.

PR #1537, opened 2020-05-06
Create user using email invitation

jesse-dds:
Replace the user creation via API with member invitations via email.

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1,6 +1,4 @@
-import adal
 import json
-import requests
 
 from flask import current_app as app
 from functools import wraps
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1113,7 +1113,7 @@ class AzureCloudProvider(CloudProviderInterface):
         msal_app = self.sdk.msal.ConfidentialClientApplication(
             self.client_id, self.secret_key, authority=authority
         )
-        scope = "https://graph.microsoft.com/.default"
+        scope = f"{self.sdk.cloud.endpoints.microsoft_graph_resource_id}/.default"
         token_response = msal_app.acquire_token_for_client(scope)
         token = token_response.get("access_token")
 
@@ -1124,7 +1124,9 @@ class AzureCloudProvider(CloudProviderInterface):
 
         # Send request
 
-        endpoint = "https://graph.microsoft.com/v1.0/invitations"
+        endpoint = (
+            f"{self.sdk.cloud.endpoints.microsoft_graph_resource_id}/v1.0/invitations"
+        )
         response = self.sdk.requests.post(
             endpoint, data=json.dumps(body), headers=headers
         )
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1095,21 +1095,11 @@ class AzureCloudProvider(CloudProviderInterface):
             UserCSPResult -- a result object containing the AAD ID.
         """
 
-        graph_api = self.sdk.cloud.endpoints.microsoft_graph_resource_id
-        login_api = self.sdk.cloud.endpoints.active_directory
-
         # Request a graph api authorization token
 
-        body = {
-            "client_id": self.client_id,
-            "grant_type": "client_credentials",
-            "client_info": 1,
-            "client_secret": self.secret_key,
-            "scope": f"{graph_api}/.default",
-        }
-
-        url = f"{login_api}/{self.tenant_id}/oauth2/v2.0/token"
-        token = self.sdk.requests.post(url, data=body).json()["access_token"]
+        token = self._get_tenant_principal_token(
+            payload.tenant_id, resource=self.graph_resource
+        )
 
         # Use the graph api to invite a user
 
@@ -1121,7 +1111,7 @@ class AzureCloudProvider(CloudProviderInterface):
             "invitedUserType": "Member",
         }
 
-        url = f"{graph_api}/v1.0/invitations"
+        url = f"{self.graph_resource}/v1.0/invitations"
         headers = {"Authorization": f"Bearer {token}"}
         response = self.sdk.requests.post(url, json=body, headers=headers)
         response.raise_for_status()
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -2,7 +2,6 @@ import json
 
 from flask import current_app as app
 from functools import wraps
-from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD as cloud
 from secrets import token_hex, token_urlsafe
 from uuid import uuid4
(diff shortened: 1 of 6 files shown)


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1099,14 +1100,42 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+        # Set up bearer token header
+
+        tenant_endpoint = (
+            f"{self.sdk.cloud.endpoints.active_directory}/{self.tenant_id}"
         )
+        auth = self.sdk.adal.AuthenticationContext(tenant_endpoint)
+        resource = self.sdk.cloud.endpoints.microsoft_graph_resource_id
+        token = auth.acquire_token_with_client_credentials(
+            resource, self.client_id, self.secret_key
+        )
```
graham-dds, line 1123: Is this doing the same thing as [`_get_service_principal_token`](https://github.com/dod-ccpo/atst/blob/53e0c8ff93c9ccab60e70375732a6b8133ea31da/atat/domain/csp/cloud/azure_cloud_provider.py#L1239:L1256)?

tests/domain/cloud/test_azure_csp.py
```diff
@@ -1262,12 +1262,16 @@ def test_update_active_directory_user_password_profile(
 
 
 def test_create_user(mock_azure: AzureCloudProvider):
+    class AuthenticaionContext(object):
+        def __init__(*a, **kw):
+            pass
```
graham-dds, line 1267: If you're able to use one of the Azure Cloud Provider private token-getting methods, you might not need to set up this application context to mock getting a token. https://github.com/dod-ccpo/atst/blob/53e0c8ff93c9ccab60e70375732a6b8133ea31da/tests/mock_azure.py#L77

atat/domain/csp/cloud/azure_cloud_provider.py
dandds: Yeah, I think we can just use `token = self._get_tenant_principal_token(payload.tenant_id, resource=self.sdk.cloud.endpoints.microsoft_graph_resource_id)`. That gets the creds for the relevant app registration and then calls `_get_service_principal_token`. 

The code here is authenticating against ATAT's home tenant. This will appear to work with the hybrid interface, but won't work in production. In prod, we need to auth against the portfolio's tenant (specifically, the one defined by `payload.tenant_id` in this case).

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1099,14 +1100,42 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+            "invitedUserEmailAddress": payload.email,
+            "inviteRedirectUrl": "https://portal.azure.com",
+            "sendInvitationMessage": True,
+            "invitedUserType": "Member",
+        }
+
+        # Set up bearer token header
+
+        tenant_endpoint = (
+            f"{self.sdk.cloud.endpoints.active_directory}/{self.tenant_id}"
```
dandds, line 1117: I think we want to use `payload.tenant_id` here. If this were production, we would be authenticating against the portfolio's tenant.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1099,14 +1100,42 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+        headers = {
+            "Authorization": f'Bearer {token["accessToken"]}',
+            "Content-Type": "application/json",
+        }
+
+        # Send request
 
-        result = self._create_active_directory_user(graph_token, payload)
-        self._update_active_directory_user_email(graph_token, result.id, payload)
+        endpoint = "https://graph.microsoft.com/v1.0/invitations"
```
dandds, line 1132: Can we get the appropriate host from `self.sdk.cloud.endpoints` like we do for other MSFT URLs?

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1099,14 +1100,42 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+
+        # Create invitation body
+
+        body = {
+            "invitedUserDisplayName": payload.display_name,
+            "invitedUserEmailAddress": payload.email,
+            "inviteRedirectUrl": "https://portal.azure.com",
+            "sendInvitationMessage": True,
+            "invitedUserType": "Member",
+        }
```
dandds, line 1112: We might want to consider updating the `UserCSPPayload` model, since it now defines a lot of fields we're not using.

dandds: Also, we should explore with CCPO whether they want to configure the [`invitedUserMessageInfo`](https://docs.microsoft.com/en-us/graph/api/resources/invitation?view=graph-rest-1.0).

tests/domain/cloud/test_azure_csp.py
```diff
@@ -1262,12 +1262,16 @@ def test_update_active_directory_user_password_profile(
 
 
 def test_create_user(mock_azure: AzureCloudProvider):
+    class AuthenticaionContext(object):
```
dandds, line 1265: Small typo: `Authenticaion`.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1093,14 +1096,38 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+
+        # Set up bearer token header
+
+        authority = f"{self.sdk.cloud.endpoints.active_directory}/{self.tenant_id}"
+        msal_app = self.sdk.msal.ConfidentialClientApplication(
```
carleton-dds, line 1113: As I mentioned in standup my ticket has moved from replacing adal with msal to replacing adal with direct rest calls, here's an example. 06b98ef

apologies for the runaround, I'm happy to push the needed changes

jesse-dds: No worries! I can do it, the example looks sufficient thanks 🙏

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1093,14 +1094,39 @@ def create_user(self, payload: UserCSPPayload) -> UserCSPResult:
...
+        body = {
+            "client_id": self.client_id,
+            "grant_type": "client_credentials",
+            "client_info": 1,
+            "client_secret": self.secret_key,
+            "scope": f"{graph_api}/.default",
+        }
+
+        url = f"{login_api}/{self.tenant_id}/oauth2/v2.0/token"
+        token = self.sdk.requests.post(url, data=body).json()["access_token"]
```
dandds, line 1112: Individual methods shouldn't be doing their own authentication; they should rely on the implementation in `_get_tenant_principal_token`. Here you could do:

```
        token = self._get_tenant_principal_token(
            payload.tenant_id, resource=self.graph_resource
        )
```

We shouldn't be using `self.client_id` and `self.secret_key` for authentication in this context. This method is for inviting a user to a portfolio's AAD tenant, so we want to authenticate as the app registration in that tenant. That's what `_get_tenant_principal_token` does for us.

PR #1538, opened 2020-05-06
Update uWSGI logging configuration.

dandds:
Addresses [AT-5251](https://ccpo.atlassian.net/browse/AT-5251). 

This change follows advice found here:

https://www.velebit.ai/blog/tech-blog-uwsgi-json-logging/

Any logs from the ATAT application are passed through from uWSGI
untouched. This fixes the issue where ATAT logs nested inside uWSGI logs
were output as escaped JSON strings, rather than JSON. The issue is
discussed in detail on this previous PR:

https://github.com/dod-ccpo/atst/pull/740

This commit moves the uwsgi.ini file to the deploy/azure directory and
uses a [kustomize configMapGenerator](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#configmapgenerator)
to create the ConfigMap. This way the config can be a regular INI file.
I've also symlinked it into the repo root so that it's available for the
Docker build. We may want to use a configMapGenerator for the NGINX
config in the future.

Notes:

- The linked blog post recommends using an additional, custom uWSGI
  plugin to escape the host and uri values, which are two fields that
  may not be JSON-escaped. I tried doing this but our Alpine base image
  does not have gcc available for compiling C code, so commands like
  `uwsgi --build-plugin escape_json.c` fail. I did not think it was
  worth the effort to resolve that issue right now.
- The above point does mean that our use of the host field in uWSGI logs
  may break the JSON encoding, but the risk seems minimal. I would
  rather solve the problem when it appears then oversolve now.
- This commit also rearranges the uWSGI logs slightly so that "source"
  appears as the first property.
- I've updated the environment variables required to run the ATAT
  container in the docker-compose.yml file.

You could test the change by running `docker-compose up --build` (it will take 7 or 8 minutes). See the notes [here](https://github.com/dod-ccpo/atst#docker-build). Once the build finishes, the app will be running on localhost:8080. You can interact with the site and watch the log output. I've also deployed the uWSGI config change to staging to test it out there.


dandds:
I'm actually not sure what in a valid URI would break JSON string encoding rules:

okaycj-atat:
> Anyone have any ideas?

I might be mistaken, but I don't think the article's author has a valid concern.  My thinking is that if:

1. All Hosts/URI are encoded as JSON Strings
2. All JSON Strings are wrapped in quotes
3. All Host/URI can't contain quotes

Then there's no issue.  What am I missing?

graham-dds:
> The blog post says
> 
> > The problem is that host and uri variables can contain characters that would mess-up JSON format.
> 
> but I'm actually not sure what in a valid URI would break JSON string encoding rules:
> 
> https://tools.ietf.org/html/rfc3986#section-2 and https://stackoverflow.com/a/1547940
> 
> Anyone have any ideas?

🤷‍♂️ The only thing I could think of would be if special characters somehow weren't escaped.

mikepw:
can anyone else build/run locally? I'm getting a 500

graham-dds:
I was able to get it up and running, but I do get the occasional 502 when I navigating around 🤔. Usually refreshing gets me back though

mikepw:
@graham-dds I didn't run the setup script

PR #1543, opened 2020-05-08
Fix aks subnet route table

mikepw:


mikepw:
ahh sorry.. ya will run `tf fmt` np! 

Are the AD and KV changes relevant to the route table fix?
  * nooo I'll back them out of this PR!
Are we just hard-coding the route table name because it can't be accessed as TF output?
  * at best, w/o re-structuring this project, we can make it a var.  This is a terrible bandaid and requires the operator to effectively mute the aks route table association, exec the entire infra, then re-exec only the aks route table association resource.
 How would we handle that when we deploy prod?
 * Above steps would work (as var w/ stepped executions)

mikepw:
closing this one in favor of fix_aks-routes

PR #1548, opened 2020-05-11
Remove ADAL library

carleton-dds:
[AT-5265](https://ccpo.atlassian.net/browse/AT-5265)

This story was originally intended to replace the ADAL library for obtaining tokens with the MSAL library, as the ADAL library is being phased out. However, after discussing it with Dan we decided to just rely on direct API calls instead of any library, to keep within the larger trend we've been following favoring removal of sdk libraries. I also updated the teardown script appropriately.

This change means we using very similar code to obtain a `client_credentials` token for different resources twice, and I briefly considered abstracting that up into a new method, but decided to leave it as is for the time being, less we become too tightly coupled, but I'm open to updating it if that's what's best.

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1013,7 +1013,7 @@ class AzureCloudProvider(CloudProviderInterface):
         """
 
         graph_token = self._get_tenant_principal_token(
-            payload.tenant_id, resource=self.graph_resource
+            payload.tenant_id, scope=self.graph_resource + "/.default"
         )
 
         # Step 1: Create an AAD identity for the user
@@ -1090,8 +1090,8 @@ class AzureCloudProvider(CloudProviderInterface):
 
         # Request a graph api authorization token
 
-        token = self._get_tenant_principal_token(
-            payload.tenant_id, resource=self.graph_resource
+        graph_token = self._get_tenant_principal_token(
+            payload.tenant_id, scope=self.graph_resource + "/.default"
         )
 
         # Use the graph api to invite a user
@@ -1105,7 +1105,7 @@ class AzureCloudProvider(CloudProviderInterface):
         }
 
         url = f"{self.graph_resource}/v1.0/invitations"
-        headers = {"Authorization": f"Bearer {token}"}
+        headers = {"Authorization": f"Bearer {graph_token}"}
         response = self.sdk.requests.post(url, json=body, headers=headers)
         response.raise_for_status()
 
@@ -1240,21 +1240,23 @@ class AzureCloudProvider(CloudProviderInterface):
 
     @log_and_raise_exceptions
     def _get_service_principal_token(
-        self, tenant_id, client_id, secret_key, resource=None
+        self, tenant_id, client_id, secret_key, scope=None
     ):
-        url = f"{self.sdk.cloud.endpoints.active_directory}/{tenant_id}/oauth2/token"
-        resource = resource or self.sdk.cloud.endpoints.resource_manager
+        url = (
+            f"{self.sdk.cloud.endpoints.active_directory}/{tenant_id}/oauth2/v2.0/token"
+        )
+        payload_scope = scope or self.sdk.cloud.endpoints.resource_manager + ".default"
         payload = {
             "grant_type": "client_credentials",
             "client_id": client_id,
             "client_secret": secret_key,
-            "resource": resource,
+            "scope": payload_scope,
         }
-        token_response = self.sdk.requests.get(url, data=payload, timeout=30)
+        token_response = self.sdk.requests.post(url, data=payload, timeout=30)
         token_response.raise_for_status()
         token = token_response.json().get("access_token")
         if token is None:
-            message = f"Failed to get service principal token for resource '{resource}' in tenant '{tenant_id}'"
+            message = f"Failed to get service principal token for scope '{payload_scope}' in tenant '{tenant_id}'"
             app.logger.error(message, exc_info=1)
             raise AuthenticationException(message)
         else:
@@ -1290,13 +1292,13 @@ class AzureCloudProvider(CloudProviderInterface):
             "root_tenant_id": self.root_tenant_id,
         }
 
-    def _get_tenant_principal_token(self, tenant_id, resource=None):
+    def _get_tenant_principal_token(self, tenant_id, scope=None):
         creds = self._source_tenant_creds(tenant_id)
         return self._get_service_principal_token(
             creds.tenant_id,
             creds.tenant_sp_client_id,
             creds.tenant_sp_key,
-            resource=resource,
+            scope=scope,
         )
 
     @log_and_raise_exceptions
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1371,26 +1371,11 @@ class AzureCloudProvider(CloudProviderInterface):
         if result.ok:
             return CostManagementQueryCSPResult(**result.json())
 
-    @log_and_raise_exceptions
-    def _get_calculator_creds(self):
-        url = f"{self.sdk.cloud.endpoints.active_directory}/{self.root_tenant_id}/oauth2/token"
-        resource = self.config.get("AZURE_CALC_RESOURCE")
-        payload = {
-            "grant_type": "client_credentials",
-            "client_id": self.config.get("AZURE_CALC_CLIENT_ID"),
-            "client_secret": self.config.get("AZURE_CALC_SECRET"),
-            "resource": resource,
-        }
-        token_response = self.sdk.requests.get(url, data=payload, timeout=30)
-        token_response.raise_for_status()
-        token = token_response.json().get("access_token")
-        if token is None:
-            message = f"Failed to get token for resource '{resource}' in tenant '{self.root_tenant_id}'"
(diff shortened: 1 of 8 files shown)


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1233,15 +1226,17 @@ def _get_root_provisioning_token(self):
     def _get_service_principal_token(
         self, tenant_id, client_id, secret_key, resource=None
     ):
-        context = self.sdk.adal.AuthenticationContext(
-            f"{self.sdk.cloud.endpoints.active_directory}/{tenant_id}"
-        )
+        url = f"{self.sdk.cloud.endpoints.active_directory}/{tenant_id}/oauth2/token"
```
dandds, line 1229: It looks like there's a v2.0 version of this endpoint. Do you think it's worth using that? (Is the URL you're using here implicitly 1.0?).

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1233,15 +1226,17 @@ def _get_root_provisioning_token(self):
...
-        )
-
-        token = token_response.get("accessToken")
+        payload = {
+            "grant_type": "client_credentials",
+            "client_id": client_id,
+            "client_secret": secret_key,
+            "resource": resource,
+        }
+        token_response = self.sdk.requests.get(url, data=payload, timeout=30)
```
dandds, line 1237: I'm a little surprised this is a GET. Do you have a docs reference for v1.0 of this API?

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1385,17 +1383,21 @@ def get_reporting_data(self, payload: CostManagementQueryCSPPayload):
 
     @log_and_raise_exceptions
     def _get_calculator_creds(self):
```
dandds, line 1385: Do you think it's worth deleting this method and adding a `_get_service_principal_token` call to the body of `get_calculator_token`?

```
token = self._get_service_principal_token(
    self.tenant_id, 
    self.config.get("AZURE_CALC_CLIENT_ID"), 
    self.config.get("AZURE_CALC_SECRET"), 
    resource=self.config.get("AZURE_CALC_RESOURCE")
)
```

poetry.lock
```diff
@@ -4,7 +4,7 @@ description = "The ADAL for Python library makes it easy for python application
 name = "adal"
 optional = false
 python-versions = "*"
-version = "1.2.2"
+version = "1.2.3"
```
dandds, line 7: I see you dropped ADAL from the pyproject.toml, so I guess this means some of the other pieces of the SDK still rely on it.

script/teardown_hybrid.py
```diff
@@ -44,15 +43,23 @@ def list_app_registrations(token):
...
+        "resource": resource,
+    }
+    token_response = requests.get(url, data=payload, timeout=30)
+    token_response.raise_for_status()
+    token = token_response.json().get("access_token")
+    if token is None:
+        message = f"Failed to get user principal token for resource '{resource}' in tenant '{tenant_id}'"
+        raise RuntimeError(message)
+    else:
+        return token
```
dandds, line 62: I wonder if it would be worth moving `_get_service_principal_token` into a function in a utils module somewhere so we could just import for things like this. What do you think?

atat/domain/csp/cloud/azure_cloud_provider.py
dandds: Forgot, here's the docs page where I saw v2.0:  https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#get-a-token

poetry.lock
graham-dds: Looks like it's a dependency of msrestazure

`poetry show --tree`
```
...
msrestazure 0.6.3 AutoRest swagger generator Python client runtime. Azure-specific module.
├── adal >=0.6.0,<2.0.0
│   ├── cryptography >=1.1.0 
...
```

tests/mock_azure.py
```diff
@@ -123,6 +107,23 @@ def mock_azure(monkeypatch):
...
+    )
+    monkeypatch.setattr(
+        azure_cloud_provider,
+        "_get_user_principal_token_for_resource",
+        Mock(return_value=MOCK_ACCESS_TOKEN),
+    )
+    monkeypatch.setattr(
+        azure_cloud_provider,
+        "_get_calculator_creds",
+        Mock(return_value=MOCK_ACCESS_TOKEN),
```
graham-dds, line 127: I noticed that all of these monkeypatched methods end up hitting `f"{self.sdk.cloud.endpoints.active_directory}/<some tenant id>/oauth2/token"`.  Would it be good to patch at the `requests` level rather than patching these private methods? Maybe something similar to [this](https://stackoverflow.com/a/38799967) method? The less we can rely on patching, the better.

poetry.lock
carleton-dds: Yup, I mentioned that in the commit message, but I probably should have called it out better in the PR description

dandds: Ah, my b, I never spend as much time in the commits as I'd like.

carleton-dds:
@okaycj-atat can you talk a little more about what your envisioning beyond the current [log_and_raise_exceptions](https://github.com/dod-ccpo/atst/blob/33341013f35d9aac1244a9985dd31c2a49b42dd6/atat/domain/csp/cloud/azure_cloud_provider.py#L87) wrapper?

graham-dds:
@okaycj-atat also interested in hearing more -- there's [this card](https://ccpo.atlassian.net/browse/AT-5248) which might address some of the things you were thinking about?

okaycj-atat:
@carleton-dds That was a terrible suggestion.  My apologies.

atat/domain/csp/cloud/azure_cloud_provider.py
carleton-dds: Good call, here's the update to 2.0. I'll do the same for `_get_user_principal_token_for_resource` too https://github.com/dod-ccpo/atst/pull/1548/commits/b093b2fd19696aeb1895165ed379c941c73caa11

atat/domain/csp/cloud/azure_cloud_provider.py
carleton-dds: As mentioned in prevous previous comment, update to 2.0: https://github.com/dod-ccpo/atst/commit/b093b2fd19696aeb1895165ed379c941c73caa11

script/teardown_hybrid.py
carleton-dds: I do think having a more generic `_get_service_principal_token` in which we inject all the config values needed would solve this problem as well as your [previous comment](https://github.com/dod-ccpo/atst/pull/1548#discussion_r423335097) regarding `_get_calculator_creds`. It was something I toyed with will working this issue, but at a certain point I just wanted to get a complete version ready for feedback. I can start tackling this now though

atat/domain/csp/cloud/azure_cloud_provider.py
carleton-dds: See [my comments below](https://github.com/dod-ccpo/atst/pull/1548#discussion_r424725426), we may have a two birds/one stone situation

script/teardown_hybrid.py
carleton-dds: @dandds  Several changes, some which include:
- I've [moved _get_user_principal_token_for_scope to a utils file](https://github.com/dod-ccpo/atst/pull/1548/commits/2f449cce1705efd8af8cbbbe47eec3195b6ec404)
- One issue with the newly structured `_get_user_principal_token_for_scope` is that I couldn't figure out how to test it properly and had to [remove its current test](https://github.com/dod-ccpo/atst/pull/1548/commits/a4418dfcc347db2cb2d1f3657e3c97ddc1343454). Any suggestions to fixing that I'm open to.
- I updated the teardown script to [use the updated method instead of making its own API calls](https://github.com/dod-ccpo/atst/pull/1548/commits/c731457f602c8706632e837607c78742a34f1910). 
- I also updated this method as well as `get_service_principal_token` to [use 2.0 endpoints](https://github.com/dod-ccpo/atst/pull/1548/commits/443802ed0c9dd3283946f51c38a4cb07966a76c5)
- Updated [_get_calculator_url](https://github.com/dod-ccpo/atst/pull/1548/commits/20067e27b979919babc4acaf5e358e68437a3d5d) to use `get_service_principal_token` instead of making its own token call

tests/mock_azure.py
carleton-dds: @graham-dds I agree that mocking is preferable to monkeypatching. I was actually  able to remove some of the monkeypatching around the `calc` methods thanks to some other changes I was making. However, I'm still monkeypatching [the following methods](https://github.com/dod-ccpo/atst/blob/a4418dfcc347db2cb2d1f3657e3c97ddc1343454/tests/mock_azure.py#L74-L89):
- `get_secret`
- `_get_keyvault_token`
- `_get_service_principal_token`

My reasoning for this is many tests call these methods in the process of testing other functionality, so instead of updating 20+ tests to mock these methods reapeatedly, I have a monkeypatched version of these methods be the "default" setting for `mock_azure` and pull in the real versions for the 1 or 2 tests that actually need to test those methods. Does that seem reasonable, or should I just go ahead with the mocking them out in all the tests that need it?

PR #1553, opened 2020-05-14
Expose build info

okaycj-atat:
To expose the build information (both the application version and git commit sha) I added two environment variables `APP_VERSION` and `GIT_SHA`.  Until we have an official app version set, I've hard coded the version in base.ini.  Circle CI has been configured to add it's environment variable `CIRCLE_SHA1` as `GIT_SHA` when the image is ran.  

Open questions:
1. Should we add a test the ghost inspector to check for this build info html comment?  Is that even possible?


config/base.ini
```diff
@@ -35,6 +36,7 @@ DEBUG_MAILER = false
 DISABLE_CRL_CHECK = false
 ENVIRONMENT = dev
 FILE_SIZE_LIMIT = 24000000
+GIT_SHA
```
jesse-dds, line 39: I don't think you need this. There's already `GITHUB_SHA` and `GITHUB_REF` in the build environment. What I think needs to happen is that you need to somehow find and replace `{{ build_info }}` in the HTML files in a "before deployment" hook if that exists.

atat/app.py
```diff
@@ -392,5 +392,9 @@ def _url_for(endpoint, **values):
             return flask_url_for(endpoint, **values)
 
     app.jinja_env.globals.update(
-        {"url_for": _url_for, "service_desk_url": app.config.get("SERVICE_DESK_URL")}
+        {
+            "url_for": _url_for,
+            "service_desk_url": app.config.get("SERVICE_DESK_URL"),
+            "build_info": f"{app.config.get('APP_VERSION')} {app.config.get('GIT_SHA', '')}",
```
jesse-dds, line 398: I don't think is going to work because the SHA information only exists at build time, this information won't be present in the host the application is running on.

okaycj-atat: I think you're correct.  I am currently working on using `--build-args` to inject this value into the environment variables.

config/base.ini
okaycj-atat: I totally missed this.  Thank you.

atat/app.py
dandds: Yeah, I think something like this strategy will work: https://vsupalov.com/docker-build-time-env-values/

templates/base.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+  <!-- {{ build_info }} -->
```
dandds, line 4: It might be nice to include a little text, like `  <!-- VERSION: {{ build_info }} -->`

atat/app.py
okaycj-atat: Using `--build-arg` doesn't work when the built image is ran on circle ci.  I resorted to using the `-e` argument on run.

config/base.ini
okaycj-atat: I was not able to find the two environment variables in the running image.  Maybe I am missing something?

.circleci/config.yml
```diff
@@ -163,6 +163,7 @@ jobs:
             docker run \
               -e PGHOST=postgres \
               -e REDIS_HOST=redis:6379 \
+              -e GIT_SHA=${CIRCLE_SHA1} \
```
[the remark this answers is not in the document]
dandds: I noticed in another comment you said that using `--build-arg` doesn't work on CircleCI. What makes you think so? Using `run` only affects the container created from the image for that single command. We need the value of the GIT_SHA persisted to the image itself so that it's available when we run new containers based on the image in Kubernetes. Using `--build-arg` is working for me locally on your branch:

```
$ docker build . --build-arg GIT_SHA=987654321 -f ./Dockerfile -t atat:latest
[a lot happens]
$ docker run -it atat:latest /bin/sh
/opt/atat/atst $ echo $GIT_SHA
987654321
```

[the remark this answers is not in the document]
dandds: Oh, I think I know what was happening: you added the test I commented on below to confirm that GIT_SHA was being set by CircleCI. The docker build takes place in two stages. We use the first stage for running tests in CI (for technical reasons) but the second stage is where GIT_SHA is set. This is why your new test fails. 

What you have in the dockerfile is correct and I wouldn't change it. I don't think you need to check that CircleCI and docker work as advertised. If you add the `--build-arg` back, we can confirm that the value is set when CD happens and the container runs in the cluster.

atat/app.py
okaycj-atat: After some conversation and explanation of multi-stage docker builds, I agree that `--build-arg` will work.

JRNPW:
Good callout about Ghost Inspector, @okaycj-atat! I'll look into that!

PR #1556, opened 2020-05-15
[AT-5261] Migrate to hybrid tenant

graham-dds:
Closes: [AT-5261](https://ccpo.atlassian.net/browse/AT-5261)

This PR reconfigures the Hybrid Cloud Provider so that integration test Azure resources are provisioned in the new tenant created solely for hybrid tests.

In making this migration, I also modified some aspects of the Azure Cloud Provider, including the Billing Owner, Initial Management Group Verification, and Admin Role Definition steps.


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -971,19 +972,21 @@ def create_admin_role_definition(self, payload: AdminRoleDefinitionCSPPayload):
         response.raise_for_status()
 
         result = response.json()
-        roleList = result.get("value")
-
-        DEFAULT_ADMIN_RD_ID = "794bb258-3e31-42ff-9ee4-731a72f62851"
```
graham-dds, line 976: I found that this ID is not constant across tenants -- it's different in the hybrid. Because of this, we shouldn't rely on a default being present and should bail out of provisioning if something goes wrong here

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1052,11 +1104,7 @@ def _assign_billing_owner_role(self, graph_token, billing_admin_role_id, user_id
 
         result = self.sdk.requests.post(url, headers=auth_header, json=request_body)
         result.raise_for_status()
-
-        if result.ok:
```
graham-dds, line 1056: We won't reach this conditional if `result.ok` is `False` because of `result.raise_for_status` above.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -255,82 +197,45 @@ def test_hybrid_user_has_tenant(self, session, csp, app_role_1, app_1, user):
...
-
         try:
             do_create_environment_role(csp, environment_role.id)
         except UserProvisioningException as e:
             if "RoleAssignmentExists" not in str(e):
                 raise e
 
 
-@pytest.mark.hybrid
-def test_create_initial_mgmt_group_verification(csp):
```
graham-dds, line 318: This test is covered by provisioning

atat/domain/csp/cloud/hybrid_cloud_provider.py
```diff
@@ -22,10 +23,12 @@ def monkeypatched(object_, name, patch):
 
 
 class HybridCloudProvider(object):
-    def __init__(self, azure: AzureCloudProvider, mock: MockCloudProvider):
+    def __init__(self, azure: AzureCloudProvider, mock: MockCloudProvider, config):
         self.azure = azure
         self.mock = mock
-        self.tenant_id = None
+        self.mock_tenant_id = str(uuid4())
```
dandds, line 29: We want the mock tenant ID to be different per call to `create_tenant`, right? It seems like this should be generated within the Hybrid's implementation of `create_tenant`. If we keep it as state on the interface instance, every portfolio the Hybrid provisions when it's deployed and running will have the same tenant ID. (I think.)

tests/domain/cloud/test_hybrid_csp.py
dandds: I think there's still value to having a separate test. The test that runs provisioning end-to-end is basically an integration test, in my opinion. This one is closer to a real unit test.

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -230,15 +181,6 @@ def test_hybrid_create_user_job(self, session, csp, app_role_1, portfolio):
 
         assert app_role_1.cloud_id
 
-    def test_hybrid_create_user_sends_email(
```
dandds, line 233: Why are we dropping this one?

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -50,7 +54,21 @@ def portfolio():
 
 @pytest.fixture(scope="function")
 def csp(app):
-    return HybridCSP(app, simulate_failures=True).cloud
+    csp = HybridCSP(app, simulate_failures=False).cloud
+
+    csp.azure.create_tenant_creds(
+        csp.mock_tenant_id,
```
dandds, line 60: It seems like you're using the stable `mock_tenant_id` as a way to consolidate test setup. That's a worthwhile goal and will work for these tests but I don't think it will work for the deployed instance of the Hybrid for the reasons described in my other comment.

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -270,9 +270,10 @@ def create_initial_mgmt_group_verification(
 
         https://docs.microsoft.com/en-us/azure/governance/management-groups/overview
         """
-        sp_token = self._get_tenant_principal_token(payload.tenant_id)
+
+        mgmt_token = self._get_elevated_management_token(payload.tenant_id)
```
dandds, line 274: Subtle change, but this means that we're using the human admin creds for this action now. I don't think you need to take any action on this PR, but it's worth noting. Our goal is to use the human admin creds for as few actions as possible, so it might be worthwhile to ensure that the app registration has appropriate perms before this point so that it can take this action instead.

atat/domain/csp/cloud/hybrid_cloud_provider.py
graham-dds: Ahh good point. I only needed this for tests (like you commented below), so I did some rearranging here https://github.com/dod-ccpo/atst/pull/1556/commits/3961d1b5cd1925d47460354b32f01a2688eb61ef

tests/domain/cloud/test_hybrid_csp.py
graham-dds: > The test that runs provisioning end-to-end is basically an integration test, in my opinion. This one is closer to a real unit test.

True, but isn't our goal with these hybrid tests to focus on integration? The only different between this test and the one in the Azure Cloud Provider unit tests is that it hits the Azure API, and we cover that in the integration test.

tests/domain/cloud/test_hybrid_csp.py
graham-dds: This is the same test as in `test_jobs.py`, and the email behavior of this job tests is not affected by the the CSP, so this felt safe to remove.

atat/domain/csp/cloud/azure_cloud_provider.py
graham-dds: 👍 When I was running through hybrid tests, this step gave me a 401 unless I used an elevated elevated access token. Maybe it's possible to add the elevated access flow to the service principal's credentials?

In any case, one thing I see missing from our setup is [removing](https://docs.microsoft.com/en-us/azure/role-based-access-control/elevate-access-global-admin#remove-elevated-access-3) this elevated access. The Azure docs call out:
> You should remove this elevated access once you have made the changes you need to make at root scope.

tests/domain/cloud/test_hybrid_csp.py
dandds: Fair point, yeah. Deleting tests makes me nervous!

atat/domain/csp/cloud/azure_cloud_provider.py
dandds: Could you add a ticket to capture this? 🙏

graham-dds: 👍 https://ccpo.atlassian.net/browse/AT-5316

Also didn't know you could filter the role assignments list: `GET https://management.azure.com/providers/Microsoft.Authorization/roleDefinitions?api-version=2015-07-01&$filter=roleName+eq+'User Access Administrator'`

This would probably allow us to refactor this pattern in the future:
```python
try:
    admin_role_def_id = next(
        (
            role.get("id")
            for role in role_list
            if role.get("displayName") == "Company Administrator"
        )
    )
    return AdminRoleDefinitionCSPResult(admin_role_def_id=admin_role_def_id)
except StopIteration:
    raise ResourceProvisioningError(
        "Azure role definition",
        "Could not locate Azure Global Admin / Company Admin role",
    )
```

PR #1559, opened 2020-05-18
Switch from *.atat.code.mil domain to atat.dev

graham-dds:
This PR addresses step 4 of `Switch sites to *.atat.dev` ([AT-5297](https://ccpo.atlassian.net/browse/AT-5297)):
> Update all Kubernetes config and application config values. You can find these by grepping for instances of the site names in the codebase.

This config has already been applied to the codebase -- we're just committing the changes to source in this PR.

In addition to updating k8s config, we also replace references to the old `*.atat.code.mil` domain in READMEs.


deploy/README.md
```diff
@@ -89,7 +89,7 @@ You may also be prompted to agree to a terms of services, which you should do.
 You will be prompted to enter the domain names you want a cert for. Enter **both** the main and auth domains as a comma separated list. For instance:
 
 ```
-jedi.atat.code.mil,jedi-auth.atat.code.mil
+atat.dev, *.atat.dev
```
dandds, line 92: Actually, @carleton-dds or @graham-dds can we update the instructions for obtaining Let's Encrypt certs now that we're doing a DNS challenge? Do you think that should happen on a separate PR? (I'm also not sure whether it needs to be documented in the codebase or elsewhere now that it happens entirely outside of the code and k8s config.)

graham-dds: > now that it happens entirely outside of the code and k8s config

Yeah if this is a case, it sounds like it would make sense to move this documentation to Confluence and maybe just drop a link to that documentation here? Also, I'd say that's enough for another PR.

dandds: We should remove those instructions here, at least, I think.

graham-dds: https://github.com/dod-ccpo/atst/pull/1559/commits/c4eb51d9e5dce2549439359250caffc5abc0c085

carleton-dds: I'll update/move over to Confluence

PR #1563, opened 2020-05-20
updates terraform azurerm provider

mikepw:
Updates the `azurerm` provider so we can use the latest released version of azure rm resources, benefit from bug fixes and improvements.


terraform/modules/k8s/main.tf
```diff
@@ -8,11 +8,11 @@ resource "azurerm_resource_group" "k8s" {
...
-  name                 = "${var.name}-${var.environment}-k8s"
-  location             = azurerm_resource_group.k8s.location
-  resource_group_name  = azurerm_resource_group.k8s.name
-  dns_prefix           = var.k8s_dns_prefix
-  private_link_enabled = var.private_link_enabled
+  name                    = "${var.name}-${var.environment}-k8s"
+  location                = azurerm_resource_group.k8s.location
+  resource_group_name     = azurerm_resource_group.k8s.name
+  dns_prefix              = var.k8s_dns_prefix
+  private_cluster_enabled = var.private_link_enabled
```
dandds, line 15: Should we change this variable name to match?

mikepw: good call. fixed

terraform/providers/pwdev/provider.tf
```diff
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version         = "=2.0.0"
+  version         = "=2.10.0"
```
jesse-dds, line 2: Is this a typo? Should this have gone from "2.0.0" to "2.1.0" or did it really move up 10 versions?

mikepw: not a typo :)
https://github.com/terraform-providers/terraform-provider-azurerm/blob/v2.10.0/CHANGELOG.md

dandds:
@mikepw I think this just needs to be rebased on staging and `script/format` to merge.

PR #1596, opened 2020-05-27
[AT-5324] Add management groups to hybrid test teardown

graham-dds:
This PR extends the hybrid teardown script by deleting management groups in addition to app registrations.

I ran this script on our tenant. It appeared to work as expected. Some management groups couldn't be deleted automatically, [because](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/delete):
> If a management group contains child resources, the request will fail.

Though I suspect some management groups with child resources were able to be deleted based on ordering after GETing all management groups. If an "application" management group was listed before it's associated "portfolio" management group, both might be able to be deleted. If the portfolio was first though, it would fail. Also, DELETEing a management groups actually kicks off an async job, so there could be race condition issues at play.

I felt that the script was safe to run on our current "hybrid" setup because we'll be switching over all hybrid operations to the new tenant by the end of [AT-5324](https://ccpo.atlassian.net/browse/AT-5324). In the new tenant though, since management groups represent portfolios, applications, and environments, deleting a management groups would be like "deleting" one of these ATAT entities from the Azure side, so to speak.

This poses an interesting question though: what happens in the ATAT UI when I try to say, add a subscription to an environment whose management group I've deleted in the Azure Portal?

### DIFF
--- a/script/teardown_hybrid.py
+++ b/script/teardown_hybrid.py
@@ -8,6 +8,7 @@ sys.path.append(parent_dir)
 from atat.app import make_config
 from atat.domain.csp.cloud.hybrid_cloud_provider import HYBRID_PREFIX
 from atat.domain.csp.cloud.utils import get_user_principal_token_for_scope
+from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD
 
 
 GRAPH_API = "https://graph.microsoft.com"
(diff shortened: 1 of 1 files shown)


dandds:
@graham-dds 

>  what happens in the ATAT UI when I try to say, add a subscription to an environment whose management group I've deleted in the Azure Portal?

For now, the answer to this is that only the portfolio point of contact (the tenant admin) is able to do that, and they shouldn't. There are going to be a lot of edge cases around disparities between ATAT and Azure that we'll have to work through post-MVP. Worth starting to document them some place.

PR #1599, opened 2020-05-28
Feature add aks bastion

mikepw:
adds aks bastion Terraform
modifies `tf_plan.sh` a little more one-liner
update of `azurerm_provider` carried over from PR `1563` that one should go first.


dandds:
@mikepw Which version of the tfvars should I be using for this?

papetti23:
Meant to drop this in the please use `DEPLOY_TAG=v0.0.5`, @dandds secret names are alpha-numeric only so before I make changes to Ansible to normalize branch names to only include [0-9a-zA-Z] I'm just patching the sem-ver.

terraform/modules/bastion/main.tf
```diff
@@ -0,0 +1,156 @@
...
+
+  name                 = "mgr-subnet"
+  resource_group_name  = var.mgmt_subnet_rg
+  virtual_network_name = var.mgmt_subnet_vpc_name
+  address_prefixes     = ["${var.mgmt_subnet_cidr}"]
+
+  enforce_private_link_endpoint_network_policies = true
+
+
+}
```
dandds, line 30: What's the difference between azure_bastion_subnet and mgmt_subnet?

terraform/modules/bastion/main.tf
```diff
@@ -0,0 +1,156 @@
...
+  name                = "jit-bastion"
+  location            = var.region
+  resource_group_name = var.rg
+
+  ip_configuration {
+    name                 = "configuration"
+    subnet_id            = azurerm_subnet.azure_bastion_subnet.id
+    public_ip_address_id = azurerm_public_ip.bastion_ip.id
+  }
+}
```
dandds, line 57: I thought we were using the AKS nodes as bastion hosts, but this is enabling the Azure bastion service. I think there's a piece I don't understand. Is the Azure bastion service what allows JIT access to the AKS VM? Could you link me to docs that explain this?

terraform/modules/bastion/main.tf
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+
+
+
+
+
+  default_node_pool {
+    name                  = "default"
+    vm_size               = "Standard_B2s"
```
dandds, line 123: Do we want the VM size variable-ized? Do you think it even needs 2vcpu/4GB?

terraform/providers/pwdev/variables.tf
```diff
@@ -4,7 +4,10 @@ variable "region" {}
 
 variable "backup_region" {}
 
-
+variable "bastion_resource_group" {}
+variable "bastion_aks_sp_secret" {}
+variable "bastion_aks_sp_id" {}
+variable "bastion_ssh_pub_key_path" {}
```
dandds, line 10: If this is a file path, does the pub key only live on your local machine right now?

terraform/modules/bastion/main.tf
mikepw: aks bastion node serial console should/could be a fallback to Azure bastion service.  Happy to put a switch on it to disable/enable bastion via parameter.  The need for bastion is for the general use case beyond AT-AT.

terraform/modules/bastion/main.tf
mikepw: B2s is the smallest aks allows.

terraform/providers/pwdev/variables.tf
mikepw: It should live in keyvault. For the demo it lived on my machine. Happy w/ next iteration or this PR for the keyvault integration.

terraform/modules/bastion/main.tf
mikepw: They're separate things, but serial console comes OOB w/ aks nodes and should be a backup for user experience reasons so I group them.  Default to bastion, fail over to serial console.  Serial console I think has a worse user experience and won't be great to use during a triage scenario. I'd spin them both up simultaneously, then spin back down when unneeded.

mikepw:
> One change please: could we get the pub key into the Key Vault like we discussed here? [#1599 (comment)](https://github.com/dod-ccpo/atst/pull/1599#discussion_r432618317)

we could .. but the it should be the operators key that gets used here. The key goes right onto the k8s node at spin up.  If additional users need access their keys should be added to `authorized_keys` not use a 'service' account key bc it makes auditing very difficult.

mikepw:
also.. this branch's rebase is currently messed up. Reworking now..

terraform/providers/pwdev/variables.tf
mikepw: I retract my 'it should live in keyvault' comment.  Operator keys should be used when spun up. Additional operator keys should be added to `authorized_keys` via Ansible or if necessary by hand.

PR #1603, opened 2020-05-28
[AT-5259] Update manual provisioning process

okaycj-atat:
Updates to the manual provisioning scripts to ensure that the mock CSP will complete without error.  

Open Questions:
- Should we extent the scope of this story to include Azure CSP?  Hybrid CSP?

To be completed in separate PR:
- Repeatable script validation
- Update documentation

[Jira Story AT-5259](https://ccpo.atlassian.net/browse/AT-5259)

### DIFF
--- a/atat/domain/csp/__init__.py
+++ b/atat/domain/csp/__init__.py
@@ -5,47 +5,34 @@ from .files import AzureFileService, MockFileService
 from .reports import MockReportingProvider
 
 
-class MockCSP:
-    def __init__(self, app, simulate_failures=False):
-        self.cloud = MockCloudProvider(
-            app.config,
-            with_delay=simulate_failures,
-            with_failure=simulate_failures,
-            with_authorization=simulate_failures,
-        )
-        self.files = MockFileService(app)
-        self.reports = MockReportingProvider()
+class CSP:
+    def __init__(self, csp, config, **kwargs):
+        azure = AzureCloudProvider(config)
+        mock = MockCloudProvider(config, **kwargs)
+        hybrid = HybridCloudProvider(azure, mock, config)
 
+        if csp == "azure":
+            self.cloud = azure
+            self.files = AzureFileService(config)
+        elif csp in ("mock-test", "mock"):
+            self.cloud = mock
+            self.files = MockFileService(config)
+        elif csp == "hybrid":
+            self.cloud = hybrid
+            self.files = AzureFileService(config)
+        else:
+            raise Exception(f"Unexpected CSP value provided: {csp}")
 
-class AzureCSP:
-    def __init__(self, app):
-        self.cloud = MockCloudProvider(app.config)
-        self.files = AzureFileService(app.config)
-        self.reports = MockReportingProvider()
-
-
-class HybridCSP:
-    def __init__(self, app, simulate_failures=False):
-        azure = AzureCloudProvider(app.config)
-        mock = MockCloudProvider(
-            app.config,
-            with_delay=simulate_failures,
-            with_failure=simulate_failures,
-            with_authorization=simulate_failures,
-        )
-        self.cloud = HybridCloudProvider(azure, mock, app.config)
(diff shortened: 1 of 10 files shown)


script/provision/a_create_tenant.py
```diff
@@ -13,21 +13,8 @@ def create_tenant(csp, inputs):
...
         **{**inputs.get("initial_inputs"), **inputs.get("csp_data")}
     )
-
-    creds = None
-
-    def get_creds(tenant_id, new_creds):
-        nonlocal creds
-        creds = new_creds
-
-    csp.update_tenant_creds = get_creds
```
dandds, line 23: It looks like this and line 26 were intended to preserve the tenant administrator credentials in the output JSON. That way the person running the scripts can log into the new tenant and perform any actions not covered by the scripts (like create the ATAT app registration, for example). I think that we need this or something akin to it.

okaycj-atat: Okay.  Thank you for knowing this.  I was going to leave it in, but lgtm.com was bouncing it back.  I'll see what I can do.

okaycj-atat: Yup, I didn't fully understand what was happening here.  After some clarification, it looks like the inputs is being written to file from within the `handle` function.

script/provision/a_create_tenant.py
```diff
@@ -22,12 +22,13 @@ def get_creds(tenant_id, new_creds):
 
     csp.update_tenant_creds = get_creds
     result = csp.create_tenant(create_tenant_payload)
-    if result.get("status") == "ok":
+
+    # Update inputs with credential information. This will be written to file inside handle().
+    # Note: The curent implementation of Mock CSP doesn't call update_tenant_creds()
+    if creds is not None:
```
jesse-dds, line 28: What is the truthy value of `creds`? Could you get away with `if creds:` here? My head always hurts with double negatives 😄

script/provision/provision_base.py
```diff
@@ -2,28 +2,48 @@
...
         def fake_source_creds(tenant_id=None):
             return KeyVaultCredentials(**creds)
 
         cloud._source_creds = fake_source_creds
 
-        return (cloud, details)
+        return cloud, details
+
+
+def get_provider(csp, config):
```
jesse-dds, line 36: This doesn't look much different from `def make_csp_provider(app, csp=None):` - do they need to be distinct? Can the original be molded in any way to suit your needs?

okaycj-atat:
@jesse-dds The "Refactor make csp function" commit might be overkill.  I don't mind removing it, if that preferable.

atat/domain/csp/__init__.py
```diff
@@ -5,47 +5,34 @@
...
-    def __init__(self, app):
-        self.cloud = MockCloudProvider(app.config)
-        self.files = AzureFileService(app.config)
-        self.reports = MockReportingProvider()
-
+class CSP:
+    def __init__(self, csp, config, **kwargs):
+        azure = AzureCloudProvider(config)
+        mock = MockCloudProvider(config, **kwargs)
+        hybrid = HybridCloudProvider(azure, mock, config)
```
tomdds, line 12: Do these all need to be initialized right away? If not, it's probably better to construct them when they are assigned to self.cloud in their respective conditional blocks.

dandds, line 12: Do we want to make instances for two classes we won't use? Could we just instantiate each of these in the appropriate conditional branch?

okaycj-atat: Yes, I wasn't sure which way was better.  This way seemed be more "DRY".  I have no issues with refactoring this so less classes would be instantiated.

okaycj-atat: Hopefully resolved in https://github.com/dod-ccpo/atst/pull/1603/commits/ad85fe83a728e55335c42af6d4dfa0f53caa9a3f.

PR #1615, opened 2020-06-03
Add AKS permissions bash script back to repo.

dandds:
This script was added to a private repository that previously contained
all of our Terraform code. I'm adding back to the repo so that it's
available to us for dry-running new instances of the infrastructure.

As noted in the comments to the script, it is a stopgap solution that
should be replaced by Terraform configuration as soon as possible.


dandds:
> assuming this was just copying the script here

Yup, all I changed was adding the comments at the top.

PR #1618, opened 2020-06-04
[AT-5229] Include best practices for Azure API integration

graham-dds:
Repurpose the Hybrid Cloud readme to be a general readme for Azure API
integration. Add a section on best practices as a result of [this spike](https://ccpo.atlassian.net/browse/AT-5229).


atat/domain/csp/cloud/README.md
```diff
@@ -1,20 +1,41 @@
...
-## Portfolios
+- include comprehensive docstrings when writing functions that interact with Azure endpoints. Link to endpoint and object documentation when possible
+
+  - For example, when [creating a management group](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/createorupdate), it might be useful to know that the response body has its [own documentation](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/createorupdate#azureasyncoperationresults)
+
+- Make our model classes the "seam" between ATAT and Azure nomenclature.
+
+  - Build payload classes so that their properties align with the parameters used in Azure documentation
+  - Build response classes so that payload data is transformed to use identifiers that make sense for ATAT
+  - Only store UUIDs or simple string identifiers in a portfolio's CSP data dictionary, and use these identifiers to build the path for "fully pathed" resources
```
tomdds, line 13: It might be worth suggesting that a comment is added somewhere (models? csp methods?) that specifies when there's an azure endpoint that switches terms for `name` and `id` (sometimes the `id` in documentation is actually what we understand as either the `name` or `object_id`)

atat/domain/csp/cloud/README.md
```diff
@@ -1,20 +1,41 @@
...
 
-In order to ensure that the provisioning of ATAT entities is happening correctly in Azure, we need a cloud provider interface that can perform as much work as is feasible during this interregnum. The Hybrid Cloud provider accomplishes this task -- in it, steps of provisioning processes are mocked or monkeypatched where needed in order to emulate as close to a production environment as possible.
+## Best Practices for integrating with the Azure REST API
 
-## Portfolios
+- include comprehensive docstrings when writing functions that interact with Azure endpoints. Link to endpoint and object documentation when possible
+
+  - For example, when [creating a management group](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/createorupdate), it might be useful to know that the response body has its [own documentation](https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/createorupdate#azureasyncoperationresults)
+
+- Make our model classes the "seam" between ATAT and Azure nomenclature.
```
okaycj-atat, line 9: I latched onto this suggestion.  I wanted to flush out if there's something more we can do here without being overwrought/over-engineered.  

Because we desire consistency with an API we don't control, I was thinking of maybe "wrapping" the api response in a class that could translate it into something more palatable.  

This is a conversation that started in slack, and might be out of scope for this PR.

graham-dds: I think this is kind of the goal with our payload / response classes in `atat/domain/csp/cloud/models.py`. For `*CSPResponse` classes, we "translate" Azure nomenclature to something more palatable. for `*CSPPayload` classes, we transform our nicely-named data into whatever format is necessary for a given endpoint. 

Does this approach address what you were envisioning?

okaycj-atat: Yes!  Actually.  My apologies for not knowing about this part of the code.  In your opinion, do these model need to be iterated to cover the "seam" idea?  Or are you saying this has already been solved?

graham-dds: > In your opinion, do these model need to be iterated to cover the "seam" idea?

Exactly. Doesn't have to be all at once, but we should modify our models so that:
 - we use better names for things -- Align with Azure Docs for payloads and what's best for ATAT in responses
- only store UUIDs or short string identifiers in CSP data, so there's never a question of "did we store the full path, or just the identifier"?

okaycj-atat: It's possible we were thinking of the same thing, but I just didn't fully understand what you were getting at.  I like this.  Thank you for explaining this to me.

PR #1619, opened 2020-06-04
[ATAT-5320] Update create database script

carleton-dds:
[ATAT-5320](https://ccpo.atlassian.net/browse/AT-5320)

While performing [database bootstrap testing](https://ccpo.atlassian.net/browse/AT-5320) for the upcoming go live dry run I encountered an error with the `script/reset_database.py` script. 

One of the migrations attempts to create the DB extension `uuid-ossp`, however extensions can only be created by admin users in postgres and the script is usual run by a non-admin user. To resolve this I updated `script/create_database.py` to also create that extension after creating the database, as that script runs as the admin user. I also had to update the `helper_functions.inc.sh` script to create the extension as well, so that `script/setup` continued to work properly for developers. It's not great that the name of the extension exists in two scripts rather than a single place (the migration), but it seemed to be the only way to make everything work properly for the moment

### DIFF
--- a/script/create_database.py
+++ b/script/create_database.py
@@ -1,6 +1,7 @@
 # Add root application dir to the python path
 import os
 import sys
+import argparse
 
 parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 sys.path.append(parent_dir)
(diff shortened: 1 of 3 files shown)


carleton-dds:
@dandds Thanks. I will create some new tickets to ensure we're reusing the same scripts as much as possible for everything. 

Related to this is that it appears that the CircleCI build is running the migrations as an admin user, but considering the future of CI could be changing I didn't want to get into it in this ticket

graham-dds:
Just curious -- should we actually remove this extension creation from [the migration](https://github.com/dod-ccpo/atst/blob/423ae1a5dbc6b95fb7b6318a31abd29a02482c3d/alembic/versions/e0c6eb21771f_reset_migrations_with_new_schema.py) that initially tripped us up, or is it okay that it now lives in a migration and setup scripts?

carleton-dds:
@graham-dds it's a good question, and I tried to do that. However CircleCI doesn't currently run the `create_database.py` script and runs the migrations as an admin user, so in order to get CI to pass I need to leave it in for the moment. I'll open up some new issues to track this work.

carleton-dds:
New Issues created:
- https://ccpo.atlassian.net/browse/AT-5350
- https://ccpo.atlassian.net/browse/AT-5351

PR #1626, opened 2020-06-08
Update dod certs

jesse-dds:
Update the certificate authority bundle. This is a potential solution to the 500s Jay and I experienced when troubleshooting https://ccpo.atlassian.net/browse/AT-5331


dandds:
@jesse-dds We need to make this change available to the k8s config, too. We should sync on this briefly tomorrow.

PR #1630, opened 2020-06-10
[AT-5242] Expand available space for total TO value

tomdds:
Addresses UI issue that makes large TO values overlap with borders/each other. This fixes displaying large numbers on the step_4 page by adding a class to override the width set by the `form-container` class on all TO steps. This same component is used for displaying the completed task order summary as well, and once we add the Total Expended field to the component, we'll have to revisit the layout as the current styling (`summary-item` in particular) does not play nice with the layout classes.

## Before
Screen Shot 2020-06-10 at 12 48 35 PM

## After
Screen Shot 2020-06-12 at 1 15 05 PM

### DIFF
--- a/templates/task_orders/fragments/task_order_view.html
+++ b/templates/task_orders/fragments/task_order_view.html
@@ -9,7 +9,7 @@
 
   <div>
     <section class="usa-grid">
-      <div class='usa-width-one-third summary-item'>
+      <div class='usa-width-one-half summary-item'>
         <h4 class="summary-item__header">
           <span class="summary-item__header-text">{{ 'task_orders.summary.total' | translate }}</span>
           {{ Tooltip(("task_orders.review.tooltip.total_value" | translate), title="", classes="icon-tooltip--tight") }}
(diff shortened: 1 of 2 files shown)


templates/task_orders/fragments/task_order_view.html
```diff
@@ -7,7 +7,7 @@
   {% set contract_amount = task_order.total_contract_amount %}
   {% set expended_funds = task_order.invoiced_funds %}
 
-  <div>
+  <div class="task-order">
```
graham-dds, line 10: image

It looks like this makes it so that there are two `<div>`s with a class of `"task-order"` in a row. I think if you got rid of [this line](https://github.com/dod-ccpo/atst/blob/64da5a9a11081b6e47465787f9e0b3e55cdcfa5b/templates/task_orders/view.html#L14), it would remove this issue and the use of this fragment would be consistent everywhere else. If this doesn't mess up the solution to this bug as it's working now, would that be good to include in this PR?

PR #1637, opened 2020-06-16
Alleviate 502 error

jesse-dds:
We are able to definitively say that a source of the 502 errors during CAC logins is due to containers exceeding their memory thresholds and trigger the OOMKiller. There is an observable leak that occurs where memory allocated to the server process is not reclaimed after multiple CAC login attempts. This is potentially due to leaks in the underlying OpenSSL crypto library.

We alleviate this issue by moving the CRL checking functionality into a separate process. This allows the operating system to reclaim the memory after the process has finished. 

However, this does not fix the overall issue, that individual login requests (depending on the CAC card) take a disproportionally large amount of resources to serve. OOMKiller will likely still be triggered if multiple people attempt to log in at once, meaning there's more investigating that needs to be done.

https://ccpo.atlassian.net/browse/AT-5331

### DIFF
--- a/atat/domain/authnid/crl/__init__.py
+++ b/atat/domain/authnid/crl/__init__.py
@@ -146,14 +146,19 @@ class CRLCache(CRLInterface):
 
     def _add_certificate_chain_to_store(self, store, issuer):
         ca = self.certificate_authorities.get(issuer.der())
+        store.add_cert(ca)
+
         while issuer != ca.get_issuer():
+            issuer = ca.get_issuer()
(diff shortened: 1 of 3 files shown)


atat/domain/authnid/crl/__init__.py
```diff
@@ -1,3 +1,4 @@
+import gc
```
dandds, line 1: I don't see where we're using this import. (And also curious why LGTM isn't squawking.)

PR #1639, opened 2020-06-18
[AT-5354] Make the fully pathed `id` derivable for `cloud_id`

jesse-dds:
https://ccpo.atlassian.net/browse/AT-5354

Co-authored By: @okaycj-atat @graham-dds 

Fixes the cloud id by deriving the management group id from a `name` rather than just using the splatted id from the polling response.

This highlights ongoing issues around standardizing our usage of names and ids. We should be working towards this:

> Only store UUIDs or simple string identifiers in a portfolio's CSP data dictionary, and use these identifiers to build the path for "fully pathed" resources

https://github.com/dod-ccpo/atst/blob/staging/atat/domain/csp/cloud/README.md#best-practices-for-integrating-with-the-azure-rest-api

### DIFF
--- a/atat/domain/csp/cloud/models.py
+++ b/atat/domain/csp/cloud/models.py
@@ -372,6 +372,7 @@ class ManagementGroupCSPPayload(AliasModel):
 
 class ManagementGroupCSPResponse(AliasModel):
     id: str
+    name: str
 
 
 class ManagementGroupGetCSPPayload(BaseCSPPayload):
@@ -395,14 +396,16 @@ class InitialMgmtGroupCSPPayload(ManagementGroupCSPPayload):
 
 
 class InitialMgmtGroupCSPResult(AliasModel):
-    root_management_group_id: str
     root_management_group_name: str
 
     class Config:
         fields = {
-            "root_management_group_id": "id",
             "root_management_group_name": "name",
         }
+
+    @property
+    def root_management_group_id(self):
+        return f"/providers/Microsoft.Management/managementGroups/{self.root_management_group_name}"
 
 
 class InitialMgmtGroupVerificationCSPPayload(ManagementGroupGetCSPPayload):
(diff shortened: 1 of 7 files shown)


atat/jobs.py
```diff
@@ -93,7 +93,9 @@ def do_create_application(csp: CloudProviderInterface, application_id=None):
         )
 
         app_result = csp.create_application(payload)
-        application.cloud_id = app_result.id
+        application.cloud_id = (
+            f"/providers/Microsoft.Management/managementGroups/{app_result.name}"
+        )
```
dandds, line 98: Doesn't this go against the "only store UUIDs or simple string identifiers" principle?

atat/domain/csp/cloud/models.py
```diff
@@ -372,6 +372,7 @@ def enforce_display_name_length(cls, name):
 
 class ManagementGroupCSPResponse(AliasModel):
     id: str
+    name: str
```
dandds, line 375: Do we need and use both of these?

atat/jobs.py
jesse-dds: It still follows the principle as we no longer are storing the "id" received from the management group polling endpoint. We are instead storing the "name" and constructing the fully-pathed identifier from that. This issue is that in this context, the "id" received everywhere `create_management` group was used is a handle to an asynchronous operation, not an identifier for a management group.

atat/domain/csp/cloud/models.py
jesse-dds: Yes, there are several model classes that inherit from `ManagementGroupCSPResponse`. `name` was added to store the simpler "name" property from management group responses. But a lot of other code depends on the `id` property being there and removing it causes more cascading changes that are out of scope for this ticket.

jesse-dds:
Maybe @graham-dds can explain better, this change was small but tricky and touches on issues with what we store from responses.

atat/jobs.py
dandds: This makes sense. I think we're talking about "storing" in different ways. I mean what we save to the database (which is the fully-pathed ID in this case) but I think you means what portion of the API response we keep as an identifier for the resource.

PR #1652, opened 2020-06-24
[AT-5367] Add appRole assignment so that the tenant service principal can invite users

graham-dds:
This PR adds a step to the portfolio provisioning process to grant the `User.Invite.All` permission to the tenant service principal via an [appRole assignment](https://docs.microsoft.com/en-us/graph/api/serviceprincipal-post-approleassignments?view=graph-rest-1.0&tabs=http). This should allow the service principal to invite users into the tenant. 

The hybrid test for this flow will be added in a separate PR after @dandds wraps up his work with management groups, since he has already started restructuring hybrid tests. For now, this is how I verified that this process is working:

Run the portfolio provisioning hybrid test on this branch. I added a print statement (not committed) in the hybrid call to create the app registration so that I could note the name of the App Registration:

```
CSP=hybrid pytest -k "test_hybrid_provision_portfolio" -m hybrid --hybrid --no-cov -s
```

image

After provisioning is completed, go to the app registration in the portal and click on `API permissions` in the sidebar. Notice that the app registration now has the `User.Invite.All` Graph permission added:

image

I'm not sure what the difference is between the `Configured permissions` and `Other permissions...` sections are, but the permission appears to be active and accepted.


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -85,6 +87,8 @@
 AZURE_SKU_ID = "0001"  # probably a static sku specific to ATAT/JEDI
 REMOTE_ROOT_ROLE_DEF_ID = "/providers/Microsoft.Authorization/roleDefinitions/00000000-0000-4000-8000-000000000000"
 
+GRAPH_API_APPLICATION_ID = "00000003-0000-0000-c000-000000000000"
```
dandds, line 90: I doublechecked: you can actually get this ID out of the portal. If you go to "Enterprise Applications", set the "Application Type" to "all" and search for "GraphAggregatorService", that's the app registration represented by this ID. It gets an instance in every AAD tenant. Might be worth noting that here in a comment, since we were looking for a reference for where the ID is derived from.

atat/models/mixins/state_machines.py
```diff
@@ -31,6 +32,7 @@ class AzureStages(Enum):
     TENANT_PRINCIPAL_CREDENTIAL = "tenant principal credential"
     ADMIN_ROLE_DEFINITION = "admin role definition"
     PRINCIPAL_ADMIN_ROLE = "tenant principal admin"
+    PRINCIPAL_APP_INVITE_PERMISSION = "grant invite permission to principal application"
```
dandds, line 35: I would make this name more generic, like `PRINCIPAL_APP_GRAPH_API_PERMISSIONS` or something. That way we can change the implementation to be "read/write the whole directory" if need be without invalidating the name.

graham-dds: 9b3b39e60ef934e254b376391942c3beef5222da

PR #1653, opened 2020-06-24
[AT-5352] Enforce nesting of application and environment management groups.

dandds:
When we create management group, we transmit information about what management group should be its parent. For an application management group, its parent should be the portfolio management group. These parameters in the request body do not seem to be respected by Azure at the moment.  It is creating application management groups as children of the tenant's root management group, which is not the hierarchy we want for our Hybrid implementation.

Making a subsequent PATCH request on the management group to update its parent fixes the issue, but we should not have to do this. I think that we should use this as a temporary fix and investigate why the Azure resource management API doesn't seem to respect the initial set of parameters we give it.

This PR introduces a new test class for the hybrid suite, `TestIntegration`. The tests are highly dependent on their order here. The first test provisions a portfolio in Azure. Subsequent tests add an application and environment, provisioning their management groups and ensuring they are nested correctly.

Notes

- The TestIntegration class implements its own version of the database session fixture. Since database entities implemented as fixtures need to be shared between tests in the class, the database session needs to be scoped to the test run session.
- The commit de4c8662d9a912b2ca45151b397518f98701fd69 introduces a parallel change to what was previously done in 60aef421e7. We save the fully qualified ID of the new environment management group as the `cloud_id` for that environment.

https://ccpo.atlassian.net/browse/AT-5352
https://ccpo.atlassian.net/browse/AT-5353

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -316,12 +316,11 @@ class AzureCloudProvider(CloudProviderInterface):
         session.headers = {
             "Authorization": f"Bearer {sp_token}",
         }
-        if parent_id is None:
-            parent_id = f"/providers/Microsoft.Management/managementGroups/{tenant_id}"
+        request_body = {"properties": {"displayName": display_name}}
 
-        request_body = {
-            "properties": {"displayName": display_name, "parent": {"id": parent_id},}
-        }
+        if parent_id:
+            request_body["properties"]["parent"] = {"id": parent_id}
+
         response = session.put(
             f"{self.sdk.cloud.endpoints.resource_manager}providers/Microsoft.Management/managementGroups/{management_group_id}?api-version=2020-02-01",
             json=request_body,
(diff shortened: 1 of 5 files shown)


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -81,60 +99,139 @@ def csp(app):
...
-    config = {"billing_account_name": "billing_account_name"}
-
-    while state_machine.state != PortfolioStates.COMPLETED:
-        collected_data = dict(
-            list(csp_data.items())
-            + list(state_machine.portfolio.to_dictionary().items())
-            + list(config.items())
+class TestIntegration:
+    @pytest.fixture(scope="session", autouse=True)
+    def session(self, db, request):
```
graham-dds, line 105: Instead of rewriting this fixture all to set the scope to `session`, you might be able to use some [pytest magic](https://docs.pytest.org/en/stable/fixture.html#dynamic-scope) to dynamically set the scope. Maybe the fixture could determine what scope to use based on the presence of a `hybrid` mark?

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -303,27 +303,76 @@ def create_initial_mgmt_group_verification(
...
+        https://docs.microsoft.com/en-us/rest/api/resources/managementgroups/update
+        Args:
+            session: a requests session object
+            parent_id: the ID of the parent for the management group
+            display_name: a display name for the management group
+            management_group_id: a simple ID for the management group, like a GUID (i.e., not fully qualified)
+        Returns:
+            True
+        """
+        request_body = {"displayName": display_name, "parentId": parent_id}
```
graham-dds, line 368: I know `displayName` is listed in the request body parameters, but did you determine if it was necessary to include in this `PATCH` since the only thing we're updating is the `parentId`?

tests/domain/cloud/test_hybrid_csp.py
dandds: Good call. Technically, the other tests in the hybrid suite should still use the function-scoped fixture, but it doesn't see to hurt them to share one. Since we intend to re-write them to all share the session-scoped fixture I think that's okay for now. 3c9f71bbff7a9cc12e2ac76232132ef634c40580

atat/domain/csp/cloud/azure_cloud_provider.py
dandds: It wasn't listed as optional, so I made the rookie mistake of trusting the docs. You're right, it works without it. 8cf310e25f7cf836160f1082e22b7f5b31474927

PR #1668, opened 2020-07-09
[AT-4889] Pass all upload config to frontend

dandds:
This updates both the code and build process so that it's no longer necessary to give details about the task order upload location to the frontend when the JS bundle is compiled. Now, all the config the JS needs to upload a file is passed to it by the `task_orders.upload_token` endpoint. This removes some complexity from the build process and makes the build less brittle.

I'm keeping this in draft mode until I can resolve conflicts with the [open PR](https://github.com/dod-ccpo/atst/pull/1661) to move the Dockerfile to a RHEL base, since I also need to make changes to the Dockerfile.

https://ccpo.atlassian.net/browse/AT-4889


atat/routes/task_orders/new.py
```diff
@@ -100,11 +100,28 @@ def update_and_render_next(
         )
 
 
+def upload_config(csp, config):
+    if csp == "azure" or csp == "hybrid":
```
graham-dds, line 104: There are starting to become enough of these CSP strings floating around that it might be good to throw them into an Enum. Maybe not something to include in this PR, but what's your take?

dandds: Yes, agreed. I think it would touch enough other spots in the code that it's out-of-scope for this PR, but it would be a good cleanup.

PR #1594, opened 2020-05-27
Refine Hybrid implementation to work with new Key Vault.

dandds:
The Terraform code was recently updated to include the Key Vault and app
registration used for managing customer AAD tenants. I ran our Hybrid
test suite using the URL for the new KV and creds for the new app
registration and encountered a few errors that are fixed here.

There were two app registrations used for running the Hybrid test suite:
1. the one that has access to the Key Vault where tenant credentials are
   stored
2. one that lives in the Hybrid tenant and exists strictly for testing
   purposes

I've added a third that has the "Invoice Section Reader" role for our
invoice section. We can't give that permission to 1 since 1 is managed
by TF, and giving that perm to 2 would require making 2 a multi-tenant
app. Instead, I've added a new app registration and accompanying config
for the purposes of running the `test_get_reporting_data` integration
test.

PR #1671, opened 2020-07-09
[AT-5277] Hybrid reporting data

graham-dds:
Addresses: https://ccpo.atlassian.net/browse/AT-5277

Configures the Hybrid Cloud Provider so that it returns reporting data retrieved by querying the Cost Management API.

Normally each portfolio would have its own billing setup, so the service principal created during portfolio provisioning would be able to query the cost management API itself. In our case though, billing setup is mocked during portfolio provisioning, and we don't want to grant the provisioned hybrid service principals the permissions to read our billing data.

In https://github.com/dod-ccpo/atst/pull/1594 we added a special app registration and granted it the "Invoice Section Reader" role for our invoice section so that we could write an integration test for the hybrid's `get_reporting_data` method. In this PR, we use that same app registration not only in the integration test, but every time we call the hybrid's `get_reporting_data` method. With this implementation, the same cost management data will be reported for every profile. This is the expected behavior for the hybrid.

Part of this PR includes a change to the K8s overlays so that the config values related to the hybrid reporting service principal and billing are available. This overlay hasn't been applied yet, but will be after this PR is merged.

image

Running the app locally, we see that the reporting section shows an `Invoiced expended funds` of `$1551.0` and `Estimated  expended funds` of `$500`. This corresponds with the "spend data" [returned by the mock cloud provider](https://github.com/dod-ccpo/atst/blob/staging/atat/domain/csp/cloud/mock_cloud_provider.py#L511:L536).

By running the app with the Hybrid CSP however (e.g. `CSP=hybrd flask run`), we see that we return real data:

image


deploy/overlays/cloudzero-dryrun-staging/flex_vol.yml
```diff
@@ -21,9 +21,9 @@ spec:
...
-              keyvaultobjecttypes: "secret;secret;secret;secret;key;secret;secret;secret;secret;secret;secret"
+              keyvaultobjectnames: "AZURE-STORAGE-KEY;MAIL-PASSWORD;PGPASSWORD;REDIS-PASSWORD;SECRET-KEY;AZURE-TENANT-ID;AZURE-CLIENT-ID;AZURE-USER-OBJECT-ID;AZURE-TENANT-ADMIN-PASSWORD;AZURE-SECRET-KEY;AZURE-HYBRID-TENANT-ID;AZURE-BILLING-ACCOUNT-NAME;AZURE-BILLING-PROFILE-ID;AZURE-INVOICE-SECTION-ID;AZURE-HYBRID-REPORTING-CLIENT-ID;AZURE-HYBRID-REPORTING-SECRET" 
+              keyvaultobjectaliases: "AZURE_STORAGE_KEY;MAIL_PASSWORD;PGPASSWORD;REDIS_PASSWORD;SECRET_KEY;AZURE_TENANT_ID;AZURE_CLIENT_ID;AZURE_USER_OBJECT_ID;AZURE_TENANT_ADMIN_PASSWORD;AZURE_SECRET_KEY;AZURE_HYBRID_TENANT_ID;AZURE_BILLING_ACCOUNT_NAME;AZURE_BILLING_PROFILE_ID;AZURE_INVOICE_SECTION_ID;AZURE_HYBRID_REPORTING_CLIENT_ID;AZURE_HYBRID_REPORTING_SECRET"
+              keyvaultobjecttypes: "secret;secret;secret;secret;key;secret;secret;secret;secret;secret;secret;secret;secret;secret;secret;secret"
```
dandds, line 26: Could you update the "Hybrid Config" section of the deployment guide with this info?

dandds: Also, uuugh, these entries.

PR #1681, opened 2020-07-15
[AT-5362] Rearrange hybrid integration tests

graham-dds:
This PR rearranges hybrid tests so that they more closely resemble real provisioning. Continuing the work of #1653, it moves all tests except subscription-related tests and a hybrid reporting test under the `TestIntegration` class.

You may have noticed that [AT-5362](https://ccpo.atlassian.net/browse/AT-5362) was not originally in this sprint and may be wondering why this PR is up right now. I originally completed the work in this PR in order to support [AT-5373](https://ccpo.atlassian.net/browse/AT-5373), but I ended up changing my approach to [AT-5373](https://ccpo.atlassian.net/browse/AT-5373) mid-ticket. 

The plan I had when I set out on completing [AT-5373](https://ccpo.atlassian.net/browse/AT-5373) was to create a subscription under the the environment created during the hybrid integration tests, using the app registration / service principal also created during those tests. I realized that this wouldn't be a viable approach since we didn't want to give those hybrid test entities the permissions to create subscriptions for our root tenant. So, I moved the subscription creation work to its own PR (#1682), and kept this work here (since it was already completed) to close out [AT-5362](https://ccpo.atlassian.net/browse/AT-5362).


graham-dds:
Right now the hybrid reporting test isn't passing because of our ongoing billing issues with Azure, but the tests in the `TestIntegration` class  should pass.

PR #1683, opened 2020-07-15
task_orders.upload_token should present "hybrid" CSP as "azure"

dandds:
In order for the frontend JS to know that it should upload files to
Azure when the backend is configured to use the hybrid provider (rather
than use a mock uploader), the backend upload_token route should present
both the hybrid and Azure CSP providers as being Azure.

### DIFF
--- a/atat/routes/task_orders/new.py
+++ b/atat/routes/task_orders/new.py
@@ -110,17 +110,23 @@ def upload_config(csp, config):
         return {}
 
 
+def upload_cloud_provider(csp):
+    if csp == "hybrid":
+        return "azure"
+    else:
+        return csp
+
+
 @task_orders_bp.route("/task_orders/<portfolio_id>/upload_token")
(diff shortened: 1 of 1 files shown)


atat/routes/task_orders/new.py
```diff
@@ -110,17 +110,23 @@ def upload_config(csp, config):
         return {}
 
 
+def upload_cloud_provider(csp):
```
graham-dds, line 113: This gets the job done, but I noticed that it's similar in spirit to the function above it. I can't decide if this would be better or not, but what do you think about having `upload_config` be something like `make_upload_args` and having it return a tuple of `(csp, upload_config)`?  It's just one fewer util function to have to keep track of.

dandds: Yeah, I couldn't decide what to do here. I'm not always a fan of tuple return types because sometimes it feels like you're smushing two unlike things together just for concision. I think what would really help clean it up is having the CSP enum you mentioned on a separate PR.

PR #1689, opened 2020-07-17
[AT-5377] Create a default subscription for environment

graham-dds:
Closes [AT-5377](https://ccpo.atlassian.net/browse/AT-5377)

Adds a subscription to an environment when it is created.

Note that for now, we don't actually track the status of subscription creation since it's a long-running async job in Azure. ATAT's only care concerning subscriptions is that a a request to create a subscription is accepted. In the future, we may track the status of this job, but that's out of scope for this PR.

The "meat" of this PR is contained in https://github.com/dod-ccpo/atst/commit/4fdcd2f037741c30c10e5d03d9769b4d0d87558e, but the prior commits were to help `do_create_environment` (where we call the function to kick off the `create_subscription` task) feel a little less crowded.


tests/models/test_environments.py
```diff
@@ -65,3 +66,76 @@ def test_environment_roles_do_not_include_deleted():
     EnvironmentRoles.disable(role_2.id)
 
     assert len(env.roles) == 2
+
+
+class TestBuildSubscriptionPayload:
```
graham-dds, line 71: These tests are mostly the same as when they lived in `test_settings`, but now they build a subscription payload via the model's method.

atat/domain/environments.py
```diff
@@ -39,12 +39,15 @@ def create_many(cls, user, application, names):
...
-        if name is not None:
-            environment.name = name
-            db.session.add(environment)
-            commit_or_raise_already_exists_error(message="environment")
-            return environment
+    def update(cls, environment, new_data):
+        if "name" in new_data:
+            environment.name = new_data["name"]
+        if "cloud_id" in new_data:
+            environment.cloud_id = new_data["cloud_id"]
```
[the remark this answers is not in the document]
dandds: I'm not suggesting you change it here, but this pattern seems a little weird to me: "Give me a dictionary with anything in it and I'll decide how to mutate the environment's state". I think a known set of kwargs might be better.

atat/jobs.py
```diff
@@ -163,23 +173,45 @@ def do_create_environment(csp: CloudProviderInterface, environment_id=None):
...
         payload = EnvironmentCSPPayload(
             tenant_id=tenant_id, display_name=environment.name, parent_id=parent_id
         )
         env_result = csp.create_environment(payload)
-        environment.cloud_id = (
-            f"/providers/Microsoft.Management/managementGroups/{env_result.name}"
+        cloud_id = f"/providers/Microsoft.Management/managementGroups/{env_result.name}"
+        Environments.update(environment, new_data={"cloud_id": cloud_id})
+
+        create_subscription.delay(environment_id=environment.id)
```
dandds, line 188: Do you think we definitely need to run this as a separate, chained background task, or would it be fine to have both happen in the same task? (If so, this might be a good use case for some more advanced Celery functionality for chaining tasks: https://docs.celeryproject.org/en/stable/userguide/canvas.html#chains). Curious what @tomdds thinks, since he's spent a lot of time pondering when/how to chain provisioning jobs.

atat/models/environment.py
```diff
@@ -69,11 +72,31 @@ def __repr__(self):
...
+        display_name = f"{self.application.name}-{self.name}-{token_urlsafe(6)}"
+
+        return SubscriptionCreationCSPPayload(
+            tenant_id=csp_data.get("tenant_id"),
+            display_name=display_name,
+            parent_group_id=parent_group_id,
+            billing_account_name=billing_account_name,
+            billing_profile_name=csp_data.get("billing_profile_name"),
+            invoice_section_name=invoice_section_name,
+        )
```
dandds, line 102: I tend to think that the models should stay as "data bags" and only know their own data from the db _or_ information that's easily derived from that data. This requires that it know implementation details about subscription creation.

I've noticed there's a gap between the Celery tasks and the Azure cloud interface: who or what should be responsible for building the payloads? It requires a lot of specific implementation knowledge and rooting around for pieces of data. For other tasks we've built the payload in the task itself, which doesn't work in your case because you also need to do it in the route. I wonder if we should make factory functions available from the cloud module, like `build_subscription_creation_payload`.

atat/domain/environments.py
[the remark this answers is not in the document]
graham-dds: > but this pattern seems a little weird to me

Agreed, but refactoring seemed to be out of scope for this PR. I feel like we should fix all instances of this pattern in one go.

atat/jobs.py
graham-dds: My goal here was to design this flow so that failing to create a subscription would not fail the entire `create_environment` job. And since management groups and subscriptions are both long-running jobs in Azure, even though we don't track subscription creation right now, I thought it would be good to define them in separate tasks in the event we do track them.

Looking into Celery chains could be interesting. What's the benefit of having a chained task vs a separete background task vs combining the logic in the same task?

atat/models/environment.py
graham-dds: > This requires that it know implementation details about subscription creation.

True, but only insofar as this property needs to know which elements of its `own data from the db or information that's easily derived from that data` to grab to hand back when accessing the property. The only data that's not accessed from data in the environment, its parent application, or parent portfolio is global config, `billing_account_name`. 

That being said, I agree that `who or what should be responsible for building the payloads?` is an important question. Not only does a subscription require "knowledge" of many ATAT entities, but it's also highly Azure-specific. So I'm definitely shying away from keeping this on the model, which is designed to be a generic cloud entity, but yeah, where does it go? I could see it being a piece of the Azure Cloud provider 👍

graham-dds: @dandds and I talked this through on Zoom and we moved this logic to the job. This also resulted in making the call to create a subscription from the routes a [celery task](https://github.com/dod-ccpo/atst/pull/1689/commits/fa0417add1a900d1c9a1f763f24a08d6c377ba36#diff-0a93d93aab584c4f5030a0e119600cbaR536-R538).

atat/jobs.py
tomdds: > What's the benefit of having a chained task vs a separete background task vs combining the logic in the same task?

Chained tasks have an association with one another, so we can see what jobs created which other jobs, which gives us better insight into the goings-ons. Combining the logic means that you don't have discrete tasks for work, also any unexpected errors in the subscription creation would then fail the environment creation as well.

Based on the API docs Dan linked, I'm not sure you can add a task to the current chain from within a task, it seems like the chain needs to be established when the current task is called. Since this task is usually triggered via a beat worker, I'm not sure how much upstream code/arch would need to change to support the chaining. The basic change I would try to make here is to figure out how to at least log the connection between this `create_environment` task and the `create_subscription` task.

tests/models/test_environments.py
```diff
@@ -6,6 +6,7 @@
 from atat.domain.environment_roles import EnvironmentRoles
 
 from tests.factories import *
+from atat.domain.csp.cloud.models import SubscriptionCreationCSPPayload
```
tomdds, line 9: Doesn't look like this is used in this file, can it be removed?

graham-dds: https://github.com/dod-ccpo/atst/pull/1689/commits/6b1102eea6f9b51261525f90989be5d72af3df9d

atat/jobs.py
tomdds: Yeah, that should give us enough connections to trace. @dandds thoughts?

dandds: LGTM!

PR #1708, opened 2020-07-31
Use Location header to get completed management group

graham-dds:
Creating management groups is an async operation. After sending
the initial request to create a management group, Azure returns a
response that contains two important headers: Azure-AsyncOperation and
Location. The former reports the status of the async job, and until now
was the only header we needed to track.

GETing the URL specified in the Azure-AsyncOperation header would return
a payload like:
```
{
    "id": "/providers/Microsoft.Management/operationResults/create/asyncOperation/status/managementGroups/<uuid>",
    "name": "<uuid>",
    "status": "Succeeded",
    "properties": {
        "tenantId": "<tenant_id>",
        "displayName": "Hybrid :: Application Name 0",
        "details": {
            "version": 1,
            "updatedTime": "2020-07-31T20:32:29.9423339Z",
            "updatedBy": "<user_uuid>",
        },
    },
}
```

Things to note here:
 - the id contains "operationResults/create/asyncOperation/status",
   which we don't want
 - there is no information about the parent, which we expect in the
   response

Even though the docs say "If your operation returns this value, always
use it (instead of Location) to track the status of the operation", I
found that by GETing the url provided by the Location header, we receive
the expected payload:
```
{
    "id": "/providers/Microsoft.Management/managementGroups/<uuid>",
    "type": "Microsoft.Management/managementGroups",
    "name": "<uuid>",
    "properties": {
        "tenantId": "<tenant_id>",
        "displayName": "Hybrid :: Application Name 0",
        "details": {
            "version": 1,
            "updatedTime": "2020-07-31T20:32:29.9423339Z",
            "updatedBy": "<user_uuid>",
            "parent": {
                "id": "/providers/Microsoft.Management/managementGroups/<tenant_id>,
                "name": "<tenant_id>",
                "displayName": "Tenant Root Group",
            },
        },
    },
}
```
I _believe_ that just relying on the Azure-AsyncOperation header url was
enough up to this point, but something must have changed on Azure's end
such that this url no longer provided the same payload.

Now, we use the Azure-AsyncOperation header URL to track the status of
the job. When we get a status of "Succeeded", we then use the Location
url to get the full payload for the created management group.

### DIFF
--- a/tests/domain/cloud/test_azure_csp.py
+++ b/tests/domain/cloud/test_azure_csp.py
@@ -1544,7 +1544,11 @@ class TestCreateManagementGroup:
         mock_session_object = Mock()
         mock_session_object.put = Mock(
             return_value=mock_requests_response(
-                status=202, headers={"Azure-AsyncOperation": "http://url.com"}
+                status=202,
+                headers={
+                    "Azure-AsyncOperation": "http://status_url.com",
+                    "Location": "http://result_url.com",
+                },
             )
         )
         mock_azure.sdk.requests.Session.return_value = mock_session_object
@@ -1561,7 +1565,7 @@ class TestCreateManagementGroup:
             "management_group_id", "display_name", "tenant_id"
         )
         mock_poll_management_group_creation_job.assert_called_once_with(
-            "http://url.com", mock_session_object
+            "http://status_url.com", "http://result_url.com", mock_session_object
         )
 
     def test_raises_exceptions(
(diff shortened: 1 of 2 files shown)


tests/domain/cloud/test_azure_csp.py
```diff
@@ -1633,9 +1645,13 @@ def make_error(status):
...
 
         with pytest.raises(ResourceProvisioningError):
-            mock_azure._poll_management_group_creation_job("url", mock_session_object)
+            mock_azure._poll_management_group_creation_job(
+                "stataus", "result", mock_session_object
+            )
         with pytest.raises(ResourceProvisioningError):
-            mock_azure._poll_management_group_creation_job("url", mock_session_object)
+            mock_azure._poll_management_group_creation_job(
+                "stataus", "result", mock_session_object
```
okaycj-atat, line 1653: Not a big deal, but "stataus" is misspelled.

graham-dds: Thanks! Fixed

PR #1714, opened 2020-08-05
Refactor root/initial management group data

graham-dds:
This PR comes as a result of this discussion https://github.com/dod-ccpo/atst/pull/1699#discussion_r464635668. It has two goals:
- Fix the fact that we would save the management group created in the `create_initial_mgmt_group` step as the `root_management_group`
- Make`root_management_group_name` an alias for`tenant_id` in CSP data

This branch is currently failing hybrid tests intermittently:
 - ~Tests will fail during portfolio provisioning during the `TENANT_PRINCIPAL_OWNERSHIP` stage due to an authorization error. This seems to happen every other hybrid test suite run, and it could be an indication that the elevated access context manager isn't working as expected~. **UPDATE**: https://github.com/dod-ccpo/atst/pull/1714/commits/00038be9d69cb6e92e1502d3ab671d75e19de400 seemed to fix these intermittent auth errors. We needed to re-obtain a token after elevated access before yielding it. This is consistent with the spike findings from AT-5316 and was the pattern used in this proof of concept script [this script](https://gist.github.com/graham-dds/a9302dbe78d24865006f3346d78348af).
 - Tests will also fail occasionally due to a `400` error during `_poll_management_group_creation_job`. I don't know what the deal is here, because simply sleeping for a little bit seems to pass this state. But sometimes 5 seconds is enough time to wait, and sometimes is isn't. There's probably a better way to handle this than sleeping, though. **UPDATE**: This is still occurring, but doesn't seem to be unique to this branch. I've seen similar errors now on the `staging` branch.


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -431,6 +431,9 @@ def _poll_management_group_creation_job(
             response_json = response.json()
             status = response_json["status"]
             if status == AsyncOperationStatus.SUCCEEDED.value:
+                # We'd occasionally get 400s here without sleeping before
+                # getting the url in the Location header
+                time.sleep(5)
```
jesse-dds, line 436: I feel that is a really long amount of time to wait. Are able to:

1) Dependably recreate the 400s? Do we know why they're happening?
2) Get away with setting the value to something lower like 1 second?

graham-dds: 1. 
> Dependably recreate the 400s? 

I haven't been able to do this, no.

> Do we know why they're happening?

We don't, but as far as I can tell, I think it's something on Azure's end? Like we grant permissions and then try to get a resource with those permissions, but Azure hasn't finished propagating the updated permissions in their system. We've encountered similar issues [elsewhere in the app](https://github.com/dod-ccpo/atst/blob/acf31f927fadb6a421adf06534ebb1ca895e6a96/tests/domain/cloud/test_hybrid_csp.py#L134)

2. I might end up just removing this sleep, as I'm still getting a 400, even with the occasional sleep. Maybe investigating point 1 further would obviate the need for this (admittedly janky) sleep stuff altogether.

graham-dds: > I might end up just removing this sleep
https://github.com/dod-ccpo/atst/pull/1714/commits/8c279d7fe03f80afa8888ab1847504cc1469779d

Look like someone `out there` is experiencing [similar issues](https://gist.github.com/oleksandrkudin/d7fb8026797bfde5287a54617ba4eb8a)

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -1753,16 +1779,17 @@ def _get_elevated_access_token(self, tenant_id, user_object_id):
         tenant_admin_token = self._get_tenant_admin_token(
             tenant_id, self.sdk.cloud.endpoints.resource_manager + "/.default"
         )
-        elevated_token = self._elevate_tenant_admin_access(tenant_admin_token)
+        self._elevate_tenant_admin_access(tenant_admin_token)
         try:
+            elevated_token = self._get_tenant_admin_token(
+                tenant_id, self.sdk.cloud.endpoints.resource_manager + "/.default"
+            )
```
tomdds, line 1786: If the second call to `_get_tenant_admin_token` fails, then `elevated_token` would be considered unassigned and throw an error. The simple way around this would just be to init `elevated_token` as `None`, but I'm also curious what drove the decision to re-authenticate after access was elevated?

Test:
```python
def test_second_token_request_fails(self, mock_azure, monkeypatch):
    monkeypatch.setattr(
        mock_azure,
        "_get_tenant_admin_token",
        Mock(side_effect=["MOCK_TOKEN", AuthenticationException("weird")]),
    )

    with mock_azure._get_elevated_access_token("tenant_id", "user_object_id") as _:
        mock_azure._remove_tenant_admin_elevated_access.assert_called_with("tenant_id", "user_object_id", token=None)
```
Result:
```
    @contextmanager
    def _get_elevated_access_token(self, tenant_id, user_object_id):
    
        tenant_admin_token = self._get_tenant_admin_token(
            tenant_id, self.sdk.cloud.endpoints.resource_manager + "/.default"
        )
        self._elevate_tenant_admin_access(tenant_admin_token)
        try:
            elevated_token = self._get_tenant_admin_token(
                tenant_id, self.sdk.cloud.endpoints.resource_manager + "/.default"
            )
            yield elevated_token
        finally:
            try:
                self._remove_tenant_admin_elevated_access(
>                   tenant_id, user_object_id, token=elevated_token
                )
E               UnboundLocalError: local variable 'elevated_token' referenced before assignment
```

graham-dds: Good catch! https://github.com/dod-ccpo/atst/pull/1714/commits/65779021a248fc5f03354972ed6b55c0ccbb80ae

> I'm also curious what drove the decision to re-authenticate after access was elevated

I guess elevating the token only adds the role assignment, so we need to re-auth so that the token we go on to use has the right permissions attached to it. I don't have it on hand, but I think Azure's error message that inspired this change said something to that effect. 

This is consistent with [what I found in the spike](https://gist.github.com/graham-dds/a9302dbe78d24865006f3346d78348af#file-elevated_access-py-L148) too.

PR #1717, opened 2020-08-06
[AT-5443] Bugfix: test suite fails based on time of day

graham-dds:
Closes [AT-5443](https://ccpo.atlassian.net/browse/AT-5443)

Test suite run at 8:54 AM UTC-4
```shell
> date -u
Thu Aug  6 12:54:17 UTC 2020
> pytest --no-cov -q
743 passed, 33 skipped in 49.14s
```
Test suite run at  9:01 PM UTC-4
```shell
> date -u
Fri Aug  7 01:01:02 UTC 2020
> pytest --no-cov -q
743 passed, 33 skipped in 49.58s
```

Changes this PR makes:

- [pendulum.now()](https://github.com/sdispater/pendulum/blob/8af1e2e8f859d5b0b8dc11e1aea6eb25b1e7aba9/pendulum/__init__.py#L197) and  [pendulum.today()](https://github.com/sdispater/pendulum/blob/8af1e2e8f859d5b0b8dc11e1aea6eb25b1e7aba9/pendulum/__init__.py#L232) use the local timezone if `None` is provided. We specified `tz=UTC` in some calls to these functions, but we should supply this kwarg to all such function calls

- certain columns use SQLAlchemy’s `Date` type, which uses [`datetime.date`](https://docs.python.org/3.7/library/datetime.html#datetime.date) internally. We should ensure that we always pass pendulum [`Date()`](https://github.com/sdispater/pendulum/blob/205a86a22cbebd1fb33c5332a05fa9c2da5a3763/pendulum/date.py#L28) objects to columns that expect `datetime.date` objects. I found that passing pendulum [`DateTime`](https://github.com/sdispater/pendulum/blob/36e8726af89b6201a6f72abe6d69863647d376ae/pendulum/datetime.py#L45) objects resulted in inaccurate conversions to `datetime.date` objects by SQLAlchemy

### DIFF
--- a/atat/domain/task_orders.py
+++ b/atat/domain/task_orders.py
@@ -1,4 +1,4 @@
-from sqlalchemy import or_
+from sqlalchemy import or_, func
 import pendulum
 
 from atat.database import db
@@ -110,7 +110,7 @@ class TaskOrders(BaseDomainClass):
           - the current date is within the CLIN's period of performance
           - the associated portfolio is done being provisioned
         """
-        today = pendulum.today(tz="UTC").date()
+
         return (
             db.session.query(CLIN)
             .join(TaskOrder)
@@ -118,8 +118,8 @@ class TaskOrders(BaseDomainClass):
             .join(PortfolioStateMachine)
             .filter(
                 CLIN.last_sent_at.is_(None),
-                CLIN.start_date <= today,
-                CLIN.end_date >= today,
+                CLIN.start_date <= func.now(),
+                CLIN.end_date >= func.now(),
                 PortfolioStateMachine.state == PortfolioStates.COMPLETED,
             )
             .all()
--- a/atat/domain/task_orders.py
+++ b/atat/domain/task_orders.py
@@ -47,7 +47,7 @@ class TaskOrders(BaseDomainClass):
     @classmethod
     def sign(cls, task_order, signer_dod_id):
         task_order.signer_dod_id = signer_dod_id
-        task_order.signed_at = pendulum.now(tz="utc")
+        task_order.signed_at = pendulum.now(tz="UTC")
         db.session.add(task_order)
         db.session.commit()
 
--- a/atat/domain/task_orders.py
+++ b/atat/domain/task_orders.py
@@ -110,6 +110,7 @@ class TaskOrders(BaseDomainClass):
           - the current date is within the CLIN's period of performance
           - the associated portfolio is done being provisioned
         """
+        today = pendulum.today(tz="UTC").date()
         return (
             db.session.query(CLIN)
             .join(TaskOrder)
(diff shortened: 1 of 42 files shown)


atat/domain/task_orders.py
```diff
@@ -110,14 +110,16 @@ def get_clins_for_create_billing_instructions(cls):
...
         return (
             db.session.query(CLIN)
             .join(TaskOrder)
             .join(Portfolio)
             .join(PortfolioStateMachine)
             .filter(
                 CLIN.last_sent_at.is_(None),
-                CLIN.start_date < pendulum.now(tz="UTC"),
+                CLIN.start_date <= today,
+                CLIN.end_date >= today,
```
tomdds, line 122: Looking over other parts of the code, this seems to be the only time that we're using `pendulum.now()` within a query. The other date queries use `func.now()` which I suspect would work fine here as well, any reason you see not to?

graham-dds: Ahh good call. I'll definitely use `func.now()` here.

Also, this made me realize that my find + replace of `.now() -> .now(tz="UTC")` might have cast too wide a next and made `func.now()` -> `func.now(tz="UTC")`. I'll make that change as well.

Speaking of `func.now()` though, reading through this [SO post](https://stackoverflow.com/a/33532154/5821620), I noticed this:
https://docs.sqlalchemy.org/en/13/core/compiler.html#utc-timestamp-function

Is this something we should implement in an effort to `make everything UTC all the time`?

graham-dds: Also, now that you mention this, I see a few places where we pass a `now` parameter to be used for a query, but that parameter is a pendulum datetime. For example:

https://github.com/dod-ccpo/atst/blob/a6a789ac8bf583b0764e2853dbde95173bc1e5b7/atat/jobs.py#L402

Calls 
https://github.com/dod-ccpo/atst/blob/a6a789ac8bf583b0764e2853dbde95173bc1e5b7/atat/domain/environments.py#L133

which uses that `now` param in the [filter](https://github.com/dod-ccpo/atst/blob/staging/atat/domain/environments.py#L108:L125):
```python
@classmethod
def base_provision_query(cls, now):
    return (
        db.session.query(Environment.id)
        .join(Application)
        .join(Portfolio)
        .join(TaskOrder)
        .join(CLIN)
        .filter(CLIN.start_date <= now)
        .filter(CLIN.end_date > now)
        # ....
```

Should we just substitue `func.now()` in all these cases?

tomdds: 🤔 For this bugfix, I think keeping the scope of the changes small would be better. But we should definitely make a datetime audit that includes some of these findings and suggestions, like defining a postgres custom utcnow() and determining how to best preserve the kind of testing support that passing dates in like that atst/atat/jobs.py example while also keeping queries efficient.

graham-dds: 4210710b2393396871d2f4b506f3d91b387bef5c to address your initial comment @tomdds

PR #1724, opened 2020-08-11
Perform GPG check when adding RHEL repositories in Dockerfiles

jesse-dds:
https://ccpo.atlassian.net/browse/AT-5381

Gets rid of `--nogpgcheck` options in our base docker files.

Requires us to use a developer subscription and provide credentials as part of the build.

### Other

 - Introduces Build Kit for the base python file for secret passing and faster build file
 - Adds a build script for the base python file
 - Update readme to show how to build the base python file
 - Copy over .gitignore contents to .dockerignore

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,6 @@
 #
 # https://docs.docker.com/engine/reference/commandline/build/#options
 ARG IMAGE
-ARG REDHAT_USERNAME
-ARG REDHAT_PASSWORD
 
 FROM $IMAGE as builder
 
@@ -33,27 +31,6 @@ RUN yum updateinfo && \
 
 # Install the `Python.h` file for compiling certain libraries.
 RUN dnf install python3-devel -y
-
-# Install dependencies for python3-saml
-RUN yum install -y \
-      libxml2-devel \
-      xmlsec1 \
-      xmlsec1-openssl
-
-# Install dev dependencies required to build the xmlsec dependency of python3-saml
-# TODO(tomdds): Find proper gpg secured sources for both of these
-
-#
-#
-# https://developers.redhat.com/articles/renew-your-red-hat-developer-program-subscription
-RUN subscription-manager remove --all
-RUN subscription-manager clean
-RUN subscription-manager register --username $REDHAT_USERNAME --password $REDHAT_PASSWORD
-RUN subscription-manager refresh
-RUN subscription-manager attach --auto
-
-# https://access.redhat.com/articles/4348511#enable
-RUN subscription-manager repos --enable codeready-builder-for-rhel-8-x86_64-rpms
 
 # https://man7.org/linux/man-pages/man1/yum-utils.1.html
 RUN yum repolist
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -93,30 +85,8 @@ RUN groupadd --system -g 101 atat
...
-RUN yum install -y \
-      libxml2-devel \
-      xmlsec1 \
-      xmlsec1-openssl
-
-# Install dev dependencies required to build the xmlsec dependency of python3-saml
-# TODO(tomdds): Find proper gpg secured sources for both of these
-RUN yum install -y ftp://ftp.redhat.com/pub/redhat/rhel/rhel-8-beta/appstream/x86_64/Packages/libtool-ltdl-devel-2.4.6-25.el8.x86_64.rpm --nogpgcheck
-RUN yum install -y http://mirror.centos.org/centos/8/PowerTools/x86_64/os/Packages/xmlsec1-devel-1.2.25-4.el8.x86_64.rpm --nogpgcheck
+RUN dnf install @postgresql -y
```
dandds, line 89: Does this need to be pinned to the postgresql10 package?

python.Dockerfile
```diff
@@ -29,13 +31,32 @@ RUN yum install -y gcc libffi-devel make wget zlib-devel
...
+RUN subscription-manager refresh
+RUN subscription-manager attach --auto
+
+# Enable the CodeReady repository.
+# Allows us to install the xmlsec1-devel package.
+# https://access.redhat.com/articles/4348511#enable
+RUN subscription-manager repos --enable codeready-builder-for-rhel-8-x86_64-rpms
+
+# Install dependencies of python3-saml.
+RUN yum install -y libxml2-devel xmlsec1 xmlsec1-openssl libtool-ltdl-devel xmlsec1-devel
```
dandds, line 56: Good call.

python.Dockerfile
```diff
@@ -29,13 +31,32 @@ RUN yum install -y gcc libffi-devel make wget zlib-devel
...
 
-# Need EPEL to install SQLLite
-# https://fedoraproject.org/wiki/EPEL
-# TODO(heyzoos): Do the GPG check.
-RUN yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm --nogpgcheck
+# Register this machine with a RedHat subscription.
+# Enables us to add the CodeReady repository.
+RUN subscription-manager remove --all
+RUN subscription-manager clean
+# Uses BuildKit preferred method of supplying secrets. These secrets will not
```
dandds, line 38: We still don't need BuildKit for the ATAT image, right? Otherwise we'd have to set the DOCKER_BUILDKIT env var in the CircleCI config.

script/build-docker-image-python-base.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+env DOCKER_BUILDKIT=1 docker build \
+    -t atst:rhel-py \
+    -f python.Dockerfile \
+    --build-arg IMAGE=cloudzerodryrunregistry.azurecr.io/rhelubi:8.2 \
```
dandds, line 6: We should probably use an env var for this since it will change with every iteration of the infra.

README.md
```diff
@@ -482,3 +482,19 @@ fi
...
+- redhat_username.secret
+- redhat_password.secret
+
+Then run the build script.
+
+```
+./script/build-docker-image-python-base.sh
+```
+
+Then publish the image.
```
dandds, line 499: Pedantic, but maybe include how to do that 🙏 .

dandds:
@jesse-dds One other question, just for my understanding: since we're relying on RHEL repos now, yum/dnf are implicitly doing a GPG check, right?

python.Dockerfile
jesse-dds: Nope because the `python.Dockerfile` isn't built during CI

dandds:
@jesse-dds Does it need the updated base image to pass CI?

PR #1725, opened 2020-08-11
[AT-5436] Script to manually run sonar scanner

okaycj-atat:
Script and properties setting to manually run SonarScanner.  Directions on how to use the script are in the REAME file. 

[Jira Story](https://ccpo.atlassian.net/browse/AT-5436)

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,23 @@
+
+
+# This is a project key that is configurable through the SonarQube UI. This
+# uniquely idetifies the project that we're attempting to scan.  This value
+# could be anything, I choose a random string.  This could easily be 
+# "atat-staging". 
+sonar.projectKey=UMhZkXGoDaqaFYssaxopPyV6
+
+# This is the project name that is shown in the UI.  This value is completely
+# for aesthetics.  The value can be change at any time with no issues.  
+sonar.projectName=ATAT
+
+# defaults to 'not provided'
+#sonar.projectVersion=1.0
+ 
+# Path is relative to the sonar-project.properties file. Defaults to .
+#sonar.sources=.
+ 
+# Encoding of the source code. Default is default system encoding
+# sonar.sourceEncoding=UTF-8
+
+# host name and port for where the SonarQube instance is running.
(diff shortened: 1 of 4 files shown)


dandds:
@okaycj-atat Are the creds for SoanarQube in 1Password? I can't find them.

sonar-project.properties
```diff
@@ -0,0 +1,23 @@
...
+#sonar.projectVersion=1.0
+ 
+# Path is relative to the sonar-project.properties file. Defaults to .
+#sonar.sources=.
+ 
+# Encoding of the source code. Default is default system encoding
+# sonar.sourceEncoding=UTF-8
+
+# host name and port for where the SonarQube instance is running.
+sonar.host.url=http://cloudzero-sonarqube.eastus.azurecontainer.io:9000
```
carleton-dds, line 23: Where and how is this instance configured?

okaycj-atat: This instance does not have permanence.  It was stood up by mike as a favor a couple weeks back.  The life of this instance should only be the 14 day trial license.  The intention is this is going away once we get our license and point the ticket for the permanent server. 

For exactly where and how, I'll have to defer to @mikepw.

okaycj-atat:
> @okaycj-atat Are the creds for SoanarQube in 1Password? I can't find them.

I'll put them in 1Password, as I agree they belong there, but they'll be changing once we get the actual license.

sonar-project.properties
dandds: Is it possible for this value to be a variable, @okaycj-atat ? Since we're talking about a new instance it sounds like the hostname might change periodically.

okaycj-atat: > Since we're talking about a new instance it sounds like the hostname might change periodically.

Unless I am not understanding something, I don't think it will.

okaycj-atat: So I think what might need to be made clear is that the _current_ instance doesn't have permanence (this is the one listed in the properties file).  The _new_ instance should have permanence, unless there's some blocker with ops I am unaware of.

okaycj-atat: @carleton-dds I added a description of how to use SonarScanner and linked a document that has notes of how I setup the SonarQube instance.

PR #1727, opened 2020-08-12
adds more operationlized terraform

mikepw:
updates the terraform to that used in dryrun2. Some high level updates include:
* AT-5368 - Improve the bootstrapping process. 
* Updates terraform to use features in TF v0.13.0 (module depends_on)
* Creates postgresql passwords
* generates outputs for ansible or other orchestrator to consume about the environment.


terraform/modules/azure_ad/ad_application.tf
```diff
@@ -9,8 +17,30 @@ resource "azuread_application" "app" {
...
   app_role_assignment_required = false
 
+}
+
+
+resource "azuread_service_principal_password" "sp_password" {
+
+  service_principal_id = azuread_service_principal.principal.id
+  value                = random_password.password.result
+  end_date             = "2099-01-01T01:02:03Z"
```
dandds, line 31: Do we have to provide values for `end_date`? Does it default to something reasonable if we don't?

terraform/modules/k8s-private/main.tf
```diff
@@ -70,60 +42,71 @@ resource "azurerm_route" "internet_route" {
   next_hop_type       = "Internet"
 }
 
-resource "azurerm_route" "vnet_route" {
-  name                = "${var.name}-${var.environment}-virtual-network"
-  resource_group_name = var.rg
-  route_table_name    = azurerm_route_table.route_table.name
-  address_prefix      = var.vpc_address_space
-  next_hop_type       = "VnetLocal"
-}
```
dandds, line 79: Can you fill me in on why we're dropping all this networking stuff for the private cluster?

terraform/modules/k8s/main.tf
```diff
@@ -41,45 +42,3 @@ resource "azurerm_kubernetes_cluster" "k8s" {
...
-      enabled = true
-    }
-  }
-  metric {
-    category = "AllMetrics"
-    retention_policy {
-      enabled = true
-    }
-  }
-}
```
dandds, line 85: Don't we need this for logging from the cluster?

terraform/modules/k8s/variables.tf
```diff
@@ -82,3 +70,15 @@ variable "private_cluster_enabled" {
   default     = false
   type        = bool
 }
+
+variable "node_resource_group" {}
+
+variable "client_id" {}
+variable "client_secret" {}
+variable "client_object_id" {}
```
dandds, line 78: I think the types/descriptions are nice, but up to you.

terraform/modules/keyvault/main.tf
```diff
@@ -97,25 +97,3 @@ resource "azurerm_key_vault_access_policy" "keyvault_admin_policy" {
...
-    }
-  }
-  metric {
-    category = "AllMetrics"
-
-    retention_policy {
-      enabled = true
-    }
-  }
-}
```
dandds, line 121: Same question as with the k8s cluster: isn't this what points the logs into our LA Workspace?

terraform/providers/dryrun2/private-k8s.tasdf
```diff
@@ -0,0 +1,24 @@
...
+  log_analytics_workspace_id = module.logs.workspace_id
+  service_dns                = var.private_aks_service_dns
+  docker_bridge_cidr         = var.private_aks_docker_bridge_cidr
+  service_cidr               = var.private_aks_service_cidr
+  subnet_cidr                = var.private_k8s_subnet_cidr
+  vnet_id                    = module.vpc.id
+  vpc_name                   = module.vpc.vpc_name
+  aks_ssh_pub_key_path       = var.aks_ssh_pub_key_path
+  vpc_address_space          = module.vpc.address_space
+}
```
dandds, line 24: What's the extension on this? 🤔

deploy/overlays/cloudzero-dryrun-staging/envvars.yml
```diff
@@ -24,21 +24,21 @@ metadata:
...
   FLASK_ENV: staging
-  PGDATABASE: cloudzero_dryrun
-  PGHOST: 191.238.6.43
-  PGUSER: atat@cloudzero-dryrun-sql
+  PGDATABASE: cloudzero_dryrun2 # TODO(heyzoos) Parameterize from terraform!
+  PGHOST: http://cloudzero-dryrun2-sql.postgres.database.azure.com/ # TODO(heyzoos) Parameterize from terraform!
+  PGUSER: atat@cloudzero-dryrun2-sql # TODO(heyzoos) Parameterize from terraform!
   PGSSLMODE: require
-  REDIS_HOST: cloudzero-dryrun-redis.redis.cache.windows.net:6380
+  REDIS_HOST: cloudzero-dryrun2-redis.redis.cache.windows.net:6380 # TODO(heyzoos) Parameterize from terraform!
```
dandds, line 41: I don't think we should be changing this overlay if we're also adding a new overlay for dryrun2. We'll still need the old value here to manage the dryrun environment.

deploy/overlays/cloudzero-pwdev-dryrun2/autoscaling.yml
```diff
@@ -0,0 +1,16 @@
...
+  minReplicas: 1
+  maxReplicas: 2
+---
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: atst-worker
+spec:
+  minReplicas: 1
+  maxReplicas: 2
```
dandds, line 16: The filename pattern I've used in the past for these is "cloudzero-[infra environment name]-[k8s namespace name]", which would make this clouzero-dryrun2-staging. "pwdev" is our original infra environment, so it doesn't really correspond to anything anymore.

terraform/modules/azure_ad/ad_application.tf
mikepw: an `end_date` must be specified.. and I'm not seeing a default. (https://www.terraform.io/docs/providers/azuread/r/service_principal_password.html)

deploy/overlays/cloudzero-pwdev-dryrun2/autoscaling.yml
mikepw: agree.. I think this is @jesse-dds change and mistakenly made it into the world.  I want to remove environment specific working overlays and instead go towards generating 'reference overlays' which would eliminate the need for multiple working directories of k8s config..

deploy/overlays/cloudzero-dryrun-staging/envvars.yml
mikepw: agree. @jesse-dds can you change back?

terraform/providers/dryrun2/private-k8s.tasdf
mikepw: haha loving that WIP label rn..

deploy/azure/crls-sync.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: crls
   namespace: atat
 spec:
-  schedule: "0 * * * *"
+  schedule: "29 * * * *"
```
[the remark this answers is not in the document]
mikepw: done

pyproject.toml
```diff
@@ -38,6 +38,9 @@ webassets = "*"
 werkzeug = "*"
 email-validator = '*'
 python3-saml = "^1.9.0"
+kubernetes = "^11.0.0"
+ansible = "^2.9.11"
+openshift = "^0.11.2"
```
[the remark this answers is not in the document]
mikepw: done

terraform/modules/azure_ad/ad_application.tf
dandds: The docs say it's optional; there's also `end_date_relative`, which might suit us better.

terraform/modules/k8s/main.tf
mikepw: we ran into a real issue that this log workspace already existed causing applies to fail.  It was very very strange.  I removed it to alleviate and alas the logs were still there.  I will re-run with this included to make sure we get what we need out of logging.

terraform/modules/azure_ad/ad_application.tf
mikepw: one or the other must be specified is what I should've said. If someone were to read an `end_date_relative` that is not familiar with the date/time of deployment they might be guessing when this service principal password expires.

terraform/modules/k8s-private/main.tf
mikepw: I will test again but I'm fairly certain that the extra networking was superfluous when the AKS Cluster becomes private. The gateway was for egress only access but since we'll need a route to the internet for ingress connections to AKS Cluster endpoints I think that doesn't make sense anymore.

ansible/roles/azure_keyvault/tasks/secret.yml
```diff
@@ -0,0 +1,27 @@
+
+
+
```
carleton-dds, line 3: What's with all the blank lines in this file? Also is this valid ansible YML? looks like you got a lot of shell commands too

terraform/providers/dryrun3/bastion.tfasdf
```diff
@@ -0,0 +1,30 @@
+
```
carleton-dds, line 1: What's with the file extension on this?

ansible/roles/terraform/tasks/setup.yml
```diff
@@ -0,0 +1,56 @@
...
+
+
+- name: tf apply (no modules)
+  terraform:
+    force_init: "{{init_tf}}"
+    project_path: "{{tf_base_dir}}/{{az_environment}}"
+    state: "present"
+    variables: "{{ app_kv_tf }}"
+    backend_config: "{{ b_end }}"
+  register: tf_run
```
jesse-dds, line 46: Potentially no longer necessary since we re-register the outputs in "show_outputs.yml"

jesse-dds:
Could we rename all of the "show_outputs.yml" files to "register_outputs.yml" to better match what they actually do?

ansible/databases.yml
```diff
@@ -0,0 +1,9 @@
+
+
+- hosts: localhost
+  roles:
+    - role: db
+      vars:
+        db_init_details: "{{ tf_app_env_outputs.outputs }}"
```
tomdds, line 7: Does `db_init_details` need to go away since it's not used in `postgres.yml` in lieu of directly referencing `tf_app_env_outputs`?

ansible/roles/db/tasks/postgres.yml
```diff
@@ -0,0 +1,98 @@
...
+
+  - name: Create (or update) PostgreSQL Database
+    azure_rm_postgresqldatabase:
+      resource_group: "{{tf_app_env_outputs.postgres_resource_group_name.value}}"
+      server_name: "{{ tf_app_env_outputs.atat_database_instance_name.value }}"
+      name: "{{ tf_app_env_outputs.atat_database_name.value }}"
+      secret: "{{ lookup('env','SUBSCRIPTION_SECRET') }}"
+      client_id: "{{ lookup('env','SUBSCRIPTION_CID') }}"
+      subscription_id: "{{ lookup('env','SUBSCRIPTION_ID')}}"
+      tenant: "{{ lookup('env','SUBSCRIPTION_TENANT')}}"
```
jesse-dds, line 23: We need to make sure we enable the GUID extension here since we're not running the .py script any longer

tomdds, line 23: What do we need to ensure the "uuid-ossp" extension is enabled in the database and done in `script/create_database.py`

tomdds:
```
conn.execute(f'CREATE EXTENSION IF NOT EXISTS "{extension_name}";')
```

tomdds: Can we use this with azure's hosted postgres? https://docs.ansible.com/ansible/2.4/postgresql_ext_module.html

ansible/roles/db/tasks/postgres.yml
```diff
@@ -0,0 +1,98 @@
...
+        SERVER_NAME: "{{ tf_app_env_outputs.atat_database_instance_name.value }}"
+        SERVER_GROUP: "{{ tf_app_env_outputs.postgres_resource_group_name.value }}"
+        SERVER_ADMIN_NAME: "{{ tf_app_env_outputs.postgres_root_user_name.value }}"
+        PGUSER: "{{ tf_app_env_outputs.postgres_root_user_name.value }}"
+        PGPASSWORD: "{{ tf_app_env_outputs.postgres_root_password.value }}"
+        POSTGRES_ROOT_USER: "{{ tf_app_env_outputs.postgres_root_user_name.value  }}"
+        POSTGRES_ROOT_PASS: "{{  tf_app_env_outputs.postgres_root_password.value  }}"
+        PGSSLROOTCERT: /tmp/pgcaroot/pgcaroot.cert
+        ATAT_PGUSER: "{{ tf_app_env_outputs.atat_user_name.value }}"
+        ATAT_PGPASSWORD: "{{  tf_app_env_outputs.atat_user_password.value  }}"
```
tomdds, line 80: Branches from the dryrun doc, can't connect using atat user creds here, but dryrun dock specifies they should be used for this.

ansible/.gitignore
```diff
@@ -0,0 +1,2 @@
+**/*.sh
```
carleton-dds, line 1: Why is there a second .gitignore? you should be able to put all this in the root one

poetry.lock
```diff
@@ -1,2379 +0,0 @@
-[[package]]
```
dandds, line 1: Wait, why are we removing the lockfile? Without a lockfile, we'll have a much harder time creating reproducible builds.

dandds (changes requested):
Happy to help poke into whatever the problem is, but I think there has to be a better solution than removing the application's lockfile.

dandds (approved):
🎉

jesse-dds (approved):

carleton-dds (approved):

tomdds (approved):
:shipit:

PR #1730, opened 2020-08-12
Rename `AZURE_ACCOUNT_NAME` to `AZURE_STORAGE_ACCOUNT_NAME`

jesse-dds:
`AZURE_ACCOUNT_NAME` is an initially confusing variable name because it implies user accounts.


graham-dds:
Love a good name change.

I noticed this line https://github.com/dod-ccpo/atst/pull/1730/files#diff-04c6e90faac2675aa89e2176d2eec7d8R367

Did we update the env var name in Circle as well?

dandds:
Good catch @graham-dds . The README is wrong, I should have updated it when I made this change: https://github.com/dod-ccpo/atst/pull/1668/files#diff-1d37e48f9ceff6d8030570cd36286a61L146-L147

PR #1738, opened 2020-08-14
[AT-5341] Handle _VERIFICATION state machine steps

graham-dds:
PR 1/2 for [AT-5341](https://ccpo.atlassian.net/browse/AT-5341)

Some API calls in Azure kick off ["async operations"](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations). In the initial call to and endpoint, we receive a response with headers that contain a URL to check on the status of the ongoing operation.  

We handle portfolio provisioning stages that encounter async operations by using `_CREATION` and `_VERIFICATION` steps in the state machine. The `_CREATION` step handles the initial call to the endpoint and saves the URL to check, and the `_VERIFICATION` step hits the this URL for the payload of the operation.

It appears we made an assumption that by the time we got to the `_VERIFICATION` step, the async operation would be complete and we'd be able to return the desired payload. But, there's a chance that we could try a verification URL and the job isn't completed yet. This would mean that we'd either falsely report that the job was successful before it completed, or possibly that the payload would not have all the expected values and a Pydantic validation error would be raised.

In this PR, we add infrastructure to the state machine to account for this case of incomplete operations. If we try a URL in the verification step and the operation is still "In Progress", we reset the state machine from the current stage's _IN_PROGRESS state to the previous stage's _CREATED state. This effectively allows for the verification URL to be tried again after the next query for portfolios pending provisioning.

For this PR, we've only applied this new pattern to the `create_task_order_billing_verification` method. If this pattern looks sound for the state machine and Azure Cloud Provider, the next PR will continue the pattern to other stages that use the `_CREATION` & `_VERIFICATION` method for handling async operations.

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -430,11 +430,14 @@ class AzureCloudProvider(CloudProviderInterface):
             response.raise_for_status()
             response_json = response.json()
             status = response_json["status"]
-            if status == "Succeeded":
+            if status == AsyncOperationStatus.SUCCEEDED.value:
                 resp = session.get(result_url)
                 resp.raise_for_status()
                 return resp.json()
-            elif status in ("Failed", "Canceled"):
+            elif status in (
+                AsyncOperationStatus.FAILED.value,
+                AsyncOperationStatus.CANCELED.value,
+            ):
                 error_message = f"{response_json['error']['message']}\nError code: {response_json['error']['code']}"
                 raise ResourceProvisioningError("management group", f"{error_message}")
             else:
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1,5 +1,6 @@
 import json
 import time
+from enum import Enum
 from contextlib import contextmanager
 from functools import wraps
 from secrets import token_hex, token_urlsafe
@@ -21,6 +22,7 @@ from .exceptions import (
     UserProvisioningException,
 )
 from .models import (
+    class_to_stage,
     AdminRoleDefinitionCSPPayload,
     AdminRoleDefinitionCSPResult,
     ApplicationCSPPayload,
@@ -163,6 +165,15 @@ class AzureSDKProvider(object):
 
         self.cloud = AZURE_PUBLIC_CLOUD
         self.requests = requests
+
+
+class AsyncOperationStatus(Enum):
+    """https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations"""
+
+    SUCCEEDED = "Succeeded"
+    FAILED = "Failed"
+    CANCELED = "Canceled"
+    IN_PROGRESS = "InProgress"
 
 
 class AzureCloudProvider(CloudProviderInterface):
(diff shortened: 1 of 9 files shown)


carleton-dds:
I'll add my unofficial ✅ too

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -742,6 +753,43 @@ def create_task_order_billing_creation(
...
+        elif response.status_code == 200:
+            status = response_json["status"]
+            if status == AsyncOperationStatus.SUCCEEDED.value:
+                return verification_response_model(**response_json)
+            elif status in (
+                AsyncOperationStatus.FAILED.value,
+                AsyncOperationStatus.CANCELED.value,
+            ):
+                error_message = f"{response_json['error']['message']}\nError code: {response_json['error']['code']}"
+                raise ResourceProvisioningError(resource, f"{error_message}")
```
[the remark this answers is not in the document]
graham-dds: Good call. https://github.com/dod-ccpo/atst/pull/1738/commits/f07aabf83bbfaa02a164cca8f597cc3323a0b1e7 and https://github.com/dod-ccpo/atst/pull/1738/commits/3d5caccbbf167ff0f921feccbadd13d2a74ab749

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -742,6 +753,43 @@ def create_task_order_billing_creation(
...
             return TaskOrderBillingVerificationCSPResult(**result.json())
 
+    def _handle_async_operation_response(
+        self, resource, response, creation_response_model, verification_response_model
+    ):
+        """Handle the response of an Async operation
+        https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations
+
+        Args:
+            resource (str): the a description of the resource being provisioned
```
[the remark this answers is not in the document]
graham-dds: 👍 https://github.com/dod-ccpo/atst/pull/1738/commits/c8f0ecab686a4ca951451a95f7520000658e7ff2

atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -165,6 +166,13 @@ def __init__(self):
         self.requests = requests
 
 
+class AsyncOperationStatus(Enum):
+    SUCCEEDED = "Succeeded"
+    FAILED = "Failed"
+    CANCELED = "Canceled"
+    IN_PROGRESS = "In Progress"
```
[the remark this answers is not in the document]
graham-dds: Good call https://github.com/dod-ccpo/atst/pull/1738/commits/bf60840baf93291755755fee7fce1cc76448e6e7

Yeah, strangely `In Progress` isn't listed [here](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations#azure-asyncoperation-request-and-response), but maybe it's a part of the `{resource provider values}`? Like if you scroll down to the example response body in [this section](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/async-operations#start-virtual-machine-202-with-azure-asyncoperation), they have the status as `InProgress`. I also see `InProgress` [over here](https://docs.microsoft.com/en-us/javascript/api/@azure/arm-frontdoor/azureasyncoperationresult?view=azure-node-latest#properties), which isn't strictly related to the docs you linked to above, but it seems similar enough 🤷.

All that being said, I did update the enum from `In Progress` to `InProgress`.

atat/domain/csp/cloud/models.py
```diff
@@ -23,6 +23,25 @@
...
+    """Given a CSP model class, derive the stage name
+
+    Args:
+        klass (pydantic.BaseModel): a pydantic CSP data model class
+
+    Returns:
+        str: the stage name
+    """
+    class_name = klass.__name__
+    payload_class_suffix = re.compile(r"(Creation|Verification)?CSP.*")
```
tomdds, line 40: Maybe worth it to move this out to a constant so we don't have to compile it every pass?

atat/domain/csp/cloud/models.py
```diff
@@ -40,10 +59,19 @@ class AliasModel(BaseModel):
...
+    reset_stage: bool = False
+
     class Config:
         alias_generator = snake_to_camel
         allow_population_by_field_name = True
 
+    def dict(self, *args, **kwargs):
+        kwargs.setdefault("exclude")
+        if kwargs["exclude"] is None:
+            kwargs["exclude"] = set()
```
tomdds, line 71: I think this would give you the same result:
```suggestion
        kwargs.setdefault("exclude", default=set())
```

graham-dds: I thought about this too, but I found that in this instance, `exclude=None` is mysteriously being passed in as a kwarg, so even if I supplied a default with `setdefault`, I had to check for `None` anyway.

atat/domain/csp/cloud/models.py
graham-dds: https://github.com/dod-ccpo/atst/pull/1738/commits/9b12936a4cc89d4b300c97cf2e3eab9364997db1

graham-dds:
Did some rebasing to tidy up the history. After I was happy with all of the changes, I ran `git diff origin/audit-verify-steps` from my local branch and it didn't return any output. So no new changes, just moving a few things around.

PR #1742, opened 2020-08-17
[AT-4902] Full verification of TLS connection to Redis server.

dandds:
This change allows us to configure a full verification of the Redis
SSL/TLS connection. Under the hood, the Python client library for Redis,
redis-py, is using the Python ssl module to make the connection. We can
pass the same configuration values we would normally provide to the ssl
module. In ATAT's config, these are represented by REDIS_SSLMODE and
REDIS_SSLCHECKHOSTNAME. The former will force the client to verify the
server's certificate against trusted certificate authorities, while the
latter verifies that the server's hostname if what's listed on the
server's certificate. These are represented as query arguments on
the URI we construct for the Redis connection.

Unfortunately, Celery requires a Redis URI for the BROKER_URL setting,
so we can't use a simple dictionary of options to make the connection.
We have to do some ugly string formatting instead.

I've updated the documentation and Kubernetes config, too.

I tested this in the Python REPL on a running instance of the site and was able to connect to our Redis instance with REDIS_SSLMODE=required and REDIS_SSLCHECKHOSTNAME=True.


config/base.ini
```diff
@@ -60,6 +60,7 @@ PORT=8000
 REDIS_HOST=localhost:6379
 REDIS_PASSWORD
 REDIS_SSLMODE
```
graham-dds, line 62: Considering this line in app.py:
```python
ssl_mode = config.get("default", "REDIS_SSLMODE") or "none"
```
Should we provide the default value of `none` here instead?

atat/app.py
```diff
@@ -264,34 +264,24 @@ def make_config(direct_config=None):
     config.set("default", "DATABASE_URI", database_uri)
 
     # Assemble REDIS_URI value
-    redis_use_tls = config["default"].getboolean("REDIS_TLS")
+    redis_use_ssl = config["default"].getboolean("REDIS_TLS")
     redis_uri = "redis{}://{}:{}@{}".format(  # pragma: allowlist secret
-        ("s" if redis_use_tls else ""),
+        ("s" if redis_use_ssl else ""),
         (config.get("default", "REDIS_USER") or ""),
         (config.get("default", "REDIS_PASSWORD") or ""),
```
graham-dds, line 271: Is there a reason we don't use the [`fallback` kwarg](https://docs.python.org/3/library/configparser.html#fallback-values) for these config `get() methods`?

atat/app.py
```diff
@@ -264,34 +264,24 @@ def make_config(direct_config=None):
...
-            "optional": "CERT_OPTIONAL",
-        }
-        celery_tls_mode_str = kombu_modes[tls_mode_str]
-        celery_uri = f"{celery_uri}/?ssl_cert_reqs={celery_tls_mode_str}"
+    if redis_use_ssl:
+        ssl_mode = config.get("default", "REDIS_SSLMODE") or "none"
+        ssl_checkhostname = "false"
+
+        if config["default"].getboolean("REDIS_SSLCHECKHOSTNAME"):
+            ssl_checkhostname = "true"
```
graham-dds, line 279: Similar to my comment above, maybe this could be:
```python
ssl_checkhostname = config["default"].getboolean("REDIS_SSLCHECKHOSTNAME", fallback="false")
```
or is this another case where we should provide a default in `base.ini`?

atat/app.py
dandds: `fallback` would work if REDIS_USER and REDIS_PASSWORD don't exist in the config object at all (similar to a key not being in a dictionary). In this case they are in the config object and are explicitly set to `None`, so we get `None` instead of whatever value we set for `fallback`.

atat/app.py
dandds: There might be a better way to handle this, but we need these values explicitly coerced to lowercase string `"false"` or `"true"`.

config/base.ini
dandds: 👍 I just amended the commit in-place.

PR #1745, opened 2020-08-19
[AT-5341] Use async handler in remaining _VERIFICATION steps

graham-dds:
PR 2/2 for [AT-5341](https://ccpo.atlassian.net/browse/AT-5341) (Initial PR: #1738 )

In 225e3f2a8 we created a handler for async operations to be used in `_VERIFICATION` state machine steps, and in 18c12cf22 we used this new method for task order billing. In this PR, we ensure that all remaining `_VERIFICATION` steps use the async handler.

### DIFF
--- a/atat/domain/csp/cloud/models.py
+++ b/atat/domain/csp/cloud/models.py
@@ -622,7 +622,7 @@ class ProductPurchaseVerificationCSPResult(AliasModel):
     premium_purchase_date: str
 
     @root_validator(pre=True)
-    def extract_premium_purchase_date(cls, values):
+    def populate_premium_purchase_date(cls, values):
         if "premium_purchase_date" in values:
             return values
         try:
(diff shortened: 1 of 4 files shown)


atat/domain/csp/cloud/models.py
```diff
@@ -621,6 +621,16 @@ class ProductPurchaseVerificationCSPPayload(BaseCSPPayload):
 class ProductPurchaseVerificationCSPResult(AliasModel):
     premium_purchase_date: str
 
+    @root_validator(pre=True)
+    def extract_premium_purchase_date(cls, values):
```
jesse-dds, line 625: This is supremely nitpicky so I'm going to approve anyways. But this isn't an `extract` function to me, it's more something that `ensure`s a property exist. So maybe `ensure_premium_purchase_date` is a more fitting name?

graham-dds: Good call. Went with `populate_<property_name>`

PR #1761, opened 2020-08-28
[AT-5257] Add isort to formatting

graham-dds:
Add import sorting to `script/format`. This will write changes for development and enforce consistent import order in CI.

For now, only enforces import ordering in the `atat` and `tests` directories.


alembic/env.py
```diff
@@ -22,19 +30,14 @@
 # my_important_option = config.get_main_option("my_important_option")
 # ... etc.
 
-import sys
-from unipath import Path
```
dandds, line 26: This hits a pretty significant hiccough I didn't foresee. In some cases (here and in some of the scripts) we explicitly alter the path so that we can import things from the `atat` module. `isort` is rearranging things so that all of the path modifications happen after all of the import statements, which breaks some of the import statements.

graham-dds: Good catch. As discussed in Slack, for now we'll just enforce sorting in `atat` and `tests`.

dandds (approved):
👍

tomdds (approved):
I ABC what you did here and I DEFinitely approve

PR #1762, opened 2020-08-28
[AT-5358] Add polling to manual provisioning

okaycj-atat:
This PR is to actually poll for async results within the manual provisioning scripts.  

Open questions:
- Should we check if the "URL" is in the payload before starting the polling?
- Is three retries enough?

### DIFF
--- a/script/provision/b_setup_billing.py
+++ b/script/provision/b_setup_billing.py
@@ -11,34 +11,21 @@ from atat.domain.csp.cloud.models import (
 from script.provision.provision_base import handle, verify_async
 
 
-def poll_billing(csp, inputs, csp_response):
-    """Polls billing endpoint for async response three times.  If verify url is not 
-    available, the csp_response payload will be returned. 
-
-    Args:
-        csp: CSP Class object
-        inputs: Json string
-        csp_response: Result from billing profile creation
-
-    Returns:
-        Response from billing profile verification or csp_response payload
-    """
-    verify_url = csp_response.get("billing_profile_verify_url")
-    csp_method = csp.create_billing_profile_verification
-    payload = BillingProfileVerificationCSPPayload(
-        **{**inputs.get("initial_inputs"), **inputs.get("csp_data"), **csp_response,}
-    )
-    return verify_async(verify_url, csp_method, payload, csp_response)
-
-
 def setup_billing(csp, inputs):
     create_billing_profile = BillingProfileCreationCSPPayload(
         **{**inputs.get("initial_inputs"), **inputs.get("csp_data")}
     )
+    result = csp.create_billing_profile_creation(create_billing_profile).dict()
 
-    result = csp.create_billing_profile_creation(create_billing_profile)
-    polling_result = poll_billing(csp, inputs, result.dict())
-    return polling_result.dict()
+    # If there is a verify URL, then we need to poll for the result.
+    if result.get("billing_profile_verify_url"):
+        csp_method = csp.create_billing_profile_verification
+        payload = BillingProfileVerificationCSPPayload(
+            **{**inputs.get("initial_inputs"), **inputs.get("csp_data"), **result,}
+        )
+        result = verify_async(csp_method, payload)
+
+    return result
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


script/provision/b_setup_billing.py
```diff
@@ -9,20 +9,36 @@
...
-        get_billing_profile = BillingProfileVerificationCSPPayload(
-            **{**inputs.get("initial_inputs"), **inputs.get("csp_data"), **csp_response}
-        )
-        try:
-            return csp.create_billing_profile_verification(get_billing_profile)
-        except GeneralCSPException:
-            time.sleep(10)
-            return poll_billing(csp, inputs, csp_response)
+        retries = 3
+        for _ in range(retries):
```
graham-dds, line 29: In these steps, I wouldn't limit polling by some number of retries. Instead, I'd just use a `while True` loop since all of our bases are covered:
- If `reset_stage` is `True`, we try it again
- If `reset_stage` is `False`, we return the payload
- If we receive an error when `GET`ing the polling URL, the `AzureCloudProvider` method that we're calling will [throw an error](https://github.com/dod-ccpo/atst/blob/staging/atat/domain/csp/cloud/azure_cloud_provider.py#L790-#L796)

script/provision/f_purchase_aadp.py
```diff
@@ -9,20 +9,36 @@
...
+                    **inputs.get("initial_inputs"),
+                    **inputs.get("csp_data"),
+                    **csp_response,
+                }
+            )
+            response = csp.create_product_purchase_verification(purchase_premium)
+            if response.reset_stage:
+                time.sleep(10)
+            else:
+                return response
```
dandds, line 41: It looks like the only things that change between the three new chunks of code you've added is the payload class and csp method. What do you think about moving this functionality to a utility function in `provision_base.py` and passing those as arguments?

script/provision/b_setup_billing.py
```diff
@@ -9,20 +9,36 @@
...
+            get_billing_profile = BillingProfileVerificationCSPPayload(
+                **{
+                    **inputs.get("initial_inputs"),
+                    **inputs.get("csp_data"),
+                    **csp_response,
+                }
+            )
+            response = csp.create_billing_profile_verification(get_billing_profile)
+            if response.reset_stage:
+                time.sleep(10)
```
graham-dds, line 39: Instead of just using a hardcoded number here, you might consider using the [`<stage_name>_retry_after` property](https://github.com/dod-ccpo/atst/blob/a74c0f1d96d525213a58bca89d856277fc9d7970/atat/domain/csp/cloud/models.py#L181) that should be present on the response model.

graham-dds:
> Open questions:
> 
>     * Should we check if the "URL" is in the payload before starting the polling?

Could you expand on this question? Not sure I'm following.

>     * Is three retries enough?

Commented on this in my review 👍

okaycj-atat:
> > Open questions:
> > ```
> > * Should we check if the "URL" is in the payload before starting the polling?
> > ```
> 
> Could you expand on this question? Not sure I'm following.

There's a check for a validation url, it's the outside if/else statment.  

```
    if csp_response.get("billing_profile_verify_url") is not None:
       ...
    else:
        return csp_response
```

Does this check still need to happen?  Is there a chance this url will not be in the payload?

graham-dds:
> > > Open questions:
> > > ```
> > > * Should we check if the "URL" is in the payload before starting the polling?
> > > ```
> > 
> > 
> > Could you expand on this question? Not sure I'm following.
> 
> There's a check for a validation url. It's the outside if/else statement.
> 
> ```
>     if csp_response.get("billing_profile_verify_url") is not None:
>        ...
>     else:
>         return csp_response
> ```
> 
> Does this check still need to happen? Is there a chance this url will not be in the payload?

Chatted about this on Zoom. This check should happen, but before we ever attempt to poll. Example:
```python
def setup_billing(csp, inputs):
    create_billing_profile = BillingProfileCreationCSPPayload(
        **{**inputs.get("initial_inputs"), **inputs.get("csp_data")}
    )
    result = csp.create_billing_profile_creation(create_billing_profile)
    if result.get("billing_profile_verify_url"):
        polling_result = poll_billing(csp, inputs, result.dict())
        return polling_result.dict()
    else:
        return result.dict()
```

PR #1764, opened 2020-08-28
[AT-5453] Validate elevated access context manager

graham-dds:
This PR contains additional tests and logging that make assertions about the behavior of the elevate access context manager. 

In #1714, we initially observed some flakiness in portfolio provisioning functions that relied on the elevated access token context manager. The goal with this PR is to provide full test coverage for this context manager so that we prevent such flakiness in the future and are sure that it works as expected.

### DIFF
--- a/tests/domain/cloud/test_hybrid_csp.py
+++ b/tests/domain/cloud/test_hybrid_csp.py
@@ -6,6 +6,7 @@ import pendulum
 import pytest
 
 from atat.domain.csp import CSP
+from atat.domain.csp.cloud.exceptions import UnknownServerException
 from atat.domain.csp.cloud.hybrid_cloud_provider import HYBRID_PREFIX
 from atat.domain.csp.cloud.models import (
     CostManagementQueryCSPPayload,
@@ -80,6 +81,10 @@ class TestIntegration:
     def tenant_id(self, portfolio):
         return portfolio.csp_data["tenant_id"]
 
+    @pytest.fixture(scope="function")
+    def user_object_id(self, portfolio):
+        return portfolio.csp_data["user_object_id"]
+
     @pytest.fixture(scope="session")
     def user(self):
         first_name = f"test-user-{uuid4()}"
@@ -114,6 +119,15 @@ class TestIntegration:
     def state_machine(self, app, csp, portfolio):
         return PortfolioStateMachineFactory.create(portfolio=portfolio, cloud=csp)
 
+    def _get_management_group(self, csp, tenant_id, management_group_id):
+        sp_token = csp.azure._get_tenant_principal_token(tenant_id)
+        response = csp.azure.sdk.requests.get(
+            f"{csp.azure.sdk.cloud.endpoints.resource_manager}{management_group_id}?api-version=2020-02-01",
+            headers=make_auth_header(sp_token),
+        )
+        response.raise_for_status()
+        return response.json()
+
     @pytest.mark.depends(name="portfolio")
     def test_hybrid_provision_portfolio(self, state_machine: PortfolioStateMachine):
         csp_data = {}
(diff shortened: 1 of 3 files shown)


tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -204,12 +204,22 @@ def test_hybrid_do_create_environment_role_job(self, session, csp, env_role):
...
 
-    @pytest.mark.depends(on=["environment_role"])
+    @pytest.mark.depends(name="disable_user", on=["environment_role"])
     def test_hybrid_disable_user(self, csp, user, tenant_id, env_role):
         env_role_cloud_id = env_role.cloud_id
-        disable_user_result = csp.azure.disable_user(tenant_id, env_role_cloud_id)
+        disable_user_result = csp.disable_user(tenant_id, env_role_cloud_id)
         assert disable_user_result["id"] == env_role_cloud_id
 
+    @pytest.mark.depends(on=["disable_user"])
```
dandds, line 213: Is there a reason to chain this to `disable_user` specifically? In my cursory read, `disable_user` doesn't require elevated access.

I know that we specifically rely on ordered tests here because of the way provisioning happens, but it might be useful for this test to be more self-contained and demonstrative. What if you got an elevated access token, made a call successfully, and then tried making the same call outside the context manager?

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -115,6 +115,15 @@ def env_role(self, app_role, environment):
...
         return PortfolioStateMachineFactory.create(portfolio=portfolio, cloud=csp)
 
+    def _get_management_group(self, csp, tenant_id, management_group_id):
+        sp_token = csp.azure._get_tenant_principal_token(tenant_id)
+        response = csp.azure.sdk.requests.get(
+            f"{csp.azure.sdk.cloud.endpoints.resource_manager}{management_group_id}?api-version=2020-02-01",
+            headers=make_auth_header(sp_token),
+        )
+        response.raise_for_status()
+        return response.json()
```
dandds, line 125: Did you move this just to consolidate tests vs helper functions?

graham-dds: Yep, exactly

tests/domain/cloud/test_hybrid_csp.py
graham-dds: Chatted about this on Zoom ☝️ 
I'll keep this test, but chain it on "portfolio"
```python
@pytest.mark.depends(on=["portfolio"])
```

And then write another test that is "more self-contained and demonstrative". It will also depend on the portfolio test, but just so we have infrastructure in place for the integrative nature of the test.

graham-dds: Both of these tests now present here: https://github.com/dod-ccpo/atst/pull/1764/commits/d77b9d80fb9484387bb913a7f33c8819122e4e7d

graham-dds:
> is elevated access limited to portfolio provisioning (which doesn't entail concurrent jobs)?

Yep, this is the case for now

tests/domain/cloud/test_hybrid_csp.py
```diff
@@ -207,7 +245,7 @@ def test_hybrid_do_create_environment_role_job(self, session, csp, env_role):
     @pytest.mark.depends(on=["environment_role"])
     def test_hybrid_disable_user(self, csp, user, tenant_id, env_role):
         env_role_cloud_id = env_role.cloud_id
-        disable_user_result = csp.azure.disable_user(tenant_id, env_role_cloud_id)
+        disable_user_result = csp.disable_user(tenant_id, env_role_cloud_id)
```
tomdds, line 248: What makes this `csp` different than the fixtures that the other tests receive?

tomdds (commented):
Just one question, looks good otherwise

tests/domain/cloud/test_hybrid_csp.py
graham-dds: All of these tests receive a hybrid cloud provider as the `csp` fixture. Since these are hybrid tests and the hybrid cloud provider has its own `disable_user` method present, this test should call it instead of the underlying Azure cloud provider method.

tomdds (approved):

PR #1771, opened 2020-09-01
Fix format string

jesse-dds:
https://ccpo.atlassian.net/browse/AT-5464

Fixes "Last Login: " being empty due to an outdated format string.

image


templates/footer.html
```diff
@@ -2,8 +2,9 @@
...
-    <div class="app-footer__login">
-      {{ "footer.login" | translate }} <local-datetime timestamp='{{ g.last_login }}' format='MMM D YYYY H:mm Z'></local-datetime>
-    </div>
+  <div class="app-footer__login">
+    {{ "footer.login" | translate }} <local-datetime timestamp='{{ g.last_login }}' format='MMM d yyyy H:mm XXX'>
+    </local-datetime>
+  </div>
   {% endif %}
-</footer>
+</footer>
```
graham-dds, line 10: I would say after adding a newline here, we'll be good to go 👍

PR #1772, opened 2020-09-03
put params at top of circle ci config so they're accessible via api

mikepw:
I scoped the original parameters too deep these bring them up to the top level as per circle ci docs:

https://circleci.com/docs/2.0/pipeline-variables/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,13 +105,13 @@ commands:
           command: sleep 3
       - run:
           name: Create database
-          command: "docker exec postgres createdb -U postgres << pipeline.parameters.pgdatabase >>"
+          command: "docker exec postgres createdb -U postgres << parameters.pgdatabase >>"
       - run:
           name: Apply migrations
-          command: docker run --network atat -e PGDATABASE=<< pipeline.parameters.pgdatabase >> << pipeline.parameters.container_env >> atat:builder .venv/bin/python .venv/bin/alembic upgrade head
+          command: docker run --network atat -e PGDATABASE=<< parameters.pgdatabase >> << pipeline.parameters.container_env >> atat:builder .venv/bin/python .venv/bin/alembic upgrade head
       - run:
           name: Apply the default permission sets
-          command: docker run --network atat -e PGDATABASE=<< pipeline.parameters.pgdatabase >> << pipeline.parameters.container_env >> atat:builder .venv/bin/python script/seed_roles.py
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -130,35 +137,31 @@ commands:
...
           name: Update Kubernetes cluster
           command: |
-            kubectl set image deployment.apps/atst atst=<< parameters.tag >> --namespace=<< parameters.namespace >>
-            kubectl set image deployment.apps/atst-worker atst-worker=<< parameters.tag >> --namespace=<< parameters.namespace >>
-            kubectl set image deployment.apps/atst-beat atst-beat=<< parameters.tag >> --namespace=<< parameters.namespace >>
-            kubectl set image cronjobs.batch/crls crls=<< parameters.tag >> --namespace=<< parameters.namespace >>
+            kubectl set image deployment.apps/atst atst=<< pipeline.parameters.tag >> --namespace=<< pipeline.parameters.namespace >>
+            kubectl set image deployment.apps/atst-worker atst-worker=<< pipeline.parameters.tag >> --namespace=<< pipeline.parameters.namespace >>
+            kubectl set image deployment.apps/atst-beat atst-beat=<< pipeline.parameters.tag >> --namespace=<< pipeline.parameters.namespace >>
+            kubectl set image cronjobs.batch/crls crls=<< pipeline.parameters.tag >> --namespace=<< pipeline.parameters.namespace >>
```
tomdds, line 161: You're still referencing the global `pipeline` scope for the `tag` and `namespace` params, which should be local

.circleci/config.yml
```diff
@@ -89,28 +105,19 @@ commands:
...
       tag:
         type: string
-        default: << parameters.azure_server_name>>/atat:latest
-      cluster_name:
-        type: string
-        default: ${CLUSTER_NAME}
-      resource_group:
-        type: string
-        default: ${RESOURCE_GROUP}
+        default: << pipeline.parameters.tag >>
```
tomdds, line 120: Seeing this makes me think it might be worth it to do this for all of the other params you migrated to the pipeline. That way each command/job will have all it's potential params documented in one place rather than throughout it's body. Thoughts?

tomdds (changes requested):
2 small changes, otherwise 🎉 parameterization!

.circleci/config.yml
mikepw: just seeing this Tom! yea I agree. For dry run we're start w/ what's in here then iterate.

jesse-dds (approved):

tomdds (approved):

PR #1777, opened 2020-09-08
[AT-5501] Move Basic Auth

dandds:
This PR moves basic auth to uWSGI and creates a new auth endpoint for local development. 

I've added a new endpoint, `/login-local`. This takes the place of `/login-dev` for local development. Developers running a copy of the app locally with the `ALLOW_LOCAL_ACCESS` setting enabled can authenticate into their local instance using the new route. By contrast, `/login-dev` will require federated authentication from the Azure Active Directory instance every time.

It's possible to make the `/login-local` endpoint available in a deployed instance of ATAT so that automated tools that can't use federated authentication mechanisms (like Ghost Inspector or load tests) can access the site. To do this, we would need to mount an additional `htpasswd` password file into the container and set `ALLOW_LOCAL_ACCESS=True`. This will make `/login-local` available in the deployed site and uWSGI will provide HTTP Basic Authentication for that route. This behavior is disabled by default for all deployed instances of the site. We do not include or deploy the password file by default since it's not part of the normal configuration for any instance of ATAT. 

*notes*
- I've added the `/dev-new-user` route to the same local-access blueprint `/login-dev` belongs to. They're both controlled by the `ALLOW_LOCAL_ACCESS` setting. This way, creating a new user on the fly is available to local developers and automated tests if needed.
- When testing the HTTP Basic Auth in the container, I discovered that uWSGI needs to be compiled with PCRE enabled in order to do its own internal routing (basic auth is a form of internal routing in uWSGI). I made some changes to the Dockerfile to enable this.
- The RHEL container does not look for uWSGI plugins the same way the previous Alpine container did. Plugins are included by default in the uWSGI bin we're using for RHEL, and I noticed that the uWSGI startup logs complained that they couldn't find the specified plugins. I removed the references to those plugins in the uWSGI config. The startup logs no longer complain and everything still works as expected.

*reference*: https://uwsgi-docs.readthedocs.io/en/latest/InternalRouting.html#basicauth


atat/routes/dev.py
```diff
@@ -188,7 +191,16 @@ def dev_new_user():
     return redirect(redirect_after_login_url())
 
 
-@bp.route("/test-email")
+@local_access_bp.route("/login-local")
+def local_access():
+    name = request.args.get("username", "amanda")
+    user = get_or_create_dev_user(name)
+    current_user_setup(user)
```
carleton-dds, line 198: This looks good for the most part, but I notice that this route no longer supports the `dod_id` param while `login-dev` does. We should probably have some parity there. If that param isn't being used by anyone that'd be good to know too

dandds: I think the only use case for that param is if you've created a new user via `/dev-new-user`. Worth maintaining parity as long as we support the latter route.

carleton-dds: My understanding is that `/dev-new-user` is mostly used for testing. Curious if @JRNPW has any thoughts, but otherwise I would suggest you add support for the `dod_id` param to this method

dandds: 69177947d673866d9a5a0b365e3620e354f52d4a

carleton-dds (approved):
Thanks for the updates

deploy/azure/atst-envvars-configmap.yml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: atst-envvars
   namespace: atat
 data:
+  ALLOW_LOCAL_ACCESS: "false"
```
jesse-dds, line 8: I'm a little fuzzy on how this value will be interpreted by the application. Will a lower case `"false"` be considered a falsy yml value and become a python `False` automatically? Or will it go through as the string `"false"` which would be interpreted as a truthy python value here:

```
if app.config.get("ALLOW_LOCAL_ACCESS"):
```

tchandler: +1 on this one, I think `ALLOW_LOCAL_ACCESS` needs to have a `getboolean` mapping added to `map_config`

dandds: Good question. It will come through to the application as the string `"false"`, but we coerce when we build the final config object for the Flask instance. Since all environment variables' values come through as strings, we use some methods on the `ConfigParser` object to coerce them as-needed: https://github.com/dod-ccpo/atst/pull/1777/files#diff-332e323788dab4b609a96832e027674bR201

https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getboolean

tomdds: Oh, hah, somehow I missed that, must've collapsed that file when looking for it 🤦

carleton-dds: I ran into this with the old `SAML_LOGIN_DEV` value which I originally set to `True`, but it should be `true`. I think `false` is the correct.

jesse-dds (approved):

PR #1779, opened 2020-09-08
Build and publish nginx images along with the app image

jesse-dds:
Builds and publishes Nginx images as part of the build pipeline. This is necessary for ops to automate the deployment of the Nginx container.

- Renames `tag` parameter to `atat-image-tag` for disambiguation
- Adds tagging, building and publishing step for the `nginx.Dockerfile` (doesn't use context)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,7 @@ commands:
           command: |
             docker load < docker-cache/atat.tar
             docker load < docker-cache/builder.tar
+            docker load < docker-cache/nginx.tar
   setup_datastores:
     parameters:
       pgdatabase:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ commands:
               --username << parameters.sp >>  
             echo "Successfully logged in to Azure CLI."
             az acr login --name << parameters.container_registry >> | grep "Succeeded"
-  cache_docker_image:
+  cache_docker_images:
     steps:
       - run:
           name: Save the docker images to a cache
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -288,15 +295,17 @@ jobs:
     steps:
       - deploy:
           namespace: staging
-          tag: ${AZURE_SERVER_NAME}/atat:staging-${CIRCLE_SHA1}
+          atat-image-tag: ${AZURE_SERVER_NAME}/atat:staging-${CIRCLE_SHA1}
```
mikepw, line 298: I think these should switch from `${AZURE_SERVER_NAME}` to `<<parameters.container_registry>>.azurecr.io` although I realize that's out of scope for this PR.

.circleci/config.yml
```diff
@@ -168,11 +175,12 @@ jobs:
           version: 18.06.0-ce
       - azlogin
       - run:
-          name: Build image
+          name: Build images
           command: |
             docker build . --build-arg GIT_SHA=${CIRCLE_SHA1} --build-arg IMAGE=<<parameters.container_registry>>.azurecr.io/rhel-py -f ./Dockerfile -t atat:builder --target builder
             docker build . --build-arg GIT_SHA=${CIRCLE_SHA1} --build-arg IMAGE=<<parameters.container_registry>>.azurecr.io/rhel-py -f ./Dockerfile -t atat:latest
-      - cache_docker_image
+            docker build --build-arg IMAGE=<<parameters.container_registry>>.azurecr.io/rhelubi:8.2 - < nginx.Dockerfile
```
dandds, line 182: One thing to consider here is that this will slow down CI for every PR. Would it be worthwhile to build the NGINX container separately as something that only happens during CD when a PR gets merged?

dandds (commented):
One question!

.circleci/config.yml
jesse-dds: Makes sense to me 👍

dandds (approved):
👍

mikepw (approved):

PR #1780, opened 2020-09-08
Poll and use latest nginx image tag

jesse-dds:
Grab the latest Nginx image to use in the k8s config.

PR #1782, opened 2020-09-08
Start populating circle ci POST with operator info

jesse-dds:
Add terraform expecting a preexisting container registry for operations use.


mikepw (approved):

carleton-dds (changes requested):
@jesse-dds This looks good for the most part. However,  What's with the duplicate commits between this and #1780?

Very excited to see some properly linted ansible files. Are you just running prettier for this? It would probably make sense to include an update to `script/format` in this PR to run prettier at the ansible folder.

tomdds:
I think you need to update this PR to point at staging now, since you already merged the `feature_operationalize-deployments` branch.

PR #1783, opened 2020-09-09
Adjust SonarQube configuration to run locally

JRNPW:
For now, we have decided to run SonarQube locally on my machine. To do this, there is a minor change to the SQ configuration, as noted in this PR.


jesse-dds (approved):

dandds (approved):
:shipit:

graham-dds:
Is there a reason we're commenting this line out instead of just deleting it?

JRNPW:
@graham-dds Reasonable question. Best answer: Because CCPO still might stand up a cloud instance. For ease of changing back-and-forth (if necessary), it's commented atm. We can probably remove it at some point in the future.

PR #1788, opened 2020-09-10
Add ability to provide first/last name when making AAD user.

dandds:
The autogenerated users created by script/add_users_to_aad_tenant.py
need to have first name and last name provided in order for them to use
federated authentication to access an instance of ATAT for which their
tenant is acting as the identity provider. Effectively, this means that
we have to provide the givenName and surname properties on the request
to create the user. I've added those as nullable fields and updated the
script.


jesse-dds (approved):

tomdds (approved):
⚙️

PR #1773, opened 2020-09-03
Remove CAC-related Python code and app config.

dandds:
Federated authentication is replacing the current authentication
strategy. This commit removes all of the Python code and application
configuration settings relevant to client certificate authentication
within ATAT.

PR #1789, opened 2020-09-10
[AT-5435] EIFS Federated Authentication

tomdds:
Follow on from https://github.com/dod-ccpo/atst/pull/1694 to introduce SAML authentication as the primary login mechanism. Closes a majority of https://ccpo.atlassian.net/browse/AT-5435, except removing the original authentication (taken care of in https://github.com/dod-ccpo/atst/pull/1773 and https://github.com/dod-ccpo/atst/pull/1774).

While this makes some changes to the `login-dev` implementation, that will still remain available for testing as either saml only or with basic auth.

Work completed in collaboration with @carleton-dds and @okaycj-atat 

# Testing Procedure

### Setup

**Note:** In order to log in against the EIFS SAML provider, you will need to be able to authenticate via CAC. This requires a card, a card reader, and the correct cert chain installed on your device.

**Environment**
The 1password vault contains the updated configuration names and values under the entry `SAML Identity Provider Local Config`. You'll need to place these entries in the relevant config file (likely dev.ini)

You'll also need [ngrok](https://ngrok.com/) to forward your local server to a public domain. You can find the ngrok auth token in 1password as `ngrok auth token`. To set up the local development tunnel, you need to run the following commands:
```
ngrok authtoken <insert token here>
ngrok http -subdomain=atat-local https://localhost:8000
```

**Certs**
The EIFS SAML identity provider requires that we run a secure server, which is provided by ngrok, but that also needs to relay to a local secure server.

First, generate some self-signed certs with `script/create_local_certs`

Then run a secure version of the dev environment with `script/secure_server`

### Testing
If you have everything set up as prescribed above, you should be able to visit `https://atat-local.ngrok.io` and see your local instance running. Once you visit that, you can either click through the `Sign in with CAC` button and modal, or visit `https://atat-local.ngrok.io/login` and you will be redirected to EIFS to select your CAC cert and enter your pin. When you return to ATAT, you'll either be on your home page or at the user creation screen (if this is your first login).


atat/routes/dev.py
```diff
@@ -134,33 +138,64 @@ def login_dev():
...
+    else:
+        persona = query_string_parameters.get("username_param") or request.args.get(
+            "username", "amanda"
+        )
+        user = get_or_create_persona(persona)
+
+    return user
+
+
+def get_or_create_persona(persona):
```
dandds, line 177: You and I added essentially the same helper function to this file in different places. You can remove one or the other and make sure they're used in the correct places. The other is called `get_or_create_dev_user` on line 115.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ webassets = "*"
 werkzeug = "*"
 email-validator = '*'
 python3-saml = "^1.9.0"
+cachetools = "^4.1.1"
```
dandds, line 41: Are you pulling this in for specific functionality, like the expiry?

dandds (changes requested):
One small cleanup change and a question. Looks great!

pyproject.toml
tomdds: Yeah, specifically for `@cachetools.func.ttl_cache` which is a "time to live" cache. The python std lib only gives you lru cache (least recently used).

atat/routes/dev.py
tomdds: Ah, yep, removed duplication in my latest rebase

atat/routes/dev.py
```diff
@@ -108,9 +112,54 @@ def message(self):
...
+def login_dev():
+    query_string_parameters = session.get("query_string_parameters", {})
+    user = None
+
+    if "sls" in request.args and request.method == "GET":
+        return redirect(url_for("atat.root"))
+
+    if (
+        "saml" in request.args or app.config.get("SAML_LOGIN_DEV", False)
+    ) and request.method == "GET":
```
dandds, line 125: I removed the `SAML_LOGIN_DEV` setting on #1777 , so I think you can remove it from this branch. This line should just check the `request.method` since `/login-dev` is only used with fed auth now.

config/base.ini
```diff
@@ -65,11 +64,13 @@ REDIS_SSLCHECKHOSTNAME=False
...
+SAML_DEV_ENTITY_ID
+SAML_DEV_IDP_URI
+SAML_DEV_SLS
 SAML_ENTITY_ID
-SAML_IDP_CERT
-SAML_IDP_ENTITY_ID
-SAML_IDP_SLS
-SAML_IDP_SSOS
+SAML_IDP_URI
+SAML_LOGIN_DEV = False
```
dandds, line 73: You can drop this.

dandds (changes requested):
Looks like `SAML_LOGIN_DEV` got re-introduced by the rebase. I might compare the `login_dev` function against the staging branch and see if any other small changes are missing.

tests/routes/test_saml_helpers.py
```diff
@@ -0,0 +1,312 @@
...
+    mock_auth = Mock(
+        get_settings=Mock(
+            return_value=Mock(
+                get_sp_metadata=Mock(return_value=Mock()),
+                validate_metadata=Mock(return_value=[]),
+            )
+        )
+    )
+
+    mock_saml_auth.return_value = mock_auth
```
graham-dds, line 107: Do you think we could somehow package some of this mock auth setup into a fixture? It's a lot of mocking to set up each time it's used. Because it we need to control some of the inner return values though, maybe we could provide parameters to the fixtures with one of the methods outlined [here](https://github.com/pytest-dev/pytest/issues/2502#issuecomment-308403299).

Also, another thing that could make this a little more streamlined is to use [`configure_mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.configure_mock) to set those nested return values.

atat/routes/dev.py
tomdds: Ah yep, didn't catch this during rebase

deploy/overlays/cloudzero-dryrun-master/envvars.yml
```diff
@@ -15,15 +14,17 @@ kind: ConfigMap
...
-  SAML_IDP_SLS: https://login.microsoftonline.com/common/wsfederation?wa=wsignout1.0
+  SAML_ENTITY_ID: https://master.atat.dev
+  SAML_ACS: https://master.atat.dev/login?acs
+  SAML_SLS: https://master.atat.dev/login?sls
+  SAML_IDP_URI: https://fs.disa.mil/federationmetadata/2007-06/federationmetadata.xml
+  SAML_DEV_ENTITY_ID: https://master.atat.dev/login-dev
+  SAML_DEV_ACS: https://master.atat.dev/login-dev?acs
+  SAML_DEV_SLS: https://master.atat.dev/login-dev?sls
+  SAML_DEV_IDP_URI: https://login.microsoftonline.com/b5ab0e1e-09f8-4258-afb7-fb17654bc5b3/federationmetadata/2007-06/federationmetadata.xml?appid=29cbbf05-89d2-4c72-a691-fe4fd0f74b3f
+  SAML_LOGIN_DEV: "true"
```
dandds, line 30: Nitpicky, sorry, but there's one more leftover.

dandds (approved):
There's one more instance of SAML_LOGIN_DEV. I approve anticipating that getting dropped! Nice work all!

graham-dds (commented):
Just a few smallish comments.

One thing I noticed throughout: since #1643 we're trying to consistently use `printf` style interpolation in logging messages. Is that something that we should use throughout the new logging that that's introduced in this PR?

tests/routes/test_saml_helpers.py
tomdds: Good call https://github.com/dod-ccpo/atst/pull/1789/commits/9c54977de5557d7440a290173604b6e4e94eb679

tests/routes/test_saml_helpers.py
```diff
@@ -0,0 +1,312 @@
...
+
+        mock_parser.parse_remote = Mock(return_value={"idp": "config"})
+
+        assert _get_idp_config(idp_uri) == "config"
+        _get_idp_config(idp_uri)
+
+        assert mock_parser.parse_remote.call_count == 1
+
+
+def test_get_user_from_saml_attributes():
```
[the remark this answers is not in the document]
tomdds: bundled them all together with a fixture https://github.com/dod-ccpo/atst/pull/1789/commits/94442e83efb4963d7042755b1faffdb0724fc4b4

atat/routes/__init__.py
```diff
@@ -119,3 +124,18 @@ def logout():
 @bp.route("/about")
 def about():
     return render_template("about.html")
+
+
+@bp.route("/login", methods=["GET", "POST"])
```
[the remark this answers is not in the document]
tomdds: Just a holdover from initial build out, split them up https://github.com/dod-ccpo/atst/pull/1789/commits/1ef7e93f4dc42a52865f5d0614663a6366ec8c4b

graham-dds (approved):

PR #1792, opened 2020-09-11
[AT-5504] Add a blank option for service branch on user edit form

tomdds:
The service dropdown on both the user onboarding form and the user edit form would default to Air Force if no other option was selected. This inserts a placeholder at the front of the list to ensure the user selects a branch themselves.

### DIFF
--- a/atat/forms/edit_user.py
+++ b/atat/forms/edit_user.py
@@ -13,6 +13,10 @@ from .fields import SelectField
 from .forms import BaseForm
 from .validators import Name, PhoneNumber
 
+SERVICE_BRANCH_CHOICES = [
+    ("", translate("fragments.edit_user_form.service_choice"))
+] + SERVICE_BRANCHES
+
 USER_FIELDS = {
     "first_name": StringField(
         translate("forms.edit_user.first_name_label"),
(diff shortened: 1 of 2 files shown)


carleton-dds (approved):
👍 When I tried to tackle this last week I was attempting to utilize `SelectField`'s `default` parameter, which gave me a lot of trouble. Just putting it first in the list of choices is a much simpler and cleaner

atat/forms/edit_user.py
```diff
@@ -28,7 +32,8 @@
     ),
     "phone_ext": StringField("Extension", validators=[Number(), Length(max=10)]),
     "service_branch": SelectField(
-        translate("forms.edit_user.service_branch_label"), choices=SERVICE_BRANCHES
+        translate("forms.edit_user.service_branch_label"),
+        choices=SERVICE_BRANCH_CHOICES,
```
dandds, line 36: I _think_ you want to add `default=""` to the kwargs here. This tells the `OptionsInput` component on the client side that the field is not valid if the user leaves that selection. It's hard for me to validate right now, though, because I hit a separate bug on this form (AT-5510).

dandds (approved):
👍 Looks good, with one caveat I noted as a comment.

PR #1795, opened 2020-09-14
[AT-5397] Overlay for bugbounty instance of ATAT.

dandds:
This adds an additional k8s kustomize overlay for the bugbounty instance of ATAT.

Keeping this in draft for now. The scaling variables will change, pending a support request to our hosting service. The configuration will also change as a result of #1789 . Depending on how things time out, we could choose to merge this first and update the bugbounty config as-needed.


deploy/overlays/cloudzero-dryrun-bugbounty/envvars.yml
```diff
@@ -0,0 +1,30 @@
...
+  CELERY_DEFAULT_QUEUE: celery-bugbounty
+  FLASK_ENV: bugbounty
+  PGDATABASE: cloudzero_dryrun_bugbounty
+  STATIC_URL: "https://bugbounty.atat.dev/static/"
+  SAML_ENTITY_ID: https://bugbounty.atat.dev/login-dev
+  SAML_ACS: https://bugbounty.atat.dev/login-dev?acs
+  SAML_SLS: https://bugbounty.atat.dev/login-dev?sls
+  SAML_IDP_ENTITY_ID: https://sts.windows.net/ab154c46-3d57-4386-9b87-8002123f3858/
+  SAML_IDP_SSOS: https://login.microsoftonline.com/ab154c46-3d57-4386-9b87-8002123f3858/saml2
+  SAML_IDP_SLS: https://login.microsoftonline.com/ab154c46-3d57-4386-9b87-8002123f3858/saml2
```
carleton-dds, line 29: I think you may have duplicated the SSOS URL. I believe the SLS needs to be:
`https://login.microsoftonline.com/common/wsfederation?wa=wsignout1.0`

tomdds: This might be a combination of my fault and Azure's. I advised Dan that he could find all these values in the metadata, but the signout url in the metadata doesn't match the one presented in the azure portal. Interestingly, using this value seems to work fine, since I tested the logout functionality using the duplicate SSOS URL.

carleton-dds: After discussing with Tom and Dan in slack, it was decided that to achieve our desired outcome of the log out URL redirecting the user back to the landing page you want to use the the same URL as the Sign On URL AND setting a Logout URL in the SAML IdP app configuration. Once EIFS is merged into staging the sign out url will come from parsing the metadata URL,  but setting manually here for now is good for now.

All of that is to say, no changes are needed to the SLS entry.

carleton-dds (approved):

tomdds (approved):

PR #1796, opened 2020-09-14
[AT-5494] Update load tests

okaycj-atat:
This PR is to update the load tests so that they work with the current version of Locust.  Additionally, completed the tests for applications and added tests for task orders and user profiles. 

Also, there was a trivial bug with the development database data that was added to this PR. 

[AT-5494](https://ccpo.atlassian.net/browse/AT-5494)

### DIFF
--- a/atat/routes/dev.py
+++ b/atat/routes/dev.py
@@ -51,7 +51,7 @@ _DEV_USERS = {
         "service_branch": random_service_branch(),
         "phone_number": "1234567890",
         "citizenship": "United States",
-        "designation": "Military",
+        "designation": "military",
     },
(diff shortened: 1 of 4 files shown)


dandds (changes requested):
Great work! I ran the tests locally and things look good.

I think we need to update the README in this directory, specifically the "To Update Image" section. The image registry it mentions doesn't exist any more. The instructions should probably be more generic and not assume a specific registry or that the load tests are already available as a container service in the CSP.

dandds (approved):
🏋️

tomdds (approved):
Nice work!

PR #1803, opened 2020-09-16
Remove SAML_IDP_CERT secret from flex_vol

tomdds:
Since we're ingesting SAML via metadata, we no longer need to store or source the IdP public cert from our key vault.


graham-dds (approved):

dandds (changes requested):
Could you remove the reference in deploy/overlays/cloudzero-dryrun-bugbounty/flex_vol.yml, too? 🙏

dandds (approved):
👍

PR #1810, opened 2020-09-18
upgrade node-forge to 0.10

steve-atat:
[Jira](https://ccpo.atlassian.net/browse/AT-5513)

- Upgrade node-forge dependency to 0.10

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "vue-template-compiler": "2.6.12"
   },
   "resolutions": {
+    "node-forge": "^0.10.0",
     "parcel-bundler/**/lodash": "^4.17.13",
     "minimist": ">=1.2.2",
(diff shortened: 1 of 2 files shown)


dandds (approved):
👍

stevesweetney:
[Explanation of the selective dependency feature in yarn](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/)

> A sub-dependency of your project got an important security update and you don’t want to wait for your direct-dependency to issue a minimum version update.

This is exactly the problem we wanted to solve

stevesweetney:
It looks like the minimum version of node-forge was bumped in a [recent PR](https://github.com/parcel-bundler/parcel/pull/5149) in the parcel repo. There's no indication of when those changes would be available in a stable release, the other alternative would be to use a nightly release of parcel v2 which does have breaking changes

PR #1815, opened 2020-09-21
Update the bugbounty overlay.

dandds:
This updates the bugbounty overlay in the following ways:

1. It was previously referencing the cloudzero-dryrun-staging overlay as its base. That overlay no longer exists since we transitioned the staging site to new infrastructure. I've updated it to reference the base config in the deploy/azure directory.
2. Because of 1, the bugbounty overlay has to explicitly define all of the configuration it was previously inheriting from the staging overlay (Postgres server URL, etc.).
3. It now has the updated configuration for fed auth against both the development and prod IdPs.
4. I've updated the scaling values for the Horizontal Pod Autoscaler since we have more room in that cluster.

Deploying these changes will also remove CAC auth from the bugbounty instance of the site.

### DIFF
--- a/deploy/overlays/cloudzero-dryrun-bugbounty/ports.yml
+++ b/deploy/overlays/cloudzero-dryrun-bugbounty/ports.yml
@@ -1,0 +1,7 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: atst-main
+  annotations:
+    service.beta.kubernetes.io/azure-load-balancer-internal-subnet: "cloudzero-dryrun-public"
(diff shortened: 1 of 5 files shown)


deploy/overlays/cloudzero-dryrun-bugbounty/ports.yml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: atst-main
+  annotations:
+    # service.beta.kubernetes.io/azure-load-balancer-internal: "true"
```
carleton-dds, line 7: Should this comment be removed?

carleton-dds (approved):
I left one quick comment (about a comment), but this looks ready to 🚢

deploy/overlays/cloudzero-dryrun-bugbounty/ports.yml
dandds: amended!

mikepw (approved):

PR #1818, opened 2020-09-21
Remove all references to htpasswd

jesse-dds:
Gets rid of uses of HTTP Password basic auth since we moved to a fed auth model.

Makes https://ccpo.atlassian.net/browse/AT-5463 no longer applicable.

Blocked by #1727


carleton-dds:
Hey what happen with this pr?

PR #1820, opened 2020-09-21
Fix hybrid tests

dandds:
Makes some small adjustments so that the hybrid tests (nearly) pass. See commits for individual details.

### DIFF
--- a/atat/domain/csp/cloud/models.py
+++ b/atat/domain/csp/cloud/models.py
@@ -639,6 +639,8 @@ class ProductPurchaseVerificationCSPResult(AliasModel):
 
 class UserMixin(BaseModel):
     password: Optional[str]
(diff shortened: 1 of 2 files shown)


graham-dds (approved):
I'm happy with these changes, but I think the root of the issue is the way we poll for management groups. As you mentioned on Slack, we'll occasionally get 400 errors during the management group polling process for applications and environments. In that process, we can always GET the URL contained in the `Azure-AsyncOperation` header, but then when we try and GET the URL contained in the `Location` header, we sometimes hit permission issues. Maybe there's a way we can rely on the data retrieved by the `Azure-AsyncOperation` URL alone, or maybe even simply receiving a success status would be enough.

carleton-dds (approved):

PR #1821, opened 2020-09-22
[AT-4977] Enumerate Postgres user perms in create_database_user.py.

dandds:
This change explicitly lists the database permissions that the application's database user requires for the database. It also deletes some of the SQL that's no longer necessary because of the order the database bootstrapping scripts are now run in.

I tested this by running the database bootstrapping scripts against a fresh Azure Postgres instance. Everything ran successfully. I connected to the new database instance as the new ATAT user via psql and was able to SELECT, INSERT, UPDATE, etc.


mikepw (approved):

mikepw (changes requested):
can you escape the the params so ansible works, like \"{dbname}\" and \"{username}\" ?

carleton-dds (approved):
Tested locally, LGTM

mikepw (approved):

PR #1822, opened 2020-09-22
[AT-5529] Generate maintenance page

okaycj-atat:
This PR will provide a script to generate an "Under Maintenance" page using the currently existing flask/jinja2 template system.  

[AT-5529](https://ccpo.atlassian.net/browse/AT-5529)


dandds (commented):
If I'm reading correctly, it looks like you're extracting all the HTML, CSS, and JS files and copying them to a known location. That's a good start, but I think we'll need the CSS and JS added to the HTML file. That might be something you can do with pyquery.

okaycj-atat:
> If I'm reading correctly, it looks like you're extracting all the HTML, CSS, and JS files and copying them to a known location. That's a good start, but I think we'll need the CSS and JS added to the HTML file. That might be something you can do with pyquery.

I had a quick talk with @jesse-dds about the limitations of the service where the "Under Maintenance" page is going.  You're correct.  I have another concern about handling binaries.  The docs suggest encoding them as base64 and adding them to the html file.  There's a bit more work todo before I am in the ball park.

graham-dds (commented):
Looks like a good start. I have few things I was hoping you could address before I do a more in-depth review:
1. It looks like you did a reformat of the `README` in addition adding the `How to generate the Under Maintenance page` section. Could you remove those formatting changes from this PR? Having standard markdown formatting could be useful, but we'd probably want to apply those changes similarly to how we handled #1761 and  #1832 .
2. I noticed that `pyquery` isn't in our `pyproject.toml`. I does look like we use it for load testing, but it it's `pip` installed in a Dockerfile in that context. We _do_ have BeautifulSoup in our `pyproject.toml` though (which we use for JS tests), and that seems like it could be used for the needs of this script. Do you think it would be possible to use BeautifulSoup instead of adding another dependency? Alternatively, if it does make sense to add `pyquery` to our dev dependencies to facilitate load testing development, maybe that could happen in this PR.
3. With the above in mind: Would it be good to include at least one test for this script, just to assert that `a page successfully is generated` ? Maybe we could also assert that the page is <= 1 megabyte, like you do in that `print` statement.

script/generate_under_maintenance.py
```diff
@@ -0,0 +1,120 @@
+import sys
+
+sys.path.append("./")
+
+from os import path, stat
+from atat.app import make_app, make_config
+from pyquery import PyQuery as pq
```
tomdds, line 7: pyquery isn't currently part of this project's dependencies, you'll want to add it as at least a dev dependecy

script/generate_under_maintenance.py
```diff
@@ -0,0 +1,120 @@
...
+
+            # encode all fonts found in css that have been whitelisted
+            urls = re.findall(r"url\([A-Za-z0-9/.\-#?]*\)", css_str)
+            urls = (u[4:-1] for u in urls)
+            for url in urls:
+                if path.basename(url) in font_whitelist:
+                    css_str = css_str.replace(url, f"url({make_base64(url)})")
+
+            # inject css into head of html
+            d("head").after(f"<style type='text/css'>{css_str}</style>")
```
tomdds, line 101: You probably want `append` and not `after` here. `after` is going to result in 2 problems:
1) The style tags will be inserted between `head` and `body`
2) The style tags will be inserted in the opposite order of their original declaration because you're iterating front to back, but prepending with `after`

README.md
```diff
@@ -262,10 +270,10 @@ All config settings must be declared in "config/base.ini", even if they are null
...
-4. Finally, ATAT will check for the presence of environment variables matching all known config values. If a matching environment variable exists, its value will override whatever value was previously set.
+4. Find the specified directory. For example, "/config"
+5. Search for files in that directory with names matching known settings. For example, a file called "AZURE_STORAGE_ACCOUNT_NAME".
+6. Read in the contents of those files and set the content as the value for that setting. These will override any values previously set. For example, if the file "AZURE_STORAGE_ACCOUNT_NAME" has the content "my-azure-account-name", that content will be set as the value for the AZURE_STORAGE_ACCOUNT_NAME setting.
+7. Finally, ATAT will check for the presence of environment variables matching all known config values. If a matching environment variable exists, its value will override whatever value was previously set.
```
tomdds, line 276: This block is supposed to be indented, not sure if it was a formatting tool or what, but 4, 5, & 6 are a sub-list of 3.

tomdds (changes requested):

script/generate_under_maintenance.py
```diff
@@ -0,0 +1,120 @@
...
+            # remove css element from html
+            link.remove()
+
+        else:
+            # this is usually the ico file...
+            link.attr.href = make_base64(link.attr.href)
+
+    # write html to file
+    args.output.mkdir(parents=True, exist_ok=True)
+    html_file = args.output / "index.html"
```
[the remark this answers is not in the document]
okaycj-atat: I like pathlib, but I don't mind going for the clearer option.

script/generate_under_maintenance.py
okaycj-atat: @graham-dds has raised the point that BeautifulSoup is already a dependency of this project.  I have never used this library before.  Should I opt for this package instead?

script/generate_under_maintenance.py
okaycj-atat: I should have checked the html file in a text editor and not in chrome.

README.md
okaycj-atat: This is embarrassing!  Also, this is something @graham-dds pointed out in a comment above.  I'll make sure to get this fixed.

script/generate_under_maintenance.py
tomdds: 👍  I don't think there's anything BeautifulSoup can't also do here. I'd go with existing dependencies rather than adding new ones when possible.

script/generate_under_maintenance.py
```diff
@@ -0,0 +1,124 @@
...
+    for img in d("img").items():
+        img.attr.src = make_base64(img.attr.src)
+
+    # link elements
+    for link in d("link").items():
+        # add css to html
+        if file_ext(link.attr.href) == ".css":
+            with open(relative_path(link.attr.href)) as fh:
+                # remove all comments, including reference to css map file
+                css_str = re.sub(r"/\*.*?\*/", "", fh.read().strip())
```
dandds, line 95: If you want this to match CSS comments spanning multiple lines, like

```css
/*
* comment
* across
* lines
*/
```
then I'm pretty sure you need to add a regex flag:

```python
css_str = re.sub(r"/\*.*?\*/", "", fh.read().strip(), flags=re.S)
```
https://docs.python.org/3/library/re.html#re.S

Find/replaces make me nervous because it's very easy to delete patterns in text you didn't anticipate, but it's probably fine as long as we vet the output really well.

templates/under_maintenance.html
```diff
@@ -0,0 +1,32 @@
...
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>{% block title %}Under Maintenance{% endblock %}</title>
+  {% assets "css" %}
+    <link rel="stylesheet" href="{{ ASSET_URL }}" type="text/css">
+  {% endassets %}
+  <link rel="icon" type="image/x-icon" href="{{ url_for('static', filename='img/favicon.ico') }} " />
+</head>
+
+<body>
```
dandds, line 16: Maybe I missed it, but is there some reason this template can't use the base template?

dandds (commented):
This is neat! I hope it was fun to write. I have a small comment and question.

When I run the script, the HTML file produced can't seem to load the fonts.

Screen Shot 2020-10-02 at 6 51 16 AM

templates/under_maintenance.html
okaycj-atat: The initial iteration of this used the public base template.  The public base template comes with the two banners at the top of the page and that is no longer a requirement.  I feel that to use a base template I would have to refactor a few templates.  Is that something I should dig into?

okaycj-atat:
> When I run the script, the HTML file produced can't seem to load the fonts.

@dandds Yup!  The cache busting got me!  I'll need to change how the whitelist works.

templates/under_maintenance.html
dandds: It seems like it could be easy to forget to update all the boilerplate in `<head>` across multiple templates if it changes; maybe pull that shared markup into a partial, or create a more basic base template for both (`base_base.html` 😁 ).

script/generate_under_maintenance.py
okaycj-atat: I assume when we get the design, this script will need some edits.  I am sure I've missed some corner cases.  About the new lines, I am assuming there aren't any.  The CSS file is minimized and the contents of the file should all be on one line.

okaycj-atat: That being said, I will make this change.

tomdds (approved):

templates/base_head.html
```diff
@@ -0,0 +1,17 @@
...
+  <title>{% block title %}{% endblock %}</title>
+  {% assets "css" %}
+    <link rel="stylesheet" href="{{ ASSET_URL }}" type="text/css">
+  {% endassets %}
+  <link rel="icon" type="image/x-icon" href="{{ url_for('static', filename='img/favicon.ico') }} " />
+</head>
+<body class="{% if g.modalOpen %} modalOpen{% endif %}">
+{% block body %}{% endblock %}
+</body>
+</html>
```
graham-dds, line 17: Missing newline here

graham-dds (commented):
Looking better! A few more things:
- could we add `#! .venv/bin/python` to the top of the script so that we can run it like `script/generate_under_maintenance.py`?
- I noticed [this commit](https://github.com/dod-ccpo/atst/pull/1822/commits/30fe133b3c271944ae5eac72fda66211ae83edef). Is there a reason you decided not to go with pathlib? It might remove the need for some of the little utility functions that you've defined in the script. For example, `file_ext()` could be replaced by the [.suffix](https://docs.python.org/3.7/library/pathlib.html#pathlib.PurePath.suffix) property of a Pathlib object. It would also _feel_ better to use [`relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) rather than using `urllib.parse` when we aren't really dealing with URLs.
- I'm getting the same error that @dandds [was earlier](https://github.com/dod-ccpo/atst/pull/1822#pullrequestreview-501030004) about fonts:
image
- The code in the `if __name__ == "__main__":` block seems to be responsible for a lot in this script. Do you think it would be worth breaking that section of the script into smaller functions with fewer responsibilities, then composing them in some `main()` function? I noticed that some of the comments above sections in the script (e.g. `# render template`,  `# encode images`, `# add css to html`) seem like natural places to split the script into smaller functions.
- Did you consider writing a small test or two like I pointed out in my last review?

I also have a question about template inheritance / extending, but I'll try to catch up with you about this in a Zoom call

script/generate_under_maintenance.py
```diff
@@ -0,0 +1,134 @@
+import os
+import sys
+
+script_directory = os.path.dirname(__file__)
+project_directory = os.path.join(script_directory, "../")
+parent_dir = os.path.abspath(project_directory)
+sys.path.append(parent_dir)
```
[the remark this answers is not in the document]
okaycj-atat: This something another review requested.  I am going to leave at is, unless you feel strongly.

okaycj-atat:
> * could we add `#! .venv/bin/python` to the top of the script so that we can run it like `script/generate_under_maintenance.py`?

I have added a shebang to the script.  Thank you.

okaycj-atat:
> * I noticed [this commit](https://github.com/dod-ccpo/atst/pull/1822/commits/30fe133b3c271944ae5eac72fda66211ae83edef). Is there a reason you decided not to go with pathlib? It might remove the need for some of the little utility functions that you've defined in the script. For example, `file_ext()` could be replaced by the [.suffix](https://docs.python.org/3.7/library/pathlib.html#pathlib.PurePath.suffix) property of a Pathlib object. It would also _feel_ better to use [`relative_to`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) rather than using `urllib.parse` when we aren't really dealing with URLs.

Another reviewer found `pathlib` to be confusing.  It was removed.

okaycj-atat:
> * I'm getting the same error that @dandds [was earlier](https://github.com/dod-ccpo/atst/pull/1822#pullrequestreview-501030004) about fonts:

Yup, I keep finding corner cases when it comes to assets found in css.  The whole matched string will change depending on how the local development site is deployed (minified vs not).  I am now using the regex grouping mechanism.  This give me access to the whole matches string and a group within it.  I am can now assume less.

okaycj-atat:
> * The code in the `if __name__ == "__main__":` block seems to be responsible for a lot in this script. Do you think it would be worth breaking that section of the script into smaller functions with fewer responsibilities, then composing them in some `main()` function? I noticed that some of the comments above sections in the script (e.g. `# render template`,  `# encode images`, `# add css to html`) seem like natural places to split the script into smaller functions.

I have refactored the script to now use a main() function. Thank you.

okaycj-atat:
> * Did you consider writing a small test or two like I pointed out in my last review?

My apologies and thank you for reminding me to do this.  It slipped my mind.  There is now a test to validate the html file size.

graham-dds (approved):
:shipit: 

Nice! Thanks for hanging with me for all the change requests. There a few more tweaks that _could_ be made, but they can be addressed when we modify the output of the script to align with Think Co. designs.

PR #1826, opened 2020-09-23
Poll for latest nginx image tag

jesse-dds:
- Adds a polling task that checks the repository for the latest built NGINX image
- Gets the single latest built NGINX image and stores it in fact called `latest_nginx_tag`
- This tag is then used to specify the image in the `NGINX_CONTAINER_IMAGE` environment variable


dandds (approved):
Seems pretty straightforward!

mikepw (approved):

PR #1828, opened 2020-09-23
Remove all references to htpasswd

jesse-dds:
- We no longer use basic auth now that we've moved to the fed auth model
- We no longer need to generate and store an http password
- Remove all references to `htpasswd`


README.md
```diff
@@ -441,21 +441,6 @@ The build assumes that you have redis and postgres running on their usual ports
...
-
-You can generate a password with the htpassword. In this example, the command will generate a "localpassword" file in the current directory for a user called "atat":
-
-```
-htpasswd -cd ./localpassword atat
-```
-
-It will prompt you to enter a password, then produce the password file. Note that the `-d` flag is necessary, per the uWSGI documentation linked above.
-
-
```
dandds, line 458: This is all still true. The `/login-local` endpoint is not available on staging or master, but can be made available this way by setting `ALLOW_LOCAL_ACCESS=true`. We may need this for automated testing some time soon.

dandds (changes requested):
The rest of the changes look good, but that section of the README is still valid.

dandds (approved):
✂️

carleton-dds (approved):

PR #1831, opened 2020-09-24
[AT-5521] Update Azure PostgreSQL CA Cert

tomdds:
The existing root certificate for Azure's PostgreSQL offering is being deactivated on Oct 26, 2020. This appends the next certificate to our config to prevent any interruption in service. Source of certificate and other information can be found: https://docs.microsoft.com/en-us/azure/postgresql/concepts-certificate-rotation#what-do-i-need-to-do-to-maintain-connectivity


dandds (approved):
📄

carleton-dds (approved):

PR #1832, opened 2020-09-24
Update Script/format to run on Ansible yml

carleton-dds:
`script/format` now runs prettier on the yaml files in `./ansible`, which should make for a more consistent experience. I had to add config to ignore the `ansible/roles/k8s/tasks/templates/envvars.yml` file because prettier really doesn't like how we nest the collections in there, but I believe its a requirement of ConfigMap

### DIFF
--- a/script/format
+++ b/script/format
@@ -8,6 +8,7 @@ if [ "$1" == "check" ]; then
   poetry run isort --check ${FILES_TO_SORT}
   yarn run prettier --list-different "js/**/*.js" --config ./prettier.config.json
   yarn run prettier --list-different "styles/**/*.scss"
+  yarn run prettier --list-different "ansible/**/*.yml"
   if [ -x "$(command -v terraform)" ]; then
     terraform fmt -recursive -check terraform/
   fi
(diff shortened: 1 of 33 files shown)


jesse-dds (approved):

script/format
```diff
@@ -8,6 +8,7 @@ if [ "$1" == "check" ]; then
   poetry run isort --check ${FILES_TO_SORT}
   yarn run prettier --list-different "js/**/*.js" --config ./prettier.config.json
   yarn run prettier --list-different "styles/**/*.scss"
+  yarn run prettier --list-different "ansible/**/*.yml"
```
graham-dds, line 11: Instead of calling `prettier` three times in a row with different globs, it looks like you can call it with multiple glob arguments

```
yarn run prettier --list-different --config ./prettier.config.json "js/**/*.js" "styles/**/*.scss" "ansible/**/*.yml" 
```
The only thing we'd need to fix with this approach is that the single quote rule we specify in `prettier.config.json` collides with our double quote convention in `*.scss` files. Looks like we can sidestep that with another entry in the prettier config:
```
{
  ...
  "overrides": [
    {
      "files": [
        "**/*.scss"
      ],
      "options": {
        "singleQuote": false
      }
    }
  ]
}

carleton-dds: Done! Yaml is also double quote so I had to had that too https://github.com/dod-ccpo/atst/pull/1832/commits/bfc156d6b89962a5f3d6a36cdaece460724e69ae

graham-dds (approved):

PR #1835, opened 2020-09-25
Updates to Ghost Inspector "helper" tests

JRNPW:
A few weeks ago, Ghost Inspector updated the way that they escape double-quotes when tests are exported to HTML files. Soon thereafter, I updated all our main tests in Github. This PR is the same updates, applied to "helper" tests (not directly executed but imported into other tests). This does not represent any functional/logical changes to these tests.


tomdds (approved):
&quot;

carleton-dds (approved):

PR #1839, opened 2020-09-29
[AT-5335] Bugfix: Detached instance error when sending PDFs

graham-dds:
Closes [AT-5335](https://ccpo.atlassian.net/browse/AT-5335)
### Recreating the bug:
First, I reset my local database
```
script/reset_database.py
```

For a sanity check, I ran the development celery queue. 
```
script/dev_queue
```
There were no errors since there were no task orders to send, which is when the bug occurs.

Next, I seeded the local database
```
python script/seed_sample.py
```

Now after running the development queue, I quickly saw an error:
```
DetachedInstanceError("Parent instance <TaskOrder at 0x10bfd5f28> is not bound to a Session; lazy load operation of attribute '_pdf' cannot proceed")
```

### Fixing the Bug
Previously observed log entries related to this error pointed to [this section](https://docs.sqlalchemy.org/en/13/errors.html#error-bhk3) of the Sqlalchemy docs. The TL;DR of that section is that this error can be mitigated with two general techniques:
- Don't close the session prematurely 
- Load everything that's needed up front

Since the error calls out the `_pdf` attribute specifically, I chose to solve this issue by eagerly loading PDF data in the query that we use to generate the list of task orders that need to be sent. After making this change, the development queue runs as expected with no `DetachedInstanceError`s, which I interpreted as solving the bug.

This PR also makes the time that we wait between celery beat scheduled tasks configurable. This is mostly intended as a developer convenience. This made it so that I had a shorter cycle time between doing things through the ATAT UI (like creating a task order) and seeing corresponding tasks pop up in Celery.


dandds (changes requested):
This definitely fixes the issue using the manual test you've described. Nice work!

One thing I ran into is that it seems like Celery 5.0 doesn't accept lowercase values for the `--loglevel` flag anymore. If I run `script/dev_queue` as-is on your branch, I get this error:

```Error: Invalid value for '-l' / '--loglevel': invalid choice: info. (choose from DEBUG, INFO, WARNING, ERROR, CRITICAL, FATAL)```

Would you mind updating "info" to "INFO" in the script?

UPDATE: Nevermind, I fixed it on https://github.com/dod-ccpo/atst/pull/1840 because the K8s cluster was broken for the same reason.

dandds (approved):
📄

carleton-dds (approved):

PR #1842, opened 2020-10-02
Minor change to selector

JRNPW:
On the TO Step 1 test, we reference the link for the PDF. It's an anchor with a unique class, so we can reference it as such, without the reference to its parent.


steve-atat (approved):
LGTM 👍

tomdds (approved):

PR #1848, opened 2020-10-06
fix(TO creation step 3): onchange -> oninput

steve-atat:
[Jira](https://ccpo.atlassian.net/browse/AT-5333?atlOrigin=eyJpIjoiMGM5NmFlMTliZGM2NDUwNzg0M2Q1NjA0ODk4NmViMmQiLCJwIjoiaiJ9)

## Why

> The difference is that the oninput event occurs immediately after the value of an element has changed, while onchange occurs when the element loses focus, after the content has been changed

From [w3Schools](https://www.w3schools.com/jsref/event_oninput.asp#:~:text=The%20difference%20is%20that%20the,works%20on%20elements.)

### DIFF
--- a/js/components/__tests__/pop_date_range.test.js
+++ b/js/components/__tests__/pop_date_range.test.js
@@ -144,13 +144,13 @@ describe('PopDateRange Test', () => {
     var endDateField = wrapper.find('fieldset[name="end_date"]')
 
     // set valid date range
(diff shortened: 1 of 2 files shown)


graham-dds (approved):
LGTM ✅

jesse-dds (approved):

PR #1853, opened 2020-10-08
[AT-5510] Fix Option Input Validation

tomdds:
Option input fields (select dropdowns and radio buttons) were not correctly reporting their validity to their parent form, causing them to be ignored for the purposes of allowing form submission.

### DIFF
--- a/js/components/options_input.js
+++ b/js/components/options_input.js
@@ -28,6 +28,21 @@ export default {
     }
   },
 
+  mounted: function () {
+    const selectEl = this.$el.querySelector('select')
+    if (selectEl) {
+      selectEl.value = this.value
+    }
+
+    const radios = this.$el.querySelectorAll('input[type="radio"]')
+    if (radios && radios.length) {
+      const initialValue = this.value
+      radios.forEach(function (radio) {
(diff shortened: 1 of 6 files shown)


js/components/options_input.js
```diff
@@ -41,5 +44,13 @@ export default {
     valid: function () {
       return this._isValid(this.value)
     },
+    showError: function () {
+      const showError =
+        (this.initialErrors && this.initialErrors.length) || false
```
stevesweetney, line 49:
```suggestion
        this.initialErrors && this.initialErrors.length
```
I think this line can be simplified. LGTM other than that

graham-dds:
Do we want to add some regression tests here to cover this bug? I noticed that our `options_input` Vue component is not currently covered by any tests.

okaycj-atat (approved):
👍

steve-atat (approved):
:shipit:

tomdds:
@graham-dds I've added some preliminary tests for the options input. I'll push another update with some tests that cover the actual bug at hand, making sure the Vue component keeps in sync with the actual input for validation.

PR #1854, opened 2020-10-08
Optimize dockerfiles

jesse-dds:
- Condense docker files into single layers, reducing build sizes and times
- Set the ops image up to depend on the python base image, greatly reducing build times
- Remove `--build-args`
- Convert entry point into a script
- Move `COPY`s to near the ends of the files, so that changing local files does not invalidate the time-consuming parts of the build


jeffdeville-pw (approved):
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

This is even better than Graham's CRAPPY FONT!

graham-dds (commented):
Is there any way we could prune that `requirements.txt` down to the bare essentials? If I'm not mistaken, it looks like there are a few packages listed there that aren't needed for Ops work.

graham-dds (approved):

PR #1868, opened 2020-10-13
[AT-5448] Bugfix: duplicate environment ids appear in environment provisioning jobs

graham-dds:
Closes [AT-5448](https://ccpo.atlassian.net/browse/AT-5448)

The source of this bug was traced to our `get_environments_pending_creation` query. In the previous iteration of the query, the same environment ID would appear in the return value once for each active CLIN on the task order. This should be fixed now with the updated query.

This PR also removes unneeded imports and unused variables in the environments domain test module.

### DIFF
--- a/atat/domain/environments.py
+++ b/atat/domain/environments.py
@@ -101,40 +101,32 @@ class Environments(object):
         return environment
 
     @classmethod
-    def base_provision_query(cls, now):
-        return (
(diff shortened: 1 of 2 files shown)


jesse-dds (approved):

okaycj-atat (approved):
Ooof, those joins...  LGTM!

PR #1873, opened 2020-10-15
fixes bad reference to bootstrap state file

mikepw:
This fixes a bad reference to terraform bootstrap state residing in a storage account.  

Prior to fix, we could not reliably re-run the deployment image because the environment specific bootstrap state file needs to get pulled into the deploy image so it knows which resources to target.


jeffdeville-pw (approved):

jesse-dds (approved):

carleton-dds:
@dependabot rebase

PR #1874, opened 2020-10-15
Fix database and saml overlay templates

jesse-dds:
Fixes two issues that occurred during the 5th dry run.


carleton-dds:
Can we also fix that line in `Kustomization.yml`? https://github.com/dod-ccpo/atst/blob/710f20266f5e921ad1ea4e9743a133604bd09d00/ansible/roles/k8s/tasks/templates/kustomization.yml#L3

carleton-dds (changes requested):
Can we also fix that line in `Kustomization.yml`? I think it should just be `../../azure/`

 https://github.com/dod-ccpo/atst/blob/710f20266f5e921ad1ea4e9743a133604bd09d00/ansible/roles/k8s/tasks/templates/kustomization.yml#L3

jeffdeville-pw (approved):

carleton-dds (approved):

PR #1876, opened 2020-10-19
[AT-5575] Update HSTS header

okaycj-atat:
Synack reported the HSTS header missing from bugbounty.  Using the supplied curl test, we found the header to be present and in duplicate.  This PR  will remove the HSTS header (Strict-Transport-Security) from the NGINX config and update it within flask's `_set_security_headers` method.  

Additionally added tests to check for this header. 

[Jira](https://ccpo.atlassian.net/browse/AT-5575)


deploy/azure/nginx-snippets.yml
```diff
@@ -6,8 +6,6 @@ metadata:
   namespace: atat
 data:
   ssl.conf: |-
-    # Guard against HTTPS -> HTTP downgrade
-    add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; always";
```
jeffdeville-pw, line 10: Out of curiosity, why did we remove this from nginx? I gather from the ticket that it was not working via nginx?

okaycj-atat: It turned out that bugbounty always had this header and the synack guys mistakenly reported it.  Bugbounty is locked down right now so the curl test isn't going to work, but it should be working on staging.

```
$ curl -s -D- https://staging.atat.dev/ | grep Strict
Strict-Transport-Security: max-age=31536000; includeSubDomains
Strict-Transport-Security: max-age=31536000; includeSubDomains; always
```

okaycj-atat: I updated the PR description to include this explanation.

graham-dds (approved):

carleton-dds (approved):

PR #1878, opened 2020-10-19
[AT-5558] Roll back Celery to version 4

graham-dds:
We've encountered a few issues since upgrading to Celery 5.0. Most recently, we noticed that tasks are being received by the worker, but not being executed. This bug appears to have popped up immediately after the Celery 5.0 upgrade.

Since Celery 4.x will be [supported until August 2021](https://docs.celeryproject.org/en/stable/whatsnew-5.0.html#long-term-support-policy) and the next LTS release for Celery won't be until 6.0, we've decided to roll back Celery to a 4.x release for now. In the future, if we require a feature that is only available in 5.0, we may consider upgrading. Until that point, the plan is to remain on Celery 4.x until 6.0 is released.


carleton-dds:
@graham-dds were you able to confirm that upcasing the log level parameters still work okay in 4.x? We had trouble with this when we upgraded to 5: https://github.com/dod-ccpo/atst/pull/1840

graham-dds:
Yep, I did confirm this. `script/dev_queue` runs locally with no issues after the downgrade. Also, it appears that the upcasing is actually what the [4.x docs recommend](https://docs.celeryproject.org/en/4.0/reference/celery.bin.worker.html?highlight=loglevel#cmdoption-celery-worker-l).

okaycj-atat (approved):

carleton-dds (approved):

PR #1898, opened 2020-10-27
[AT-5643] Update sonar script

okaycj-atat:
Made some changes to the sonar-scanner script.  Additionally, added some directories to the exclusion list in the properties file.  

[Jira](https://ccpo.atlassian.net/browse/AT-5643)


graham-dds (commented):
When I first ran this script, it failed some of the JS tests. I think this was because I hadn't run them in a while, and the templates that the tests needed to reference didn't exist yet.

I think we can prevent this kind of thing in the future by including [run_python_render_vue_component](https://github.com/dod-ccpo/atst/blob/8e17260436b5e9d698b407698f3941cfbaef759a/script/include/test_functions.inc.sh#L27) in this script?

graham-dds (commented):
Sorry, one thing I missed the first time I reviewed:

Similar to how Jests tests might have failed before you included that `run_python_render_vue_component` function, could you see a scenario where Python tests fail because the test database isn’t up to date?

Do you think we should somehow include this [snippet](https://github.com/dod-ccpo/atst/blob/staging/script/include/run_test#L12:L25) so that we run the test DB migrations before we run tests for the coverage report?

sonar-project.properties
```diff
@@ -6,5 +6,5 @@ sonar.projectKey=dod-ccpo_atst
 sonar.organization=dod-ccpo
 sonar.javascript.lcov.reportPaths=coverage/lcov.info
 sonar.python.coverage.reportPaths=coverage.xml
-sonar.exclusions=uitests/**
+sonar.exclusions=uitests/**,tests/**
```
graham-dds, line 9: Why are we excluding tests? This seems like a good thing to scan. I remember previously in pointed out cases where we asserted two objects were equal, e.g.`expected_list` == `expected_list` instead of `expected_list` == `new_list`

okaycj-atat: Valid point.  This was requested by CCPO as it was causing some false positives.  let me circle back with @JayNewlin and see if we can find some middle ground.

okaycj-atat:
> Similar to how Jests tests might have failed before you included that `run_python_render_vue_component` function, could you see a scenario where Python tests fail because the test database isn’t up to date?
> 
> Do you think we should somehow include this [snippet](https://github.com/dod-ccpo/atst/blob/staging/script/include/run_test#L12:L25) so that we run the test DB migrations before we run tests for the coverage report?

Sure.  I can make this change.  I don't mean to push back, but wouldn't it be easier to check that the user is on the staging branch and run `script/setup`?

sonar-project.properties
JRNPW: @graham-dds and @okaycj-atat Very valid point. At the moment, it's an attempt to reduce the noise-to-signal ratio. I have created AT-5706 to turn it on after we've had enough time to digest the "real issues" that SonarCloud can identify (probably after go-live).

Noise-to-signal: This report will be available to auditors and others who are not "in the know." We need to have a way to ensure that we are catching valid security issues, but we also have to document every high (or higher) issue that SonarCloud identifies, which is a time-consuming task. Given the current project time constraints, turning this off to focus on other areas of code is necessary.

graham-dds (approved):
Thanks for bearing with me through some of my comments. 

I notice that the script was updated to only run on a clean `staging` branch. Though we might want to remove this requirement eventually, the script as is works well enough for our purposes now.

tomdds (approved):
📡

PR #1899, opened 2020-10-27
[AT-5588] Make small optimizations to application Dockerfile

graham-dds:
This commit does not represent a comprehensive Docker optimization
effort. Rather it grabs low hanging fruit to make the build of the
application container slightly faster:
1. Like in 8df2cf6dd5840a4534f87cc78366e075969a7383, we reduce the number
of layers by combining RUN steps where possible
2. One place where we actually split RUN and COPY steps apart involves
   installing project dependencies. Now, we selectively COPY Python and
Javascript dependency lockfiles, install the dependencies, and create a
production JS bundle before COPYing the rest of the source. This way, we
limit invalidating the cached dependency layers to when we acutally
change dependencies.

There are probably some dependencies still present in the application
Dockerfile that are uncecessary or could be moved to the base image. The
issue is that we install some dependenices here, but then install
other tools directly in CI/CD pipelines. Since the story for this ticket
focuses specifically on our Dockerfiles, and since we don't know what
our CI/CD setup will look like for production, we held off on making
changes that would reqiure modifying our Circle CI config. There are
still plenty of gains to be made to optimize our Docker build process
and pipelines, but they'll come later when some of these questions are
answered.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,29 +27,29 @@ RUN yum install -y \
       unzip \
       util-linux \
       wget \
-      zlib-devel
+      zlib-devel \
+    # Install yarn.
+    # https://linuxize.com/post/how-to-install-yarn-on-centos-8/
+    && curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo \
+    && rpm --import https://dl.yarnpkg.com/rpm/pubkey.gpg \
+    && dnf install -y \
+         yarn \
+         # Install the `Python.h` file for compiling certain libraries.
+         python3-devel \
+    && pip3 install poetry
 
-# Install the `Python.h` file for compiling certain libraries.
-RUN dnf install python3-devel -y
+# download Python and JS dependnencies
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,29 +27,29 @@ RUN yum install -y \
...
+    && pip3 install poetry
+
+# download Python and JS dependnencies
+COPY pyproject.toml poetry.lock poetry.toml package.json yarn.lock ./
+RUN poetry install --no-root --no-dev && yarn install
+
+# genrate a production JS bundle
+COPY ./static ./static
+COPY ./styles ./styles
+COPY ./js ./js
```
jeffdeville-pw, line 48: I think you can do:
```
COPY static/ styles/ js/ ./
```

jeffdeville-pw (approved):

okaycj-atat (approved):
LGTM

Dockerfile
graham-dds: I had thought that too, but [this SO answer](https://stackoverflow.com/a/30316600) leads me to believe that combining those `COPY` layers would copy the  _contents_ of `static`, `styles`, and `js` into the workdir.

PR #1910, opened 2020-10-29
[AT-5531] Script SAML provider creation

jesse-dds:
Scripts the process of SAML provider creation described in this [document](https://github.com/dod-ccpo/atst/blob/staging/docs/SAMLIDP.md). It accomplishes "A development IdP is automatically created for non-prod environments" of this [ticket](https://ccpo.atlassian.net/browse/AT-5449).

This script primarily:
- Creates a new app registration
- Enables SAML single sign on
- Adds a self-signed certificate
- Adds several users from a `user_service_principals.txt` files to the newly created app registration

### DIFF
--- a/ops/create-saml-idp.py
+++ b/ops/create-saml-idp.py
@@ -3,6 +3,7 @@
 import base64
 import click
 import datetime
+from os import path
 import requests
 import time
 import uuid
@@ -263,6 +264,13 @@ def assign_user_to_saml_idp(user_object_id: str, sp_object_id: str, access_token
 @click.argument("password")
 @click.argument("hostname")
 def main(tenant_id: str, application_id: str, password: str, hostname: str) -> NoReturn:
+    if not path.exists("user_service_principals.txt.txt"):
+        print("Could not locate user_service_principals.txt")
+        print(
+            "This file must exist and contain one user object id per line, each to be registered with the identity provider."
+        )
+        exit(1)
+
     access_token = authenticate(tenant_id, application_id, password)
 
     # Initialize templated app and service principle
--- a/ops/create-saml-idp.py
+++ b/ops/create-saml-idp.py
@@ -225,7 +225,6 @@ def register_self_signed_certificate(sp_object_id: str, access_token: str):
 
     headers = {
         "Authorization": f"Bearer {access_token}",
-        "Content-Type": "application/json",
     }
 
     response = requests.patch(url, json=data, headers=headers)
@@ -242,7 +241,6 @@ def assign_user_to_saml_idp(user_object_id: str, sp_object_id: str, access_token
     }
     headers = {
         "Authorization": f"Bearer {access_token}",
-        "Content-Type": "application/json",
     }
 
     response = requests.post(url, json=data, headers=headers)
--- a/ops/create-saml-idp.py
+++ b/ops/create-saml-idp.py
@@ -40,11 +40,10 @@ def create_templated_app_registration(access_token: str):
     # application template id is prescribed for SSO SAML providers
     # https://docs.microsoft.com/en-us/graph/application-saml-sso-configure-api?tabs=http#create-the-gallery-application
     url = "https://graph.microsoft.com/beta/applicationTemplates/8adf8e6e-67b2-4cf2-a259-e3dc5476c621/instantiate"
-    data = '{ "displayName": "ATAT SAML Auth" }'
+    data = {"displayName": "ATAT SAML Auth"}
 
     headers = {
         "Authorization": f"Bearer {access_token}",
-        "Content-Type": "application/json",
     }
 
     response = requests.post(url, data=data, headers=headers)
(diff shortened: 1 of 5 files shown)


graham-dds (changes requested):
Do you think we could break the `main()` function up into sub-functions? It seems like places where you included comments to explain what the function is doing at that point in the script (`# Authenticate with the provided service principle.`, `# Create a new app registration from a template.`, etc) could be good places to split things up. This will make the script easier to follow and test.

ops/create-saml-idp.py
```diff
@@ -0,0 +1,263 @@
...
+        }
+        poll_response = requests.get(
+            f"https://graph.microsoft.com/beta/servicePrincipals/{sp_object_id}",
+            headers=headers,
+        )
+
+        try:
+            poll_response.raise_for_status()
+            sp_exists = True
+        except requests.HTTPError:
```
graham-dds, line 77: This while loop makes me a bit nervous. Could we maybe only retry on certain HTTP errors? I think the only error that tells us the provisioning _might_ still be running is a 401, right? And with that, maybe we add a timeout via a max number of iterations for this loop so we're not stuck here forever?

ops/create-saml-idp.py
```diff
@@ -0,0 +1,263 @@
...
+        .public_key(key.public_key())
+        .serial_number(x509.random_serial_number())
+        .not_valid_before(datetime.datetime.utcnow())
+        .not_valid_after(
+            # Our certificate will be valid for 1 year
+            datetime.datetime.utcnow()
+            + datetime.timedelta(days=365)
+        )
+        .add_extension(
+            x509.SubjectAlternativeName([x509.DNSName("localhost")]),
```
graham-dds, line 165: is DNSName always going to be `localhost`?

ops/create-saml-idp.py
```diff
@@ -0,0 +1,263 @@
...
+    response = requests.patch(url, data=data, headers=headers)
+    response.raise_for_status()
+
+    # Register login URL with the new service principal.
+
+    url = f"https://graph.microsoft.com/beta/servicePrincipals/{sp_object_id}"
+    data = '{ "loginUrl": "https://localhost:8000/login" }'
+    headers = {
+        "Authorization": f"Bearer {access_token}",
+        "Content-Type": "application/json",
```
carleton-dds, line 123: Instead of `localhost` These need to be the FQDN of the new site. `dryrun.atat.dev`

ops/create-saml-idp.py
```diff
@@ -0,0 +1,263 @@
...
+@click.command()
+@click.argument("tenant-id",)
+@click.argument("application-id")
+@click.argument("object-id")
+@click.argument("password")
+def main(
+    tenant_id: str, application_id: str, object_id: str, password: str
+) -> NoReturn:
+
+    # Authenticate with the provided service principle.
```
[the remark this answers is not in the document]
tomdds: Not going to use this util because it assumes a configured app/csp and this script runs entirely in the terraform/ansible context, so the ATAT app isn't available

ops/create-saml-idp.py
tomdds: Added a comment about it, we're generating this cert for signing SAML payloads, so the DNS name is actually irrelevant since we don't use it for TLS or anything, just basic crypto

reset_db.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
graham-dds, line 1: Is this file meant to be in this PR? I'm not sure how it relates to the SAML stuff.

ops/create-saml-idp.py
```diff
@@ -0,0 +1,283 @@
...
+    response.raise_for_status()
+
+
+def assign_user_to_saml_idp(user_object_id: str, sp_object_id: str, access_token: str):
+    url = f"https://graph.microsoft.com/v1.0/servicePrincipals/{sp_object_id}/appRoleAssignments"
+    data = {
+        "principalId": user_object_id,
+        "principalType": "User",
+        "resourceId": sp_object_id,
+        "appRoleId": "18d14569-c3bd-439b-9a66-3a2aee01d14f",
```
graham-dds, line 241: Is there any documentation we can point to for this magic UUID?

graham-dds (approved):
Looks good! I left a few more questions and comments, but nothing stopping an approval.

Also, I noticed a few places were we provide a `"Content-Type": "application/json",` header when posting / patching json. I don't think this is necessary if you're using the `json` kwarg, but I guess it doesn't hurt to keep it.

reset_db.sh
tomdds: oh oops, not sure how this got caught up

ops/create-saml-idp.py
tomdds: The only place I'm seeing it is in an example in their docs: https://docs.microsoft.com/en-us/graph/application-proxy-configure-api?tabs=http

@jesse-dds any ideas?

carleton-dds (commented):
This all looks good 🙌. 2 quick changes would be great though:

1. Can you log the application ID of the newly created app?  It will make debugging any potential issues quicker
2. The script assumes`user_service_principals.txt` exists in the same directory. Can you add a check for that when the script runs, as well as some detail, perhaps in the help text of the script, that the file must be present and contain one user_object_id per line?

carleton-dds (approved):

PR #1911, opened 2020-10-29
[AT-5653] Add Dockerhub authentication to Docker image pulls in CircleCI

graham-dds:
Docker recently announced that [on November 1st, they will start rate limiting anonymous container pulls as well as authenticated pulls on the free plan](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/).

CircleCI responded to this news with [the following](https://circleci.com/docs/2.0/private-images/):
> CircleCI has partnered with Docker to ensure that our users can continue to access Docker Hub without rate limits. As of November 1st 2020, with few exceptions, you should not be impacted by any rate limits when pulling images from Docker Hub through CircleCI.

We can see this in action in a workflow that was run before this PR. In [this step](https://app.circleci.com/pipelines/github/dod-ccpo/atst/5932/workflows/194cbc89-409b-4e5c-85ab-755e2e12a873/jobs/26024/parallel-runs/0/steps/0-0), we see the following message:
```
Warning: No authentication provided, using CircleCI credentials for pulls from Docker Hub.
```

Nevertheless, CircleCI's announcement continues:
> However, these rate limits may go into effect for CircleCI users in the future. That’s why we’re encouraging you and your team to add Docker Hub authentication to your CircleCI configuration and consider upgrading your Docker Hub plan, as appropriate, to prevent any impact from rate limits in the future.

With this in mind, this PR adds DockerHub authentication to our CircleCI workflow so that we can mitigate rate limiting if/when it goes into effect.

Now, [ steps](https://app.circleci.com/pipelines/github/dod-ccpo/atst/5934/workflows/21f75836-282a-486e-845e-c4fcc70b1a5a/jobs/26029/parallel-runs/0/steps/0-0) in workflows that pull docker image no longer contain the warning message.

The credentials we use to authenticate against DockerHub in CircleCI are on a free tier at the time of this PR. This means that instead of being rate limited to 100 pulls every 6 hours, our limit is bumped to 200 pulls every 6 hours. We should be able to survive on this limit, but we should pay attention to our consumption rates and upgrade our plan if needed so that we don't encounter rate limiting.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -445,47 +445,47 @@ jobs:
           command: docker push << pipeline.parameters.container_registry >>.azurecr.io/ops:latest
 
   deploy-atat:
-     docker:
-       - image: docker:19
-         auth:
+    docker:
+      - image: docker:19
+        auth:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
-     environment:
-       DOCKER_BUILDKIT: 1
-     steps:
-       - checkout
-       - setup_remote_docker:
-           docker_layer_caching: false
-           version: 19.03.13
-       - install_azure_cli
-       - log_into_ops_registry
-       - run:
-           name: Build ops deployment image and deploy to container instances
-           command: |
-             docker build \
-             --cache-from <<pipeline.parameters.ops_registry >>.azurecr.io/ops-cache:latest \
-             -t ops:latest \
-             --build-arg operator_sp_client_id=${AZURE_SP} \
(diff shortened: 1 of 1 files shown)


carleton-dds (approved):
LGTM. I'll note for posterity that $DOCKERHUB_USERNAME and $DOCKERHUB_PASSWORD are now required to be part of the circleci project ENV vars

JRNPW:
Security note, also for posterity: These variables are set and stored in project environment variables in CircleCI. They are outside the repository and cannot be read or edited in the app once they are set.

PR #1912, opened 2020-10-29
Fix static folder permissions error

jesse-dds:
Fixes the fact that the project folder was owned by `root`. This was due to a mistake introduced in the last PR, where `chown` was moved before the `COPY` statements.


graham-dds (approved):
TFW you optimize the Dockerfile so much it stops working

carleton-dds (approved):

carleton-dds (approved):

PR #1928, opened 2020-11-03
[AT-5709] Automate new tenant bootstrapping

jesse-dds:
https://ccpo.atlassian.net/browse/AT-5709

Builds the operator environment in a new tenant. This PR is responsible for creating and configuring:

- A new resource group
- Virtual Network
- Container Registry
- Storage Account

I've also updated how the providers are specified according to the new preferred way in the terraform docs. https://www.terraform.io/docs/configuration/providers.html#version-an-older-way-to-manage-provider-versions

### DIFF
--- a/terraform/providers/bootstrap_new_tenant/main.tf
+++ b/terraform/providers/bootstrap_new_tenant/main.tf
@@ -21,18 +21,6 @@ provider "azurerm" {
   features {}
 }
 
-variable "namespace" {
-  type        = string
-  default     = "dev"
-  description = "Namespace of provisioned resources."
-}
-
-variable "location" {
-  type        = string
-  default     = "East US"
-  description = "Azure region in which resources are provisioned."
-}
-
 resource "azurerm_resource_group" "bootstrap_resource_group" {
   name     = "cloudzero-ops-${var.namespace}"
   location = var.location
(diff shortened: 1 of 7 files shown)


tomdds (approved):

terraform/providers/bootstrap_new_tenant/main.tf
```diff
@@ -0,0 +1,45 @@
...
+  resource_group_name      = azurerm_resource_group.operations_resource_group.name
+  location                 = var.operations_location
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+
+  network_rules {
+    default_action             = "Deny"
+    virtual_network_subnet_ids = [azurerm_subnet.deployment_subnet.id]
+  }
+}
```
jeffdeville-pw, line 38: Don't we also need to create a container to store the state?

jeffdeville-pw (approved):

jeffdeville-pw:
I'm assuming that we'll have another PR that stores this state remotely, and then another that will wire these outputs to the inputs of the next terraform command, right?

terraform/providers/bootstrap_new_tenant/main.tf
jesse-dds: Yes there is a container in the #1937 pull request 👍

PR #1937, opened 2020-11-05
Wire up to deployment (less) config values

jesse-dds:
### Wire up outputs between terraform stages

Dynamically generate the `backend` of `application_env` and `bootstrap` so that these terraform modules can read *directly* from the outputs of earlier stages. This allows configuration values to be passed through directly.

### Reorganize modules

Reorganize the module structure to follow the structure specified [here](terraform.io/docs/modules/index.html#module-structure):

- `main.tf`
- `variables.tf`
- `outputs.tf`

The addition of the `versions.tf` file is due to this file being auto-created by running `terraform 0.13upgrade` leading me to believe that is an appropriate file to put in the versioning, backend, and remote state information.

### Remove some glue code

Since the terraform stages can read from each other directly, we don't need the ansible glue that does `tfstate` management or variable management.

### Miscellaneous

- Sectioned off *pass-through variables*.  These are variables that come in through `variables` and are passed straight to `outputs.tf`. Not configuring the terraform in any appreciable way.
- Add the `recover` key permission to the key vault administrator.
- Removed unused `images` folder
- Add terraform `plan` and lock data to the `.gitignore`


mikepw:
how do you handle restarting the deployment during container deploys?

PR #1948, opened 2020-11-09
[AT-5708] Terraform for deployment VM

carleton-dds:
This Terraform will spin up a Ubuntu/XFCE4 virtual machine in its own vnet/subnet that should have all the necessary installations and configurations (az, kubectl, Ansible, etc)  to drive a dry run deployment. You can test yourself by setting your `az` to one of the `pwsandbox` subscriptions and doing the usual `terraform plan/apply` routine. You should be able to SSH and RDP into the machine with the credentials supplied in the TF output and find all the tools good to go. (Microsoft Remote Desktop is a good RDP app if you need one)

In order to do most of the installs and configurations we utilize the [cloud-init](https://cloudinit.readthedocs.io/en/latest/) formatted `cloud_config.yml` file, which can take up to 10 minutes after the Terraform has completed to also complete. You know it's complete when the `/var/log/cloud-init-output.log` file on the VM displays a message the looks like the following:
```
cloud-init v. 20.3-2-g371b392c-0ubuntu1~18.04.1 finished at Mon, 09 Nov 2020 15:50:01 +0000. Datasource DataSourceAzure [seed=/dev/sr0].  Up 4052.43 seconds
```
This message will be logged even if the cloud-init encountered problems, so you may need to check earlier in the log for any error messages.

I placed these files in the `/ops` directory but it may need to be moved and have the exact vnet setting changed to work best with the dry run deployment

### DIFF
--- a/ops/vm/cloud_config.yml
+++ b/ops/vm/cloud_config.yml
@@ -1,0 +1,50 @@
+#cloud-config
+package_update: true
+packages:
+  - software-properties-common
+  - apt-transport-https
+  - wget
+  - firefox
+  - xmlsec1
+  - curl
+  - vim
+  - gettext
+  - unzip
+  - libbz2-dev
+  - libssl-dev
+  - libpq-dev
+  - jq
+  - gnupg2
+  - xrdp
+  - xfce4
+  - libxml2-dev
+  - libxmlsec1-dev
+  - libxmlsec1-openssl
+apt_sources:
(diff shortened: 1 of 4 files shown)


jeffdeville-pw (approved):
Nice!

jesse-dds (approved):

jesse-dds (approved):

PR #1955, opened 2020-11-12
[AT-5688] A few changes to clean up the sonar cloud script

okaycj-atat:
This PR will cover a couple of additional tweaks to the SonarCloud config.

- Make sure that the local branch is the same as origin.
- Add a couple of directories to the duplication exclusion list. 

[Jira](https://ccpo.atlassian.net/browse/AT-5688)


script/sonar_scanner
```diff
@@ -34,6 +34,8 @@ fi
 # exit when any command fails
 set -e
 
+git pull origin ${TARGET_BRANCH}
```
graham-dds, line 37: Maybe we could provide a default value for `TARGET_BRANCH` of `git branch --show-current`, instead of having to remember to supply an environment variable?

okaycj-atat: My apologies, maybe I could have made this more clear.  This is a variable that is set in this script. 

https://github.com/dod-ccpo/atst/blob/4e314fad85bf9416f5830e5aea5f05a1287b4e7e/script/sonar_scanner#L18

graham-dds (approved):

carleton-dds (approved):

PR #1958, opened 2020-11-13
[AT-5744] ops phase 1

jeffdeville-pw:


python.Dockerfile
```diff
@@ -36,8 +37,8 @@ RUN --mount=type=secret,id=redhat_username \
     subscription-manager remove --all && \
     subscription-manager clean && \
     subscription-manager register \
-        --username $(cat /run/secrets/redhat_username) \
-        --password $(cat /run/secrets/redhat_password) && \
+        --username PromptworksRHEL \
```
carleton-dds, line 40: Credentials should never be committed. We'll need to rotate these

carleton-dds:
Please run `script/format` on this branch. looks like Circle is complaining about some formatting tweakings

ops/phase1/run.sh
```diff
@@ -0,0 +1,94 @@
...
+
+echo "Terraform Bootstrap"
+cd ../bootstrap
+terraform init
+terraform plan -out out.plan
+terraform apply out.plan
+
+echo "Building RHEL"
+cd ../../../ops/phase1
+[ ! -d "atat-rhel-image" ] && git clone git@github.com:dod-ccpo/atat-rhel-image.git
```
carleton-dds, line 48: As you know, there is the problem of accessing the private repo when `git` hasn't been set up for  the right user. What we did on Thursday was clone over HTTPS, which will prompt the user to input their username and password, avoiding the need to generate ssh keys or anything.

Due to 2FA simply using their password will fail, and the user will need to generate a [Personal Access Token](https://github.com/settings/tokens), which isn't hard to do, but we may want to prompt the user to do with an echo

```suggestion
if [ ! -d "atat-rhel-image" ]; then
	echo "Use a Personal Access Token instead of your password to auhenticeated with Github (https://github.com/settings/tokens)"
	git clone https://github.com/dod-ccpo/atat-rhel-image.git
fi
`````

File: ops/phase1/run.sh [hunk 3]
```diff
@@ -0,0 +1,94 @@
...
+  --image rhel-py \
+  --file python.Dockerfile \
+  .
+
+echo "Build Ops Dockerfile"
+export COMMIT_SHA=$(git rev-parse HEAD)
+az acr build --registry ${REGISTRY_NAME} \
+  --build-arg IMAGE=${REGISTRY_NAME}/rhel-py:latest \
+  --image ops:${COMMIT_SHA} \
+  --file python.Dockerfile \
```
carleton-dds, line 67: Shouldn't this be `ops.Dockerfile`?

carleton-dds:
Overall I think this all looks really good. I was able to get it to mostly work on my machine. A few outstanding topics:
- We should have some sort of guard clause for the needed variables at the top. We could also include a quick check for `which terraform` to and install if needed
- We may also want to echo the current subscription to remain folks of where it's working
- Something that came up on Thursday was both `tfvars` and `atatdev.pem` files, we should address those as needed
- When I used `carleton1` as my namespace I got an error about max length. We should add a caution about that:
```
Error: name ("czopsstorageaccountcarleton1") can only consist of lowercase letters and numbers, and must be between 3 and 24 characters long
```

graham-dds (approved):
Approving this as-is since we're merging into @jesse-dds 's feature branch. This will keep ops work moving. The PR that eventually merges @jesse-dds 's branch to staging should have a more careful review.

ops/phase1/run.sh
jeffdeville-pw: Yup, fixed

PR #1966, opened 2020-11-17
[AT-5659] Change file access for nginx-secrets volume

steve-atat:
[Jira](https://ccpo.atlassian.net/browse/AT-5659)

Update `nginx-secrets` volume to fit these criteria

- Read/write only by root

- Readable only by nginx group

- No access for everyone

### DIFF
--- a/deploy/azure/azure.yml
+++ b/deploy/azure/azure.yml
@@ -28,7 +28,7 @@ spec:
         fsGroup: 101
       initContainers:
         - name: change-file-access
-          image: busybox
+          image: $CONTAINER_IMAGE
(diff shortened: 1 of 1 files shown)


deploy/azure/azure.yml
```diff
@@ -26,6 +26,13 @@ spec:
     spec:
       securityContext:
         fsGroup: 101
+      initContainers:
+        - name: change-file-access
+          image: busybox
```
carleton-dds, line 31: Using an image straight from docker hub may be a problem down the line. Would it be possible to use $CONTAINER_IMAGE to make this change?

tomdds (approved):

okaycj-atat (approved):
LGTM!

PR #1976, opened 2020-11-18
[AT-5710] ops phase2

jeffdeville-pw:


ops/phase1/run.sh
```diff
@@ -19,38 +19,40 @@ export TF_VAR_operator_client_id=$ARM_CLIENT_ID
...
-echo $TF_VAR_operator_tenant_id
+# echo "ARM_CLIENT_ID=$ARM_CLIENT_ID"
+# echo "ARM_CLIENT_SECRET=$ARM_CLIENT_SECRET"
+# echo "ARM_SUBSCRIPTION_ID=$ARM_SUBSCRIPTION_ID"
+# echo "ARM_TENANT_ID=$ARM_TENANT_ID"
+# echo "-----------"
+# echo $TF_VAR_operator_subscription_id
+# echo $TF_VAR_operator_client_id
+# echo $TF_VAR_operator_client_secret
+# echo $TF_VAR_operator_tenant_id
```
graham-dds, line 30: Why are these commented out? It seems like we need them when we `az container create` below

graham-dds (commented):
A few questions off the bat:
- Instead of adding another `pyproject.toml` in a subdirectory just to add `Click`, could we add additional dependencies we need to the root `pyproject.toml`'s dev dependencies or [rely on the standard library](https://docs.python.org/3/library/argparse.html)? 
- Similarly, any reason to start a new gitignore, rather than dump everything into the root gitginore? 
- Is there a reason we want to leave commented out code in here? Can we safely delete what's commented out?

ops/phase1/run.sh
```diff
@@ -0,0 +1,95 @@
...
+  --registry-password ${TF_VAR_operator_client_secret} \
+  --registry-username ${TF_VAR_operator_client_id} \
+  --memory 4 \
+  --cpu 4 \
+  --secure-environment-variables "OPS_RESOURCE_GROUP=$TF_VAR_resource_group_name" "OPS_STORAGE_ACCOUNT=$TF_VAR_storage_account_name" "SUBSCRIPTION_ID=$TF_VAR_operator_subscription_id" "SP_CLIENT_ID=$TF_VAR_operator_client_id" "SP_CLIENT_SECRET=$TF_VAR_operator_client_secret" "TENANT_ID=$TF_VAR_operator_tenant_id" "OPS_REGISTRY=$REGISTRY_NAME" "NAMESPACE=$1" "LOGGING_WORKSPACE=$LOGGING_WORKSPACE" \
+  --command-line "tail -f /dev/null" \
+  --log-analytics-workspace $LOGGING_WORKSPACE \
+  --restart-policy Never
+
+echo "export OPS_RESOURCE_GROUP=$TF_VAR_resource_group_name" >> ../../.envrc
```
jeffdeville-pw, line 84: Delete

jesse-dds (approved):

PR #1977, opened 2020-11-18
[AT-5744] Toggle subnet storage restriction

graham-dds:
In ops phase 1, by the end of the `bootstrap` terraform provider, we create (among other things) a vnet, subnet, storage account, and two storage containers. 

By the end of "phase 1", we want to limit access to the storage account to only the subnet that was created by the `bootstrap` provider. But if we specify this rule in terraform, the provider cannot create the two storage containers since it isn't apart of the vnet/ subnet. So, we run the provider and then use `az` to set this rule on the storage account after the fact. 

This creates another problem, however. In the `bootstrap` provider, we dynamically set one of the containers that is created as a remote terraform backend. If we try to run further `terraform` commands after the script is complete (before we move to phase 2), we cannot access our remote state because we aren't in the vnet/subnet that the container allows access from.

To solve this issue, this PR adds logic to the `run` script to remove the subnet access restriction on the storage account, if it exists. After `bootstrap` provider runs, we lock the account down again.


graham-dds:
Closing this 
1. Because it's pretty behind at this point
2. It probably needs to be reopened (if this work is still needed) against staging

PR #1982, opened 2020-11-23
[AT-5703] Fix CLIN dropdown edit bug

tomdds:
This adds a new option to the OptionsInput macro to change how the data is fed into the vue template. Because we need data literals for doing automated testing, we need a way to also provide string literals to the options input.

### DIFF
--- a/templates/components/options_input.html
+++ b/templates/components/options_input.html
@@ -8,12 +8,19 @@
   label=True,
   show_validation=True,
   disabled=False,
-  optional=True) -%}
+  optional=True,
+  data_literal=False) -%}
   <optionsinput
     name='{{ field.name }}'
     inline-template
     {% if field.errors %}v-bind:initial-errors='{{ field.errors | list }}'{% endif %}
-    {% if field.data and field.data != "None" %}v-bind:initial-value="{{ field.data }}"{% endif %}
+    {% if field.data and field.data != "None" %}
+      {% if data_literal %}
+        v-bind:initial-value="{{ field.data }}"
+      {% else %}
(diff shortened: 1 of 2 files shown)


templates/components/options_input.html
```diff
@@ -8,12 +8,19 @@
...
   <optionsinput
     name='{{ field.name }}'
     inline-template
     {% if field.errors %}v-bind:initial-errors='{{ field.errors | list }}'{% endif %}
-    {% if field.data and field.data != "None" %}v-bind:initial-value="{{ field.data }}"{% endif %}
+    {% if field.data and field.data != "None" %}
+      {% if data_literal %}
+        v-bind:initial-value="{{ field.data }}"
+      {% else %}
+        v-bind:initial-value="'{{ field.data }}'"
```
carleton-dds, line 21: @tomdds can you help me understand this if statement? Seems like it's actually not doing anything

tomdds: yeah, it looks REALLY similar, the basic idea is that if `data_literal` is `True` you'll get output like:
`v-bind:initial-value="FIELD_DATA"`
which, when the component mounts, it'll attempt to look up `FIELD_DATA` as a variable, rather than a string literal. This enables us to inject test values from our component tests, if `data_literal` is `False` it'll wrap the value in a second set of quotes:
`v-bind:initial-value="'FIELD_DATA'"`
which will pass the string literal `FIELD_DATA` as the initial value, rather than a variable reference.

okaycj-atat (approved):
LGTM!

graham-dds (approved):

PR #1993, opened 2020-11-30
Fix privatek8s fw routes

jeffdeville-pw:


mikepw:
additional fixes;
* remove selector from atet-main private load balancer
* adds back `---` block separators to k8s flex-vol template generation.
* Manual Failover to maintenance page is simple. Just manually reconfigure the firewall rule's `translated IP address` value  from aks load balancer to the maintenance page container instance ip in the adjacent, less prioritized firewall rule (that IP is also available in portal, terraform outputs and terraform state).

After deploy to test;
* internally via `curl -k -v https://LB_IP` from bastion container
* externally via (reconfigure NAT rule in firewall to allow desired source IPs for both rules), use a browser to hit the firewall's public IP address as in `https://FW_PUBLIC_IP`, you can use /etc/hosts on your local machine to resolve a more atat-ish hostname.

Doc to configure asymmetric routing for a public load balancer (when/if needed) are available in confluence.

mikepw:
additional fix:
* The AKS service principle is/has been created dynamically for sometime but was being inputted via `app.tfvars.json` with an Azure AD app that was created back in April. That service principal didn't have the necessary network permissions to create the internal load balancer needed for a truly private instance of ATAT.  A few commits back this was fixed.
* All service principals created in terraform now include the `environment` or `random_pet` identifier so they're more easily identifiable in the portal.

PR #1998, opened 2020-12-01
[AT-5734] Under maintenance page

okaycj-atat:
Update style for generated Under Maintenance page.  

Also, from the [LGTM docs](https://lgtm.com/help/lgtm/alert-suppression#python), I have supressed the alert for a few lines at the request of reviewers.  

Screen Shot 2020-12-01 at 11 51 50 AM
Screen Shot 2020-12-01 at 11 50 07 AM
Screen Shot 2020-12-01 at 11 49 44 AM

[Jira](https://ccpo.atlassian.net/browse/AT-5734)


graham-dds (commented):
Looks great! 

Do you think we could write styles like we do with the rest of the app, rather that in-lining them in the HTML? It would also allow you to write them in a `sass` syntax, which is nice.

You can define those styles as an scss file (maybe in the `sections` directory) and then import it into `atat.scss`. You'll need to `yarn build` to have the styles show up before you generate the page. [Here's](https://gist.github.com/graham-dds/9d5b1bdef78091a78067ede531ecfed2) a diff of what I was thinking.

graham-dds (approved):
🔨 🔧

alembic/env.py
```diff
@@ -33,7 +33,7 @@
 app_config = make_config()
 config.set_main_option("sqlalchemy.url", app_config["DATABASE_URI"])
 
-from atat.database import db
+# from atat.database import db
```
graham-dds, line 36: I'm a _little_ scared about commenting this line out, just because it's lived here for so long. Are we sure we don't need to have `db` in scope for some weird reason? It looks like tests pass, but I wonder if this will bite us when we deploy or something. 

If we're sure that we don't need db in scope, can we just delete this line rather than comment it out?

ipython_setup.py
```diff
@@ -1,5 +1,4 @@
 from atat.app import make_config, make_app
-from atat.database import db
```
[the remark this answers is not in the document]
okaycj-atat: I am sure there is.  I'll make the update.

alembic/env.py
okaycj-atat: Sure thing.  I meant to, it was an error on my part.  Nice catch.

okaycj-atat:
@graham-dds I am not sure what's up with LGMT, but the errors aren't there anymore and I've reverted the code where I removed imports.  Let me know if this is sufficient.

steve-atat (approved):
👍

PR #2009, opened 2020-12-08
refactor: Update a few URLs to use HTTPS/TLS

steve-atat:
[Jira](https://ccpo.atlassian.net/browse/AT-5745)

### DIFF
--- a/ansible/utils/__init__.py
+++ b/ansible/utils/__init__.py
@@ -14,7 +14,7 @@ def acquire_token():
     token = None
     try:
         token_res = requests.get(
-            "http://169.254.169.254/metadata/identity/oauth2/token",
+            "https://169.254.169.254/metadata/identity/oauth2/token",
             params=token_params,
             headers=token_headers,
             timeout=10,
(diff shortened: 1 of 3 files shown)


carleton-dds (approved):

okaycj-atat (approved):

ansible/utils/__init__.py
```diff
@@ -14,7 +14,7 @@ def acquire_token():
     token = None
     try:
         token_res = requests.get(
-            "http://169.254.169.254/metadata/identity/oauth2/token",
+            "https://169.254.169.254/metadata/identity/oauth2/token",
```
whit1206, line 17: @steve-atat What is this hardcoded URL? Can it be moved to a config?

graham-dds: @whit1206 This looks like it's a [special URL](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service).

```
[Azure Instance Metadata Service (IMDS)] is a REST endpoint that's available at a well-known, non-routable IP address (169.254.169.254). You access it only from within the VM. Communication between the VM and IMDS never leaves the host. Have your HTTP clients bypass web proxies within the VM when querying IMDS, and treat 169.254.169.254 the same as 168.63.129.16.
```

PR #2012, opened 2020-12-09
ATAT-5514 adding support link

gottlieb-intevity:


graham-dds:
I believe this functionality was added in #1499 

Could you try setting an environment variable `SERVICE_DESK_URL` with some url and seeing if that generates a link in the navbar that points to it?

PR #2018, opened 2020-12-10
[AT-5815] Integrate SAML IDP creation script into deployment code

jesse-dds:
Run the SAML IDP creation script and pass it into the environment of the application.

### DIFF
--- a/deploy/azure/secret_provider.yml
+++ b/deploy/azure/secret_provider.yml
@@ -74,6 +74,10 @@ spec:
           objectName: AZURE-HYBRID-REPORTING-SECRET
           objectAlias: AZURE_HYBRID_REPORTING_SECRET
           objectType: secret
+        - |
+          objectName: SAML-DEV-IDP-URI
+          objectAlias: SAML_DEV_IDP_URI
+          objectType: secret
 
 ---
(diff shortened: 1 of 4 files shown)


carleton-dds:
What you've got here is great 🙌 

It just occurred to me though that the `user_service_principals.txt` script is required by the SAML script, which means that file has to be present, which means we'll probably have to make a few changes:
- In [Initial Deployment Doc](https://ccpo.atlassian.net/wiki/spaces/AT/pages/501940394/Initial+Deployment) we'll now need to instruct deployment drivers to create `user_service_principals.txt` and fill it one per line with the user object ID of all developers that need dev access to the front end. (User object IDs can be found on [the profile page](https://portal.azure.com/#blade/Microsoft_AAD_IAM/UserDetailsMenuBlade/Profile/userId/3a9cf02e-9dbe-416a-897e-1a4b18059fbd) of each user in the `Users` module)
- This file will need to be uploaded to the storage container along with the tfvars file and the others
- `ops/phase2/provisioner.py` will need to download this file along with the others

Also:
- Looks like a `script/format` run is needed to catch a couple of things
- now that #2019 has been merged into staging, your changes probably should be made against that branch.

PR #2034, opened 2020-12-14
[AT-5815] Generate SAML IDP as part of deployment

jesse-dds:
Run the SAML IDP creation script and pass it into the environment of the application.

## How to test

- Go to an existing provisioner container instance
- Clone down the branch
- Run provisioner.py
- You’d then have to jump into the bastion
- Clone the new deploy/azure/secret_provider.yml overlay
- `kubectl apply`
-  that as well
- Then you should be able to see in the `kubectl describe`


tomdds (approved):

carleton-dds (approved):

jeffdeville-pw:
Note for Hunter. This branch we _think_ works, but the team did not have time to run a full deployment to test it. If you're picking it up, the magic happens in the provision step. Keep an eye on that one, and if the SAML_DEV_IDP_URI is set correctly, you should be in good shape.

jasonburkert-ccpo:
Closing per discussion with @nkingsley-ccpo and @kylelaker-ccpo

PR #2045, opened 2020-12-16
Delete all jobs prior to attempting to run a migration

jesse-dds:
As a way of mitigating migration job failures, delete other competing jobs. This is a pretty jank way of fixing this issue, hopefully a more robust CD solution for deploying to the private cluster gets rid of this.


steve-atat (approved):
👍

carleton-dds (approved):

PR #2046, opened 2020-12-16
Add a prefix to keep psql server names unique

jeffdeville-pw:
Hunter hit an issue where the server name was not unique because their env name was 'test'. This adds cz-


tomdds (approved):
:cz:

steve-atat (approved):
:shipit:

JRNPW (approved):
Change and its reason is clear to me.

PR #1918, opened 2020-10-30
Merge to Master

carleton-dds:
This will update master with all changes since the last master merge: #1823

PR #2051, opened 2020-12-18
Merge Staging Into Master

carleton-dds:
This represents the final merge from Promptworks.
This will update master with all changes since the last master merge: #1918


JRNPW (approved):
1. I have been able to login successfully multiple times with my CAC and as `/dev-login` users.
2. I have been able to navigate through and around the UI without any problems.
3. Ghost Inspector also performs regular UI tests of even more scenarios, and those are all passing.
4. While AT-5558 is still open, the provisioning process itself works as has been proven with other environments. There has not been any major change to provisioning in quite a while. The bug is not an indication of a problem with provisioning, but it appears to be a configuration issue. See the bug ticket for further details.

This code is ready for handoff.

jeffsegal-ccpo (approved):

PR #2076, opened 2021-01-25
Ensure dev DOD IDs are 10 digits

kylelaker-ccpo:
More expressedly choosing the a set of 9 digits seems like the intent is
clearer than choosing a random number between 0 and a magic number based
on the digits. The previous behavior was to always have at least one
leading zero so that constraint is preserved.

Resolves: AT-5914

### DIFF
--- a/atat/routes/saml_helpers.py
+++ b/atat/routes/saml_helpers.py
@@ -1,5 +1,6 @@
+import random
 import re
-from random import randint
+import string
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


jeffsegal-ccpo (approved):

jasonburkert-ccpo (approved):
Ran generating code in a REPL to create three sample ID values (0386717380, 0191623599, 0856598569).  All three sample IDs match the regex used in the unit test.

PR #2084, opened 2021-02-03
Feature/AT-5899

nkingsley-ccpo:
Activating sonarqube scanning.


jeffsegal-ccpo (approved):
Went over changes with Nate over the phone, looks good.

PR #2085, opened 2021-02-03
AT-4691 Created banner that indicates 'Classifier' status

hiz-ccpo:


nkingsley-ccpo (changes requested):
There are a lot of items in here that should be in the .gitignore file:

atst2.code-workspace
celerybeat-schedule.*
*.crt
postgresql
redis-server

we should add those in to the .gitignore file, and remove from here before continuing with the review

PR #2086, opened 2021-02-03
Feature/at 5944 flask debug toolbar

jGonzalez-CCPO:
https://ccpo.atlassian.net/browse/AT-5944


nkingsley-ccpo:
@jGonzalez-CCPO, I would recommend that we have a separate app.py for this (e.g. app-debug.py) and update the Procfile to use that instead, so we don't have a bleed-over of this that could propagate into prod.

nkingsley-ccpo:
Also, flask_debugtoolbar needs to be added as a python dev dependency

jGonzalez-CCPO:
> @jGonzalez-CCPO, I would recommend that we have a separate app.py for this (e.g. app-debug.py) and update the Procfile to use that instead, so we don't have a bleed-over of this that could propagate into prod.

done

atat/app.py
```diff
@@ -88,6 +89,11 @@ def make_app(config):
     if ENV != "prod":
         app.register_blueprint(dev_routes)
 
+    if (ENV == "dev" or ENV == "development") and app.config[
+        "DEV_DEBUG_TOOL"
+    ] == "True":
+        debug_tools_bar(app)
+
```
kylelaker-ccpo, line 96: Let's break out these names to make them slightly less "magic". Or use a regex and accept any environment that starts with `dev`.

```suggestion
    dev_env_names = ("dev", "development")
    if (ENV in dev_env_names) and app.config["DEV_DEBUG_TOOL"] == "True":
        debug_tools_bar(app)

```

atst2.code-workspace
```diff
@@ -5,6 +5,6 @@
 		}
 	],
 	"settings": {
-		"workbench.colorTheme": "Visual Studio Light"
```
kylelaker-ccpo, line 8: This modification will impact all other developers on the project using VS Code. It may be best to not commit any changes to this file. I think #2090 actually deletes this file.

config/base.ini
```diff
@@ -93,4 +95,4 @@ AZURE_HYBRID_TENANT_ID
 AZURE_INVOICE_SECTION_ID
 AZURE_TENANT_ADMIN_PASSWORD
 AZURE_TENANT_ADMIN_USERNAME
-AZURE_USER_OBJECT_ID
+AZURE_USER_OBJECT_ID
```
kylelaker-ccpo, line 98: Removing the EOL at the end of this file is likely a mistake of the text editor or IDE. Please restore the newline.

kylelaker-ccpo (changes requested):
Could you provide some context on why so many changes are being made to the `yarn.lock` file? It seems like all of the `integrity` lines with the SHA hashes are being removed which doesn't seem ideal.

atat/debug.py
```diff
@@ -0,0 +1,15 @@
...
+
+def debug_tools_bar(app):
+    """
+    Set debuger tool bar
+    """
+    # the toolbar is only enabled in debug mode:
+    app.debug = True
+    # set a 'SECRET_KEY' to enable the Flask session cookies
+    app.config["SECRET_KEY"] = (
+        app.config["DEV_DEBUG_TOOL_SECRET_KEY"] or "jfhdfjshkjskjdjdhhe"
```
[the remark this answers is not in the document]
jGonzalez-CCPO: thanks. 

so, we don't need this step then.

atst2.code-workspace
jGonzalez-CCPO: Agree. I would reverse this.

atat/app.py
jGonzalez-CCPO: The standard name for the environment for development by the documentation of Flask should be `development.` 

But for some reason, everywhere in the code, the old team uses `dev.` 

I am trying to have a working theory of why they name this in a custom matter. At this moment, I don't know why. But if you can give me some inside about that, maybe in the ticket. 

> base on https://flask.palletsprojects.com/en/1.1.x/cli/#watch-extra-files-with-the-reloader

So, for me would be better to only have it on `development` (that exact word only), but if you read in the code there are some service been provided for the `dev` (also an exact word), so I try to make available for both.  

I am also open to plan this better with you later, probably a 15 min meeting.

kylelaker-ccpo: Yep! I see that the goal is to make it work for both `dev` and `development` as the code is written; this small refactoring would accomplish the same thing but actually assign a "name" to those strings.

I feel we should have a discussion as part of a separate ticket whether we can use `development` instead of `dev`.

atst2.code-workspace
jGonzalez-CCPO: done

kylelaker-ccpo (commented):
Couple recommendations on renaming as well in addition to moving this to a dev dependency

atat/debug.py
```diff
@@ -0,0 +1,13 @@
+from flask_debugtoolbar import DebugToolbarExtension
+
+
+def debug_tools_bar(app):
```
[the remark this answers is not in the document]
jGonzalez-CCPO: done

atat/app.py
```diff
@@ -14,6 +14,7 @@
 
 from atat.assets import environment as assets_environment
 from atat.database import db
+from atat.debug import debug_tools_bar
```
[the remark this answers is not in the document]
jGonzalez-CCPO: done

atat/debug.py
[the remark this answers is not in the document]
jGonzalez-CCPO: done

jGonzalez-CCPO:
METADATA FIX RE-BASE on https://github.com/dod-ccpo/atat/pull/2110

PR #2090, opened 2021-02-03
At 4961 - Added Configurable Classifier Banner

hiz-ccpo:


jeffsegal-ccpo (approved):
Looks good!

styles/components/_classifier_banner.scss
```diff
@@ -0,0 +1,20 @@
...
+  position: fixed;
+  text-align: center;
+  padding: ($gap / 2) $gap;
+  z-index: 15;
+  height: $classifier-banner-height;
+  width: 100%;
+  z-index: 15;
+  color:#ffffff;
+
+  &.classified{
```
kylelaker-ccpo, line 11: It seems like this may not be used since the colors are set in the application configuration. Was it intentional to keep this in the SCSS?

kylelaker-ccpo (commented):
Looks good! A quick pass through `script/format` should fix up the CI build! It looks like the formatter prefers spaces between colons (`:`) and values.

styles/components/_classifier_banner.scss
hiz-ccpo: good catch. I'll remove these styles.

PR #2108, opened 2021-02-08
Update to Python 3.8

kylelaker-ccpo:
Adding support and fixing breakages with Python 3.8 as well as using it wherever possible.

Ticket: AT-5936


kylelaker-ccpo:
*Note*: Any users with a local `.venv/` will like need to `rm -rf .venv/ && ./script/setup`; otherwise the virtual environment will still by Python 3.7.3 which may cause issues.

nkingsley-ccpo (approved):

jasonburkert-ccpo (approved):
Appreciate the detailed explanations.  

Re: commit 6163b29ab17963d533be1e16ba51d9a9e5c57f4e about ops.Dockerfile -- Do we need a follow-up ticket to investigate whether the Python link is needed and can perhaps be removed?

kylelaker-ccpo:
@jasonburkert-ccpo: I believe that there is larger work happening on the Dockerfiles as well as the infrastructure code in general; we should probably sync up with @nkingsley-ccpo.

PR #2117, opened 2021-02-11
At 5946 remove public pages

jGonzalez-CCPO:
ticket: https://ccpo.atlassian.net/browse/AT-5946
doc: https://ccpo.atlassian.net/wiki/spaces/AT/pages/1296171468/AT-5946+-+Remove+unwanted+public-facing+pages

### DIFF
--- a/atat/domain/auth.py
+++ b/atat/domain/auth.py
@@ -14,7 +14,7 @@ UNPROTECTED_ROUTES = [
     "atat.logout",
     "atat.unauthorized",
     "static",
-    "atat.about",
+    # "atat.about",
     "atat.login",
     "atat.handle_login_response",
 ]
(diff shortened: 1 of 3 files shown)


nkingsley-ccpo (changes requested):
Rather than commenting these items out, we should just remove them, as since this project is stored in git we have a historical trail of what was added/removed and can revert these changes if it were ever needed.

jGonzalez-CCPO:
> .

jGonzalez-CCPO:
> Rather than commenting these items out, we should just remove them, as since this project is stored in git we have a historical trail of what was added/removed and can revert these changes if it were ever needed.

I agree, but that was explicitly asked on the ticket.

jGonzalez-CCPO:
replaced by: https://github.com/dod-ccpo/atat/pull/2126/files

PR #2130, opened 2021-02-17
Exlcuding items from code coverage that don't need to be tracked

nkingsley-ccpo:


jasonburkert-ccpo (approved):
This was a result of a discussion about code coverage between @nkingsley-ccpo @kylelaker-ccpo @anhtran-ccpo and myself.

PR #2148, opened 2021-02-19
Fix SonarCloud Bug finding in create-saml-idp.py

kylelaker-ccpo:
SonarCloud reports the fact that we're passing all arguments by name as
a Blocker Bug. This is actually a false positive since we were passing
all the correct parameters in the correct order; however, since the code
is functional without the names we can go ahead and omit them.


jasonburkert (approved):
Verified removal of named arguments in favor of positional arguments for the purpose of overcoming the SonarCloud false positive.

jasonburkert-ccpo (approved):
Lulz. Verified removal of named arguments in favor of positional arguments for the purpose of overcoming the SonarCloud false positive.

PR #2149, opened 2021-02-19
AT-5985 Chage Home link to ATAT

jGonzalez-CCPO:
ticket: https://ccpo.atlassian.net/browse/AT-5985


jasonburkert-ccpo (approved):
Verified on local dev

PR #2152, opened 2021-02-19
AT-5977 header over content fix

jGonzalez-CCPO:
ticket:
https://ccpo.atlassian.net/browse/AT-5977


jasonburkert-ccpo (approved):
Verified on local dev

PR #2160, opened 2021-02-22
AT-5994 new seal logo added

jGonzalez-CCPO:
ticket: https://ccpo.atlassian.net/browse/AT-5994


jasonburkert-ccpo (approved):
Verified in local dev.
seal before and after

PR #2162, opened 2021-02-22
Use System Random to for random numbers

kylelaker-ccpo:
This attempts to use the system random source for operations that
require random numbers, whether those operations are cryptographically
sensitive or not.


jasonburkert-ccpo (approved):
Decided on this change after discussing SonarCloud Security HotSpot with @kylelaker-ccpo 
Since this is used to generate an id, we decided to improve this use of random.

PR #2171, opened 2021-02-25
AT-6030 Fix user interface incompatibility with IE 11

jGonzalez-CCPO:
https://ccpo.atlassian.net/browse/AT-6030?focusedCommentId=310288

making all dependencies look version fix compatibility problem with IE


jasonburkert-ccpo (changes requested):
I've found that I can safely re-introduce carets in _devDependencies_ and _resolutions_ without impacting functionality on IE 11.  However, attempting to do the same for _dependencies_ breaks IE 11.  Would like to perform additional testing to identify a specific dependency that's causing the IE 11 breakage.  Plan to introduce one caret at a time and test.

Please re-introduce carets in _devDependencies_ and _resolutions_ and commit.

Additionally, we must troubleshoot and resolve the CI test error for SonarCloud prior accepting and merging this PR.

jGonzalez-CCPO:
> I've found that I can safely re-introduce carets in _devDependencies_ and _resolutions_ without impacting functionality on IE 11. However, attempting to do the same for _dependencies_ breaks IE 11. Would like to perform additional testing to identify a specific dependency that's causing the IE 11 breakage. Plan to introduce one caret at a time and test.
> 
> Please re-introduce carets in _devDependencies_ and _resolutions_ and commit.
> 
> Additionally, we must troubleshoot and resolve the CI test error for SonarCloud prior to accepting and merging this PR.

I follow the rabbit hole to a specific file but not sure how it gets affected by the JS packages. since it is on the python side. 

```
script/generate_under_maintenance.py:82
```

This is until now my minimal required reverts versions. so now we only have like 6 dependencies to look at. 

I am still testing.

kylelaker-ccpo (approved):
LGTM but will hold off on merging to confirm @jasonburkert-ccpo's concerns have been addressed

jasonburkert-ccpo (approved):

jasonburkert-ccpo:
Reviewing.  Don't merge this yet.

package.json
```diff
@@ -25,7 +25,7 @@
     "svg-innerhtml": "^1.1.0",
     "text-mask-addons": "^3.8.0",
     "uswds": "^1.6.13",
-    "v-tooltip": "^2.0.3",
+    "v-tooltip": "~2.0.3",
```
jasonburkert-ccpo, line 28: ^2.0.3 allows 2.x.y, in this case 2.1.2 from January 2021 - this is problematic for IE 11 support
~2.0.3 allows 2.0.x, in this case 2.0.3 from January 2020 - this works on IE 11

### [v-tooltip release history](https://yarnpkg.com/package/v-tooltip)
v-tooltip 2.1.2 is latest release January 2021
v-tooltip 2.1.0 was released late December 2020
v-tooltip 2.0.3 was released January 2020

jasonburkert-ccpo:
Restricting javascript library v-tooltip to 2.0.x fixes the IE 11 incompatibility.  Using 2.1.x causes support for IE 11 to break.

Reviewed on local dev.
- `git checkout AT-6030-IE-compatibility-bug`
- `script/update`
- `script/test`
- `script/server`

script/test passed.  IE 11 incompatibility fixed.
image

PR #2184, opened 2021-03-02
Revert "AT-5988: Removed "Baltimore" root CA from CA Chain "

kylelaker-ccpo:
Reverts dod-ccpo/atat#2131

This broke deployment and communication with the DB. Per the [Microsoft Docs](https://docs.microsoft.com/en-us/azure/postgresql/concepts-certificate-rotation#what-change-was-performed-on-february-15-2021-02152021) the Baltimore cert should not be removed yet.

Specifically:

>  During this change, the BaltimoreCyberTrustRoot root certificate was not replaced with DigiCertGlobalRootG2 and that change is deferred to allow more time for customers to make the change.

and

> Please do not drop or alter Baltimore certificate until the cert change is made. We will send a communication once the change is done, after which it is safe for them to drop the Baltimore certificate.

So this removal, while removing the correct cert may have been a bit premature.


kylelaker-ccpo:
Sorry for the branch name, that's how the GitHub UI handles revert naming. Should be short-lived anyway.

PR #2192, opened 2021-03-04
Dangerously delete less dangerously

kylelaker-ccpo:
This script used to just take a prefix and delete everything without
prompting or validating; however, it largely seems like it is invoked in
interactive contexts, so showing the user what is going to be deleted
and asking if it is correct is probably the best way to do this. That
way we can still dangerously delete but maybe with slightly less danger.

This another PR related to the work on AT-5711 but is not tightly
integrated. These changes were made to support testing.


jasonburkert-ccpo (approved):
Looks good.  Much better to ask permission in this case.

PR #2194, opened 2021-03-05
AT-5578 xss attack validate name for dev new user

jGonzalez-CCPO:
https://ccpo.atlassian.net/browse/AT-5578

### DIFF
--- a/atat/routes/dev.py
+++ b/atat/routes/dev.py
@@ -181,7 +181,7 @@ def get_or_create_dev_persona(persona):
     return user
 
 
-@local_access_bp.route("/dev-new-user")
+@dev_bp.route("/dev-new-user")
 def dev_new_user():
     first_name = request.args.get("first_name", None)
     last_name = request.args.get("last_name", None)
--- a/atat/routes/dev.py
+++ b/atat/routes/dev.py
@@ -192,29 +192,32 @@ def dev_new_user():
     last_name = request.args.get("last_name", None, type=str)
     dod_id = request.args.get("dod_id", None, type=str)
 
-    # This interphase convert dictionaries into Objects instance.
+    # 1: It need to check fist that have all the parameters
+    if None in [first_name, last_name, dod_id]:
+        raise IncompleteInfoError()
+
+    # This interphase convert dictionaries into Objects instance. It is Require for step 2.
     class Struct(object):
         def __init__(self, **entries):
             self.__dict__.update(entries)
 
-    # using standard validator to validate inputs or throw a error
+    # 2: Using standard validator to validate inputs or throw a error
     is_name({}, Struct(**{"data": first_name}))
     is_name({}, Struct(**{"data": last_name}))
     is_number({}, Struct(**{"data": dod_id}))
 
-    if None in [first_name, last_name, dod_id]:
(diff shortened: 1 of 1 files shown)


atat/routes/dev.py
```diff
@@ -183,23 +188,36 @@ def get_or_create_dev_persona(persona):
...
+    first_name = request.args.get("first_name", None, type=str)
+    last_name = request.args.get("last_name", None, type=str)
+    dod_id = request.args.get("dod_id", None, type=str)
 
+    # 1: It need to check fist that have all the parameters
     if None in [first_name, last_name, dod_id]:
         raise IncompleteInfoError()
 
+    # This interphase convert dictionaries into Objects instance. It is Require for step 2.
+    class Struct(object):
```
kylelaker-ccpo, line 200: Was there a specific reason to create a `Struct` class over using the standard library--provided `types.SimpleNamespace`? It works almost exactly the same way as what has been done here.

atat/routes/dev.py
```diff
@@ -183,23 +188,36 @@ def get_or_create_dev_persona(persona):
...
+        def __init__(self, **entries):
+            self.__dict__.update(entries)
+
+    # 2: Using standard validator to validate inputs or throw a error
+    is_name({}, Struct(**{"data": first_name}))
+    is_name({}, Struct(**{"data": last_name}))
+    is_number({}, Struct(**{"data": dod_id}))
```
kylelaker-ccpo, line 207: The `entries` parameter for `Struct.__init__` already has `**` so doing this on both ends makes the code harder to understand. The symbols are confusing so maybe consider:

```suggestion
    is_name({}, Struct(data=first_name))
    is_name({}, Struct(data=last_name))
    is_number({}, Struct(data=dod_id))
```

atat/routes/dev.py
```diff
@@ -183,23 +188,36 @@ def get_or_create_dev_persona(persona):
 
 @local_access_bp.route("/dev-new-user")
 def dev_new_user():
-    first_name = request.args.get("first_name", None)
-    last_name = request.args.get("last_name", None)
-    dod_id = request.args.get("dod_id", None)
+    first_name = request.args.get("first_name", None, type=str)
+    last_name = request.args.get("last_name", None, type=str)
+    dod_id = request.args.get("dod_id", None, type=str)
```
kylelaker-ccpo, line 193: Unless I am mistaken, URL query parameters are always going to be strings. Further, `None` is the default value for the `default` parameter. So the equivalent code is
```suggestion
    first_name = request.args.get("first_name")
    last_name = request.args.get("last_name")
    dod_id = request.args.get("dod_id")
```

kylelaker-ccpo (changes requested):
Nice! Validating these inputs is going to be really great for the local dev environment. I have a changes to request though. Nice comments to explain the validation process.

kylelaker-ccpo (approved):
Looks good to me. Since we did some pair programming on parts of this, I would like to get @jasonburkert-ccpo to review as well. Locally validated the same test cases used in the Jira ticket.

jasonburkert-ccpo (approved):
Good back and forth with @kylelaker-ccpo

PR #2202, opened 2021-03-15
AT-4926 Improve accuracy of validation message for input field Application Name

jGonzalez-CCPO:
Improve accuracy of validation message for input field `Application Name`

Previously the validation message only spoke to the expected length of the input.  However, the regex that validates the input field also checks content allowing only certain characters.  This change corrects the validation message to more accurately describe the characters allowed in the `Application Name` in addition to describing the accepted length.

Ticket: AT-4926


jasonburkert-ccpo (approved):
Looks good.  Tested in local development.

image

PR #2203, opened 2021-03-15
AT-6063 Define ApplicationEnvironment enum to standardize named environments

jGonzalez-CCPO:
Enum ApplicationEnvironment defines named environments.
Configure environment variable FLASK_ENV to one of the following:
1) development
2) production
3) test
4) ci

For example:
`export FLASK_ENV="development"`
Will allow login via route `/login-local`

Ticket: AT-6063


README.md
```diff
@@ -265,7 +265,7 @@ To generate coverage reports for the Javascript tests:
 All config settings must be declared in "config/base.ini", even if they are null. Configuration is set in the following order:
 
 1. Settings from "config/base.ini" are read in and applied.
-2. If FLASK_ENV is set as an environment variable and there is an INI file with a matching name, that INI file is read in and applied. For instance, if FLASK_ENV is set to "prod" and a "prod.ini" file exists, those values will be applied and will override any values set by the base.ini.
+2. If FLASK_ENV is set as an environment variable and there is an INI file with a matching name, that INI file is read in and applied. For instance, if FLASK_ENV is set to "production" and a "prod.ini" file exists, those values will be applied and will override any values set by the base.ini.
```
jasonburkert-ccpo, line 268: @jGonzalez-CCPO In the case of setting FLASK_ENV="production", wouldn't the file that we look for be called `production.ini` ?

terraform/providers/bootstrap/variables.tf
```diff
@@ -6,7 +6,7 @@ variable "operations_location" {
 
 variable "namespace" {
   type = string
-  # default     = "dev"
+  # default     = "development"
```
jasonburkert-ccpo, line 9: This appears to be a comment.  Is it accurate after changing?  Should it be removed instead?

README.md
jasonburkert-ccpo, line 268: In the case of FLASK_ENV="production", wouldn't the the expected file be `production.ini` ?

atat/app.py
```diff
@@ -118,7 +118,7 @@ def make_flask_callbacks(app):
     @app.before_request
     def _set_globals():
         g.current_user = None
-        g.dev = os.getenv("FLASK_ENV", "dev") == "dev"
+        g.dev = os.getenv("FLASK_ENV", "development") == "development"
```
jasonburkert-ccpo, line 121: Should this be conditional be `!= "production"` ?

jasonburkert-ccpo (changes requested):
Please review my comments.
Is this the intended functionality?

This looks good.
image

This looks good.
image

Should this be more explicit about defaulting to "development"?  Is that, in fact, what has happened?

If FLASK_ENV is not production or development, I suggest some output, perhaps WARNING-level output, that says value 'foobar' was found for FLASK_ENV and you are defaulting to "development".  Then output "Environment: foobar" (in screenshot below), should instead say "Environment: development".

image

terraform/providers/bootstrap/variables.tf
jGonzalez-CCPO: It is a piece of documentation for reference on the terrace template.

README.md
jGonzalez-CCPO: I am not sure that functionality even exists in the current code. I have not used, or see any indication of a prod.ini file been called, or used. I can be wrong but looks to me that this one is out of the scope of the ticket.

atat/app.py
```diff
@@ -39,11 +39,17 @@
...
 from atat.utils.session_limiter import SessionLimiter
 
-ENV = os.getenv("FLASK_ENV", "dev")
+ENV = os.getenv("FLASK_ENV", "production")
+
+# Env is development by default.
+if not (ENV in ["development", "production", "test", "ci"]):
+    ENV = "production"
+
+print(">>> Environment name set: %s", (str(ENV)))
```
kylelaker-ccpo, line 48: This particular line has a few concerning things:
1. It prints directly to standard output without using the logging framework (as all our other messages do)
2. It will run _every time_ this file gets `import`ed which isn't ideal
3. As written, the `%` gets printed every time.

Can this move in to `make_app` _after_ `app` has been initalized? (Currently, that'd be Line 62)

So then the line would be:

```python
app.logger.info("Environment name set: %s", ENV)
```

`print` uses different semantics than the functions in the `logging` module and does not perform `%`-style interpolation. Additionally, all our output to the console should happen via the app logger.

atat/app.py
```diff
@@ -39,11 +39,17 @@
 from atat.utils.notification_sender import NotificationSender
 from atat.utils.session_limiter import SessionLimiter
 
-ENV = os.getenv("FLASK_ENV", "dev")
+ENV = os.getenv("FLASK_ENV", "production")
+
+# Env is development by default.
+if not (ENV in ["development", "production", "test", "ci"]):
```
kylelaker-ccpo, line 45: I would really prefer if our allowed values were defined as an `enum.Enum` rather than "magic strings".

```python
class ApplicationEnvironment(enum.Enum):
    DEVELOPMENT = "development"
    TEST = "test"
    CI = "ci"
    PRODUCTION = "production"

ENV = ApplicationEnvironment(os.getenv("FLASK_ENV", "development"))
```

And then everywhere, we can compare to `ApplicationEnvironment.DEVELOPMENT` or `ApplicationEnvironment.PRODUCTION` or something like that.

We'd even be able to define methods or attributes as well which may be nice. But in general, if we want to enumerate allowed values, we should use an enumeration :)

atat/app.py
jGonzalez-CCPO: @jasonburkert-ccpo agree to remove that log

jGonzalez-CCPO (commented):
changed

atat/app.py
jGonzalez-CCPO: @kylelaker-ccpo SO, if I get this well do you want this to fail instead of having a default ENV?

jGonzalez-CCPO: the criteria until now are to default to production

jGonzalez-CCPO: suggested changes 104 failed, 703 passed, 35 skipped, 1 warning in 51.70s

jeffsegal-ccpo: Hey guys - yes please always fail fast on things like this. Definitely never ever want to default to dev because that becomes a security risk.

kylelaker-ccpo: I hold the following fairly strong opinions related to the questions raised here:

1. It is important to always fail as early and descriptively as possible. We don't necessarily know what breaks or bad (and insecure) assumptions could be made in an "invalid" environment. We should fail quickly upfront. We should let users (users means "operators" and "developers" here) know _why_ there was a failure, even if it's implicit log messages
2. Enumerations provide additional safety, easier comparisons, and greater flexibility. You also get a usable error if you try to use an invalid value.

In addition to updating the initialization of `ENV`, it is important to replace all other uses of any of the literal strings with the `Enum` name. My suggestion here cannot be applied in isolation; it is the beginnings of a larger suggestion to "use an enum." If that will require further discussion and conversation, reach out and perhaps we can do a pair programming session or quick chat on this!

jGonzalez-CCPO: this is no longer part of the implementation some changes were need to comply with the coverages.

jGonzalez-CCPO (commented):
In order to comply with the coverage requirement for testing, I finish rewriting the atat/app.py implementation of the environment name and the uses.

jGonzalez-CCPO (commented):
this no longer apply

jGonzalez-CCPO:
List vs enum: 


Screen Shot 2021-03-19 at 14 36 32

jGonzalez-CCPO:
as you see the list performed better and allows for cover the criteria of a default "production" environment.

jeffsegal-ccpo (approved):
Looks good to me, thanks Jesus.

kylelaker-ccpo (approved):
The next time we get a chance to touch some of this code, we should see whether it'd be possible to avoid some of this roundtripping to/from `.value` with `configparser`. Looks good though. Thanks for the work, Jesus!

jasonburkert-ccpo (approved):

jasonburkert-ccpo:
@jGonzalez-CCPO Please rebase with 'develop'.  I attempted but there is a conflict to be resolved.

PR #2204, opened 2021-03-15
AT-6081 Remove file `templates/components/date_picker.html` and any references

jGonzalez-CCPO:
Remove file `templates/components/date_picker.html` and any references

This file came under scrutiny due to an included hidden field.  However, it was found this file is no longer in use.  Removing the file and any references.

Ticket: AT-6081


jGonzalez-CCPO:
https://github.com/dod-ccpo/atat/pull/2204

jasonburkert-ccpo (approved):
Few changes are as productive as removing 97 lines of code that we never have to maintain again -- and the functionality is unchanged.  Nice job.

PR #2211, opened 2021-03-24
Feature/at 5398

DGabriel-CCPO:
The query causing the error already has error handling, which is why the 500 error is raised. The actual error that we receive is a 401 error because we are unauthorized to query that URL, and the server sends us back a 500 error. To my knowledge, HTTP errors cant be flashed in which is what the ticket is calling for.

If we don’t want the 500 error to occur, we have to fix the query; otherwise, the application functions how it is intended.

the code change that I will push will involve commenting out the query to see if the reports page will show without it. I could not test this in my local environment because I need the CSP to provision a portfolio, application, and task order to test.

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -1605,40 +1605,40 @@ class AzureCloudProvider(CloudProviderInterface):
         raw_creds = self.get_secret(hashed)
         return KeyVaultCredentials(**json.loads(raw_creds))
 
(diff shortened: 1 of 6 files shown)


jasonburkert-ccpo:
Looked briefly at the changes.  Uncertain about the intent.  I believe ticket AT-5398 is confusing and could be clarified.  Leaning toward closing this PR and refocusing in a new feature branch.

jasonburkert-ccpo:
After team discussion we've decided to clarify ticket and go a slightly different direction.

PR #2214, opened 2021-03-25
Remove dev dependency 'cryptography'

jasonburkert-ccpo:
This explicit Python dependency is unnecessary.  Other libraries we use may have a transitive dependency on versions of the library, but ATAT doesn't require it directly.


jasonburkert-ccpo:
I've tested this and I think it's ready for review.

jeffsegal-ccpo (approved):

PR #2219, opened 2021-03-29
AT-5239 Create enum for Microsoft Azure API versions

anhtran-ccpo:
Application is communicating with several Azure APIs.  This will standardize invocations, identify which versions are used where, and facilitate future changes.

Ticket: AT-5239


jasonburkert-ccpo (approved):
Thumbs up.

PR #2221, opened 2021-03-29
AT-5398 Temporarily comment call to function get_reporting_data to prevent HTTP error on '/reports' route

DGabriel-CCPO:
An Azure API call was failing while servicing requests to `/reports` route only if a portfolio had already been provisioned.  Thus far we've only had provisioning working on MCA environments and Billing APIs are altogether different under EA environments.

Commenting out the call to function `get_reporting_data` which invokes the specific Azure API will allow the other reporting features to be demonstrated while giving us time to get things working on EA environments.

Ticket: AT-5398

### DIFF
--- a/atat/domain/reports.py
+++ b/atat/domain/reports.py
@@ -1,10 +1,10 @@
-# import pendulum
-# from flask import current_app
+import pendulum
+from flask import current_app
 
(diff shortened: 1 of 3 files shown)


jasonburkert-ccpo (approved):

image

PR #2228, opened 2021-03-30
Bulk dependency update

kylelaker-ccpo:
Bulk update of JavaScript and Python dependencies for the _application_. The `ops/` directory and `ansible/` directory are not modified. Tests passed locally. Validated `login-local` works in local environment as well as `script/test`. There seems to be an incompatibility with SQLAlchemy 1.4 so that is now pinned to 1.3.x.


jasonburkert-ccpo:
BTW, the CI failure mentioned above is to be addressed in AT-6079.

jasonburkert-ccpo (approved):
Peer reviewed.  Tested in local development.

PR #2229, opened 2021-03-31
Delete ops/requirements.txt

kylelaker-ccpo:
This folder is no longer necessary with the upcoming
`atat-infrastructure` work but we don't want to remove the entire folder
until the ticket can be implemented thoughtfully and in earnest. Since
the `requirements.txt` already has conflicting dependencies and a lack
of documentation, deleting it to avoid dependabot PRs that will cause
further breakage and noise may be the best option.

This is a compromise solution until we can remove the `ops/` folder
completely.


kylelaker-ccpo:
Not ready for merge yet.

```
ERROR: Cannot install ansible[azure]==2.9.2 and azure-cli-core~=2.21.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested azure-cli-core~=2.21.0
    ansible[azure] 2.9.2 depends on azure-cli-core==2.0.35

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```

kylelaker-ccpo:
Above comment is resolved by the suggestion to just delete the `requirements.txt` file.

jasonburkert-ccpo (approved):
Verified.  No more `ops/requirements.txt`.  Bye.

PR #2230, opened 2021-03-31
Reduce GitHub Action pipeline time

kylelaker-ccpo:
This relies more on the GitHub Runner rather than running in a container. This allows us to better leverage the tools cache and pre-installed packages and software. We also remove some duplicate runs of the tests by just emitting coverage reports by default. Additionally, we cache our dependencies which will help improve our run time because we've previously wasted quite a lot of time there. For Python, we only cache within a particular Python version.

Workflow run times seem to be hovering around 6m 30s to 8m.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,6 +9,7 @@ omit =
   atat/models/audit_event.py
   atat/domain/csp/cloud/hybrid_cloud_provider.py
 branch = True
+relative_files = True
 
 [report]
 exclude_lines =
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -8,12 +8,12 @@ on:
       - "develop"
       - "release/**"
 
-
 jobs:
   sonarcloudScan:
     name: SonarCloud Scan
     runs-on: ubuntu-20.04
-    container: "ubuntu:focal"
+    env:
+      PYTHON_VERSION: "3.8.8"
     services:
       redis:
         image: redis:6
@@ -22,6 +22,8 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
+        ports:
+          - 6379:6379
       postgres:
         image: postgres:10
         env:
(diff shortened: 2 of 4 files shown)


kylelaker-ccpo:
Linked to the README for the Ubuntu 20.04 runner: https://github.com/actions/virtual-environments/blob/ubuntu20/20210318.0/images/linux/Ubuntu2004-README.md

kylelaker-ccpo:
@nkingsley-ccpo and @jasonburkert-ccpo I think I am content calling this sufficiently optimized at this point. Switching to using the `sonarsource` action will be helpful in ensuring we're always using the latest version available of the tool as well and means that we don't have to unzip or use our `script/sonar_scanner` script anymore.

Using relative paths in the pytest coverage reports will be good for making the coverage tests work better regardless of where they're being run, especially since the changes add a `coverage.xml` file every time we run the tests. This makes it a bit easier for everyone to run the sonar scanner themselves locally if they wanted to.

We've made it easier to update the version of Python we'll use going forward in the pipeline and we won't reuse a cached virtual environment if we update the version.

We're now seen a few sub 6 minute runs and largely top out around the mid-7m mark. Of course, that will increase a bit if we can't reuse the cache but it is still pretty good. Somewhere close enough to half of the time that we were seeing before.

This does introduce 2 changes to folks who run the scripts locally:
 - The `node_modules/` folder is not deleted during `script/setup`. We don't do the equivalent of this for the Python `.venv` folder locally so this doesn't feel like a significant issue. In general, the team uses `script/update` anyway and they can always delete the folder manually if needed
 - A `coverage.xml` is always emitted running `script/test`. This file is already ignored by `.gitignore`, doesn't impact how the tests run, and might provide useful information anyway (since we usually blow away the Python coverage info in the CLI output with a flurry of `jest` deprecation warnings)

This gives us a good foundation for adding future jobs or workflows with how we can have a "minimal" direct route from checkout to tests of the application codebase. Other jobs could also theoretically reuse the same cache which will be really nice. We know that UI tests and deployments are potential upcoming feature requests of the GHA pipeline so hopefully this is helpful there.

At this point, I don't think there are any further significant improvements to make where the juice is worth the squeeze. I have looked over this and I don't think there's anything that would cause false failures or successes in the pipeline. I am also not aware of any tools that consume our `coverage.xml` beyond SonarCloud, so nothing should be impacted by it switching to using relative paths.

jasonburkert-ccpo (approved):
@kylelaker-ccpo Walked me through the changes.  Can see time improvements on test runs on the feature branch.

PR #2231, opened 2021-03-31
AT-5947 fix portfolio funding text

megansilvester-ccpo:
Made change to translations.yaml.  There is only one space between the end of the subject sentence and the beginning of the next.


jeffsegal-ccpo (approved):
*First PR woot woot*

PR #2236, opened 2021-04-05
AT-6128 Links under Application Environments are fixed for all browsers

jGonzalez-CCPO:
On IE 11, the location of Members and Edit links are different from the other browsers.

 

This behavior was seen running locally using code as of this commit:https://github.com/dod-ccpo/atat/commit/b708a0c3df9daf9ba058030cd9dea4f8cfb38cbb

https://ccpo.atlassian.net/browse/AT-6128


jasonburkert-ccpo (approved):

image

PR #2249, opened 2021-04-14
Update "Sonar Scans" GHA name

kylelaker-ccpo:
I don't want to bikeshed over the naming here too much but I think that
we're running into confusion over what exactly the goal of the pipeline
does. We do quite a bit more than just a static analysis of the code.
Over time, of course we want to make this better. But today, we can at
least be clearer about exactly what's happening. Perhaps that will
reduce some of the complexity to understanding failures and where tests
should go.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,5 @@
 ---
-name: SonarCloud Scan Jobs
+name: Full CI Build
 on:
   pull_request:
   push:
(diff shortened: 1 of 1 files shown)


jeffsegal-ccpo (approved):
Thanks! Seems silly but hopefully this will actually keep one or two people from being confused when they look at it in GHA. :)

jasonburkert-ccpo (approved):
Love the increased clarity.

kylelaker-ccpo:
@jeffsegal-ccpo -- merging this will likely require a change to branch protection rules since we're changing the name of one of the workflows. Otherwise, "SonarCloud Scan" will continue to appear with "Expected" and the new CI job may not be sufficient to permit merges.

PR #2255, opened 2021-04-15
Bump SQLAlchemy version to 1.4.8

kylelaker-ccpo:
This is based off #2254 with a fix for the tests. This does not fix the warnings introduced in the new version.


tests/models/test_user.py
```diff
@@ -14,10 +14,18 @@ def test_profile_complete_with_all_info():
...
-    assert not user.profile_complete
+def test_profile_complete_with_missing_info(missing_field, session):
+    # Missing the dod_id field is still cause for the profile to not be
+    # completed and is therefore in REQUIRED_FIELDS; however, the DB
+    # session will autoflush by default and try to write that field back
+    # to the database and for dod_id and others, that could result in an
+    # IntegrityError. Ensuring that we don't autoflush during this test
+    # will allow us to test the `profile_complete` logic without breaking
+    # database constraints.
+    with session.no_autoflush:
```
jasonburkert-ccpo, line 25: This is the new line.  SQLAlchemy changed its flushing behavior in v1.4.  So we're changing this test to adapt.

jasonburkert-ccpo (approved):
Peer reviewed.  Ran `script/test` locally and performed UI smoke test.  Looks good.

PR #2256, opened 2021-04-16
Refactor smoke tests to improve speed and validation points

ehoward-ccpo:
The issue with IE11 seems to have been resolved.  The initial smoke tests had time.sleep written in the code which is now removed.  The page assertions were scanning the page for text and have been replaced with validation that points to specific areas/elements on the page to ensure not only the text is displayed but displaying in the right location. 
Ticket: AT-6159

### DIFF
--- a/uitests/framework/page_objects/new_portfolio_page.py
+++ b/uitests/framework/page_objects/new_portfolio_page.py
@@ -82,7 +82,10 @@ class AddNewPortfolioPages:
         self.driver.find_element_by_css_selector(self.select_checkbox_css).click()
 
     def click_save_portfolio_btn(self):
-        self.driver.find_element_by_css_selector(self.save_portfolio_btn_css).click()
+        # self.driver.find_element_by_css_selector(self.save_portfolio_btn_css).click()
+        WebDriverWait(self.driver, 30).until(
+            EC.presence_of_element_located(
+                (By.CSS_SELECTOR, self.save_portfolio_btn_css))).click()
         # saveBtn = self.driver.find_element_by_css_selector(self.save_portfolio_btn_css)
         # saveBtn_text = saveBtn.text
         # if saveBtn_text == 'Save Portfolio':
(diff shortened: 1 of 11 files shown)


uitests/framework/page_objects/new_portfolio_page.py
```diff
@@ -39,7 +39,7 @@ class AddNewPortfolioPages:
     btn_confirm_revoke = "button.action-group__action.usa-button.usa-button-primary"
     btn_resend_invite = "tr:nth-child(2) > td.toggle-menu__container > div.toggle-menu > div > a:nth-child(2)"
     btn_resend_invite_b = "td.toggle-menu__container > div > div > a:nth-child(2)"
-    btn_resend_invite_confirm = "//html/body/div/div[3]/div[2]/div/div/div[2]/div[2]/div[3]/div/div/div/div/form/div[2]/input"
+    btn_resend_invite_confirm = "input[value='Resend Invite']"
```
jasonburkert-ccpo, line 42: That's much better.

uitests/framework/test_cases/test_login.py
```diff
@@ -10,31 +10,27 @@
 
 
 @pytest.mark.smoke
-class TestLogin:
+class Test_Login:
```
jasonburkert-ccpo, line 13: According to [PEP 8 -- Style Guide for Python Code](https://www.python.org/dev/peps/pep-0008/#class-names) the previous CapWords style is preferred for Class Names.

uitests/framework/page_objects/new_portfolio_page.py
jasonburkert-ccpo: Just thinking out loud...
I wonder if it would make sense for UI text such as _Resend Invite_ above to refer directly to the source of truth -- file `translations.yaml`.  In this case _Resend Invite_ is on line 470 of that file.

Tests could load the yaml, and refer to that node the same way that the application does in file `members.html` on line 145 as node `portfolios.applications.members.menu.resend` which has the value _Resend Invite_.  This could apply to any text supplied by that yaml, and then the tests would remain in sync with UI with respect to such strings.

jasonburkert-ccpo (approved):
Looks good.  
I'd recommend following the PEP 8 naming conventions in future updates.

PR #2259, opened 2021-04-19
AT-6121 Allow only PDF files for Task Order upload

jGonzalez-CCPO:
System requests and accepts only PDF files for Task Order upload.

Ticket: AT-6121

### DIFF
--- a/atat/forms/validators.py
+++ b/atat/forms/validators.py
@@ -98,9 +98,7 @@ def file_length(max_length=50000000, message=None):
     return _file_length
 
 
-def alpha_numeric_name_pdf(
-    message=translate("forms.task_order.filename_error"),
-):
+def alpha_numeric_name_pdf(message=translate("forms.task_order.filename_error"),):
     return Regexp(
         regex=r"^([A-Za-z0-9\-_ \.\/\\]*(\.pdf))?$",
         message=message,
--- a/atat/forms/validators.py
+++ b/atat/forms/validators.py
@@ -98,12 +98,12 @@ def file_length(max_length=50000000, message=None):
     return _file_length
 
 
-def alpha_file_pdf(
+def alpha_numeric_name_pdf(
     message=translate("forms.validators.alpha_numeric_message")
     + " and File name must end on '.pdf'.",
 ):
     return Regexp(
-        regex=r"^[A-Za-z0-9\-_ \.\/\\]*.pdf$", message=message, flags=re.IGNORECASE
+        regex=r"^[A-Za-z0-9\-_ \.\/\\]*\.pdf$", message=message, flags=re.IGNORECASE
     )
 
 
--- a/atat/forms/validators.py
+++ b/atat/forms/validators.py
@@ -98,8 +98,13 @@ def file_length(max_length=50000000, message=None):
     return _file_length
 
 
-def alpha_file_pdf(message="File name must end on '.pdf'. and content only valid characters."):
-    return Regexp(regex=r"^[A-Za-z0-9\-_ \.\/\\]*.pdf$", message=message, flags=re.IGNORECASE)
+def alpha_file_pdf(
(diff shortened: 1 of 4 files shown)


atat/forms/validators.py
```diff
@@ -98,5 +98,13 @@ def _file_length(_form, field):
     return _file_length
 
 
+def alpha_numeric_name_pdf(message=translate("forms.task_order.filename_error"),):
+    return Regexp(
+        regex=r"^([A-Za-z0-9\-_ \.\/\\]*(\.pdf))?$",
```
kylelaker-ccpo, line 103: I think there are a few improvements that can be made to this regex. As is, the empty string (`""`) would be accepted. So would `.pdf` without anything before the extension. I'd recommend more like:

```
^[A-Za-z0-9\-_ \.\/\\]+\.pdf$
```

This enforces that prior to the `.` we have at least one of the allowed characters. The empty string is not accepted. There aren't any capture groups to add complexity either.

See https://regex101.com/r/D4UfKa/1 and especially the "Unit Tests" on the sidebar on the left.

kylelaker-ccpo (changes requested):
Nice addition! Small request to get the wording more in-sync between the two files and a question on the regex

atat/forms/validators.py
kylelaker-ccpo: Saw the Jira comment. If we need to accept the empty string, we can do:
```
^([A-Za-z0-9\-_ \.\/\\]+\.pdf)?$
```

jGonzalez-CCPO: Read my comment on the ticket, due to various tests that depends on this to be an empty string i was forced to modify it to accept that alternative.

jasonburkert-ccpo (approved):
Peer reviewed.  Tested in local development.  Looks good.
Upload dialog behaves as expected on supported browsers, and files with file extensions of mixed case are accepted and can be uploaded.

image

js/lib/input_validations.js
```diff
@@ -139,9 +139,9 @@ export default {
   },
   restrictedFileName: {
     mask: false,
-    match: /^[A-Za-z0-9\-_ \.]+$/,
+    match: /^[A-Za-z0-9\-_ \.\/\\]*\.pdf$/i,
     unmask: [],
     validationError:
-      'File names can only contain the characters A-Z, 0-9, space, hyphen, underscore, and period.',
+      'File names can only contain the characters A-Z, 0-9, space, hyphen, underscore, and period. And end in .pdf',
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Confirmed this was addressed and suggested change adopted.

translations.yaml
```diff
@@ -316,7 +316,7 @@ forms:
   task_order:
     upload_error: There was an error uploading your file. Please try again. If you encounter repeated problems uploading this file, please contact CCPO.
     size_error: "The file you have selected is too large. Please choose a file no larger than {file_size_limit}MB."
-    filename_error: File names can only contain the characters A-Z, 0-9, space, hyphen, underscore, and period.
+    filename_error: File names must end on 'pdf' and can only contain the characters A-Z, 0-9, space, hyphen, underscore, and period.
```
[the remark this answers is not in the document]
jasonburkert-ccpo: Confirmed this was addressed and suggested change adopted.

jasonburkert-ccpo (approved):
Good.  This change will require a base file name.  Simply `.pdf` will no longer be accepted.

image

kylelaker-ccpo (approved):
Requests have been addressed. Thanks!

jGonzalez-CCPO:
@kylelaker-ccpo @jasonburkert-ccpo i just re-test, format and rebase all the requested changes :)

jasonburkert-ccpo:
Tested locally again. Looks good.

PR #2261, opened 2021-04-20
AT-5596 Remove Secondary "Log in" button from sign-in page

jGonzalez-CCPO:
Remove "login" button from the top right-hand corner of the home page.

Ticket: AT-5596


jasonburkert-ccpo (approved):

image

PR #2263, opened 2021-04-21
AT-5546 Removed /test-email route

anhtran-ccpo:


kylelaker-ccpo (changes requested):
I think for completeness we may also want to remove this from `_NO_ACCESS_CHECK_REQUIRED` in `tests/test_access.py`.

anhtran-ccpo:
> I think for completeness we may also want to remove this from `_NO_ACCESS_CHECK_REQUIRED` in `tests/test_access.py`.

Good catch. Just removed

kylelaker-ccpo (approved):

PR #2265, opened 2021-04-21
Use `date-fns` for nearly all date calculations

kylelaker-ccpo:
Using a switch based on the month is unnecessary when February is the
only month that changes. This code is largely intended, it seems, to
properly handle date calculations when `this.year` is `null` (or
falsey); however, `new Date(null, this.month)` would yield a month in
the year 1900. This seems fine for calulating non-February. For February
it does cause an issue but continuing to special case that and returning
29 keeps with the current spirit of the code (and keeps the existing
test cases passing).

Related to #2253

### DIFF
--- a/js/components/date_selector.js
+++ b/js/components/date_selector.js
@@ -134,24 +134,34 @@ export default {
     },
 
     daysMaxCalculation: function () {
-      switch (parseInt(this.month)) {
-        case 2: // February
-          if (this.year) {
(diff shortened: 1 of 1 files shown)


jasonburkert-ccpo (approved):
The code-level comments are a big improvement as to purpose.  The UI has dates being entered in three parts, and this supports reasonable behavior for these elements.

image
image

PR #2273, opened 2021-04-23
AT-6035 Improve the customer flow through ATAT

DGabriel-CCPO:
Description
As a customer, I have a disjointed experience in the overall process required to get access to an environment within the underlying CSP.  The most natural flow for the application would provide me with an overview of the process, walk me through each step of the process, and then point me to the underlying CSP once all of my environments had been created. 

Acceptance Criteria
1) Improve the authenticated "landing page" to read from left to right, top to bottom
2) Ensure the customer flow from Portfolio creation through Application creation is not disjointed
3) Provide some method for me to go to the underlying CSP once my Application has been created

https://ccpo.atlassian.net/browse/AT-6035

### DIFF
--- a/styles/elements/_buttons.scss
+++ b/styles/elements/_buttons.scss
@@ -31,7 +31,4 @@ input[type="button"].button-danger-outline {
     box-shadow: inset 0 0 0 2px $color-red-darkest;
   }
 }
-.next_button {
-  display: grid;
-  justify-content: end;
-}
+
--- a/styles/elements/_buttons.scss
+++ b/styles/elements/_buttons.scss
@@ -31,3 +31,7 @@ input[type="button"].button-danger-outline {
     box-shadow: inset 0 0 0 2px $color-red-darkest;
   }
 }
+.next_button {
+  display: grid;
+  justify-content: end;
+}
--- a/styles/elements/_buttons.scss
+++ b/styles/elements/_buttons.scss
@@ -31,4 +31,3 @@ input[type="button"].button-danger-outline {
     box-shadow: inset 0 0 0 2px $color-red-darkest;
   }
 }
-
--- a/templates/applications/fragments/environments.html
+++ b/templates/applications/fragments/environments.html
@@ -38,11 +38,11 @@
                         {% if not env["pending"] -%}
                           <span>
                             <a
-                              href='https://portal.azure.com/'
+                              href='{{ url_for("applications.access_environment", environment_id=env.id)}}'
                               target='_blank'
                               rel='noopener noreferrer'>
                               {{ env['name'] }} {{ Icon('link', classes='icon--medium icon--primary') }}
-                              {{applications}}
+                              {{ applications }}
                             </a>
                           </span>
                         {% else -%}
--- a/templates/applications/fragments/environments.html
+++ b/templates/applications/fragments/environments.html
@@ -42,7 +42,6 @@
                               target='_blank'
                               rel='noopener noreferrer'>
                               {{ env['name'] }} {{ Icon('link', classes='icon--medium icon--primary') }}
-                              {{ applications }}
                             </a>
                           </span>
                         {% else -%}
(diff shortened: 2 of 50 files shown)


styles/elements/_buttons.scss
```diff
@@ -31,3 +31,7 @@ input[type="button"].button-danger-outline {
     box-shadow: inset 0 0 0 2px $color-red-darkest;
   }
 }
+.next_button {
```
jeffsegal-ccpo, line 34: My only question is whether there's an over-arching set of styles for buttons that we want to hook into vs just hard-coding one specifically for this button. Not saying not to do it but I want to make sure that we're applying a well organized approach here.

DGabriel-CCPO: I could have done it like this ` <div class="action-group-footer action-group-footer--expand-offset">
    <div class="action-group-footer--container">
      <a href="{{ url_for("applications.view_new_application_step_1", portfolio_id=portfolio.id) }}" class="usa-button usa-button-primary">
        Next: {{ "portfolios.applications.create_button"|translate }}
      </a>
    </div>
  </div>` which would have used the styles we already have and the appearance would be 
Screen Shot 2021-04-27 at 8 46 15 AM . i can do it this way i didn't initially because i didn't want to add a footer just to add a button when i could just add a class to a button. I do see your point and maybe i should convert it to how i did it here.

jeffsegal-ccpo (approved):

templates/applications/fragments/environments.html
```diff
@@ -38,10 +38,11 @@ <h3>{{ "portfolios.applications.settings.environments" | translate }}</h3>
                         {% if not env["pending"] -%}
                           <span>
                             <a
-                              href='{{ url_for("applications.access_environment", environment_id=env.id)}}'
+                              href='https://portal.azure.com/'
                               target='_blank'
                               rel='noopener noreferrer'>
                               {{ env['name'] }} {{ Icon('link', classes='icon--medium icon--primary') }}
+                              {{applications}}
```
kylelaker-ccpo, line 45:
```suggestion
                              {{ applications }}
```

Small style thing

templates/applications/fragments/environments.html
```diff
@@ -38,10 +38,11 @@ <h3>{{ "portfolios.applications.settings.environments" | translate }}</h3>
                         {% if not env["pending"] -%}
                           <span>
                             <a
-                              href='{{ url_for("applications.access_environment", environment_id=env.id)}}'
```
kylelaker-ccpo, line 41: So the code for `applications.access_environment` always redirects to `https://portal.azure.com` as-is. It's an extra redirect, sure, but it removes this hard-coded reference from the front-end and HTML. The existing code also has an access check to make sure the user has permission to access the environment.

The same `url_for` is used in `templates/applications/index.html` and should probably be kept consistent, correct?

Without both of these uses `atat/routes/applications/__init__.py::access_environments` is dead code and would need to be removed.

But I think we should _really_ try to actually use that code if possible. Because it will allow us to do deep linking in the future as well as access control. And it makes it easier to be cloud-agnostic.

kylelaker-ccpo: This might be possible by also checking for the proper permission sets in `atat/routes/applications/__init__.py::wrap_environment_role_lookup` in addition to just raw environment access but I am not 100% sure about that.

kylelaker-ccpo (commented):
There are a few thoughts that I have on the link to the Portal. This introduces some inconsistencies in how we do this linking and may create some dead code.

templates/applications/fragments/environments.html
jGonzalez-CCPO: It look like the original issue was relate to a bad setting on the watcher of that route: atat/routes/applications/__init__.py::access_environments 

I guess we not need a watcher for outbound link

jGonzalez-CCPO: we make it work you just need to assign the current user to the environment and then validate your user. https://localhost:8000/applications/invitations/<token>


and then when you click on it it work as it should be.

jGonzalez-CCPO (approved):

kylelaker-ccpo:
Thanks for your work on this (and to @jGonzalez-CCPO for his support as well). This looks great! Validated the UI changes are present as described in the ticket.

PR #2281, opened 2021-04-27
Feature/at 6194 UI tests fix random failures

palaparthi-ccpo:
In UI Test Cases : Under Smoke test suite the following two test cases are randomly failing when the test cases are been executed as a suite because both are using the str(time_run)so removed from the code .
1.test_reports_basic
2.test_create_to
Ticket: AT-6194

### DIFF
--- a/uitests/framework/test_cases/test_create_to.py
+++ b/uitests/framework/test_cases/test_create_to.py
@@ -4,9 +4,15 @@ import random
 import string
 
 from selenium.common.exceptions import TimeoutException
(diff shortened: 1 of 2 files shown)


kylelaker-ccpo (approved):
Changes look good and validated the changes locally running against the BrowserStack Automate API. Thanks for the fixes!

kylelaker-ccpo:
Thanks for your work on fixing these @prasanthialaparthi-ccpo and for getting the branch updated with develop so we can merge. Looks good!

PR #2283, opened 2021-04-28
AT-6164 Change the Portfolio Primary PoC 

jGonzalez-CCPO:
ATAT needs to have the capability for Mission Owners to change the primary point-of-contact. A CCPO Administrator should also have the capability to change the Mission Owner PPoC.  There must always be one and only one primary point-of-contact.
 

 
Current Portfolio Permissions:

Acceptance Criteria
The Primary PoC, or CCPO admin, can change the PoC to any individual who is already a portfolio manager for the portfolio
The UI should update and show the change
The Primary PoC can only be assigned to a single user

https://ccpo.atlassian.net/browse/AT-6164


templates/portfolios/fragments/portfolio_members.html
```diff
@@ -127,23 +149,28 @@ <h3>Portfolio Managers</h3>
...
-                    <a
-                      {% if not member.ppoc %}v-on:click="openModal('{{ perms_modal }}')"{% endif %}
-                      class="{% if member.ppoc %}disabled{% endif %}">
-                      Edit Permissions
-                    </a>
+                      <a
+                        {% if not member.ppoc %} v-on:click="openModal('{{ perms_modal }}')"{% endif %}
+                        class="{% if member.ppoc %}disabled{% endif %}">
+                        Edit Permissions
+                      </a>
```
kylelaker-ccpo, line 159: I think this indentation was more correct and inline with the surrounding style previously; however, I am open to being told I am mistaken

atat/routes/portfolios/admin.py
```diff
@@ -113,26 +114,35 @@ def admin(portfolio_id):
...
-#             "portfolios.admin",
-#             portfolio_id=portfolio.id,
-#             fragment="primary-point-of-contact",
-#             _anchor="primary-point-of-contact",
-#         )
-#     )
+@portfolios_bp.route("/portfolios/<portfolio_id>/update_ppoc", methods=["POST"])
+@user_can(Permissions.EDIT_PORTFOLIO_POC, message="update portfolio ppoc")
+def update_ppoc(portfolio_id):  # pragma: no cover
+    role_id = http_request.args.get("role_id")
```
kylelaker-ccpo, line 120: I'd really prefer if this came via the form rather than via a query string. I don't love that we're already doing a POST to send the result and the CSRF token but that this is a query string parameter. All that data, ideally, would be encapsulated in the form or POST request body instead. It looks like that's what the commented out code originally did. Is there a technical reason why this isn't possible?

kylelaker-ccpo (changes requested):
Overall this looks good! I have validated that it's possible to update the PPOC in the expected conditions and that the CSRF token properly ensures that a user cannot be "tricked" into updating a PPOC. There are a few things that ideally would be cleaned up. The `pragma: nocover` means that our coverage tool won't properly report the code coverage for that function. Additionally, we should prefer using the POST request body when we have one rather than using query string parameters.

Otherwise, it's largely style fixes and cleanup. 

Thanks for your work on this!

atat/routes/portfolios/admin.py
```diff
@@ -113,26 +114,35 @@ def admin(portfolio_id):
...
-#         url_for(
-#             "portfolios.admin",
-#             portfolio_id=portfolio.id,
-#             fragment="primary-point-of-contact",
-#             _anchor="primary-point-of-contact",
-#         )
-#     )
+@portfolios_bp.route("/portfolios/<portfolio_id>/update_ppoc", methods=["POST"])
+@user_can(Permissions.EDIT_PORTFOLIO_POC, message="update portfolio ppoc")
+def update_ppoc(portfolio_id):  # pragma: no cover
```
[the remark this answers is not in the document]
jGonzalez-CCPO: let see

atat/routes/portfolios/admin.py
```diff
@@ -113,26 +114,35 @@ def admin(portfolio_id):
...
+                "portfolios.admin",
+                portfolio_id=portfolio.id,
+                fragment="primary-point-of-contact",
+                _anchor="primary-point-of-contact",
+            )
+        )
+
+    return redirect(
+        url_for(
+            "portfolios/admin.html",
```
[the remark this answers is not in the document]
jGonzalez-CCPO: I think you right, I copy the mistake of line 94, should I fix all this. if that change works?

atat/routes/portfolios/admin.py
jGonzalez-CCPO: Checking on that

templates/portfolios/fragments/portfolio_members.html
jGonzalez-CCPO: I move it to be along to the other element on the same form.

kylelaker-ccpo:
There are bunch of tests in `tests/routes/portfolios/test_admin.py` that are decorated with `@pytest.mark.skip` that define the expected behavior of this route. Those decorations should be deleted so that the tests are no longer skipped. That may help improve coverage.

atat/routes/portfolios/admin.py
jGonzalez-CCPO: I just follow the existing patterns, I will follow your recommendation instead because it also called my attention. but notice that all the other forms on that temple are also bypassing the form for the role_id. I guess that is out of the scope of this ticket.

kylelaker-ccpo (changes requested):
Looks good! Nice job getting to 100% coverage. I am not concerned about the introduced code smell reported by SonarCloud since it's trivial and I am actually not convinced that creating a variable actually makes things cleaner here; it also doesn't impact passing the quality gate.

There are a few other tests in the file mentioned in my previous comment that are still being skipped. Since all of them seem like reasonable tests, I don't see a compelling reason to delete any of them or to continue skipping them. For each of them, we should validate that they pass when no longer skipped. If for any reason they fail, the test or the application code should be corrected accordingly.

Thanks for your work on improving this!

DGabriel-CCPO (approved):

kylelaker-ccpo (approved):
Great work on this! I appreciate you working with me to get the comments addressed. Validated running this locally and was able to reassign PPOC as expected. Also ensured that it wasn't possible to end up without a PPOC. Proper test coverage, and as mentioned, the additional code smell is acceptable at this time and does not impact the quality gate.

PR #2287, opened 2021-04-28
Cleanup use of hard coded strings in favor of referencing values from base.ini

jasonburkert-ccpo:
1. Removed  `AZURE_GRAPH_RESOURCE` from _base.ini_ instead referencing `AZURE_PUBLIC_CLOUD.endpoints.microsoft_graph_resource_id` from package `msrestazure`
2. Adjustments from the above because the new value `https://graph.microsoft.com/` includes a trailing forward-slash, and the old value did not
3. Similar to the above, made use of `AZURE_PUBLIC_CLOUD.endpoints.resource_manager` from package `msrestazure`
4. Improved description of `AZURE_POWERSHELL_CLIENT_ID` in readme.  There is no meaningful use of the value 1950a258-227b-4e31-a9cf-717495945fc2 that can be replaced with a reference to the definitive occurrence in file base.ini

Within file _azure_cloud_provider.py_ the referenced endpoints from package `msrestazure` are already conveniently available from `self.sdk.cloud.endpoints`.  We should make more use of these.

In some cases the forward-slashes before default scopes had to have been duplicated.  I imagine we'll find more of this. 

Ticket: AT-6033

### DIFF
--- a/atat/domain/csp/cloud/azure_cloud_provider.py
+++ b/atat/domain/csp/cloud/azure_cloud_provider.py
@@ -198,8 +198,6 @@ class AzureCloudProvider(CloudProviderInterface):
         self.root_tenant_id = config["AZURE_TENANT_ID"]
         self.vault_url = config["AZURE_VAULT_URL"]
         self.powershell_client_id = config["AZURE_POWERSHELL_CLIENT_ID"]
-        self.graph_resource = self.sdk.cloud.endpoints.microsoft_graph_resource_id
-        self.graph_scope = self.graph_resource + ".default"
         self.default_aadp_qty = config["AZURE_AADP_QTY"]
         self.roles = {
             "owner": config["AZURE_ROLE_DEF_ID_OWNER"],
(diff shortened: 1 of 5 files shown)


atat/domain/csp/cloud/azure_cloud_provider.py
```diff
@@ -197,7 +196,7 @@ def __init__(self, config, azure_sdk_provider=None):
         self.root_tenant_id = config["AZURE_TENANT_ID"]
         self.vault_url = config["AZURE_VAULT_URL"]
         self.powershell_client_id = config["AZURE_POWERSHELL_CLIENT_ID"]
-        self.graph_resource = AZURE_PUBLIC_CLOUD.endpoints.microsoft_graph_resource_id
+        self.graph_resource = self.sdk.cloud.endpoints.microsoft_graph_resource_id
```
jasonburkert-ccpo, line 199: All of the endpoints in package `msrestazure` are already conveniently accessible here (because `self.sdk.cloud` = `AZURE_PUBLIC_CLOUD` from that package).  We just need to make use of them.

kylelaker-ccpo (approved):
Great changes! I checked over `base.ini` for any other string values that are present in both that file and hardcoded in the application and I couldn't find any (except for _very_ trivial things like the number `5` used in unrelated contexts). This is awesome and it's great to see this all get unified. Love to see that we've also dropped a config value to just pull it directly from the Azure SDK. Great stuff!

PR #2289, opened 2021-04-29
Fixing run instructions to include the right host

jeffsegal-ccpo:


anhtran-ccpo:
Is running against a non local instance simple as changing the host address to that server? i.e `demo.atat.dev`

kylelaker-ccpo (changes requested):
I think there's some wonkiness with `host.docker.internal` with WSL2. Let me take a look this AM. But while I can resolve the hostname, the port doesn't seem to get forwarded to WSL2.

jeffsegal-ccpo:
> Is running against a non local instance simple as changing the host address to that server? i.e `demo.atat.dev`

Yep

kylelaker-ccpo (approved):
Looks like I was wrong -- this works fine with WSL since it's all within Docker. Ran tests against `localhost` through Locust and everything worked fine.

PR #2290, opened 2021-04-29
Update to GitHub-native dependabot

kylelaker-ccpo:
https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates

The new dependabot does not support auto-merging which is fine because we don't do that anyway. That's why it's not present in the new config. Additionally, we now check for updates to our GitHub Actions actions.

Ticket: AT-6132


jasonburkert-ccpo (approved):
Swaps old config file for new config file.  Looks good.
Continues checking Python and JavaScript dependencies, and adds checking of GitHub Actions dependencies.

PR #2291, opened 2021-04-29
Update cssnano to remove is-svg

kylelaker-ccpo:
The update to `cssnano` 4.1.11 removes the dependency on `is-svg`. Previously, it depended on v3.0.0 which has a security vulnerability. It seems that updating `yarn upgrade parcel-bundler` also fixes this but it changes way more packages than I was comfortable with for a targeted dependency update. Changing the `resolution` for `cssnano` to specify 4.1.11 lets us use the new version without the dependency with the fewest number of changes to other packages that I could find.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "minimist": ">=1.2.2",
     "acorn": ">=5.7.4 <6.0.0 || >=6.4.1 <7.0.0 || >=7.1.1",
     "uswds/**/acorn": "^6.4.1",
-    "kind-of": ">=6.0.3"
+    "kind-of": ">=6.0.3",
+    "cssnano": "4.1.11"
   },
   "browserslist": [
     "last 3 version",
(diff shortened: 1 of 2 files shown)


jasonburkert-ccpo (approved):
Looks good.  Checked out branch to inspect dependencies.
Confirmed removes package `is-svg` which is the subject of a current high priority dependabot alert.
```
:~/atat$ git status
On branch feature/update-is-svg
Your branch is up to date with 'origin/feature/update-is-svg'.

nothing to commit, working tree clean
:~/atat$ yarn why is-svg
yarn why v1.22.4
[1/4] Why do we have the module "is-svg"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
error We couldn't find a match!
Done in 0.58s.
:~/atat$ yarn why cssnano
yarn why v1.22.4
[1/4] Why do we have the module "cssnano"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "cssnano@4.1.11"
info Reasons this module exists
   - "parcel-bundler" depends on it
   - Hoisted from "parcel-bundler#cssnano"
   - Hoisted from "parcel-bundler#htmlnano#cssnano"
info Disk size without dependencies: "48KB"
info Disk size with unique dependencies: "992KB"
info Disk size with transitive dependencies: "22.95MB"
info Number of shared dependencies: 40
Done in 0.84s.
```

PR #2292, opened 2021-04-29
Fix accessibility issue focusing on the first portfolio link  

anhtran-ccpo:
The sidenav link had issues focusing on the first link when the second portfolio is selected. This brings the focus up to the front when focused on through tabbing in the keyboard.

AT-5116

### DIFF
--- a/styles/elements/_sidenav.scss
+++ b/styles/elements/_sidenav.scss
@@ -113,7 +113,7 @@
 
     &:focus {
       position: relative;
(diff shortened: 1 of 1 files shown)


kylelaker-ccpo (approved):
Looking good! image

This fixes _both_ the issue where the header covers the top of the focus dashes and where the blue selected item covers the bottom of it. The item below is highlighted just fine too. 
image

PR #2294, opened 2021-04-29
Integrate UI tests to CI pipeline

kylelaker-ccpo:
This breaks out the previous workflow into several different workflows and jobs that can run concurrently. Additionally, the UI tests are added to the pipeline, running the smoke tests in Chrome as an initial target that hopefully can run quickly but can also provide feedback on any likely changes. It is easy to change to run different tests or browsers over time but that may have an impact on how long it takes to run the pipeline to complete.

Ticket: AT-6185

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,13 @@ on:
   pull_request:
   push:
     branches:
-      - "master"
+      - "main"
       - "develop"
       - "release/**"
 
 jobs:
   cibuildscript:
-    name: Pre-merge Checks / Unit Tests / Sonar Scan
+    name: Unit Tests and Sonar Scan
     runs-on: ubuntu-20.04
     env:
       PYTHON_VERSION: "3.8.8"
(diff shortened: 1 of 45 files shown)


kylelaker-ccpo:
Quick update on status; These seem to work! That's exciting. The commit messages have the more significant details on the implementation and the decisions that have been made so far. Documentation changes for the UI Test README to ensure the document reflects the code changes is forthcoming.

kylelaker-ccpo:
One thing I would like early feedback on is:
 - which UI test suites should the pipeline run (smoke only, all, regression, etc)
 - which browser(s) should be used? (chrome + ie, all, chrome only, ie only, etc)

Remember that the more tests that we run and the more browsers we test with, the longer this will take; and these tests will be run on _every_ PR. Not only does that impact developer experience but also could cause scenarios where we start tripping over our concurrency limit.

jGonzalez-CCPO (approved):

kylelaker-ccpo:
@jeffsegal-ccpo since this renames "Pre-merge Checks / Unit Tests / Sonar Scan" to "Full CI Build / Unit Tests and Sonar Scan", it is going to require your merge and for the branch protection rules to be updated to reference the new rule. This is the same thing we had to do in #2249. We'll also need to identify which/how many of these should become required for branch protection.

PR #2298, opened 2021-04-29
AT-6014 Update JavaScript tests to eliminate use of deprecated functions from Vue Test Utils (VTU) library

jGonzalez-CCPO:
Replaced deprecated methods from Vue Test Utils (VTU) library with recommended methods.

The remaining deprecation warning is related to the use of [this method](https://vue-test-utils.vuejs.org/api/wrapper/setmethods.html).

Ticket: AT-6014

### DIFF
--- a/js/components/__tests__/multi_checkbox_input.test.js
+++ b/js/components/__tests__/multi_checkbox_input.test.js
@@ -79,8 +79,8 @@ describe('Multicheckbox shows validation states correctly', () => {
     await checkboxA.setChecked()
     await checkboxA.setChecked(false)
 
-    expect(wrapper.contains('.usa-input--error')).toBe(true)
-    expect(wrapper.contains('.usa-input--success')).toBe(false)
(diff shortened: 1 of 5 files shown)


jasonburkert-ccpo (approved):
Ran Jest tests on feature branch.  Confirmed all deprecation warnings but one have been eliminated.
image

jGonzalez-CCPO:
I proposing that the last one be its own ticket since require to redefine, re-plan and recreate this test.

The feature in depreciation allows for a very unhealthy practice of modifying the same components methods that you are trying to test. Defeating the reason for the test in the first place. The same Vue team is taking off this feature for that exact reason.

In this test, the feature was used to emulate the interaction of a human, which is a better test on a live UI test than in a unit component test.  

Making it a bug.  


PD:
https://github.com/dod-ccpo/atat/pull/2298
`Link - Anti-pattern. Vue does not support arbitrarily replacement of methods, nor should VTU. If you need to stub out an action, extract the hard parts away.`

PR #2270, opened 2021-04-22
Add additional test cases for BrowserStack testing framework

ehoward-ccpo:
Additional test cases for BrowserStack testing framework.

Developed the remaining 6 test cases to complete UI tests.  Removed functions, methods, variables from 3 pageObjects that were unused in the framework.  Class names were updated to the python standard CapWords style. All tests have been converted from Ghost Inspector to the automation test framework in Python. All tests can run in BrowserStack on the supported browsers and show pass or fail.

1. Add CCPO User
2. Application Index with App
3. Edit App Member
4. New Portfolio Member
5. New User Flow
6. Edit Portfolio Member

Ticket: AT-6163

PR #2299, opened 2021-04-29
Revert "feature/AT-6163 removed unused markers"

kylelaker-ccpo:
This partially reverts commit a34a979defba88c64226f0ec2afb988bcdf9ff11 from #2270.

This change seems to have been introduced because the markers are not used in the UI tests; however, they are used in the regular application unit tests (in the `tests/` directory). Without these, `pytest` will emit a warning about unregistered markers. This is the cause of more verbose output in the recent CI runs. Adding these back in removes the warnings and ensures that any `PytestUnknownMarkWarning` we see are actually due to misspellings or other errors.

The initial commit correctly added the `daily` marker which is not removed during this revert.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -5,6 +5,11 @@ env =
 addopts = --ignore=tests/acceptance/ --cov=atat --cov-report term-missing --cov-fail-under 90
 
 markers =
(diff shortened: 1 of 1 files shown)


ehoward-ccpo (approved):
Understood, thanks for correcting my mistake.

PR #2300, opened 2021-04-30
Fix typo to appease dependabot

kylelaker-ccpo:
Dependabot seems to be grumbling on all our PRs because of this typo. I
have run the config file as-is through the validator at
https://dependabot.com/docs/config-file-beta/validator/ and it does not
report any issues with this updated config


jasonburkert-ccpo (approved):
Thumbs up.

PR #2301, opened 2021-04-30
Added tab index to the log in button

anhtran-ccpo:
Added tab index to the log in button in order to make it focusable through the keyboard
AT-5117

### DIFF
--- a/templates/fragments/dod_login.html
+++ b/templates/fragments/dod_login.html
@@ -1,6 +1,6 @@
 <dodlogin inline-template key="notice">
   <div>
-    <div class="usa-button usa-button-primary usa-button-big login-banner__button" v-on:click="displayNotice()">
+    <div class="usa-button usa-button-primary usa-button-big login-banner__button" v-on:click="displayNotice()" tabindex="0">
       {{ "login.login_button" | translate }}
     </div>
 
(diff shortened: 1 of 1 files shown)


kylelaker-ccpo (changes requested):
While this does make the button focusable, it does not make it so that the button can actually be triggered by the keyboard. Is there a reason this can't just be a `<button>`?  Or maybe the [button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)?

anhtran-ccpo:
> While this does make the button focusable, it does not make it so that the button can actually be triggered by the keyboard. Is there a reason this can't just be a `<button>`? Or maybe the [button role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)?

Just changed to a button to make it focusable and clickable through the keyboard. For some reason I thought it was already a button and I missed that. Thanks

kylelaker-ccpo (changes requested):
If this is already covered under another ticket, that's all good and lmk. But now when you trigger the "Sign in with CAC" using the keyboard, you get the "You are accessing a USG IIS" and you can't "Continue" with the keyboard. Can we fix that button as well or create a ticket to do so? It's still impossible to login without a mouse from my experience.

anhtran-ccpo:
> If this is already covered under another ticket, that's all good and lmk. But now when you trigger the "Sign in with CAC" using the keyboard, you get the "You are accessing a USG IIS" and you can't "Continue" with the keyboard. Can we fix that button as well or create a ticket to do so? It's still impossible to login without a mouse from my experience.

After digging into this for a bit. I found that the anchor tags do not interact with the space bar. Only the enter button is usable on anchor tags.

kylelaker-ccpo (approved):
Using Enter/Return does work on the "Continue" page. Looks good! Thanks for the clarification

PR #2305, opened 2021-04-30
Feature/AT 6197 relocate generators to page objects

palaparthi-ccpo:
Removed all the duplicate code from the test cases and moved to the PO s file.
AT-6197

### DIFF
--- a/uitests/framework/page_objects/application_page.py
+++ b/uitests/framework/page_objects/application_page.py
@@ -1,3 +1,7 @@
+import pytest
+import random
+import string
+
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
@@ -39,7 +43,7 @@ class CreateApplicationPages:
     def __init__(self, driver):
         self.driver = driver
 
-    def select_portfolio(self):  ###Needs to be moved to new portfolio page objects
+    def select_portfolio(self):  # Needs to be moved to new portfolio page objects
         self.driver.find_element_by_css_selector(self.btn_select_portfolio_css).click()
 
     def click_create_app(self):
@@ -293,3 +297,7 @@ class CreateApplicationPages:
                 "Set Portfolio Permissions",
             )
         )
+
+
+def random_dod_id_generator(size=10, chars=string.digits):
+    return "".join(random.choice(chars) for x in range(size))
--- a/uitests/framework/test_cases/test_add_ccpo_user.py
+++ b/uitests/framework/test_cases/test_add_ccpo_user.py
@@ -1,6 +1,4 @@
 import pytest
-import random
-import string
 
 from selenium.webdriver.common.by import By
 from selenium.common.exceptions import TimeoutException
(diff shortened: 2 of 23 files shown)


uitests/framework/page_objects/application_page.py
```diff
@@ -1,3 +1,7 @@
+import pytest
```
jasonburkert-ccpo, line 1: I believe the import of `pytest` can be removed.  It doesn't appear to be referenced.

uitests/framework/test_cases/test_add_ccpo_user.py
```diff
@@ -33,9 +31,6 @@ def test_new_ccpo_user(self, setup):
         self.login = Login(self.driver)
         self.cm = PageObjectMethods(self.driver)
 
-        # Generator to create unique DOD ID number
-        self.dodid = random_no_generator()
-
         self.login.validate_ccpo_user_displayed()
         self.login.click_add_new_user()
         self.login.enter_new_dod_id("3456789012")
```
jasonburkert-ccpo, line 36: This static dod id is repeated throughout the tests.
For example: https://github.com/dod-ccpo/atat/search?q=3456789012

As a potential future improvement, consider replacing this with a reference to _DEV_USERS['brandon']['dod_id'] from [atat/routes/dev.py](https://github.com/dod-ccpo/atat/blob/fa0cd96601ac3f6c84f05bd28f12cde327b7d3a1/atat/routes/dev.py#L76)

jasonburkert-ccpo: For example:
image

uitests/framework/test_cases/test_create_draft_to.py
```diff
@@ -35,6 +34,7 @@ def test_create_draft_task_order(self, setup):
         self.port.click_new_portfolio()
         self.port.validate_new_portfolio()
         self.port.validate_name_desc()
+        # Random Generator for unique Portfolio Name
         self.pName = "Test Portfolio" + random_generator()
```
jasonburkert-ccpo, line 38: A single space here between the prefix and the random number might improve the looks of the test output.

jasonburkert-ccpo: This could apply anywhere that a portfolio/application/environment prefix is being combined with a random generated content.

uitests/framework/page_objects/application_page.py
```diff
@@ -293,3 +297,7 @@ def validate_port_permission(self):
                 "Set Portfolio Permissions",
             )
         )
+
+
+def random_dod_id_generator(size=10, chars=string.digits):
+    return "".join(random.choice(chars) for x in range(size))
```
jasonburkert-ccpo, line 303: This implementation is really no different than `random_no_generator()` in `task_order_page.py`, except that they have a different default length.

They could be combined and called thusly:
`random_no_generator(size=10)` for DoD ids
and
`random_no_generator(size=random.randrange(13,18))` for TO numbers
Note that the code shown is correct for TOs' expected range of 13 to 17 characters.

jasonburkert-ccpo: This seems to be generally true of all the random generator functions.  They could all be replaced with a single implementation in a central place.  Then commonly used aliases can delegate with relevant parameters...

```
def generate_dod_id():
    return random_no_generator(size=10, chars=string.digits)

def generate_task_order_no():
    return random_no_generator(size=random.randrange(13,18), chars=string.digits+string.ascii_letters)
```

jasonburkert-ccpo (approved):
I made several comments about potential improvements, but none I think that should hold up the ticket or merging this to develop.

@palaparthi-ccpo Please take some time to review my comments.

PR #2310, opened 2021-05-03
Fix initial application invite success message to use first and last name

DGabriel-CCPO:
Changed the initial application invite success message from using email address to using first and last name.
This makes the initial message consistent with the re-send message.

Ticket: AT-4990


jasonburkert-ccpo (approved):
Looks good.

PR #2313, opened 2021-05-04
Attempt to fix Sonar Scans for Dependabot

kylelaker-ccpo:


jeffsegal-ccpo (approved):

jeffsegal-ccpo:
Doing an admin merge because there's no easy way of testing this update short of this.

PR #2320, opened 2021-05-05
Added logs for user account events and for Redis deleting a session

anhtran-ccpo:
Provided logs around user accounts and when Redis deleting a session

Ticket: AT-5017

### DIFF
--- a/atat/domain/auth.py
+++ b/atat/domain/auth.py
@@ -64,7 +64,7 @@ def get_last_login():
 def _nullify_session(session):
     session_key = f"{app.config.get('SESSION_KEY_PREFIX')}{session.sid}"
     app.redis.delete(session_key)
-    app.logger.info("%s is deleted from Redis.", session_key)
+    app.logger.info("The session [%s] has been destroyed.", session_key)
     request.cookies = ImmutableTypeConversionDict()
     request_ctx_stack.top.session = app.session_interface.open_session(app, request)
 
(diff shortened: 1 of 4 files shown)


atat/domain/auth.py
```diff
@@ -64,6 +64,7 @@ def get_last_login():
 def _nullify_session(session):
     session_key = f"{app.config.get('SESSION_KEY_PREFIX')}{session.sid}"
     app.redis.delete(session_key)
+    app.logger.info(f"{session_key} is deleted from redis")
```
kylelaker-ccpo, line 67: f-strings are awesome and I am glad to see you're using them; however, there is one place in Python that we actually prefer to not use them, and that is with the `logging` framework. This is usually a performance thing. In this scenario, the string interpolation will always be performed, even when we aren't logging info-level messages. With the `printf`-style,:

```
app.logger.info("%s is deleted from redis", session_key)
```

The new string only gets created if we actually need to log the message since the log message is created lazily. Our pylint config probably should have caught this but it doesn't seem to detect the `app.logger` as being a logger.

anhtran-ccpo: Changed to % string interpolation

kylelaker-ccpo (approved):

kylelaker-ccpo (approved):
Thanks! Logging fixes look good!

kylelaker-ccpo:
Alright cool, must have been a temporary failure on the SonarCloud-side.

kylelaker-ccpo (approved):
I like these new wording updates. Thanks!

PR #2322, opened 2021-05-05
Add keyboard accessibility to Application/Add Team Member

jGonzalez-CCPO:
Currently when viewing the Application there is an "Add Team Member" button, but that button does not receive keyboard focus.  There are a few items that need to be adjusted in order to make it accessible to persons with disabilities such as:

1. tabindex=02. role="button"3. The event needs to be changed from click to onclick4. aria-haspopup="true"

### DIFF
--- a/templates/applications/fragments/add_new_environment.html
+++ b/templates/applications/fragments/add_new_environment.html
@@ -27,8 +27,8 @@
       </form>
     </div>
     <a class="usa-button usa-button-secondary add-new-button"
-       tabindex=0
        role="button"
+       tabindex=0
        v-on:click="toggle">
       {{ "portfolios.applications.add_environment" | translate }}
     </a>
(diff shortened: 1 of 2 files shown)


tomarnold-ccpo (changes requested):
Hey, looks like you added the aria-haspopup="true" to both the "add new environment" button as well as the "add new user" button.  Only the "add new user" opens a modal window.  The add new environment is actually an expand/collapse and shouldn't have the attribute aria-haspopup.

jGonzalez-CCPO:
Only the "add new user" opens a modal window. The add new environment is actually an expand/collapse and shouldn't have the attribute aria-haspopup. change was done

tomarnold-ccpo (commented):
Looks good to me.

tomarnold-ccpo (approved):

PR #2330, opened 2021-05-06
Feature/at 6191

DGabriel-CCPO:
There needs to be an administrative dashboard that exists within ATAT.  The page should be available for CCPO administrators only and should leverage the admin console framework (Site Navigation, Admin Navigation, Global Header, Page Title Header)

### DIFF
--- a/styles/components/_admin_console_layout.scss
+++ b/styles/components/_admin_console_layout.scss
@@ -1,0 +1,80 @@
+.admin-header {
+  flex-direction: column;
+  margin: $gap * 2 0;
+  max-width: $max-panel-width;
+
+  @include media($small-screen) {
+    flex-direction: row;
+  }
+
+  .col--grow {
+    overflow: inherit;
+    align-self: center;
+  }
+
+  &__name {
+    @include h1;
+
+    h1 {
+      margin: 0;
+      font-size: 3.5rem;
+    }
+
+    p {
+      font-size: $small-font-size;
+      margin: 0 0 (-$gap * 0.5);
+      color: $color-gray-medium;
+      max-width: 100%;
+    }
+  }
+
+  .links {
+    justify-content: center;
+    font-size: $small-font-size;
+
+    .icon-link {
+      padding: 0;
+      border-radius: 0;
+      color: $color-blue-darkest;
+      min-width: 10rem;
+      min-height: 10rem;
+
+      .col {
+        margin: 0 auto;
+      }
+
+      &:hover {
+        background-color: $color-aqua-lightest;
+      }
+
+      .icon {
+        @include icon-color($color-blue-darkest);
+        @include icon-size(30);
+      }
+
+      &.active {
+        color: $color-blue;
+        background-color: $color-gray-lightest;
+        text-decoration: none;
+
+        &:hover {
+          background-color: $color-aqua-lightest;
+        }
+
+        .icon {
+          @include icon-color($color-blue);
+        }
+      }
+
+      &--icon {
+        text-align: center;
+      }
+    }
+  }
+}
+.admin-panel-container {
+  @include media($large-screen) {
+    @include grid-row;
+    min-height: 500px;
+  }
+}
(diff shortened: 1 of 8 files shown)


styles/components/_admin_console_layout.scss
```diff
@@ -0,0 +1,80 @@
+.admin-header {
```
jGonzalez-CCPO, line 1: MAy be this style can be it now generic since is used in more than one component

jGonzalez-CCPO (changes requested):

jGonzalez-CCPO (approved):

templates/admin/header.html
```diff
@@ -0,0 +1,26 @@
+
+<div class='admin-header row'>
+  <div class='col col--grow'>
+    <div class='admin-header__name'>
+      <p>ADMIN CONSOLE</p>
```
kylelaker-ccpo, line 5: Does this need to be here? 
image

templates/admin/admin_sidenav.html
```diff
@@ -0,0 +1,36 @@
...
+        </div>
+        <ul class="sidenav__list" v-if="isVisible">
+            {{ SidenavItem('Dashboard',
+              href=url_for("admin.dashboard"),
+              active=request.url_rule.endpoint == "admin.dashboard",
+              ) }}
+          {{ SidenavItem('Users',
+              href=url_for("admin.all_users_page"),
+              active=request.url_rule.endpoint == "admin.all_users_page",
+              ) }}
```
kylelaker-ccpo, line 31: Can this indentation be cleaned up a bit?

kylelaker-ccpo (changes requested):

templates/admin/admin_sidenav.html
jGonzalez-CCPO: I rewrite that, I hope to address your concern.

templates/admin/header.html
jGonzalez-CCPO: Not sure, but I guess that is @DGabriel-CCPO  responding to criteria #3: "Add Admin page header"

DGabriel-CCPO: this is only There because it was present in the screenshot. I can remove it if need be but I would need to know what to replace it with.

kylelaker-ccpo, line 5: Now that we're keeping this and have style applied, thoughts on making the case normal 
```suggestion
      <p>Admin Console</p>
```

And in the CSS applying:
```
text-transform: uppercase
```
https://caniuse.com/mdn-css_properties_text-transform

The upper case is more of a style/presentation thing than actual meaning and feels more appropriate to have in the CSS.

templates/admin/header.html
```diff
@@ -0,0 +1,26 @@
...
+      <h1>Dashboard</h1>
+    </div>
+  </div>
+  <div class="col">
+    <div class="action-group">
+      <a role="button"
+         tabindex="0"
+         href=""
```
kylelaker-ccpo, line 13: We recently merged some code to get this working for environments but I guess that code can't really be used at a global level, correct? And that's why we can't really wire this up right now?

templates/admin/admin_sidenav.html
```diff
@@ -0,0 +1,43 @@
...
+          </template>
+          <template v-else>
+            <a href="#" v-on:click="toggle" class="sidenav__toggle col">
+              <span>Show</span>
+              {{ Icon('angle-double-right-solid', classes="sidenav__toggle-arrows icon--primary") }}
+            </a>
+          </template>
+        </div>
+        <ul class="sidenav__list" v-if="isVisible">
+
```
kylelaker-ccpo, line 24: Should we drop this empty line? There's not one at the bottom of the block.

```suggestion
```

templates/admin/header.html
jGonzalez-CCPO: My understanding is that the functionality of those buttons is out of the scope of the ticket as we agree on the pocket game.

kylelaker-ccpo (approved):

PR #2338, opened 2021-05-07
Always run all checks

kylelaker-ccpo:
The branch protection rules require all these checks

### DIFF
--- a/.github/workflows/javascript.yml
+++ b/.github/workflows/javascript.yml
@@ -7,21 +7,21 @@ on:
       - "main"
       - "develop"
       - "release/*"
-    paths:
-      - "**.js"
-      - ".github/**"
-      - "package.json"
-      - "yarn.lock"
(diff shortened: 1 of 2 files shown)


kylelaker-ccpo:
The original PR tried to be too "smart" by only running the necessary checks for code that had changed; however, the branch protection rules require all of our checks to pass, which was an oversight in my recommendation on what rules we should require. This comments out the paths because 1) they're fairly intentional and the list isn't one that I want to lose and 2) because it implies that we do want the lines present, just not right now.

This makes it so that all the rules are triggered on every pull request, regardless of path.

jasonburkert-ccpo (approved):
@kylelaker-ccpo Explained that branch protection rules require All checks to run.  That means that the JavaScript and Python checks being modified here must have the qualifying paths removed (commented actually) so that they are always triggered and not just when something is modified in those paths.

PR #2343, opened 2021-05-10
Ensure the status is set at the end of test

kylelaker-ccpo:
The set_session_name call was being invoked (incorrectly) when
set_session_status should have been invoked. The call to set the name
would have failed because the wrong arguments were being supplied. So
this just resulted in unmarked tests in the BrowserStack Automate
console.

Ticket: AT-6185


jasonburkert-ccpo (approved):
Looks good.

PR #2344, opened 2021-05-10
update pyenv init command due to recent upstream change

jasonburkert-ccpo:
pyenv has changed recently.  updated pyenv init command to set PATH.

The old command issues this warning:
```
WARNING: `pyenv init -` no longer sets PATH.
Run `pyenv init` to see the necessary changes to make to your configuration.

```
https://github.com/pyenv/pyenv/issues/1906


kylelaker-ccpo (approved):

kylelaker-ccpo:
The unfortunate thing here is that if folks don't have the latest `pyenv` they may not have the `pyenv init --path` option. We should probably recommend, for the existing users, that folks:
```
git -C ~/.pyenv pull
```
and _then_ make the changes in this PR.

This only applies to the existing users, not to new use (who this doc is geared towards). But something to keep in mind if we need to provide transition advice to the team.

jasonburkert-ccpo:
For users with existing environments, in addition to the git pull below to get the latest pyenv, you may also need to edit any dot files such as  `~/.bashrc` or `~/.bash_profile` where you might have the old command `pyenv init -` and update it to the new command `pyenv init --path`.
 
> The unfortunate thing here is that if folks don't have the latest `pyenv` they may not have the `pyenv init --path` option. We should probably recommend, for the existing users, that folks:
> 
> ```
> git -C ~/.pyenv pull
> ```
> 
> and _then_ make the changes in this PR.
> 
> This only applies to the existing users, not to new use (who this doc is geared towards). But something to keep in mind if we need to provide transition advice to the team.

PR #2346, opened 2021-05-10
AT-6181 Step-by-Step [Next/Cancel] walkthrough of application is not accessible

jGonzalez-CCPO:
Description
• Unsaved changes
Throughout the application there is a Next button and a Cancel link at the bottom of each of the steps.  

 
There are a couple of things broken with these buttons and links.  First, the button needs to receive a role of button as it is actually an input.  This will inform a screen reader user that the object can be interacted with as opposed to typing in information.
The cancel link needs to receive tabindex=0, because although it is an anchor element it does not have an href and as such does not receive focus natively.  
Also, the visual order of the objects needs to match the order of the DOM, or vice versa.  Currently the logical flow is Next and then Cancel, which from an application perspective makes sense however then the cancel button should be to the right of the Next button not to the left.

Acceptance Criteria
1. Cancel link receives tabindex=0 attribute
2. Next button receives role="button" attribute


jGonzalez-CCPO:
@tomarnold-ccpo summiting my changes now. to address your comment. 

@click is the view way or asserting the click and keyboard click behavior on veu.

tomarnold-ccpo (approved):
Everything looks good, including the `aria-haspopup="true"`, thanks for adding that in.

PR #2351, opened 2021-05-11
Automate re-running failed Dependabot jobs

kylelaker-ccpo:
This automates the workaround for the current situation with Dependabot. It re-runs all failed workflow runs on open pull requests authored by Dependabot. This allows us to manually validate that the dependencies and updates look reasonable before triggering the re-run.

For an example see: https://github.com/kylelaker-ccpo/atat/runs/2559939595?check_suite_focus=true

Ticket: AT-6132


jasonburkert-ccpo:
Love the improvements.

PR #2357, opened 2021-05-12
Add Full UI Test Suite

kylelaker-ccpo:
This adds an automated run of the full UI test suite every day at 21:30
UTC. This is an approximate time; GitHub Actions may delay running this
workflow at peak times per the documentation. It is also possible to
trigger a manual run using the workflow_dispatch mechanism.

The Build Name in BrowserStack is set to the output of the `-I` flag to
`date(1)` which looks like: "2021-05-12".

Hopefully this helps ensure that we run the tests daily while also
avoiding running them at times where we are already making heavy use of
BrowserStack.


kylelaker-ccpo:
**Feedback Received**: The database needs to be cleared and re-seeded between each browser.

Maybe it'd be possible to run all three browsers concurrently against different DBs? We can look into the `matrix` definition for workflows.

ehoward-ccpo:
There are random failures with the following tests that need to be skipped in order to have successful UI test runs.  We are working on solving the problems. 
uitests/framework/test_cases/test_resend_app_mem_invite.py
uitests/framework/test_cases/test_resend_port_mem_invite.py
uitests/framework/test_cases/test_revoke_app_mem_invite.py
uitests/framework/test_cases/test_revoke_env.py
uitests/framework/test_cases/test_revoke_port_mem_invite.py

PR #2359, opened 2021-05-13
unsecure two routes for load testing

jasonburkert-ccpo:
This is not to be merged.  Just for short term load testing.

Ticket: AT-6168


deploy/azure/uwsgi.ini
```diff
@@ -8,8 +8,9 @@ chmod-socket = 666
 chown-socket = atat:atat
 
 ; basic auth for local login; not enabled in all environments
-route = ^/login-local basicauth:Local Login,/config/localpassword
-route = ^/dev-new-user basicauth:Local Login,/config/localpassword
+; AT-6168 - Disable for load testing
+; route = ^/login-local basicauth:Local Login,/config/localpassword
+; route = ^/dev-new-user basicauth:Local Login,/config/localpassword
```
jasonburkert-ccpo, line 13: @jGonzalez-CCPO points out that this has been moved to the `atat-infrastructure` repo, so what I'm doing here has no effect.

kylelaker-ccpo:
Per the review comments and with #2363 having been opened, it seems reasonable to close this PR (which we will not merge) and to keep the branch for reference as far as changes that will need to be made within `atat-infrastructure` to support this change.

PR #2360, opened 2021-05-13
Ensure BrowserStack Local is always stopped

kylelaker-ccpo:
`if: ${{ always() }}` signifies that this step should run, even if a
previous step failed. This is important to ensure that logs get provided
as an artifact correctly and to cleanup. Currently, this step gets
skipped if any previous step failed.


jasonburkert-ccpo (approved):
Good idea.

PR #2361, opened 2021-05-13
fixed lang accessibility issue on login page

DGabriel-CCPO:
Per Siteimprove*
Compliance Level:  A
The webpage is missing a definition of the natural language.
*Possible Ways to Mediate*
Provide a 'lang' attribute for HTML documents or an xml:lang attribute for xhtml documents to indicate the natural language of the page.

Ticket: AT-5146


jasonburkert-ccpo (approved):
Good!

PR #2362, opened 2021-05-13
Update to Python 3.9

kylelaker-ccpo:
In general, this updates only the code that is not OBE by the coming
transition to atat-infrastructure. This means that the `ops/`,
`ansible/`, and other directories have largely been ignored. Further,
where possible, we try to accept _any_ Python 3.9 version for the
tooling in this repository. Whereas it is important to ensure that we
are deploying the latest approved patch version, specifying tight
requirements on patches for this repository has generally caused issues
with compatibility in the past.

Additionally, `.tool-versions` which is not used by ATAT is removed and
added to the .gitignore so it does not get added back.

We also now specify our version as `^3.9`. This will allow us to, in the
future, adopt new versions of Python (such as 3.10) in the
infrastructure repository without necessarily needing to make changes to
the application project. We could do `^3.8`; however, Python 3.9 gives
us some new nice new features and it would be nice to adopt them. And we
can't do that if we declare 3.8 compatibility still.

Ticket: AT-6161


kylelaker-ccpo:
The merge for this should be coordinated with messaging to the team about the need to re-create virtual environments.

kylelaker-ccpo:
These are the instructions to re-create the virtual environment that should be shared with the team:
```
git -C ~/.pyenv pull
pyenv install 3.9.5
pyenv shell 3.9.5
pip install poetry
rm -rf ./.venv
./script/setup
```

kylelaker-ccpo:
It looks like the version of Selenium that we are using has a bug (https://github.com/SeleniumHQ/selenium/issues/9460) in the functionality of it's upload method that breaks compatibility with Python 3.9. The recommendation seems to be to adopt a pre-release (beta 3) version of Selenium 4 to resolve this. It is supposed to be a drop-in replacement.

uitests/framework/test_cases/conftest.py
```diff
@@ -29,6 +30,52 @@
...
+                return filename
+            elif '{"status":405,"value":["GET","HEAD","DELETE"]}' in e.__str__():
+                return filename
+            else:
+                raise e
+    # fmt: on
+
+    monkeypatch.setattr(WebElement, "_upload", patched_upload)
+
+
```
jasonburkert-ccpo, line 78: Looks fair enough.  Want to give it a whirl before merging.

kylelaker-ccpo:
If we want to loosen the requirement to allow Python 3.8 _or_ 3.9 (not just 3.9, as the PR is currently written) we should identify when we are going to drop support for 3.8.

uitests/framework/test_cases/conftest.py
jasonburkert-ccpo: Tried it locally.  Works fine.  Ran script/test with 3.9.5, and smoke test of application in local development.  Only stumbled in one place using instructions above because I had a version set with `pyenv global`.  Perhaps replace `pyenv shell 3.9.5` with `pyenv global 3.9.5`

jasonburkert-ccpo:
Tried feature branch and application in local development.  Works fine.  Ran script/test with 3.9.5, and smoke test.  Only stumbled in one place using instructions above because I had a version set with `pyenv global`.  Perhaps replace `pyenv shell 3.9.5` with `pyenv global 3.9.5`

kylelaker-ccpo:
@jasonburkert-ccpo the instructions have been updated. Will make sure to include that in any announcements to the team.

jasonburkert-ccpo:
Tested patch to `conftest.py`.

Baseline on 'develop' with Python 3.8.8
image

This feature branch with Python 3.9.5
image

jasonburkert-ccpo (approved):
Tested `script/test`, and `script/secure_server` with smoke test.  Also ran BrowserStackLocal tests with chrome.

PR #2363, opened 2021-05-13
Remove infrastructure code

kylelaker-ccpo:
Remove code that's been moved to `dod-ccpo/atat-infrastructure` repo.


jasonburkert-ccpo (approved):
I'd maybe handle edits to `README.md` separately.  I agree with the removal of everything else.

jasonburkert-ccpo:
You're right. `script/secure_server` fails without `Procfile`.
> 2021-05-13 17:14:54 [10854] [ERROR] Procfile does not exist or is not a file

kylelaker-ccpo:
I'd recommend that we hold off on merging this until after tomorrow's deployment, just due to the possibility there may be something unforeseen here and I don't want to cause any grief to that process.

PR #2367, opened 2021-05-14
Avoid reading attributes of missing PDFs

kylelaker-ccpo:
Presently, the code attempts to read the object name and ID of the PDF
to generate a download link, even if the TO is in draft status. When in
draft status, we don't use the download link because we either skip
directly to the page in the workflow where fields are missing or we
prompt for an upload (because there is not a PDF).

Here, we refactor the route in such a way that we only generate the
download URL if it is going to be used.

Ticket: AT-6203


jGonzalez-CCPO (approved):

jGonzalez-CCPO:
This is a very complete solution, I tested and it worked. code is clean.

PR #2370, opened 2021-05-18
Revert "Bump node-sass from 5.0.0 to 6.0.0"

kylelaker-ccpo:
Reverts dod-ccpo/atat#2340

This causes issues for some local environments where only Node 10 is available. The pipeline and local runs did not catch this.


kylelaker-ccpo:
image

This is an example of the issues users are seeing.

jasonburkert-ccpo (approved):
Revert this until we are ready.

PR #2371, opened 2021-05-18
AT-6180 Display Admin link for administrative users only

jGonzalez-CCPO:
Description
As an administrative user I want the ability to have a link in the header that will direct me to the Admin portal page.  The link should be next to the username, and should have an icon.
 

Acceptance Criteria
Admin link directing administrators to /ccpo-users is made available in the global header
This should only be available to Administrators of ATAT
Non-copyrighted Icon should be identified and added to the left of the text


DGabriel-CCPO (approved):

templates/navigation/topbar.html
```diff
@@ -11,22 +14,29 @@
...
-        <span class="topbar__link-label">Support</span>
-      </a>
-      {% endif %}
-      <a href="{{ url_for('atat.logout') }}" class="topbar__link"
-        title='{{ "navigation.topbar.logout_link_title" | translate }}' v-on:click="logout">
-        <span class="topbar__link-label">Logout</span>
-        {{ Icon('logout', classes='topbar__link-icon') }}
-      </a>
+        {% set current_user = g.current_user -%}
+        {% set can_manage_ccpo_users = current_user.is_permission_sets('manage_ccpo_users') -%}
```
kylelaker-ccpo, line 18: This effectively bypasses the way we have traditionally done checks based on permissions and implements a new method with unnecessary additional code on the `User` class. Instead, see how `user_can` is added via context processor for the `portfolios` and `atat` blueprints. That uses the existing `user_can_access` infrastructure (or basically the same code) as the rest of the application, in the way that we protect routes. This is handled in `atat.utils.context_processors`. Please change to either adopting a context processor, or alternatively, I am open to making `user_can_view` and `permissions` available throughout the application using `app.jinja_env.globals` (see how this is set in atat.app`). It would also mean that we could remove the custom context processor for the `atat` blueprint.

kylelaker-ccpo (changes requested):

templates/navigation/topbar.html
```diff
@@ -1,5 +1,8 @@
 {% from "components/icon.html" import Icon %}
 
+
+
+
```
kylelaker-ccpo, line 5: This additional whitespace is unnecessary.

templates/navigation/topbar.html
jGonzalez-CCPO: user_can, user_can_access are not accessible in table components  :(

PR #2374, opened 2021-05-18
Reuse user access logic across templates

kylelaker-ccpo:
Now that the Admin link in the header navigation bar requires the
ability to determine the permissions that a user has, it seems
reasonable to add the user_can_view function (accessible as user_can for
consistency) and the Permissions enum to the global Jinja environment.
Previously, this was done via two specific context processors, one for
the portfolios blueprint and another for the admin and ccpo blueprints,
so there is precedence for exposing this functionality.

Further, making it easier to reuse consistently ensures that we do not
have divergent behavior for permissions modeling and that we can make
changes to permissions check logic in as few places as possible.

Ticket: AT-6180
See also: #2371


kylelaker-ccpo:
Prior to marking this as ready for review I need to:
- write a better commit message and PR body
- cleanup the lint errors
- ensure that there's not a way to replace `user_can_view` with the canonical `user_can` implementation without breaking the portfolios, ccpo, and admin blueprints

atat/utils/context_processors.py
```diff
@@ -84,16 +84,15 @@ def assign_resources(view_args):
...
-        return Authorization.has_atat_permission(g.current_user, permission)
+    if g.application and Authorization.has_application_permission(
+        g.current_user, g.application, permission
+    ):
+        return True
+    if g.portfolio and Authorization.has_portfolio_permission(
+        g.current_user, g.portfolio, permission
+    ):
+        return True
+    return Authorization.has_atat_permission(g.current_user, permission)
```
kylelaker-ccpo, line 95: The usage of `g` is really what makes it harder to use the canonical `user_can` function. It _might_ be possible to pass `g.application` and `g.portfolio` and then have them check for `None` better but I am unsure of the impact that may have across the application.

What is important is that this function needs to always return a boolean, not raise an exception. An exception here would result in either a 404 or 500 for the end user. It really should only be used to hide/show elements; all routes, of course, need continue doing their own validation using `@user_can` decorator regardless.

kylelaker-ccpo:
The earlier issues have been resolved. This is ready for review.

jGonzalez-CCPO (approved):
I test the branch locally and it work as expected. 
Only waiting on approved changes on criteria by Tom

kylelaker-ccpo:
ACs were updated. Merging this PR based on the approval.

PR #2377, opened 2021-05-19
Restructure /login-dev to operate like /login

kylelaker-ccpo:
The structure of these two routes should largely be the same; however,
they use fairly different implementations. Reorganizing such that it
follows the same logic used in `/login` provides more consistency.

These two routes should now largely operate in the same way except for
the IdP that gets used and the `login_method` on the session.


atat/routes/dev.py
```diff
@@ -121,34 +121,24 @@ def message(self):
         return "You must provide each of: first_name, last_name and dod_id"
 
 
-@dev_bp.route("/login-dev", methods=["GET", "POST"])
+@dev_bp.route("/login-dev", methods=["GET"])
```
jasonburkert-ccpo, line 124: Good.

jasonburkert-ccpo (approved):
Thanks for the quick turn-around.

PR #2380, opened 2021-05-21
Feature/at 6187

DGabriel-CCPO:
Application Environment links do not receive keyboard focus

The template that was used to render the spans for the Application Environments needs tabindex and other aria attributes to make them keyboard-only accessible. this will result in a bad user experience for those who are unable to navigate the webpage with a mouse.

Ticket: AT-6187


jGonzalez-CCPO (approved):

templates/components/toggle_list.html
```diff
@@ -1,9 +1,9 @@
 {% macro ToggleButton(action, open_html, close_html, section_name, classes="", active_style=False) %}
-  <span v-on:click="toggleSection('{{ section_name }}')">
-    <span v-if="selectedSection === '{{ section_name }}'" class="accordion-table__item__toggler {{ classes }}{% if active_style -%} accordion-table__item__toggler--active{%- endif %}">
+  <span @keypress="toggleSection('{{ section_name }}')" @click="toggleSection('{{ section_name }}')">
+    <span tabindex="0" aria-controls="1" aria-expanded="true" role="region" v-if="selectedSection === '{{ section_name }}'" class="accordion-table__item__toggler {{ classes }}{% if active_style -%} accordion-table__item__toggler--active{%- endif %}">
```
jGonzalez-CCPO, line 3: I think it would be more readable if these properties are enditated, Whenever we have more than 3 on a tag, and we inserting data with jinja.

jGonzalez-CCPO (changes requested):

jGonzalez-CCPO (approved):

jGonzalez-CCPO:
code is looking good

PR #2385, opened 2021-05-24
Allow script/setup to _not_ reset the DB

kylelaker-ccpo:
Currently running `script/setup` always results in the database being
reset which can be an annoyance when changes require that the script be
run (such as major Python updates). The script currently enforces that
the database be reset through the `RESET_DB` variable; it is always set
to `"true"`.

Changing this to _default_ to `"true"` but to allow it to be configured
would make it slightly easier for folks to decide for themselves whether
the database should be reset. All that needs to be done is to set the
`RESET_DB` environment variable to `"false"`.

A warning is emitted when the database is not being reset, just to
remind users that this isn't necessarily the ideal behavior.


kylelaker-ccpo:
Example run of the script:

```
$ export RESET_DB=false
$ ./script/setup
Creating virtual environment using Python version 3.9.5...
Installing dependencies from lock file

No dependencies to install or update
Requirement already satisfied: pip==21.* in ./.venv/lib/python3.9/site-packages (21.1.2)
Installing dependencies from lock file

No dependencies to install or update
yarn install v1.22.10
[1/4] Resolving packages...
success Already up-to-date.
Done in 1.56s.
WARNING: The database (atat) will not be reset
yarn run v1.22.10
$ parcel build js/index.js -d static/assets --public-url /static/assets -o index.js
✨  Built in 4.03s.
<remainder of output removed>
```

kylelaker-ccpo:
@jasonburkert-ccpo Any thoughts on whether we actually want this? I'd like to either close or merge this to work towards clearing the PR backlog :)

jasonburkert-ccpo (approved):
This looks good to me.  Tried it locally.  Works great.  Doesn't alter base behavior, but allows you to retain your database if you want to.

PR #2387, opened 2021-05-24
Remove non-actionable TODOs

zachclark-ccpo:
Code base contained 40+ instances of the TODO markers, and these TODOs were decided to be non-actionable, deleting them to prevent further confusion.

Confluence Page discussing Non-actionable TODOs: https://ccpo.atlassian.net/wiki/spaces/AT/pages/1299087643/AT-5959+-+Code+base+contains+40+instances+of+the+TODO+markers
Ticket: https://ccpo.atlassian.net/browse/AT-5959


kylelaker-ccpo (changes requested):
Let's talk about removing a31aaa0 from this PR. Since the only usage is in the function that tests it, we probably can remove both the function and it's test. Therefore, we can either remove both in this PR or create a ticket to do so.

And since we'll be modifying history anyway, let's talk about doing a rebase to squash 2c60325 into b0cbe48.

The former is a blocker to approval; the latter is a nice-to-have.

jasonburkert-ccpo (approved):
These removals look reasonable.

jasonburkert-ccpo:
Re: https://github.com/dod-ccpo/atat/commit/a31aaa040a5c2da0511f16cd3ca0b0e41e9a3c96  Agree.  Please remove.

https://github.com/dod-ccpo/atat/search?q=log_system_event

image

zachclark-ccpo:
> Re: [a31aaa0](https://github.com/dod-ccpo/atat/commit/a31aaa040a5c2da0511f16cd3ca0b0e41e9a3c96) Agree. Please remove function `log_system_event`
> 
> https://github.com/dod-ccpo/atat/search?q=log_system_event
> 
> image

Hey Jason, for removing log_system_event , wont it mess up this pagination function? 
Should I remove it and leave it as:
```
@pytest.mark.audit_log
def test_paginate_audit_log():
    events = AuditLog.get_all_events(pagination_opts={"per_page": 25, "page": 2})
    assert len(events) == 25
```
Or is it better to remove the pagination test entirely? Please advise.

tests/domain/test_audit_log.py
```diff
@@ -31,23 +31,13 @@ def developer():
 
 @pytest.mark.audit_log
 def test_paginate_audit_log():
-    user = UserFactory.create()
-    for _ in range(100):
-        AuditLog.log_system_event(user, action="create")
-
     events = AuditLog.get_all_events(pagination_opts={"per_page": 25, "page": 2})
     assert len(events) == 25
```
kylelaker-ccpo, line 35: This test doesn't get run in the pipeline but would fail as-is. We're not logging any events so there's no way that there can be 25 of them :)

tests/domain/test_audit_log.py
```diff
@@ -31,23 +31,13 @@ def developer():
...
     portfolio = PortfolioFactory.create()
-    application = ApplicationFactory.create(portfolio=portfolio)
-    for _ in range(100):
-        AuditLog.log_system_event(
-            resource=application, action="create", portfolio=portfolio
-        )
-
     events = AuditLog.get_portfolio_events(
         portfolio, pagination_opts={"per_page": 25, "page": 2}
     )
```
kylelaker-ccpo, line 43: Same concerns as other comment.

kylelaker-ccpo:
@zachclark-ccpo and @jasonburkert-ccpo looking at this further, I think we need to discuss the value of the `AuditLog` class. We always set `USE_AUDIT_LOG` to `false`; therefore it doesn't get tested (or used in deployed environments). If we remove these tests, we will lose code coverage for the `get_` functions on the `AuditLog` class which I don't want to do. But reworking them to hackily call action that will trigger audit events is suboptimal as well.

@zachclark-ccpo can you provide information on when this class and the related functions/methods were added (including PRs and potentially tickets) and perhaps we can have a discussion to evaluate the proper path forward on this?

I don't want to suggest removing tests for code that does get run; I don't want to keep dead code around; and I don't want to add too much complexity to what is largely a removal ticket. We may need to create a separate ticket to handle the deletion/updating of the `AuditLog` class but even that should have a conversation around desired outcomes. Sorry for asking different things on this and thank you for raising good questions about the removal.

zachclark-ccpo:
> @zachclark-ccpo and @jasonburkert-ccpo looking at this further, I think we need to discuss the value of the `AuditLog` class. We always set `USE_AUDIT_LOG` to `false`; therefore it doesn't get tested (or used in deployed environments). If we remove these tests, we will lose code coverage for the `get_` functions on the `AuditLog` class which I don't want to do. But reworking them to hackily call action that will trigger audit events is suboptimal as well.
> 
> @zachclark-ccpo can you provide information on when this class and the related functions/methods were added (including PRs and potentially tickets) and perhaps we can have a discussion to evaluate the proper path forward on this?
> 
> I don't want to suggest removing tests for code that does get run; I don't want to keep dead code around; and I don't want to add too much complexity to what is largely a removal ticket. We may need to create a separate ticket to handle the deletion/updating of the `AuditLog` class but even that should have a conversation around desired outcomes. Sorry for asking different things on this and thank you for raising good questions about the removal.

AuditLog base.ini toggle PR: https://github.com/dod-ccpo/atat/pull/1127

kylelaker-ccpo (approved):
Thanks for your work and attention to detail here. I really appreciate it. This is great for cleaning up a good chunk of the code base.

PR #2390, opened 2021-05-24
Make it easier to document the CSP header policies

kylelaker-ccpo:
Previously the CSP header was a single string, which makes sense but it
was different for both development and non-development environments.
Additionally, there were some less-than-ideal configurations that are
not properly documented. These are required either by Parcel or other
libraries that we're using, but they were all tacked on to `default-src`
with minimal documentation.

This splits out the directives and allowed sources into a dictionary
that maps the directive to a list of policies. These policies can then
be documented inline. This makes future additions easier to track and to
document.


atat/app.py
```diff
@@ -160,15 +160,60 @@ def _set_security_headers(response):
...
-            "default-src 'self' 'unsafe-eval' 'unsafe-inline'; connect-src *"
-            if environment_name is ApplicationEnvironment.DEVELOPMENT
-            else f"default-src 'self' 'unsafe-eval' 'unsafe-inline' {blob_storage_url} {static_url}",
+            response, build_csp_header(environment_name, blob_storage_url, static_url)
         )
 
         return response
 
 
+def build_csp_header(environment, blob_storage_url, static_url):
```
jasonburkert-ccpo, line 169: I think it might be a good idea to add a unit test for this new function.  That will maintain a check on future changes to this security-sensitive area.

kylelaker-ccpo:
Looks like the meat of this change was included in #2417; however, that PR did not add the test coverage that we were looking for. I am going to keep this branch around to ensure we add adequate tests.

PR #2397, opened 2021-05-26
Add script/refresh_develop

jasonburkert-ccpo:
This script will refresh your ATAT dev environment. It does the following...

- Update Linux OS
- Install and update python tooling like pyenv, python, pip, and poetry
- Install and update linux packages like redis, postgresql, nodejs, yarn
- Start or restart services Redis and Postgres
- Update local ATAT git repo with latest from 'origin/develop'
- Install and update python and javascript packages
- Update Postgres database schema, wipe, and seed with sample data
- Create self-signed SSL certificate to support https
- Display a comprehensive list of versions used in your dev environment


jasonburkert-ccpo:
There are some Yes/No prompts to skip some of the steps, and avoid some more aggressive options.
I've only run this on Ubuntu under WSL.

I'm wondering if the name could be improved and whether it should be adapted to macOS also.

PR #2398, opened 2021-05-27
Improve test code and failure messages

kylelaker-ccpo:
This handles three related issues to improving the test code in the ATAT codebase. First, we have not been running `pylint` and related tooling on the test code. This has resulted in unused imports and variables and other issues. Those issues have been corrected and the quality checks have been enabled for the Python checks.

Second, in the Python `assert` statement a "fail message" can be provided that is printed when the expression evaluates to False; however, when this is done in tests run with pytest, it breaks assert introspection, per [the documentation](https://docs.pytest.org/en/6.2.x/assert.html). We use these messages in a few modules throughout our code base. This adds a new linter to ensure that we do not introduce any further occurrences (and removes existing occurrences as well).

This will also lay a foundation for adding additional `pylint` checkers in the future.


pylint_custom/assert_introspection.py
```diff
@@ -0,0 +1,41 @@
...
+            "Asserts in tests should not have a fail message.",
+        ),
+    }
+
+    def __init__(self, linter=None):
+        super(AssertIntrospectionChecker, self).__init__(linter)
+        self._module_stack = []
+
+    def visit_module(self, node):
+        self._module_stack.append(is_a_test(node))
```
jasonburkert-ccpo, line 25: @kylelaker-ccpo Just your own words coming back to you...
```suggestion
    def visit_module(self, node):
        """
        Tracks whether the module we're "in" is a "test" module.
        """
        self._module_stack.append(is_a_test(node))
```

pylint_custom/assert_introspection.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def visit_assert(self, node):
+        if self._module_stack[-1] and node.fail:
+            self.add_message("assert-message-present", node=node)
```
jasonburkert-ccpo, line 29:
```suggestion
    def visit_assert(self, node):
        """
        Ensure that if we're in a test module that no assert statements have a fail attribute.
        """
        if self._module_stack[-1] and node.fail:
            self.add_message("assert-message-present", node=node)
```

pylint_custom/assert_introspection.py
kylelaker-ccpo: Rewrote the function and made the comment slightly different. I believe I met the spirit of this suggestion.

pylint_custom/assert_introspection.py
kylelaker-ccpo: Rewrote the function and made the comment slightly different. I believe I met the spirit of this suggestion.

jeffsegal-ccpo (approved):
Thanks Kyle, this definitely helps keep us even more tidied up. Don't see any reason not to be linting & formatting our tests.

jasonburkert-ccpo (approved):
Reviewed each change.  Ran tests locally.  Looks good.

PR #2399, opened 2021-05-27
Update Actionable TODOs to include link to Jira for tracking purposes

zachclark-ccpo:
As a request from Chris Whitten, updated all actionable TODOs with the appropriate JIRA ticket number for easier tracking.
Each TODO now has the number next to it (e.g. `TODO: (AT-XXXX)`) as well as clickable link to the actual JIRA ticket.


kylelaker-ccpo (changes requested):
Thanks for getting these added in! Make sure that you're putting the full ticket ID for each TODO. The structure is `AT-XXXX` (a few are missing `AT-`). And if you've got a simple process for modifying this, `TODO(AT-XXXX):` would be a little cleaner in some ways. I almost wonder if it'd let us do some automation around reporting if we had such a predictable convention for our TODOs.

Separately, if we do develop a convention for `TODO` format in this ticket, let's add that to our coding styles.

To be clear, the full ticket ID (including `AT-`) must be present on all TODOs before I'll feel comfortable approving. Changing the structure to `TODO(AT-XXXX):` is a nice-to-have and not the reason for me selecting "Request changes".

atat/models/portfolio_state_machine.py
```diff
@@ -177,7 +177,7 @@ def is_csp_data_valid(self, event):
         if self.portfolio.csp_data is None or not isinstance(
             self.portfolio.csp_data, dict
         ):
-            # TODO: Should this throw an exception and fail the portfolio? Is
+            # TODO: (6214) Should this throw an exception and fail the portfolio? Is
```
whit1206, line 180: can you make this AT-6214 please

atat/models/utils.py
```diff
@@ -89,7 +89,7 @@ def claim_many_for_update(resources: List, minutes=30):
         .update({"claimed_until": claim_until}, synchronize_session="fetch")
     )
     if rows_updated < 1:
-        # TODO: Generalize this exception class so it can take multiple resources
+        # TODO: (6217) Generalize this exception class so it can take multiple resources
```
whit1206, line 92: please make this AT-6217

atat/routes/portfolios/invitations.py
```diff
@@ -115,7 +115,7 @@ def invite_member(portfolio_id):
                 "error.html", message="There was an error processing your request."
             )
 
-    # TODO: flash error message if form is invalid
+    # TODO: (6218) flash error message if form is invalid
```
whit1206, line 118: Please make this AT-6218

atat/models/portfolio_state_machine.py
zachclark-ccpo: Will do.

atat/models/utils.py
zachclark-ccpo: Will do.

atat/routes/portfolios/invitations.py
zachclark-ccpo: Will do, good catch.

kylelaker-ccpo (approved):
Thanks, @zachclark-ccpo! Using this consistent format will make it easier in the future to both create a linter rule to ensure that TODOs have a ticket and to create reports on the code base to locate TODOs. Once again, thanks for the depth and detail and willingness to iterate on this ticket.

PR #2403, opened 2021-05-28
Feature/at 6241 UI tests fix failures 

palaparthi-ccpo:
These tests are using weak CSS selectors, and need to be more specific.
1.test_settings
2.test_create_app
Updated to strong CSS Selector to the above tests and also removed the duplicated function in other Page Objects file and updated the test case to use the function from uitests/framework/page_objects/__init__.py  file.
Ticket: AT-6241


anhtran-ccpo:
pull the latest code from develop. there seems to be a conflict that needs to be resolved

jasonburkert-ccpo (approved):
Looks good.

PR #2411, opened 2021-06-01
AT-5883 Drop down menu to edit portfolio permissions

jGonzalez-CCPO:
Description
As a CCPO admin, I want to be able to view and edit permissions, so that I can make corrections when personnel changes arise or when users have issues with their permissions.

See below for an image of the component and see Manager Menu/PPOC in Abstract for design specs.

Note: Abstract/Invision links will only be available until June 2021. A PNG version of the view from the above Abstract link is retained for reference on Confluence here: Manager Menu-PPOC.png. Please see Design Files and Assets  in Confluence for instructions on how you can view design specs using Sketch.

Acceptance Criteria
when a user clicks the … button, a dropdown menu should appear

the dropdown menu should include:

edit permissions

remove manager

when a user hovers over one of the options, it should be brought into focus by being the only font in black (others in gray)

if “edit permissions” is clicked, the Edit Permissions Modal should appear (modal is out of scope for this ticket, see AT-5884: Edit portfolio permissions modalTO DO)

if “remove manager” is clicked, a confirmation box should appear asking the admin to confirm they want to remove a manager

note: Promptworks created this requirement as a best practice, but this ticket wasn’t written until  the last week of Think Co’s contract, meaning Think Co did not have time to design for this. Future development will need to include a confirmation box design and accompanying implementation ticket. 

it neither option is clicked, and the user clicks … again, no changes are made and the menu collapses


kylelaker-ccpo (changes requested):
Not sure why this is removing `yarn.lock`. But that file _must_ remain in the repository.

kylelaker-ccpo:
Sounds like this is still a work in progress. Converting to a draft PR.

PR #2421, opened 2021-06-08
Feature/AT-6251 update browserstack readme page

ehoward-ccpo:
The BrowserStack README page to set up the environment to run test suites and test cases is out of date and needs to be updated and show which terminals are used for each line execution.  

Ticket: AT-6251


jasonburkert-ccpo (approved):
Awesome.

PR #2433, opened 2021-06-14
AT-6233 Upgrade to Jest 27

anhtran-ccpo:
Upgraded to Jest 27 along with other various package updates to fix broken unit tests from the upgrade

There was some changes made to the jest config to get a particular test to work. The default environment the tests run in are node and not a browser like environment.  I have made it default to the jsdom instead. More information can be found here.

https://jestjs.io/docs/configuration#testenvironment-string

If a node environment is desired, they can use the doc block specifying the jest environment to be node instead like so:

```
/**
 * @jest-environment node
 */
```

AT-6233


jasonburkert-ccpo (approved):
Ran `script/update` and ran all tests with `script/test`.  Looks good.
```
:~/atat$ yarn list | grep jest/core
├─ @jest/core@27.0.4
│  ├─ @jest/core@^27.0.4
│  ├─ @jest/core@^27.0.4
```

PR #2435, opened 2021-06-15
AT-6242 Added STDIN to the script for adding ccpo users from a yaml

anhtran-ccpo:
Provided we are taking the yaml file and its content into standard input, this script has been updated to take stdin and leverage the yaml safe load to supply the users into the database. 

If you have the yaml file the command can be tested locally via the following command line
```cat <path_to_ccpo_users_file> | script/add_ccpo_users.py -```

### CODE REVIEW

File: script/add_ccpo_users.py [hunk 1]
```diff
@@ -41,6 +42,10 @@ def _load_yaml(file_):
     config = make_config({"default": {"DEBUG": False}})
     app = make_app(config)
     with app.app_context():
-        ccpo_user_file = sys.argv[1]
-        ccpo_users = _load_yaml(ccpo_user_file)
-        add_ccpo_users(ccpo_users)
+        if sys.argv[1] == "-":
```
kylelaker-ccpo, line 45:
```suggestion
        if len(sys.argv) == 1:
```

Thinking about this, let's make stdin the "default". This also prevents any issues in the very unlikely scenario that the file `-` exists in the current directory.

script/add_ccpo_users.py
```diff
@@ -41,6 +42,10 @@ def _load_yaml(file_):
...
+            ccpo_users = yaml.safe_load(sys.stdin)
+            add_ccpo_users(ccpo_users)
+        else:
+            ccpo_user_file = sys.argv[1]
+            ccpo_users = _load_yaml(ccpo_user_file)
+            add_ccpo_users(ccpo_users)
```
kylelaker-ccpo, line 51:
```suggestion
        else:
            ccpo_user_file = sys.argv[1]
            ccpo_users = _load_yaml(ccpo_user_file)
        add_ccpo_users(ccpo_users)
```

kylelaker-ccpo (changes requested):
Minor requests after thinking about how to capture the intent of stdin a little more. Thanks for your work on this!

script/add_ccpo_users.py
anhtran-ccpo: makes sense. its very unorthodox but do you think it would be better to make it `> 0`

kylelaker-ccpo: `len(sys.argv)` will always be `>0`. `sys.argv[0]` is the name of the program itself. So `== 1` checks that the _only_ argument is the name of the program and that no additional CLI arguments were provided. So we either want `== 1` or `< 2`. And 1 always feels like less of a magic number than 2 :)

script/add_ccpo_users.py
anhtran-ccpo: Updated

kylelaker-ccpo (approved):

PR #2439, opened 2021-06-16
At 6157

danielgottlieb-ctr:
Second attempt for updating JEDI to ATAT for branding


kylelaker-ccpo (changes requested):
As-is, this "breaks" a deployed environment and yields the maintenance page. Will investigate to provide more specific information, but also going to hit request changes now to prevent this from being merged

jasonburkert-ccpo:
It's probably best that we not alter the files in the `/alembic` directory in this manner, since those are versioning files for the database schema and are effectively applied one after the other.  To effect the desired change the column name(s) in question could be revised as yet another alembic change to be applied at the tail end of the chain.

Of course that won't eliminate the mentions of the branding in question from the file system unless the changes can be squashed like git commits.  Perhaps [this](https://stackoverflow.com/a/34492022) could help.

PR #2454, opened 2021-06-23
AT-6274-Value is no more exists on UI so Updated with New Value

palaparthi-ccpo:
Data using in Test no longer exists in the UI ,so updated the value with the existing data.AT-6274


jasonburkert-ccpo (approved):
Simply adjusting to available data

PR #2455, opened 2021-06-23
Pass correct arguments to black Action

kylelaker-ccpo:
It seems that either passing `args` was always a mistake or the Action
recently changed the name of the argument to `black_args`. This was
resuling in a warning on all runs of the workflow.


kylelaker-ccpo:
Example of the previous warning:
image
https://github.com/dod-ccpo/atat/actions/runs/964804740

It never impacted the actual results but it's good to catch it now rather than later.

kylelaker-ccpo:
Oops. Apparently this argument is deprecated too and we need to `src` and `options`.

PR #2457, opened 2021-06-23
Feature/at 5813 chart hover states

DGabriel-CCPO:
Addresses tooltip in charts 

onHover the tooltip would cut off when it reached the end of the div with this PR a custom tooltip has been introduced to mitigate that issue

Ticket: 5813


js/components/bar-taskorder-by-org.vue
```diff
@@ -83,7 +83,77 @@ export default {
...
+
+              innerHtml += '</tbody>';
+
+              var tableRoot = tooltipEl.querySelector('table');
+              tableRoot.innerHTML = innerHtml;
+            }
+
+            var position = this._chart.canvas.getBoundingClientRect();
+
+
```
tonysumner-ccpo, line 145: Don't do it now, but when we revisit this, put all these styles in a class and then add the class to the element, with something like tooltipEl.className += << customClassName >>

PR #2460, opened 2021-06-23
remove orphaned 673k photo

jasonburkert-ccpo:
No references to this in repo.  Doesn't belong here.


kylelaker-ccpo:
Those test failures are kinda sad. Are we hardcoding references to files we're not using?

